PR #6, opened 2016-03-10
First code reorganization

oesteban:
A first polishing to make code more readable. Also added install script and set a modular structure.


rwblair:
Now why did I do that....

Issue #22, opened 2016-04-05
BIDS reader function


oesteban:
Write a function that you provide with a subject id and returns a dict with the full bids information about that subject (not only files, but also parameters)

oesteban:
Hi @rwblair, I think this is already solved in your settings. Do you think we can merge it?

PR #24, opened 2016-04-19
Updates to how dataset is read and workflows executed on it.

rwblair:
Expanded on gather bids data to allow for workflows to be executed per subject session. Still some issues with executing EPI Unwrap Workflow.


oesteban:
Close #22.

PR #26, opened 2016-04-22
Updates to collect bids to match original workflows gathered files

rwblair:
Theres some unit tests to boot, but these still need to be integrated into circle ci.


rwblair:
Also added empty file mirroring the aa connectivity dataset. Had to remove subjects S8848XEU, S8576CQQ, and S6086STG since they only had T2s in their anatomical directory.

PR #36, opened 2016-05-11
[FIX] relative imports

oesteban:
Fixes some problems from #35.


oesteban:
It works now. Test failed because the subject I picked up for tests does not have T1 :S

Issue #62, opened 2016-06-07
Change workflow name: se_pair_workflow


oesteban:
Use a more appropriate name.

The naming ```se_pair_workflow``` happens to be incorrect. Actually, in the AA database, the fieldmap is computed from 8 images (four pairs of SE acquisitions). We should think of a different way to name all of this because _pair_ is particularly misleading: it is easily confounded with computing the fieldmap using a _pair_ of GRE acquisitions (#20) and looking at the phase difference. Actually, the latter case is a lot more appropriate for the _pair_ word, since it is always computed with two (you could use more, but I don't think anybody is doing that).

On the other case (our current ```se_pair_workflow```) it is appropriate to have more than two images. Results improve significantly.

shoshber:
@oesteban do you have a suggestion for what the new name should be? Perhaps SpinEchoWorkflow (if I changed it to a class, which I kinda want to do) or spinecho_workflow (if we keep it as a function)?

PR #71, opened 2016-06-22
WIP: add tests/refactor fieldmap code

shoshber:


oesteban:
@shoshber, maybe this would fit in this PR since you are refactoring the fieldmaps machinery.

The naming ```se_pair_workflow``` happens to be incorrect. Actually, in the AA database, the fieldmap is computed from 8 images (four pairs of SE acquisitions). We should think of a different way to name all of this because _pair_ is particularly misleading: it is easily confounded with computing the fieldmap using a _pair_ of GRE acquisitions and looking at the phase difference. Actually, the latter case is a lot more appropriate for the _pair_ word, since it is always computed with two (you could use more, but I don't think anybody is doing that).

On the other case (our current ```se_pair_workflow```) it is appropriate to have more than two images. Results improve significantly.

oesteban:
Actually, this was already open: #62

shoshber:
Circle ci test is failing because the docker image does not rebuild unless preprocessing-workflow/master is changed, but it needs to be rebuilt to read the new dependencies.

oesteban:
> Circle ci test is failing because the docker image does not rebuild unless preprocessing-workflow/master is changed, but it needs to be rebuilt to read the new dependencies.

Are you sure? I'd say you are not not using the docker image to run your tests... But I might be wrong because I just took a very brief look into this.

circle.yml
```diff
@@ -20,6 +20,9 @@ dependencies:
     - docker build -t oesteban/fmriprep .
 test:
   override:
+    # run unit tests
+    - pwd
+    - python -m unittest discover test
```
oesteban: This is not using the docker image to run your tests...

fmriprep/workflows/fieldmap/se_pair_workflow.py
```diff
@@ -101,41 +97,6 @@ def se_pair_workflow(name='Fieldmap_SEs', settings=None):  # pylint: disable=R09
 
     return workflow
 
-def fieldmap_to_phasediff(name='Fieldmap2Phasediff'):
```
oesteban: I'm not totally convinced that we need to have fieldmap submodules, what is the rationale behind this?

circle.yml
shoshber: thank you! I had not looked into this very much but I think you just saved me a lot of time

fmriprep/workflows/fieldmap/se_pair_workflow.py
shoshber: Based on this (copied from our email thread): "The sepair_wf workflow was conceived to have the same outputs that any other fieldmap correction workflow would have."

This means that all fieldmap workflows will have similarities (outputs) that make them easy to sub-in/out for each other. It is generally good programming practice, as you know, to factor code so that nothing is repeated. I am most comfortable writing object-oriented style code, so (in my head) the natural way to organize the code would be with an abstract FieldmapWorkflow.py class that defines these outputs, and several fieldmap subclasses that actually implement the necessary workflows. This is just a middle step toward that goal, so I can see how it would not feel justified. The immediate goal here was just to split the fieldmap.py file up into smaller ones, as it was getting monstrously long and was about to get longer.

Does that clarify my thoughts here? Do you have any objections?

oesteban: Ok no problem, it is true it gets too long.

Regarding a more object oriented approach, I am totally in favor of that. This is why I started a never ended PR for nipype with ```InterfacedWorkflows``` that, in practice, solved this problem. I see a lot of code duplicated from that PR here if you decide to dive into that. Feel free to look at it first.

shoshber: https://github.com/nipy/nipype/pull/882

circle.yml
```diff
@@ -20,6 +20,11 @@ dependencies:
     - docker build -t oesteban/fmriprep .
 test:
   override:
+    # run unit tests
+    - source activate crnenv
```
oesteban: the workflow is containerized in a docker image. Thus, anything you want to do should be inside the image. To run code inside the image there are several strategies (see https://github.com/poldracklab/preprocessing-workflow/blob/master/build/circle_tests.sh#L6).

Basically, you may want to write an entry point for the docker image (like https://github.com/poldracklab/preprocessing-workflow/blob/master/build/files/run_fmriprep) in which you source the environment and call the tests. Finally, you will call the fmriprep docker image setting this new entrypoint (with something like ```--entrypoint=/usr/local/bin/run_shoshber_tests```). This will replace the default entrypoint ```run_fmriprep```).

Anything you place in that directory with the preffix ```run_``` will be placed and granted permissions in the docker image automatically.

build/circle_tests.sh
```diff
@@ -3,4 +3,6 @@
 set -x
 set -e
 
+docker run -t --entrypoint="/usr/bin/run_unittests" -w "/root/src/preprocessing-workflow" oesteban/fmriprep
```
oesteban: You are almost there. Before the image name ```oesteban/fmriprep``` you need to specify the entrypoint. In your case it is ```--entrypoint=/usr/bin/run_unittests```. And you're done!

oesteban: BTW, if you want CircleCI to understand this test as a separate test (and run always the second line of this file), then you will need to put this line as a new entry in the ```circle.yml``` file. Actually, having this file is not necessary here, we could also get the smoke test out to the yml file and remove this build/circle_tests.sh

shoshber: Great suggestion! I won't act on it now because not trying to find a perfect solution here, especially because I'm supposed to be working on fieldmap stuff--I just want to close this pull request because long-running pr's are bad form in my book. It is true that the test infrastructure (e.g., files in the `build` directory) in this package is a little wonky. Maybe create an issue for un-wonkifying the test infrastructure?

oesteban: @shoshber Ok, now I see that the entrypoint was there :). Sorry about that.

fmriprep/workflows/fieldmap/se_pair_workflow.py
shoshber: InterfacedWorkflows looks really nice--what are the blockers to getting it into nipype?

shoshber:
Yussssssss

PR #73, opened 2016-06-23
make settings['workdir'] always accessible; absolute imports

shoshber:
@oesteban @rwblair I would like confirmation that this is ok from one of you before I merge

### DIFF
--- a/fmriprep/run_workflow.py
+++ b/fmriprep/run_workflow.py
@@ -100,6 +100,12 @@ def main():
         print('fmriprep version ' + __version__)
         exit(0)
 
+    # Warn for default work/output directories
+    if (opts.work_dir == parser.get_default('work_dir') or
+          opts.output_dir == parser.get_default('output_dir')):
+        logging.warning("work-dir and/or output-dir not specified. Using " +
+                        opts.work_dir + " and " + opts.output_dir)
+
     settings = {
         'bids_root': op.abspath(opts.bids_root),
         'write_graph': opts.write_graph,
--- a/fmriprep/run_workflow.py
+++ b/fmriprep/run_workflow.py
@@ -16,6 +16,7 @@ from __future__ import unicode_literals
 from argparse import ArgumentParser
 from argparse import RawTextHelpFormatter
 from multiprocessing import cpu_count
+import logging
 import os
 import os.path as op
 import matplotlib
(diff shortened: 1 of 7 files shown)


oesteban:
This is a great idea. Maybe you can also address this problem on mriqc: https://github.com/poldracklab/mriqc/issues/111

I'm pretty sure this will also happen in the preprocessing pipeline.

rwblair:
I think it would be best to fail with an error message and force the user to choose an explicit work directory. The output to the work directory can take up a non trivial amount of space so it would be good to make the user think about where its all going to end up...

But a warning message saying that no work directory was provided and that the current directory is being used may be enough. If something bad happens to their current working directory/partition they can at least go back and see what happened.

fmriprep/run_workflow.py
```diff
@@ -116,18 +116,19 @@ def main():
...
-        log_dir = op.join(settings['work_dir'], 'log')
-        if not op.exists(log_dir):
-            os.makedirs(log_dir)
-
-        # Set nipype config
-        ncfg.update_config({
-            'logging': {'log_directory': log_dir, 'log_to_file': True},
-            'execution': {'crashdump_dir': log_dir}
-        })
+    settings['work_dir'] = (op.abspath(opts.work_dir)
```
oesteban: This is too verbose, IMHO. For this very same behavior, it is enough to:
1. Define a ```default=os.getcwd()``` in the argument (line 95)
2. Do ```settings['work_dir'] = op.abspath(opts.work_dir)

shoshber: I see what you're saying. I get too excited about ternary operators!

That said, I'm about to re-write this anyway, to follow Ross's recommendation (output a warning, and proceed).

fmriprep/run_workflow.py
```diff
@@ -97,7 +96,7 @@ def main():
     opts = parser.parse_args()
 
     if opts.version:
-        print('fmriprep version ' + __version__)
+        logging.info('fmriprep version ' + __version__)
```
oesteban: Why should we use logging for an option that only shows the current version? Isn't it a little bit of overkilling the problem?

oesteban:
@shoshber I have added #78 for a more consistent setting of output and working directories, including locks for multiprocessing operation. I suggest to merge #78 first and then this one, maybe adding the warning when default settings are to be used as regards outputs and working directories.

All the stuff about absolute imports is OK.

fmriprep/run_workflow.py
shoshber: reverted

PR #86, opened 2016-07-22
changes for splitting epi workflow and more!

rwblair:


rwblair:
Not sure that last merge I did is quite rght.

PR #113, opened 2016-08-16
Update to use grabbit.

rwblair:
Grabbit now being used to pull in data in utils.misc.collect_bids_data.

Dockerfile updated to install freesurfer.


oesteban:
This is so cool! Thanks!

PR #117, opened 2016-08-19
make cli BIDS-Apps compliant

shoshber:
https://github.com/poldracklab/preprocessing-workflow/issues/106

### DIFF
--- a/fmriprep/run_workflow.py
+++ b/fmriprep/run_workflow.py
@@ -16,13 +16,13 @@ from __future__ import unicode_literals
 from argparse import ArgumentParser
 from argparse import RawTextHelpFormatter
 from lockfile import LockFile
+
 import logging
 from multiprocessing import cpu_count
 import os
 import os.path as op
 import pkg_resources as pkgr
-
-from pprint import pprint as pp
+import pprint
 
 def main():
     """Entry point"""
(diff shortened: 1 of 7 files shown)


oesteban:
can you rebase? thx!

shoshber:
merged

fmriprep/run_workflow.py
```diff
@@ -28,14 +30,24 @@ def main():
     from nipype.pipeline import engine as pe
     from fmriprep import __version__
     from fmriprep.workflows import fmriprep_single
-
+    from fmriprep.utils.misc import collect_bids_data
```
oesteban: import should go away

fmriprep/run_workflow.py
```diff
@@ -124,10 +131,13 @@ def main():
 
     # Determine subjects to be processed
     subject_list = opts.subject_id
-    if not subject_list or subject_list is None:
+
+    if not subject_list or len(subject_list) == 0:
         subject_list = [op.basename(subdir)[4:] for subdir in glob.glob(
             op.join(settings['bids_root'], 'sub-*'))]
 
+    logger.info("subject list: " + pprint.pformat(subject_list))
```
oesteban: subject_list is just a list, so we can also remove the pprint: ```logger.info("subject list: {}", ', '.join(subject_list))```

oesteban:
I'm still not convinced by changing the collect_bids_data return object because the tests expect it to be in a certain way. I would prefer to adapt the tests, regardless we eventually move to a result object more similar to what you expect now in the tests.

shoshber:
Yeah I forgot this pr had those commits in it too. I will cherrypick and make a new pr for this as well.

Issue #107, opened 2016-08-14
Fix smoke test on circle ci


oesteban:
Right after @suyashdb finishes #104 . Also depends on #108

shoshber:
#108 is closed

oesteban:
Yes, that means that the docker image including FreeSurfer is ready. Now this only depends on #104

shoshber:
* for now just the one with phase diff

shoshber:
(summary of live discussion w/ @oesteban )
https://circleci.com/gh/shoshber/preprocessing-workflow/2 Downloading the test data has taken almost 3 hours at this point and it's still not done--@oesteban will look at making the test data smaller

oesteban:
@shoshber  https://3552243d5be815c1b09152da6525cb8fe7b900a6.googledrive.com/host/0BxI12kyv2olZVUswazA3NkFvOXM/ds054_downsampled.tar.gz

shoshber:
[ ] set cached_directories in circle ci

oesteban:
Ok, as we can see in https://circleci.com/gh/poldracklab/preprocessing-workflow/206, the smoke test is functional again. It's not passing, but that's something to be solved in #124 . This one is finished.

shoshber:
I severely misunderstood the point of this issue then. We should talk about that

oesteban:
https://circleci.com/gh/poldracklab/preprocessing-workflow/209

Issue #104, opened 2016-08-12
Create test dataset of subsampled images


oesteban:
Create a dataset and downsample it to implement smoke tests. This small dataset would include one subject from ds005 (no fieldmaps), one subject from HCP in which only the multiple-SEs fieldmaps is found (multiple-SEs fieldmaps) and one subject from HCP in which only the magnitude&phase fieldmaps are found (two-GRE fieldmaps).

suyashdb:
@oesteban - i have copied two datasets to '/corral-repl/utexas/poldracklab/openfmri/test_datsets_hcp/'.

ds 0005 - No fieldmaps
ds054 - magnitude&phase fieldmaps are found (two-GRE fieldmaps)
HCP - In process (4: Multiple phase encoded directions)

suyashdb:
@oesteban @chrisfilo 
Looks like there is naming nomenclature issue in hcp/fmaps .. 
If i take - LR as index - 01
RL as index - 02
and latter LR/RL in filename is encoded in .json as -i, +i .. Event then two diff fmap files will have same name as follows -

rfMRI_REST1_LR_285345_3T_SpinEchoFieldMap_LR.nii.gz ->  sub-285345_dir-01_epi.nii.gz

rfMRI_REST1_LR_285345_3T_SpinEchoFieldMap_RL.nii.gz -> sub-285345_dir-01_epi.nii.gz

There .json file will have filed intended for field pointing to two diff func-files but how can we have fmap files with same name...?

suyashdb:
I am using case 4 for fmap naming under BIDS Std:
8.9.4 Case 4: Multiple phase encoded directions (topup)

fmap/sub<label>[_acq<label>]_dir<index>_epi.nii[.gz]
fmap/sub<label>[_acq<label>]_dir<index>_epi.json

oesteban:
rfMRI_REST1_LR_285345_3T_SpinEchoFieldMap_RL.nii.gz -> sub-285345_dir-0**2**_epi.nii.gz

suyashdb:
then rfMRI_**REST2_LR**_285345_3T_SpinEchoFieldMap_RL.nii.gz -> sub-285345_dir-02_epi.nii.gz

suyashdb:
The solution to this will be using the optional acq tag:
sub-285345_acq-REST1LR_dir-01_epi.nii.gz

But do you want that ..?

oesteban:
ah ok, I see.

[...]

Issue #103, opened 2016-08-12
test elegant failure & report on bad/variable BIDS data


oesteban:
Create several fake BIDS structures of empty files, with different combinations of subjects, modalities, sessions, tasks, acquisitions, and runs, and exercise #102 against it:
- [ ] Create the fake BIDS structures. Test a matrix of valid cases: `[one bold, several bold, no fieldmap, se-fmap, gre-fmap, se&gre-fmap, no sbref, sbref, no t1, t1, several t1, no t2, t2, several t2] x [one bold, several bold, no fieldmap, se-fmap, gre-fmap, se&gre-fmap, no sbref, sbref, no t1, t1, several t1, no t2, t2, several t2]`
- [ ] Create fake BIDS structures with problems:
  - [ ] there is SBRef but no corresponding bold
  - [ ] there is only one PE direction for SE-fieldmaps
  - [ ] there is only phasediff and no magnitude in fmap/
- [x] Write corresponding unit tests (automated, so that we can flexibly add more fake structures in the future without need of modifying the test code)
- [x] Remove any dependency in real data (https://github.com/shoshber/preprocessing-workflow/blob/nipypedep87/test/utils/test_misc.py#L15)

oesteban:
We're moving this to pybids, is that correct @chrisfilo ?

shoshber:
exactly which module is this supposed to test?

shoshber:
also, I don't think this can be *fully* automated, unless I'm missing something important

oesteban:
Well, the structures of fake files (```touch filename```) can be hard-coded or written in runtime from a tree or dictionary.

Particularly, I'd like this interface (https://github.com/oesteban/preprocessing-workflow/blob/master/fmriprep/interfaces/bids.py#L43) to catch all the problematic cases, discriminating by subjects (meaning, if only one subject is missing the bold image, then we shouldn't break all) and runs (meaning, if one subject has 5 runs whole but 1 is light of the bold then this subject can be processed on the 5 correct runs and we need to notice that one was omitted for having one image missing).

shoshber:
Based on previous discussions, I think we will not worry about handling of datasets with varying subject data for now.

Let's wait on this issue until after it's moved to pybids, since that might change the code organization significantly

PR #118, opened 2016-08-19
depend on latest nipype master

shoshber:


oesteban:
can you rebase, so I can merge this?

shoshber:
I pulled from poldracklab's master--that's what you mean, right?

chrisfilo:
[Merging master is slightly different from rebasing](https://www.atlassian.com/git/tutorials/merging-vs-rebasing/workflow-walkthrough), but will solve the problem as well.

fmriprep/run_workflow.py
```diff
@@ -28,7 +30,7 @@ def main():
     from nipype.pipeline import engine as pe
     from fmriprep import __version__
     from fmriprep.workflows import fmriprep_single
-
+    from fmriprep.utils.misc import collect_bids_data
```
oesteban: The collect_bids_data is not used here anymore, please remove.

fmriprep/run_workflow.py
```diff
@@ -80,8 +76,12 @@ def main():
         'work_dir': op.abspath(opts.work_dir)
     }
 
+    # set up logger
+    logger = logging.getLogger('cli')
+
     if opts.debug:
         settings['ants_t1-mni_settings'] = 't1-mni_registration_test'
+        logger.setLevel(logging.DEBUG)
```
oesteban: :+1: this is a great idea :)

fmriprep/utils/misc.py
```diff
@@ -63,38 +61,44 @@ def collect_bids_data(dataset, subject, spec=None, session=None, run=None):
...
+
+        session_data = copy.deepcopy(INPUTS_SPEC)
+        fieldmap_files = [x.filename for x in layout.get(**queries['fmap'])]
+        session_data['fmap'] = fieldmap_files
+        t1_files = [x.filename for x in layout.get(**queries['t1w'])]
+        session_data['t1w'] = t1_files
+        sbref_files = [x.filename for x in layout.get(**queries['sbref'])]
+        session_data['sbref'] = sbref_files
+
+        loop_on = ['session', 'run', 'acquisition', 'task']
```
oesteban: I think @rwblair was serializing sessions as well with this loop. I think this file deserves a different PR where he can check it out and you can go through it together.

test/utils/test_misc.py
```diff
@@ -11,11 +12,12 @@ class TestCollectBids(unittest.TestCase):
...
-            all_subjects = misc.collect_bids_data(c.DATASET)
+            subject_data = misc.collect_bids_data(c.DATASET, self.subject_id,
+                                                  c.BIDS_SPEC)
             self.imaging_data = {
-                self.subject_id: all_subjects[self.subject_id]
+                self.subject_id: subject_data
             }
-        except Exception as e:
+        except IOError as e:
             url = "http://googledrive.com/host/0BxI12kyv2olZbl9GN3BIOVVoelE"
```
oesteban: Please update the link with the new one (see #104)

oesteban:
This PR is great. My only concern is ```fmriprep/utils/misc.py```. I think @rwblair had already given support for multisession datasets in it. So let's split it, hold this file off the PR and submit it in another different PR that Ross will review when he is back with you. Sounds good?

test/utils/test_misc.py
shoshber: I will do this in another pr

fmriprep/utils/misc.py
shoshber: I had to do this to get the tests to work. Can we commit this for now?

fmriprep/run_workflow.py
shoshber: done

oesteban:
Ok, I see.

The problem here is that tests are anchored to the old downsampled dataset, right? (https://github.com/shoshber/preprocessing-workflow/blob/nipypedep87/test/utils/test_misc.py#L15). Let's think a bit about this.

We basically need to test two things:

  1. We need to make sure that fmriprep correctly builds up the list of subjects x sessions x image-sets to be passed on the workflow. This means testing the ```collect_bids_data``` function. This is why #103 exists.
  2. Smoke tests on the pipeline itself (#107) - BTW I'm going to edit this issue to make it more representative of the idea.

Basically, you need to split the current output list from ```collect_bids_data``` into sessions, so that it works with the old dataset, which was intended for the smoke tests in 2).

I would propose to forget about fixing circleci right now (we'll keep the annoying reds for a while), and merge this PR which intention was only _depend on latest nipype master_.

Next steps would be: #103 (there is no need for real/downsampled data to make this real) and then #107 (this needs the dataset that @suyashdb arranged).

shoshber:
Alright, that makes sense. I think I can cherry-pick just the relevant commits to leave unit tests/collect_bids_data untouched. Didn't see #103 until now.

shoshber:
closing this, see new pr

PR #122, opened 2016-08-23
depend on latest nipype master 2.0

shoshber:
replaces old pr

### DIFF
--- a/fmriprep/run_workflow.py
+++ b/fmriprep/run_workflow.py
@@ -16,11 +16,14 @@ from __future__ import unicode_literals
 from argparse import ArgumentParser
 from argparse import RawTextHelpFormatter
 from lockfile import LockFile
+
 import logging
 from multiprocessing import cpu_count
 import os
 import os.path as op
 import glob
+import pkg_resources as pkgr
+import pprint
 
 def main():
     """Entry point"""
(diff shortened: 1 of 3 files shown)


fmriprep/run_workflow.py
```diff
@@ -16,11 +16,14 @@
 from argparse import ArgumentParser
 from argparse import RawTextHelpFormatter
 from lockfile import LockFile
+
 import logging
 from multiprocessing import cpu_count
 import os
 import os.path as op
 import glob
+import pkg_resources as pkgr
```
oesteban: pkgr is not used... why this import?

fmriprep/run_workflow.py
```diff
@@ -128,6 +136,8 @@ def main():
         subject_list = [op.basename(subdir)[4:] for subdir in glob.glob(
             op.join(settings['bids_root'], 'sub-*'))]
 
+    logger.info("subject data is " + pprint.pformat(subject_list))
```
oesteban: Same as in the other PR, this is just a list so pprint is overkill the problem and adds one more import.

oesteban:
any news on these two small comments?

fmriprep/run_workflow.py
shoshber: cherrypicking is hard haha

PR #125, opened 2016-08-27
pull onto temp docker branch for docker testing

shoshber:


circle.yml
```diff
@@ -12,12 +12,12 @@ dependencies:
     - bash build/circle_setup.sh
   override:
     - if [[ -e ~/docker/image.tar ]]; then docker load -i ~/docker/image.tar; fi
-    - docker build -t oesteban/fmriprep:latest .
-    - mkdir -p ~/docker; docker save oesteban/fmriprep:latest > ~/docker/image.tar
+    - docker build -t poldracklab/preprocessing-workflow:latest .
```
oesteban: why change the name? I prefer poldracklab/fmriprep better...

fmriprep/run_workflow.py
```diff
@@ -76,6 +76,11 @@ def main():
         'skull_strip_ants': opts.skull_strip_ants,
         'output_dir': op.abspath(opts.output_dir),
     }
+    settings['work_dir'] = settings['output_dir'] # other wfs assume this is set
```
oesteban: This does not make any sense. We have now an orphaned option.

Additionally, settings['work_dir'] is necessary along the process. We need to roll this back.

fmriprep/run_workflow.py
```diff
@@ -138,7 +134,7 @@ def main():
         subject_list = [op.basename(subdir)[4:] for subdir in glob.glob(
             op.join(settings['bids_root'], 'sub-*'))]
 
-    logger.info("subject list: {}", ', '.join(subject_list))
+    logger.info("subject list: " + str(subject_list))
```
oesteban: logging should be done with ```logger.info('some message with format %s tokens %s', string1, string2)```.

You are right, the former version didn't work because of the curly braces, so please replace this with ```logger.info("subject list: %s", ', '.join(subject_list))```

Finally, I would avoid the explicit use of ```str``` whenever possible, since it is different in python 2 and 3

build/circle_setup.sh
```diff
@@ -7,9 +7,9 @@ set -e
 if [[ ! -d ${HOME}/scratch/data/aa_conn ]]; then
     # Folder for downloads
     mkdir -p ${HOME}/downloads
-    wget -c -O ${HOME}/downloads/AA_Conn.tar "http://googledrive.com/host/0BxI12kyv2olZY0ZkZVB4TUVtSWM"
+    wget -c -O ${HOME}/downloads/testdata.tar "https://3552243d5be815c1b09152da6525cb8fe7b900a6.googledrive.com/host/0BxI12kyv2olZVUswazA3NkFvOXM/ds054_downsampled.tar.gz"
```
oesteban: The new file is compressed. Also, you can remove the ``` -O ${HOME}/downloads/testdata.tar``` because the link I provided has filename (it didn't before, that's why there is that ```-O``` argument.

build/circle_setup.sh
```diff
@@ -7,9 +7,9 @@ set -e
...
     mkdir -p ${HOME}/scratch/data/
-    tar xf ${HOME}/downloads/AA_Conn.tar -C ${HOME}/scratch/data
+    tar xf ${HOME}/downloads/testdata.tar -C ${HOME}/scratch/data
```
oesteban: ```tar xzf``` you need to add the ```z``` to uncompress.

PR #129, opened 2016-08-28
stuff from trying to complete iss107

shoshber:
just little things


shoshber:
I don't want all of it to go to waste. What would you like to keep, @oesteban

PR #143, opened 2016-08-30
[WIP] Issue44

rwblair:
I'm still working on testing this. Opening PR in case @oesteban has any comments while its being tested.


oesteban:
I don't see the circleCI for some weird reason. But the PR looks good to me.

rwblair:
I think its because my fork of the project has a different repo name. I'll try and change it and see if that fixes it.

oesteban:
Looks like you are right, the name change is probably affecting your builds

PR #145, opened 2016-08-30
quicker docker image rebuilds

chrisfilo:


oesteban:
Interesting, LGTM

oesteban:
@rwblair do you want to have the honor?

PR #151, opened 2016-09-09
[WIP] Issue147

rwblair:
Attempt to make both subject_list and epi files for subjects iterable. Not complete, interested to see what circleci has to say about it.

### DIFF
--- a/fmriprep/data/utils.py
+++ b/fmriprep/data/utils.py
@@ -23,6 +23,14 @@ import hashlib
 import subprocess as sp
 from fmriprep.data.compat import _urllib, md5_hash
 
+try:
+    from urllib.parse import urlparse, urlencode
+    from urllib.request import build_opener, urlopen, Request
+    from urllib.error import HTTPError
+except ImportError:
+    from urlparse import urlparse
+    from urllib import urlencode
+    from urllib2 import build_opener, urlopen, Request, HTTPError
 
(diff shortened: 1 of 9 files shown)


fmriprep/data/utils.py
```diff
@@ -23,6 +23,14 @@
 import subprocess as sp
 from fmriprep.data.compat import _urllib, md5_hash
 
+try:
```
oesteban: I generally prefer this solution http://python-future.org/compatible_idioms.html#urllib-module:
```
from future.standard_library import install_aliases
install_aliases()

from urllib.parse import urlparse, urlencode
from urllib.request import urlopen, Request
from urllib.error import HTTPError
```

setup.py
```diff
@@ -8,6 +8,9 @@
 import os
 import sys
 
+from fmriprep.info import (__version__, __author__, __copyright__, __credits__,
```
oesteban: This should not be necessary, we have this: https://github.com/rwblair/fmriprep/blob/3866c4e555f4ebef0d1fc2dab0c4013126a60846/setup.py#L23 for that.

The reason is: if you import fmriprep in the setup.py, then you need to have fmriprep installed for setup.py to work correctly.

rwblair: Kk, I'll remove it. When I was attempting to do an install of fmriprep locally it was failing because it couldn't get at the information in fmriprep.info. I'll try and rework it to not import from a library that isn't installed yet.

PR #152, opened 2016-09-13
[FIX] setup.py in python 3

oesteban:
To be merged AFTER #151


oesteban:
@chrisfilo @rwblair @shoshber please review :)

fmriprep/run_workflow.py
```diff
@@ -3,33 +3,26 @@
 # @Author: oesteban
 # @Date:   2015-11-19 16:44:27
 # @Last Modified by:   oesteban
-# @Last Modified time: 2016-08-30 16:52:12
+# @Last Modified time: 2016-09-13 13:37:43
```
shoshber: Do we actually need this stuff at the top? We have git!

fmriprep/run_workflow.py
```diff
@@ -74,6 +67,14 @@ def main():
...
 
     opts = parser.parse_args()
+    create_workflow(opts)
+
+
+def create_workflow(opts):
+    import logging
+    from nipype import config as ncfg
+    from fmriprep.workflows import base as fwb
+    from fmriprep.workflows.base import base_workflow_enumerator
```
shoshber: I've been meaning to ask this--how come sometimes we import at the top of the file, and other times inside the function?

requirements.txt
```diff
@@ -1,11 +1,2 @@
...
 -e git+https://github.com/incf/pybids.git@master#egg=pybids
-numpy
-lockfile
-future
-scikit-learn
-matplotlib
-nilearn
-sklearn
-nibabel
-niworkflows>=0.0.3a3
```
shoshber: This breaks `pip install -r requirements.txt`. Is there another solution?

setup.py
```diff
@@ -3,69 +3,55 @@
...
-    if REQUIREMENTS is None:
-        REQUIREMENTS = []
+    # Python 3: use a locals dictionary
+    # http://stackoverflow.com/a/1463370/6820620
+    ldict = locals()
+    # Get version and release info, which is all stored in fmriprep/info.py
+    module_file = op.join(this_path, 'fmriprep', 'info.py')
+    with open(module_file) as infofile:
+        pythoncode = [line for line in infofile.readlines() if not line.strip().startswith('#')]
+        exec('\n'.join(pythoncode), globals(), ldict)
```
shoshber: since info.py is a python file, can we avoid `exec`?

fmriprep/run_workflow.py
oesteban: You import inside the function when you want to delay those imports until the function is executed. This can be of interest:

- In niype, when you are wrapping functions in a function node (there's no way to import things otherwise).
- In this case: if you import nipype at the top, when you run ```fmriprep -v``` to see the version you get all the warnings that come with nipype, which is pretty annoying.

setup.py
oesteban: I actually started doing that but moved to this exec. Particularly, I was doing that in ```niworkflows```. The result was a chicken-and-egg problem: to get setup.py to the point of understanding nipype as a dependency, niworkflows required itself to be installed. Looking around this was the most effective solution.

If you try to import info.py from the setup.py you are assuming that the code has python package structure (which can change for various reasons). This is a more agnostic way to load those variables.

requirements.txt
oesteban:
```
pip install -e .
python setup.py install
```

Will both install those dependencies if they are missing.

Please see [these recommendations](https://packaging.python.org/requirements/) and [this discussion](https://www.reddit.com/r/Python/comments/3uzl2a/setuppy_requirementstxt_or_a_combination/). This is my interpretation it seems that ```requirements.txt``` should be written for development environments, and production (pypi, mostly) dependencies should be handled with ```setuptools```. Let me know if I misinterpreted these sources

fmriprep/run_workflow.py
shoshber: Cool, those seem like good reasons

requirements.txt
shoshber: Good reading, thanks. I think you are right. Why don't we do what this user suggested: https://www.reddit.com/r/Python/comments/3uzl2a/setuppy_requirementstxt_or_a_combination/cxka1a4

so both approaches will still work?

setup.py
shoshber: Makes sense. If this is a standard approach then 👍

requirements.txt
oesteban: Yes, I think that is one possibility. However, I don't think it makes a lot of sense: the project becomes editable mode always. The only case for which I've seen the requirements.txt is really necessary is autobuilding documentation in RTD.

oesteban:
@rwblair can you have a look?

rwblair:
Changes look good to me. Was able to install via pip install -e . and setup.py install in a python 3.5 environment. Merging.

PR #154, opened 2016-09-14
automatic pep8 linter

shoshber:
The annoying part is that you have to run setup.py after cloning. Couldn't find a better way--I hope that is an acceptable solution.

Also I set the limit really low (-8) because fmriprep is really not close to pep8 at all.


shoshber:
this is on hold until #151 gets done

setup.py
```diff
@@ -63,6 +63,20 @@ def main():
         ],
     )
 
+    # if this is being set up in a git repo
+    if glob('.git'):
+        enforce_pep8()
+
+def enforce_pep8():
```
oesteban: What is this for?

shoshber: it's a cheaty but easy way to check that this is a developer environment using git.

shoshber:
closed after discussion in tech meeting; alternate implementation proposed

PR #158, opened 2016-10-04
[ENH] Remove data downloader, use niworkflows

oesteban:


oesteban:
The downloads problem is kind of blocking everything: sorry for merging without second opinions.

PR #178, opened 2016-11-18
[RTM] Issue176 Interface reports.

rwblair:
Interface reports were originally developed in fmriprep directly. They have since been moved to niworkflows. Current state of the code uses these new report capable interfaces, but leaves the old reporting code in place. No data sinks have been made at this time to collect outputs of the report capable interfaces. These should be added before merging.


fmriprep/info.py
```diff
@@ -62,7 +61,8 @@
 
 LINKS_REQUIRES = [
     'git+https://github.com/nipy/nipype.git@master#egg=nipype',
-    'git+https://github.com/incf/pybids.git@master#egg=pybids'
+    'git+https://github.com/incf/pybids.git@master#egg=pybids',
+    'git+https://github.com/poldracklab/niworkflows.git@master#egg=niworkflows',
```
oesteban: For this to work, you need to keep the old ```'niworkflows'``` entry in the ```REQUIRES``` list. Make sure that there is no version selector though.

fmriprep/run_workflow.py
```diff
@@ -82,7 +82,8 @@ def create_workflow(opts):
         'debug': opts.debug,
         'skull_strip_ants': opts.skull_strip_ants,
         'output_dir': op.abspath(opts.output_dir),
-        'work_dir': op.abspath(opts.work_dir)
+        'work_dir': op.abspath(opts.work_dir),
+        'remove_unnecessary_outputs': False
```
oesteban: I would honestly prefer to have this in your ~/.nipype/nipype.cfg, but whatever you want :). Especially bc you are now connecting the outputs corresponding to the generated reports into a subsequent node. This options will not have any effect in practice.

fmriprep/viz/overlay_3d_report.tpl
```diff
@@ -0,0 +1,34 @@
+<div id='{{unique_string}}'>
```
oesteban: Shouldn't this file belong in ```niworkflows``` ?

requirements.txt
```diff
@@ -1,2 +1,3 @@
 nipype
```
oesteban: I just spotted this: why are we not using the github version with -e?

rwblair: Not sure, I'll try and see how and why this is the way it is.

fmriprep/run_workflow.py
rwblair: Oh yeah this was a hold over from testing something. Removed.

fmriprep/info.py
rwblair: Added niworkflows to requires in info.py

oesteban:
@rwblair all done?

rwblair:
Let me add some datasinks for the new style report nodes.

rwblair:
Ok ready to merge

rwblair:
ok ready to merge.

shoshber:
Have you pulled all my changes into your branch yet? It would be better if you did that before merging this PR to minimize potential merge conflicts.

Also I noticed a few indentation weirdnesses, if you would run pylint one more time to fix up the little things that would be 👍

fmriprep/workflows/sbref.py
```diff
@@ -47,7 +47,7 @@ def sbref_preprocess(name='SBrefPreprocessing', settings=None):
 
     mean = pe.Node(fsl.MeanImage(dimension='T'), name='SBRefMean')
     inu = pe.Node(ants.N4BiasFieldCorrection(dimension=3), name='SBRefBias')
-    bet = pe.Node(BETRPT(frac=0.6, mask=True), name='SBRefBET')
+    bet = pe.Node(BETRPT(generate_report=True, frac=0.6, mask=True), name='SBRefBET')
```
oesteban: Since these interfaces are "report capable", there is no need for this ```generate_report``` right? That should be ```True``` always, WDYT?

oesteban: (I meant to change the base classess, I was not referring to this particular line which is necessary with the current implementation).

rwblair: I think keeping the flag is good if we are going to pass these upstream to nipype at some point. Turning reports on only when needed will probably be desired there.

shoshber:
LGTM but someone else should look at it too

PR #183, opened 2016-12-02
[RTM] store scratch directories in artifacts

chrisfilo:
This PR increases the verbosity of our tests by generating [workflow graphs](https://67-66973489-gh.circle-artifacts.com/0/home/ubuntu/ds005/scratch/workflow_enumerator/graph.dot.svg) and saving scratch directories (however this is only restricted to txt rst svg and html files to save time).

Saving scratch comes at increaseing test time to 9 minutes.


chrisfilo:
I managed to bring it down to 6 minutes. This is ready to merge.

Issue #179, opened 2016-12-01
Docker: consider moving to poldracklab/neuroimaging-core:freesurfer-0.0.2


oesteban:
That docker image is better maintained and it is versioned. The current base image oesteban/crn_nipype:freesurfer is kind of stale (and too heavy).

That said, it should still work, although it is making it difficult for CircleCI to build in #178.

For reference, we can use the Dockerfile of niworkflows. On top of poldracklab/neuroimaging-core:freesurfer-0.0.2 we should just install conda.

oesteban:
Note - this is not urgent, we can address it after the BEAST sprint.

oesteban:
For a reference on how the new dockerfile would look like, we have niworkflows https://github.com/oesteban/niworkflows/blob/master/Dockerfile

chrisfilo:
Superseded by #273

PR #186, opened 2016-12-02
[WIP] speed up docker builds

chrisfilo:


oesteban:
When #206 is merged (incl. https://github.com/chrisfilo/fmriprep/pull/2), we'll be in a good position to make #179 happen. I think that will speed up the building process and I think will also make the caching of docker a bit less unstable.

PR #192, opened 2016-12-02
[RTM] Minor changes to correct anatomical workflow.

rwblair:
so far: update template being used by robust normilization in anatomical workflow


rwblair:
From local tests the update to the template being used resolved issues seen on some datasets

PR #198, opened 2016-12-03
[RTM] Some love to sdc_unwarp

oesteban:
Several improvements to make this workflow a bit more reliable:

- Updated the ReadSidecarJSON interface with the latest version from mriqc (fixes several problems of the former version).
  - To consider: moving in niworkflows.
- Updated the function to create Topup encoding files
  - To consider: submit a complete interface to nipype, since that'd is extremely useful for people if we do it correctly).
  - NOTE: I haven't implemented the "EchoTime" case yet, since we are aiming at Dec 15. In a future, that should be also implemented, along with a better logic (i.e. what to do if a file has all sorts of ways to calculate the echo time).
- Remove the old JSONFileGrabber to get the settings for ants, using the ```from_file``` new feature of nipype.
- By the title of the PR I wouldn't expect less from @rwblair than heart'ing this.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -29,9 +29,9 @@ dependencies:
 test:
   override:
     - docker run -ti --rm --entrypoint="/usr/bin/run_unittests" poldracklab/fmriprep:latest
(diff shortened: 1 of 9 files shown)


oesteban:
I am removing ```--write-graph``` from tests (see #216)

chrisfilo:
Write graph [works without problems in master](https://circleci.com/gh/poldracklab/fmriprep/379). Could you share the graphviz error you are getting? I can try looking into it (I tried checking on Circle, but the logs are truncated at 4Mb). One thing I noticed before was that graphiz was failing when nodes were given names with hyphens in them. This does not seem to be the case here though.

Issue #64, opened 2016-06-07
Replace BBR with bbregister


oesteban:
Depends on #63

PR #206, opened 2016-12-03
[RTM] use nibabel to reorient to avoid unecessary interpolation

chrisfilo:


oesteban:
@chrisfilo - please have a look at the patch I just submitted to your branch, let's get rid of the freesurfer dependency (we were only using mri_convert!)

chrisfilo:
Thanks for the patch! FreeSurfer will come back though (see #63 and #64).

PR #210, opened 2016-12-04
[RTM] remove unnecessary fslroi

chrisfilo:


chrisfilo:
I have verified that the two files have the same affine matrices. I'll merge it because I'm about to propose some changes that build on top of this one.

PR #223, opened 2016-12-05
[RTM] memory consumption driven scheduling

chrisfilo:
This PR is attempting to avoid situations where we processing fails because we are trying to run too many tasks consuming a lot of memory at the same time. Now the size of the biggest _bold file is read and three times this much RAM is requested for each critical 4D jobs.

I did not implement this for the sbref/fieldmaps sub workflows since I know @oesteban is actively refactoring them in #198

### DIFF
--- a/fmriprep/workflows/confounds.py
+++ b/fmriprep/workflows/confounds.py
@@ -40,17 +40,27 @@ def discover_wf(settings, name="ConfoundDiscoverer"):
     signals = pe.Node(nilearn.SignalExtraction(include_global=True, detrend=True,
                                                class_labels=FAST_DEFAULT_SEGS),
                       name="SignalExtraction")
+    signals.interface.estimated_memory_gb = settings[
+                                              "biggest_epi_file_size_gb"] * 3
     # DVARS
     dvars = pe.Node(confounds.ComputeDVARS(save_all=True, remove_zerovariance=True),
                     name="ComputeDVARS")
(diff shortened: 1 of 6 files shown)


rwblair:
LGTM

fmriprep/workflows/confounds.py
```diff
@@ -40,17 +40,27 @@ def discover_wf(settings, name="ConfoundDiscoverer"):
...
     # CompCor
     tcompcor = pe.Node(confounds.TCompCor(components_file='tcompcor.tsv'), name="tCompCor")
+    tcompcor.interface.estimated_memory_gb = settings[
+                                              "biggest_epi_file_size_gb"] * 3
     acompcor_roi = pe.Node(mask.BinarizeSegmentation(
         false_values=[FAST_DEFAULT_SEGS.index('GrayMatter') + 1, 0]),  # 0 denotes background
                            name="CalcaCompCorROI")
     acompcor = pe.Node(confounds.ACompCor(components_file='acompcor.tsv'), name="aCompCor")
+    acompcor.interface.estimated_memory_gb = settings[
+                                              "biggest_epi_file_size_gb"] * 3
```
shoshber: why `* 3`? Does it make sense to put it into a variable/constant?

chrisfilo:
This value can vary for each of those interfaces and does not always have
to be set to 3. This value is however a sensible starting point - we can
fine tune each of those Nodes separately later.

Issue #177, opened 2016-11-09
PV of segmentation projected from T1 impedes aCompCor and CSF signal extraction


shoshber:
Improve segmentation. See screenshot; voxels labeled "CSF" overlap the cortex
![screen shot 2016-11-08 at 4 41 05 pm](https://cloud.githubusercontent.com/assets/1402948/20123248/b281c8de-a5d2-11e6-8f7a-55cd7b8056e9.png)

@chrisfilo

oesteban:
We'll need a little bit of context here. What is presented in the screenshot? Where the segmentation comes from? Was it running with the ```--debug``` flag?

chrisfilo:
I think a few things might've been confused when creating this screenshot.
We should reeveluate the segmentation maps after transformation to EPI
space (for the purpose of signal extraction) when we run one of the beast
subjects.

oesteban:
Ok, I guess we can close this one?

chrisfilo:
Not until someone runs FMRIPREP on a full resolution subjects and checks if
the segmentation in the EPI space looks ok.

chrisfilo:
This is still a problem:

![image](https://cloud.githubusercontent.com/assets/238759/20856332/36d84b3e-b8c1-11e6-9067-cfa674e4a32a.png)

It seems that when we convert the segmentation from T1 to EPI we get severe partial volume issues. The voxels corresponding to CSF around the brain overlap with grey matter to a great extent. I would not recommend using aCompCorr until we fix this.

One solution would be using better segmentation - switching to FreeSurfer will allow us to grab a mask for the ventricles alone - without the CSF around the brain.

oesteban:
While FreeSurfer is not integrated, we could transform the posterior probability maps (or the PVEs, even better) and then apply a MAP criterium to obtain the segmentation. That will probably improve a lot the great amount of pixels classified as GM taking over the WM.

chrisfilo:
We can just use tCompCorr for now.

oesteban:
Ok, I'll update the title of the issue.

chrisfilo:
This issue also impacts the CSF SignalExtration - I've updated the title accordingly.

oesteban:
? - I don't see the new title :D

chrisfilo:
There is a way to fix it without using FreeSurfer.

From the original paper http://www.sciencedirect.com/science/article/pii/S1053811907003837:
[...]

PR #227, opened 2016-12-06
RTM Don't use aCompCor for now

shoshber:
Because segmentation isn't good enough for it to be useful: https://github.com/poldracklab/fmriprep/issues/177


chrisfilo:
I'm afraid that reverting this change when we decide to fix aCompCor will be complicated. If you are worried that users will accidentally use compcor why don't just comment out the lines linking it to the confounds concatenator? Adding a link to https://github.com/poldracklab/fmriprep/issues/177 above the commented lines would make the change more clear.

rwblair:
LGTM

shoshber:
why would reverting the commit be difficult? I prefer to rely on versioning to keep old code rather than commenting stuff out.

chrisfilo:
Because other changes will be merged in the meantime and when you try to
revert this change you will run into a conflict.

shoshber:
But it's the same if you comment stuff out, the commented code becomes out of date so quickly there's no point in keeping it in the file

PR #230, opened 2016-12-06
[RTM] option for explicitly limiting memory

chrisfilo:


shoshber:
Is this limiting memory for each node or for the entire workflow?

chrisfilo:
Entire workflow - more details here:
http://nipype.readthedocs.io/en/latest/users/resource_sched_profiler.html

shoshber:
Perhaps a more explicit description is warranted.

PR #234, opened 2016-12-07
[RTM] add an option to skip native space output

chrisfilo:


oesteban:
There seems to be some conflicts here, will merge right after.

Issue #235, opened 2016-12-07
ds054 outputs contain "report" and "reports" folders


chrisfilo:

rwblair:
ah looks like the sbref derivatives data sinks are all being dumped into report and everything else is going into reports. Will fix.

chrisfilo:
Fixes #239

PR #239, opened 2016-12-07
[WIP]change outpath for sbref report derivatives to reports

rwblair:


chrisfilo:
This seems to fix #235. Unless you want to add anything merge when ready.

PR #241, opened 2016-12-07
[RTM]search for reportlets from directory where subject id was found not r…

rwblair:
…oot of reports dir

### DIFF
--- a/fmriprep/viz/reports.py
+++ b/fmriprep/viz/reports.py
@@ -142,8 +142,9 @@ def run_reports(out_dir):
         dir = root.split('/')[-1]
         try:
             subject = re.search('^(?P<subject_id>sub-[a-zA-Z0-9]+)$', dir).group()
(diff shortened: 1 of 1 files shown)


fmriprep/viz/reports.py
```diff
@@ -142,8 +142,9 @@ def run_reports(out_dir):
         dir = root.split('/')[-1]
         try:
             subject = re.search('^(?P<subject_id>sub-[a-zA-Z0-9]+)$', dir).group()
+            print(subject)
```
chrisfilo: left over debug print

chrisfilo:
I still cannot see the final aggregated HTML file in the CircleCi artifacts. What am I missing?

rwblair:
this commit also in #242 will merge from there once reports in circle are showing up.

chrisfilo:
thanks!

PR #242, opened 2016-12-07
[RTM]Fix sbref dupe

rwblair:
There were two images with different titles but the same image in reports for ds054.

Need to confirm these changes don't mess with ds005 in any way.


chrisfilo:
There are no sbrefs in ds005. Did you mean ds054?

rwblair:
The change in the config.json potentially effects how/when mean epi registration is displayed, which is used by ds005.

PR #247, opened 2016-12-07
[WIP] Fieldmaps

oesteban:


chrisfilo:
From @satra: TOPUP scripts used at MIT - https://github.com/gablab/openfmri/pull/15

oesteban:
Quick guide to review this PR. List of changes:

- Improved registration (```fmriprep/data/fmap-any_registration.json``` and _testing version) from the fieldmap reference (magnitude image or corrected SE) to the target image (typically EPI or SBRef).
- Added a TOPUP config file for SBRef-to-SBRef registration (```fmriprep/data/sbref2sbref.cnf```)
- Added fast calculation of tensor BSpline coefficients in Cython (includes the pyx file, and some changes in info.py and setup.py to include numpy and cython as pre-requisites before calling setup.py).
- Added a module of interfaces to handle images. Particularly, the ```IntraModalMerge``` interface, which calculates the average of a bunch of input images of the same modality/contrast.
- Added a module of interfaces to assist topup.
- Refactored fieldmap processing:
  - Added a base module that creates all the workflows that work with the input images. This module plays the role of the former fieldmap_decider, but now it does not make any decision. I aggregates all the possible ways to estimate the fieldmap given the input data.
  - Clearly defined the boundaries between fieldmap estimators and unwarping:
     - Created three specific modules for the three types of fieldmaps: phdiff (two GRE echoes to estimate the phase difference map), fmap (one fieldmap in Hz), pepolar (two EPI images with different encoding blips).
     - Now the unwarp module will take those fieldmaps and apply the correction.

TODO list:
  - phdiff.py: this workflow needs to be revised. It basically should call PRELUDE to do the phase unwrapping and then we just convert it to Hz. Once that's done, the workflow calls the fmap_workflow to finish up the processing.
  - unwarp.py: for now, this is only taking the last fieldmap of the input list. I'll figure out a nice way to average all the input estimations and perform the best correction possible.
  - Memory: there are some (mostly MCFLIRT) interfaces here and there that need setting the estimated memory.

chrisfilo:
Is the goal of this PR to run smoothly on the Beast? Do you want me to rerun the analysis on all subjects and post my findings here?

oesteban:
If you don't mind doing so, that'd be great!

chrisfilo:
There is some good stuff on fieldmaps in this [presentation](https://fsl.fmrib.ox.ac.uk/fslcourse/lectures/reg.pdf) - especially for diagnosing if everything went well
image

oesteban:
First green flags in a long time!. This is not ready yet, but this is a good step forward. A couple of handy links:

- The new documentation: https://638-53175327-gh.circle-artifacts.com/0/home/ubuntu/_build_html/workflows.html (the graphs help understand what is done here).
- The reports: https://638-53175327-gh.circle-artifacts.com/0/home/ubuntu/ds054/out/fmriprep/sub-100185.html

PR #256, opened 2016-12-09
[RTM] move reorient to beginning of sbref processing

### LINKED ISSUE #205: SBRef reorientation should be moved to sbref_preprocess()
Currently sbref reorientation is performed just before sbref to T1 coregistration. However, for the EPI to SBref coregistration a version before applying reorientation is used. This can cause problem in the final images after resampling to MNI.

chrisfilo:
Closes #205 .


chrisfilo:
This is not a crucial change. My reasoning is that a) reorient can mess up with detecting the right phase encoding direction b) we do not apply it to the EPI image which leads to minor inconsistency.

chrisfilo:
ACQ @oesteban

PR #255, opened 2016-12-09
[RTM] enable bias field correction of the mean EPI in ds005

chrisfilo:

PR #257, opened 2016-12-09
[RTM] improved EPI masking and EPI T1 coregistration

### LINKED ISSUE #252: BET skulls tripping cuts into the cerebellum
image

We should investigate BET parameters or other skull stripping options.

chrisfilo:
- Fixes #252 (EPI skullstripping problems)
- Switches FLIRTs to using masks as weights instead of taking masked files as inputs


chrisfilo:
Includes changes from #255 and #256.

chrisfilo:
FYI - this works well on all Beast subjects

PR #272, opened 2016-12-13
[RTM] Deleting unused file

chrisfilo:
Bleeding edge requirements are installed in the Dockerfile.


shoshber:
Can we keep this requirements.txt and have the Dockerfile read from it? I like to install these requirements from `pip install -r requirements.txt` in my native environment so I can run unit tests without needing Docker.

chrisfilo:
Sure - I'll merge this and you can submit a new PR with modifications to the Dockerfile.

PR #285, opened 2016-12-16
[RTM] fix for derivatives sink when multiple t1s are present (ds000051)

chrisfilo:
I had to pick new source name for derivatives obtained from a combination of T1s. I also had to turn off checking if the `source_file` file exists for this to work - rather controversial, but maybe we can address this when refactoring derivatives datasink.


oesteban:
This is also happening for ds115

PR #293, opened 2016-12-26
[RTM] niworkflows update

### LINKED ISSUE #266: T1 -> MNI normalization failure
image

That's subject 100003 from ds054

chrisfilo:
Fixes #266


chrisfilo:
I've tested this on all subjects listed in https://github.com/poldracklab/fmriprep/issues/266#issuecomment-269181677 - most of them show very good results. The only exception is ds109 sub 48 which has very unusual anatomy (almost non-existent ventricles and one-third of the corpus callosum missing).

PR #294, opened 2016-12-26
[RTM] Reverting to using explicitly masked images

### LINKED ISSUE #292: EPI -> T1 coregistration failure
image

image

chrisfilo:
Fixes #292. Turns out that using brain mask as weight images was a fundamentally stupid idea since it removed the brain edge feature from the optimization.


chrisfilo:
I can confirm that this solves the problem for ds114. I also tested example subjects from ds005 ds051 ds054 and the beast to make sure it does not introduce any problems.

oesteban:
all set?

chrisfilo:
Yup.

Issue #66, opened 2016-06-07
Show errors for missing images in reports.


rwblair:
If an image fails to generate report should put in additional information about failures from image generation workflows.

oesteban:
I think this is currently well covered with the new reports.

oesteban:
Yep, the problem is that the out_report output trait is never populated: the node fails and, even though a reportlet is generated, the nipype engine will not consider it.

chrisfilo:
The reports should include the content of all generated crashfiles. Here's the code for reading crashfiles: https://github.com/nipy/nipype/blob/master/nipype/scripts/crash_files.py (could be easily adapted with mocking `print`).

We briefly discussed adding optional plain text crashfiles to nipype (see https://github.com/nipy/nipype/issues/1184), but I doubt this can happen before the next nipype release (rc1 was just released).

rwblair:
In niworkflows for the definition of ReportCapableInterface there is check to see if the interface ran successfully and if not to call _generate_error_report: https://github.com/poldracklab/niworkflows/blob/master/niworkflows/common/report.py#L59

I'm looking into writing the crash information out to the reportlet here. That being said in the reports that we have on afterhours I'm not seeing any instances of the html from _generate_error_report.

chrisfilo:
I would suggest a different approach. After running the workflow that exits with errors 
- crawl the crashfile directory and identify all crashfiles belonging to each subject analyzed in this run (you might need additional daytime restrictions to avoid capturing crashfiles from historical runs)
- append error information to corresponding subject specific HTML reports report

This way we will be able to capture all errors (not only those coming from Report Capable interfaces). It will also avoid having to deal with the problematic case of a failed node that populates error reportlet that should be send to a datasink (what @oesteban mentioned in https://github.com/poldracklab/fmriprep/issues/66#issuecomment-267456545).

chrisfilo:
Closed by #298

PR #298, opened 2017-01-06
[RTM]Fix for Issue #66

rwblair:
addresses some concerns in #66 Present commits pull in all present crashfiles. May limit this to all crash files generated in the last 24hours. Or we may consider cleaning up crash files on each new run.

### DIFF
--- a/fmriprep/run_workflow.py
+++ b/fmriprep/run_workflow.py
@@ -13,6 +13,7 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 import os
 import os.path as op
 import glob
+import sys
 from argparse import ArgumentParser
 from argparse import RawTextHelpFormatter
 from multiprocessing import cpu_count
@@ -84,6 +85,8 @@ def create_workflow(opts):
     from fmriprep.utils import make_folder
     from fmriprep.viz.reports import run_reports
     from fmriprep.workflows.base import base_workflow_enumerator
+
+    errno = 0
 
     settings = {
         'bids_root': op.abspath(opts.bids_dir),
@@ -160,13 +163,15 @@ def create_workflow(opts):
     try:
         preproc_wf.run(**plugin_settings)
     except RuntimeError:
-        pass
+        errno = 1
 
     if opts.write_graph:
         preproc_wf.write_graph(graph2use="colored", format='svg',
                                simple_form=True)
 
     run_reports(settings['output_dir'])
+
+    sys.exit(errno)
 
 if __name__ == '__main__':
     main()
--- a/fmriprep/run_workflow.py
+++ b/fmriprep/run_workflow.py
@@ -121,8 +121,7 @@ def create_workflow(opts):
     # Set nipype config
     ncfg.update_config({
         'logging': {'log_directory': log_dir, 'log_to_file': True},
-        'execution': {'crashdump_dir': log_dir, 
-                      'remove_unnecessary_outputs': False}
+        'execution': {'remove_unnecessary_outputs': False}
     })
 
     # nipype plugin configuration
@@ -158,7 +157,10 @@ def create_workflow(opts):
     preproc_wf = base_workflow_enumerator(subject_list, task_id=opts.task_id,
                                           settings=settings)
     preproc_wf.base_dir = settings['work_dir']
-    preproc_wf.run(**plugin_settings)
+    try:
+        preproc_wf.run(**plugin_settings)
+    except RuntimeError:
+        pass
 
     if opts.write_graph:
         preproc_wf.write_graph(graph2use="colored", format='svg',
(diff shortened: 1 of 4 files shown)


fmriprep/viz/reports.py
```diff
@@ -120,24 +118,59 @@ def index(self):
...
+        if os.path.isdir(error_dir):
+            self.index_error_dir(error_dir)
+
+    def index_error_dir(self, error_dir):
+        for root, directories, filenames in os.walk(error_dir):
+            for f in filenames:
+                try:
+                    crash_data = loadcrash(os.path.join(root, f))
+                except ValueError:
+                    continue
```
chrisfilo: What situation does this account for?

fmriprep/viz/reports.py
```diff
@@ -120,24 +118,59 @@ def index(self):
...
+                error = {}
+                node = None
+                node_str = ''
+                node_dir_str = ''
+                inputs_str = ''
```
chrisfilo: Are those `_str` variables used anywhere?

fmriprep/viz/report.tpl
```diff
@@ -82,10 +83,43 @@ body {
...
+    <h1 class="sub-report-title">Errors</h1>
+    <ul>
+    {% for error in errors %}
+        <li>
+        <div class="nipype_error">
+            Node Name: <a href="#" onclick="toggle('{{error.file|replace('.', '')}}_details_id');">{{ error.node }}</a><br>
+            <div id="{{error.file|replace('.', '')}}_details_id" style="display:none">
+            File: {{ error.file }}<br>
+            Working Directory: {{ error.node_dir }}<br>
+            Inputs: {{ error.inputs }}<br>
```
chrisfilo: Isn't `inputs` a Bunch object? Or a traited class? If so you should be able to iterate over it fields and format inputs one by one. This might be helpful to iterate over traited spec: https://github.com/nipy/nipype/blob/bbc93ae5f4efee89a2d61541fbce6d8d7ab779a5/nipype/interfaces/base.py#L874

fmriprep/run_workflow.py
```diff
@@ -121,8 +124,7 @@ def create_workflow(opts):
     # Set nipype config
     ncfg.update_config({
         'logging': {'log_directory': log_dir, 'log_to_file': True},
-        'execution': {'crashdump_dir': log_dir, 
-                      'remove_unnecessary_outputs': False}
+        'execution': {'remove_unnecessary_outputs': False}
```
chrisfilo: Left over from debugging?

fmriprep/workflows/base.py
```diff
@@ -29,6 +32,11 @@ def base_workflow_enumerator(subject_list, task_id, settings):
         generated_workflow = base_workflow_generator(subject, task_id=task_id,
                                                      settings=settings)
         if generated_workflow:
+            generated_workflow.config['execution']['crashdump_dir'] =(
+                os.path.join(settings['output_dir'], 'log', subject)
```
chrisfilo: I'm afraid that this way we will be catching crashfiles from previous executions. Why not use a tempdir?

rwblair: I'll have it make a directory with subject + timestamp, then in run reports have it look for the newest directory.

fmriprep/viz/reports.py
rwblair: Left over from a previous implementation, will remove.

fmriprep/viz/reports.py
rwblair: If for any reason there is a file in the crash directory that is not a pklz this will skip it from being read in.

fmriprep/run_workflow.py
rwblair: Had removed it since crashdumpdir is being set in base.py for every node. I'll re-add it as a fallback in case the new way of setting the crash directory fails for any reason we can consistently find crash files and know that the directory wasn't set to the subject directory.

chrisfilo: Sorry I misread the changelog. I thought you added the `'execution': {'remove_unnecessary_outputs': False}`. I'll open a new PR about this.

fmriprep/viz/reports.py
chrisfilo: I would prefer to avoid silencing all ValueError exceptions and limit the files that are attempted to be read to those that begin with "crash" and end with ".pklz"

fmriprep/workflows/base.py
chrisfilo: I see. Seems ok to me (although I still feel that a unique tempdir style folder would be more robust). I was trying to think of cases when the timestamp solution would break - when a user sends multiple jobs using the same output directory and the job happen to start at the same time. This is, however,  not a very likely scenario and many more things related to report generation would break in such case.

fmriprep/run_workflow.py
rwblair: Looks like I added it here d04fdca5627f7f7bc86b4520fe85920afe1d95ab

PR #299, opened 2017-01-08
WIP check brain volume of aCompCor mask

shoshber:


effigies:
Is this something we're planning on reviving/finishing at some point?

chrisfilo:
Yes - there is still visible variance is relative aCompCor mask sizes and
their accuracy. Since we now have FreeSurfer - I would recommend switching
to its parcellation.

On Wed, May 10, 2017 at 12:02 PM, Chris Markiewicz <notifications@github.com

effigies:
If aCompCor masks seem to have problems again, we can come back to this.

PR #312, opened 2017-01-10
Add notes for starting a development environment

### LINKED ISSUE #309: Write documentation for developers
Focus on developing with docker scenario.

effigies:
Added a sub-heading to Docker, since most of the prep is just getting the Docker image to run at all. Is this the sort of thing you were looking for, @chrisfilo?

Closes #309.

### DIFF
--- a/docs/contributors.rst
+++ b/docs/contributors.rst
@@ -1,0 +1,118 @@
+.. include:: links.rst
+
+------------------------
+Contributing to FMRIPREP
+------------------------
+
+This document explains how to prepare a new development environment and
+update an existing environment, as necessary.
+
+Development in Docker is encouraged, for the sake of consistency and
+portability.
+By default, work should be built off of `poldracklab/fmriprep:latest
+<https://hub.docker.com/r/poldracklab/fmriprep/>`_ (see the
+installation_ guide for the basic procedure for running).
+
+It will be assumed the developer has a working repository in
+``$HOME/projects/fmriprep``, and examples are also given for
+`niworkflows <https://github.com/poldracklab/niworkflows>`_ and
+`NIPYPE <https://github.com/nipy/nipype>`_.
+
+Patching working repositories
+=============================
+In order to test new code without rebuilding the Docker image, it is
+possible to mount working repositories as source directories within the
+container.
+In the docker container, the following Python sources are kept in
+``/root/src``: ::
+
+    /root/src
+    ├── fmriprep/
+    ├── nipype/
+    └── niworkflows/
+
+To patch in working repositories, for instance contained in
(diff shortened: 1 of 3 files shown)


chrisfilo:
Yup this is great. Some comments:

1. Explain how to build the container.
2. Explain how to add python and other dependencies.
3. When you mount external sources I was under the impression that it only works if a egg-info file is present see: https://github.com/poldracklab/fmriprep/wiki/Mounting-niworkflows-for-development

effigies:
Updated. Think maybe with the amount of content, should move to a new file. What do you think?

Comments on the content itself are welcome, of course.

rwblair:
A dedicated contributors file for this documentation would be good. I think its common in many larger projects.

docs/installation.rst
```diff
@@ -23,6 +23,111 @@ For example: ::
...
+    root@03e5df018c5e:~# cd ~/src/fmriprep/
+    root@03e5df018c5e:~/src/fmriprep# pip install -e .
+    root@03e5df018c5e:~# cd ~/src/niworkflows/
+    root@03e5df018c5e:~/src/niworkflows# pip install -e .
+    root@03e5df018c5e:~# cd ~/src/nipype/
+    root@03e5df018c5e:~/src/nipype# pip install -e .
+
+Adding dependencies
+```````````````````
+New dependencies to insert into the Docker image will either be Python
```
rwblair: Maybe rewrite:
"New dependencies to be inserted into"

effigies:
Ready for review.

docs/contributors.rst
```diff
@@ -0,0 +1,118 @@
...
+Contributing to FMRIPREP
+------------------------
+
+This document explains how to prepare a new development environment and
+update an existing environment, as necessary.
+
+Development in Docker is encouraged, for the sake of consistency and
+portability.
+By default, work should be built off of `poldracklab/fmriprep:latest
+<https://hub.docker.com/r/poldracklab/fmriprep/>`_ (see the
```
oesteban: This looks like a link (with that underscore) but seems to be broken.

docs/contributors.rst
```diff
@@ -0,0 +1,118 @@
...
+installation_ guide for the basic procedure for running).
```
oesteban: This looks like a link (with that underscore) but seems to be broken.

effigies: Building it locally (make -C docs html), it points to installation.html. Is there a better way to do this?

oesteban: Sorry, now I realize that I was looking at the github advanced diff... should be fine.

chrisfilo:
Great work!

PR #317, opened 2017-01-11
[RTM] fmriprep-docker wrapper script

effigies:
*Edited*

Current behavior (image not pulled):

```
$ ./wrapper/fmriprep.py -h
usage: fmriprep.py [-h] [-v] [-i IMG] [-f PATH] [-n PATH] [-p PATH] [--shell]
                   [--scratch PATH]
                   [bids_dir] [output_dir] [{participant}]

fMRI Preprocessing workflow

positional arguments:
  bids_dir
  output_dir
  {participant}

optional arguments:
  -h, --help            show this help message and exit (default: False)
  -v, --version         show program's version number and exit (default:
                        False)
  -i IMG, --image IMG   image name (default: poldracklab/fmriprep:latest)

Developer options:
  Tools for testing and debugging FMRIPREP

  -f PATH, --patch-fmriprep PATH
                        working fmriprep repository (default: None)
  -n PATH, --patch-niworkflows PATH
                        working niworkflows repository (default: None)
  -p PATH, --patch-nipype PATH
                        working nipype repository (default: None)
  --shell               open shell in image instead of running FMRIPREP
                        (default: False)
  --scratch PATH        use scratch directory (default: None)

Image 'poldracklab/fmriprep:latest' is missing
Would you like to download? [Y/n]
```

If image is downloaded:

```
Status: Downloaded newer image for poldracklab/fmriprep:latest
usage: fmriprep.py [-h] [-v]
                   [--participant_label PARTICIPANT_LABEL [PARTICIPANT_LABEL ...]]
                   [-s SESSION_ID] [-r RUN_ID] [-t TASK_ID] [--debug]
                   [--nthreads NTHREADS] [--mem_mb MEM_MB]
                   [--use-plugin USE_PLUGIN]
                   [--ignore {fieldmaps,slicetiming} [{fieldmaps,slicetiming} ...]]
                   [--skip-native] [--ants-nthreads ANTS_NTHREADS]
                   [--skull-strip-ants] [--no-skull-strip-ants]
                   [--no-freesurfer] [--no-submm-recon] [-w WORK_DIR]
                   [--reports-only] [--write-graph] [-i IMG] [-f PATH]
                   [-n PATH] [-p PATH] [--shell] [--scratch PATH]
                   bids_dir output_dir {participant}

FMRIPREP: fMRI PREProcessing workflows

positional arguments:
  bids_dir              the root folder of a BIDS valid dataset (sub-XXXXX folders should be found at the top level in this folder).
  output_dir            the output path for the outcomes of preprocessing and visual reports
  {participant}         processing stage to be run, only "participant" in the case of FMRPREP (see BIDS-Apps specification).

optional arguments:
  -h, --help            show this help message and exit (default: False)
  -v, --version         show program's version number and exit (default:
                        False)
  -i IMG, --image IMG   image name (default: poldracklab/fmriprep:latest)

Options for filtering BIDS queries:
  --participant_label PARTICIPANT_LABEL [PARTICIPANT_LABEL ...]
                        one or more participant identifiers (the sub- prefix can be removed)
  -s SESSION_ID, --session-id SESSION_ID
                        select a specific session to be processed
  -r RUN_ID, --run-id RUN_ID
                        select a specific run to be processed
  -t TASK_ID, --task-id TASK_ID
                        select a specific task to be processed

Options to handle performance:
  --debug               run debug version of workflow
  --nthreads NTHREADS   number of threads
  --mem_mb MEM_MB       upper bound memory limit for FMRIPREP processes
  --use-plugin USE_PLUGIN
                        nipype plugin configuration file

Workflow configuration:
  --ignore {fieldmaps,slicetiming} [{fieldmaps,slicetiming} ...]
                        ignore selected aspects of the input dataset to disable relatedparts of the workflow
  --skip-native         don't output timeseries in native space

Specific options for ANTs registrations:
  --ants-nthreads ANTS_NTHREADS
                        number of threads that will be set in ANTs processes
  --skull-strip-ants    use ANTs-based skull-stripping (default, slow))
  --no-skull-strip-ants
                        don't use ANTs-based skull-stripping (use  AFNI instead, fast)

Specific options for FreeSurfer preprocessing:
  --no-freesurfer       disable FreeSurfer preprocessing
  --no-submm-recon      disable sub-millimeter (hires) reconstruction

Other options:
  -w WORK_DIR, --work-dir WORK_DIR
                        path where intermediate results should be stored
  --reports-only        only generate reports, don't run workflows. This will only rerun report aggregation, not reportlet generation for specific nodes.
  --write-graph         Write workflow graph.

Developer options:
  Tools for testing and debugging FMRIPREP

  -f PATH, --patch-fmriprep PATH
                        working fmriprep repository (default: None)
  -n PATH, --patch-niworkflows PATH
                        working niworkflows repository (default: None)
  -p PATH, --patch-nipype PATH
                        working nipype repository (default: None)
  --shell               open shell in image instead of running FMRIPREP
                        (default: False)
  --scratch PATH        use scratch directory (default: None)
```

Example run:

```
$ fmriprep.py /data/bids/ds000114 /data/out/ds000114/derivatives participant
RUNNING: docker run --rm -it -v /data/bids/ds000114:/data:ro -v /data/out/ds000114/derivatives:/out poldracklab/fmriprep:latest /data /out participant
[...]
```


fmriprep.py
```diff
@@ -0,0 +1,55 @@
...
+                        default=os.path.join(data_base, 'ds005'))
+    parser.add_argument('--out_path', type=str,
+                        default=os.path.join(working, 'out'))
+    parser.add_argument('--shell', action='store_true')
+    parser.add_argument('-i', '--image', type=str,
+                        default='poldracklab/fmriprep:latest')
+    parser.add_argument('ARGS', type=str, nargs='*')
+    opts = parser.parse_args(argv)
+
+    command = ['docker', 'run', '--rm']
```
chrisfilo: Don't you need '-it' here as well?

effigies: It's inserted if the `--shell` option is given.

chrisfilo: Does `Ctrl-C` work without those flags?

effigies: Oh. That's why that didn't work. I used `docker kill`.

fmriprep.py
```diff
@@ -0,0 +1,56 @@
...
+working = '/data/fmriprep/working'
+
+
+def main(cmd, *argv):
+    parser = argparse.ArgumentParser()
+    parser.add_argument('-f', '--patch_fmriprep', action='store_true')
+    parser.add_argument('-w', '--patch_niworkflows', action='store_true')
+    parser.add_argument('-n', '--patch_nipype', action='store_true')
+    parser.add_argument('--fmriprep_path', type=str,
+                        default=os.path.join(code_base, 'fmriprep'))
```
chrisfilo: I would avoid defaults here since they are specific to a particular machine (in this case your are setting those folders to `/data/chris/...`

fmriprep.py
```diff
@@ -0,0 +1,56 @@
...
+                                        '/root/src/nipype', 'ro'))])
+
+    command.extend(['-v', ':'.join((opts.out_path, '/out')),
+                    '-v', ':'.join((opts.data_path, '/data', 'ro'))])
+
+    if opts.shell:
+        command.append('--entrypoint=bash')
+
+    command.append(opts.image)
+    ret = subprocess.run(command + opts.ARGS)
```
chrisfilo: It would be nice to print the command before running it. It will help debugging.

effigies:
Current help:

```Shell
usage: fmriprep.py [-h] [-i IMAGE] [-f PATCH_FMRIPREP] [-n PATCH_NIWORKFLOWS]
                   [-p PATCH_NIPYPE] [--shell]
                   [bids_dir] [output_dir] [{participant}]

positional arguments:
  bids_dir
  output_dir
  {participant}

optional arguments:
  -h, --help            show this help message and exit
  -i IMAGE, --image IMAGE
  -f PATCH_FMRIPREP, --patch-fmriprep PATCH_FMRIPREP
  -n PATCH_NIWORKFLOWS, --patch-niworkflows PATCH_NIWORKFLOWS
  -p PATCH_NIPYPE, --patch-nipype PATCH_NIPYPE
  --shell
```

Doc strings need to be added. Would probably also be good to grab the help from the within-image executable -- e.g., running `docker run --rm $IMAGE -h` -- and parsing its help file enough to show all the options.

effigies:
In response to some comments by @rwblair, I made it tolerant of relative paths, and have it ask if you want to pull down the image if you're just running `-h/-v` without an image. (Better description of errors, in the process.)

Also added a merged help that basically adds the "developer options" to the flag list and their description to the bottom of the regular help.

Some questions:

1. Currently doesn't try to pull 'poldracklab/fmriprep:latest' if the image exists locally. This will have stable behavior, so things that run won't break from running again, but users may want `:latest` to mean `:latest`.
1. If we do want `:latest` to be pulled, is there a quick way to check without running `pull`? My googling mostly finds people complaining about it not pulling by default. (If someone knows how to get the container ID from DockerHub, I could do the rest.)

effigies:
Any more comments on this?

And how best to publish? I remember somebody suggesting a gist, but we could just as easily keep it in the tree, and have the docs point to https://github.com/poldracklab/fmriprep/blob/master/fmriprep.py, and then updates are easily incorporated. Or perhaps some other location. Should the name still be `fmriprep.py`?

fmriprep.py
```diff
@@ -0,0 +1,235 @@
...
+
+    if opts.patch_fmriprep is not None:
+        command.extend(['-v', ':'.join((opts.patch_fmriprep,
+                                        '/root/src/fmriprep'))])
+    if opts.patch_niworkflows:
+        command.extend(['-v', ':'.join((opts.patch_niworfklows,
+                                        '/src/niworkflows'))])
+    if opts.patch_nipype:
+        command.extend(['-v', ':'.join((opts.patch_nipype,
+                                        '/src/nipype'))])
```
rwblair: Current docker image appears to be using the pipy package for nipype not. Directory for the source code is:
`/usr/local/miniconda/lib/python3.5/site-packages/nipype/`

Caveat, the files in site packages are from the module level not the root of the project, so if I was patching into it I would use `/home/rwblair/projects/nipype/nipype` not `/home/rwblair/projects/nipype`

effigies: Hmm. This does seem to be a problem, because when pinning with `requirements.txt`, the commit is checked out into `/src/nipype`. So this behavior is going to vary based on whether you're developing a branch that requires a pin or not.

rwblair: Could mount it into both places by default. Not an elegant solution but it guarantees we get the code where it needs to be without having to inspect the container/requirements.txt file.

effigies:
Updated the wrapper I've been working with. This patches in source directories (e.g. fmriprep/fmriprep), not the base repository, now, due to #381.

@chrisfilo was saying some people have been having issues with Docker commands (particularly the `-w` on both sides of the image), so bumping this. However we want to distribute it is fine with me.

effigies:
Now:

* Prints `--help` when there are no arguments
* Runs `free -m` in the docker image and warns (does not fail) on <8GB
* Correctly sets paths in Windows
* Uses `-w` to set the scratch path (was previously using `--scratch` to avoid overriding internal flags)

Learned a bit about PyPI to the extent that a version number once claimed can't be replaced. So 0.3.2 is missing because I assumed I could delete it and reuse. The current uploaded version is `0.3.2.1` as far as pip is concerned, but appears on the website and runs as `0.3.2`. Once we settle on a final version (if this is not it), I will use `0.3.2.0` which I think will permit `fmriprep-docker==0.3.2` to be pulled in pip.

effigies:
Python 2 support and docs updated. Current version posted as `0.3.2.2`. (`pip install --upgrade fmriprep-docker` gets it.)

PR #324, opened 2017-01-12
have all svgs in elem-image div take on width 100%

rwblair:


rwblair:
If the latest commit fixes issues in chrome then https://github.com/poldracklab/niworkflows/pull/116 will need to be merged, and the niworkflows version updated a final time before this can be merged.

PR #340, opened 2017-01-18
[RTM] Pybids extension regex fix

rwblair:


rwblair:
This does not need the change in pybids to work, the first commit was undone and another method was found.

PR #343, opened 2017-01-21
Use a meaningful dev version tag

effigies:
`0.2.1-dev` reflects that the latest release was `0.2.0`, and the next bugfix release would be `0.2.1`.

The following ordering holds, so pip version constraints are usable: `0.2.0 < 0.2.1-dev < 0.2.1 < 0.2.2-dev < 0.3`

Also, should just note that the [version number in 0.2.0 tag/release](https://github.com/poldracklab/fmriprep/blob/0.2.0/fmriprep/info.py#L9) is `dev`.

### DIFF
--- a/fmriprep/info.py
+++ b/fmriprep/info.py
@@ -6,7 +6,7 @@ Base module variables
 """
 from __future__ import unicode_literals
 
-__version__ = '99.99.99'
+__version__ = '0.2.1-dev'
 __author__ = 'The CRN developers'
 __copyright__ = 'Copyright 2016, Center for Reproducible Neuroscience, Stanford University'
 __credits__ = ['Craig Moodie', 'Ross Blair', 'Oscar Esteban', 'Chris Gorgolewski', 'Shoshana Berleant',


chrisfilo:
Ok let me clarify this a little bit. When I set up the release automation I wanted it to require as little intervention as possible. So when you make a new release on GitHub or push a new tag things are deployed to pypi and Docker Hub for you. This deployment process modifies the version number to match the tag. So even though version of the 0.2.0 release is set to `dev` on github it is set to `0.2.0` on pypi (the code is for this is [here](https://github.com/poldracklab/fmriprep/blob/master/circle.yml#L59)).

The problem with using `-dev` as a version is that you have to remember to update this each time you do a release. What do we gain in return?

effigies:
Oh, I see. I missed that machinery going in. 

The reason I suggested this is because, when I pulled the `poldracklab/fmriprep:latest`, the version said `dev`, which seemed uninformative. Same thing with `poldracklab/fmriprep:0.2.0`:

```Shell
$ docker run --rm -it poldracklab/fmriprep:0.2.0 -v
fmriprep vdev
```

This will make bug reports hard.

chrisfilo:
My bad I forgot to update the version for the the containers pushed to docker hub. I've opened a PR to fix this #344

effigies:
Sounds good. Closing in favor of #344.

PR #345, opened 2017-01-21
Update FreeSurfer bundle URL

effigies:
Apparently 6.0.0 was released properly yesterday, and 6.0.0rc was removed from the server.


effigies:
Should fix #344 build.

PR #346, opened 2017-01-23
[RTM] Use Bias field correction from ants if possible.

rwblair:


oesteban:
Is this ready @rwblair ?

rwblair:
Yes.

Issue #101, opened 2016-08-12
Add surface reconstruction


chrisfilo:
- Use FreeSurfer 6 beta
- Run FS by default but allow turning it off on the command line
- Make sure to use all available cores for FreeSurfer
- Use `-highres` if the input T1 resolution is <1mm
- Use WM and ventricle parcels produced by FS in aCompCor
- Feed multiple T1s to recon-all if available (but only from one session and only if they have the same dimensions/resolution - for longitudinal datasets each session should be analyzed separately)
- Feed T2s to recon-all if available (but only if they have resolution along the coarsest dimension < 1.2mm)
- Replace internal Freesurfer skullstripping with something better (for example ANTs) - see https://github.com/nipy/nipype/blob/master/nipype/workflows/smri/freesurfer/recon.py#L12
- If T2s are available use sqrt(T1*T2) bias field correction method (see HCP pipelines for details)
- Use FreeSurfer implementation of BBR for EPI to T1 coregistration
- Output: 
  - SUBJECTS_DIR (with all subjects)
  - fMRI timeseries in fsaverage space

Implement and use reportlets outlined in https://github.com/poldracklab/niworkflows/issues/102 and https://github.com/poldracklab/niworkflows/issues/103

oesteban:
Duplicate of #63. But here we have a clear description. I close the old one.

effigies:
Focus group findings
----------
- Verifying white/pial surface alignment with structural was :+1:
  - As a summary stat, noting spikes in cortical thickness (distance from white to pial that is different from surrounding vertices) would be a good addition. (Greg F)
- Interest shown in gifti/cifti outputs (Greg F)
- Mixed response on FreeSurfer 5.3.0 vs 6.0.0
  - Positive response to v6
  - For v5.3, concern was established labs would be more interested in using the same tools, although if they're doing longitudinal, they probably have existing pipelines and aren't going to go back to use FMRIPREP.
  - So our focus should mainly be people setting up new pipelines, so v6.0 is the smarter move

effigies:
Will note here that there weren't any particularly strong opinions about surface sampling techniques (vertex, voxel or pixel based) among the focus group participants I talked to. Not sure if this should be its own issue, given that there's no recommendation to be made.

chrisfilo:
Let's stick with freesurfer defaults and revisit if this comes up again.

effigies:
[...]

PR #347, opened 2017-01-26
[RTM] Add recon-all preprocessing

effigies:
The goal of this PR is to get recon-all capabilities into FMRIPREP. Using its outputs in other steps in the pipeline will be addressed in other PRs.

Goals from #101:
- [x] Use FreeSurfer 6
  - [x] #338/#345
  - [x] Depends on resolution of nipy/nipype#1782
- [x] Command-line flag
- [x] Use `-hires` for T1 voxel size < 1mm
- [x] Accept multiple T1s
  - [x] Basic: Use inputs from BIDS
  - [x] Check dimension/resolution compatibility first
- [ ] Accept T2s
  - [ ] Require coarsest dimension to have resolution < 1.2mm
- [x] Use ANTS skull-stripping
  - Currently `mri_convert`ing ANTS output to `mri/brainmask(.auto).mgz`
- [ ] `sqrt(T1*T2)` bias field correction

#101 issues for a separate PR (for easy reference later):
- Use FreeSurfer WM/Ventricle parcellations in aCompCor
- Use `bbregister` to register functionals to T1


fmriprep/workflows/anatomical.py
```diff
@@ -75,10 +79,46 @@ def t1w_preprocessing(name='t1w_preprocessing', settings=None):
...
+        niu.Function(
+            function=inject_skullstripped,
+            input_names=['subjects_dir', 'subject_id', 'skullstripped'],
+            output_names=['subjects_dir', 'subject_id']),
+        name='InjectSkullstrip')
+
+    reconall = pe.Node(
+        freesurfer.ReconAll(
+            args='-parallel -openmp {:d}'.format(cpu_count())),
+        name='Reconstruction2')
```
effigies: Having an issue here:

    traits.trait_errors.TraitError: The trait 'subjects_dir' of a ReconAllInputSpec instance is an existing
    directory name, but the path '/scratch/workflow_enumerator/01/t1w_preprocessing/Reconstruction' does not exist.

But... it's still running `Reconstruction2`, and I'm getting outputs. Is this a nipype bug?

fmriprep/workflows/anatomical.py
```diff
@@ -75,10 +79,46 @@ def t1w_preprocessing(name='t1w_preprocessing', settings=None):
...
-    # should not be necesssary byt does not hurt - make sure the multiproc
+    # should not be necesssary but does not hurt - make sure the multiproc
     # scheduler knows the resource limits
     t1_2_mni.interface.num_threads = settings['ants_nthreads']
 
+    # 6. FreeSurfer reconstruction
+    autorecon1 = pe.Node(
+        freesurfer.ReconAll(
+            directive='autorecon1',
+            args='-noskullstrip -parallel -openmp {:d}'.format(cpu_count())),
```
chrisfilo: Please use the value from --nthreads

chrisfilo: Also set the number of expected threads for the scheduler (See https://github.com/poldracklab/fmriprep/blob/master/fmriprep/workflows/anatomical.py#L253)

fmriprep/workflows/anatomical.py
effigies: Nope, actually it's a segfault.

effigies: The erroring bit:

```
#-----------------------------------------
#@# Parcellation Stats 2 lh Thu Jan 26 17:41:39 UTC 2017
/scratch/workflow_enumerator/01/t1w_preprocessing/Reconstruction/recon_all/scripts

 mris_anatomical_stats -th3 -mgz -cortex ../label/lh.cortex.label -f ../stats/lh.aparc.a2009s.stats -b -a ../label/lh.aparc.a2009s.annot -c ../label/aparc.annot.a2009s.ctab recon_all lh white

#-----------------------------------------
#@# Parcellation Stats 2 rh Thu Jan 26 17:41:39 UTC 2017
/scratch/workflow_enumerator/01/t1w_preprocessing/Reconstruction/recon_all/scripts

 mris_anatomical_stats -th3 -mgz -cortex ../label/rh.cortex.label -f ../stats/rh.aparc.a2009s.stats -b -a ../label/rh.aparc.a2009s.annot -c ../label/aparc.annot.a2009s.ctab recon_all rh white

Waiting for PID 509 of (509 512) to complete...
/opt/freesurfer/bin/reconbatchjobs: line 81:   509 Segmentation fault      (core dumped) exec $JOB >> $LOG 2>&1
Waiting for PID 512 of (509 512) to complete...
/opt/freesurfer/bin/reconbatchjobs: line 81:   512 Segmentation fault      (core dumped) exec $JOB >> $LOG 2>&1
PIDs (509 512) completed and logs appended.
Linux 336cda4cfeae 4.4.0-59-generic #80-Ubuntu SMP Fri Jan 6 17:47:47 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
```

I've seen `mris_anatomical_stats` segfault issues before. I suspect it's because of the huge chunk of skull left by AFNI skull stripping (which is probably because of the low-res test data).

Short term: we'll kill the `mris_anatomical_stats` outputs, at least for testing. The outputs aren't used downstream in recon-all.

Medium term: I'll try to get a minimal reproduction (and better logs) for the FreeSurfer folks. Even if it fails, it shouldn't be segfaulting.

Somewhat relevant: There's no way to actually run these on CircleCI. I'll hook up the `--no-freesurfer` option for the tests, but we should figure out a way to test and validate FreeSurfer bits (at least) before releases.

effigies:
Using https://openfmri.org/dataset/ds000005/ with `--participant_label sub-01` to test FreeSurfer.

Could someone have a look at this error?

```
170131-17:48:08,337 niworkflows INFO:
	 Generating report for ReconAll (subject recon_all)
170131-17:48:08,338 niworkflows INFO:
	 Generating visual report
170131-17:48:47,327 niworkflows INFO:
	 Successfully created report (/scratch/workflow_enumerator/sub-01/t1w_preprocessing/Reconstruction2/reconall.svg)
170131-17:48:47,404 workflow INFO:
	 [Job finished] jobname: Reconstruction2 jobid: 95
170131-17:48:47,514 workflow INFO:
	 Executing: ReconAll_Report ID: 96
170131-17:48:47,531 workflow INFO:
	 Executing node ReconAll_Report in dir: /scratch/workflow_enumerator/sub-01/t1w_preprocessing/ReconAll_Report
170131-17:48:47,586 workflow INFO:
	 [Job finished] jobname: ReconAll_Report jobid: 96
  File "/usr/local/miniconda/bin/fmriprep", line 9, in <module>
    load_entry_point('fmriprep', 'console_scripts', 'fmriprep')()
  File "/root/src/fmriprep/fmriprep/run_workflow.py", line 87, in main
    create_workflow(opts)
  File "/root/src/fmriprep/fmriprep/run_workflow.py", line 185, in create_workflow
    run_reports(settings['output_dir'])
  File "/root/src/fmriprep/fmriprep/viz/reports.py", line 190, in run_reports
    report = Report(root, config, out_dir, out_filename)
  File "/root/src/fmriprep/fmriprep/viz/reports.py", line 89, in __init__
    self._load_config(config)
  File "/root/src/fmriprep/fmriprep/viz/reports.py", line 104, in _load_config
    self.index()
  File "/root/src/fmriprep/fmriprep/viz/reports.py", line 125, in index
    self.index_error_dir(error_dir)
  File "/root/src/fmriprep/fmriprep/viz/reports.py", line 147, in index_error_dir
    crash_data = loadcrash(os.path.join(root, f))
  File "/src/nipype/nipype/utils/filemanip.py", line 502, in loadcrash
    return loadpkl(infile)
  File "/src/nipype/nipype/utils/filemanip.py", line 527, in loadpkl
    unpkl = pickle.load(pkl_file)
  File "/usr/local/miniconda/lib/python3.5/site-packages/traits/has_traits.py", line 1441, in __setstate__
    self.trait_set( trait_change_notify = trait_change_notify, **state )
  File "/usr/local/miniconda/lib/python3.5/site-packages/traits/has_traits.py", line 1544, in trait_set
    setattr( self, name, value )
  File "/src/nipype/nipype/interfaces/base.py", line 2047, in validate
    value = super(MultiPath, self).validate(object, name, newvalue)
  File "/usr/local/miniconda/lib/python3.5/site-packages/traits/trait_types.py", line 2337, in validate
    return TraitListObject( self, object, name, value )
  File "/usr/local/miniconda/lib/python3.5/site-packages/traits/trait_handlers.py", line 2314, in __init__
    raise excp
  File "/usr/local/miniconda/lib/python3.5/site-packages/traits/trait_handlers.py", line 2306, in __init__
    value = [ validate( object, name, val ) for val in value ]
  File "/usr/local/miniconda/lib/python3.5/site-packages/traits/trait_handlers.py", line 2306, in <listcomp>
    value = [ validate( object, name, val ) for val in value ]
  File "/src/nipype/nipype/interfaces/traits_extension.py", line 89, in validate
    self.error(object, name, value)
  File "/usr/local/miniconda/lib/python3.5/site-packages/traits/trait_handlers.py", line 173, in error
    value )
traits.trait_errors.TraitError: Each element of the 'in_file' trait of a DerivativesDataSinkInputSpec instance must be an existing file name, but a value of '/scratch/workflow_enumerator/01/t1w_preprocessing/Reconstruction2/report.html' <class 'str'> was specified.
```

It produces the expected output file, but I can't figure out why it seems to think there should exist a `/scratch/workflow_enumerator/01/t1w_preprocessing/Reconstruction2/report.html` file, given that the enumerator uses `sub-01`.

Going to try running again in an absolutely clean scratch directory, but that's going to take a while, so if you see something right away, cool.

effigies:
Okay, so I'm having two major issues. The first is the above, where for whatever reason the reporting works fine, but throws an error.

The second is I can't run [`-parcstats`, `-parcstats2`, `-parcstats3`, `-segstats`, `-wmparc` or `-balabels`](https://github.com/poldracklab/fmriprep/pull/347/files#diff-759e56fac1f65a08378d5d91a25fa84dR122), given an ANTS-skullstripped brainmask without segfaulting.

Though that may be because ANTS skullstripping seems a bit over-zealous.
ants_mask

effigies:
Report: [sub-01.html](https://qnl-linux1.bu.edu/~effigies/sub-01.html)

BIDS directory after running:

```Shell
$ tree -L 4 /data/bids/ds000005_R2.0.1 | head -n 27
/data/bids/ds000005_R2.0.1
├── CHANGES
├── dataset_description.json
├── derivatives
│   └── freesurfer
│       └── sub-01
│           ├── label
│           ├── mri
│           ├── scripts
│           ├── stats
│           ├── surf
│           ├── tmp
│           ├── touch
│           └── trash
├── participants.tsv
├── README
├── sub-01
│   ├── anat
│   │   ├── sub-01_inplaneT2.nii.gz
│   │   └── sub-01_T1w.nii.gz
│   └── func
│       ├── sub-01_task-mixedgamblestask_run-01_bold.nii.gz
│       ├── sub-01_task-mixedgamblestask_run-01_events.tsv
│       ├── sub-01_task-mixedgamblestask_run-02_bold.nii.gz
│       ├── sub-01_task-mixedgamblestask_run-02_events.tsv
│       ├── sub-01_task-mixedgamblestask_run-03_bold.nii.gz
│       └── sub-01_task-mixedgamblestask_run-03_events.tsv
...
```

effigies:
At this point, we have reconstruction, saving derivatives, and reporting.

At present, I'm accepting whatever external skullstripping is used, ANTs or AFNI. If we want to only swap out the FreeSurfer skullstripping if ANTs is specified, we could do that.

The following features are coded but not tested:

* Multiple T1w images (haven't yet dug up a dataset to test this on)
* T2 images (no <1.2mm T2s to test on)
  * Also, at present accepting `inplaneT2` (because those were available in the above dataset), but I think we should actually be looking for images labeled `T2w`?
* `-hires` (Waiting in a Sherlock queue to run on BEAST data)

Talking with @chrisfilo, perhaps the T2 stuff should be stripped out for now, and another PR can pull them in when we can get some proper testing. The same could go for multiple T1s or `-hires`, if we just want to get basic support in as quickly as possible.

The whole thing depends on nipy/nipype#1790, which serves as a decent test of that PR.

chrisfilo:
Yes - lets keep T2s for another PR. We should get some test data soon.

chrisfilo:
Are you planning to map the EPI timeseries to fsaverage in a separate PR?

effigies:
Yes, since it's a separate process to `recon-all`, that seemed to make more sense to me.

fmriprep/workflows/anatomical.py
```diff
@@ -74,10 +76,130 @@ def t1w_preprocessing(name='t1w_preprocessing', settings=None):
...
+            mridir = os.path.join(subjects_dir, subject_id, 'mri')
+            t1 = os.path.join(mridir, 'T1.mgz')
+            bm_auto = os.path.join(mridir, 'brainmask.auto.mgz')
+            bm = os.path.join(mridir, 'brainmask.mgz')
+
+            img = nib.load(t1)
+            mask = nib.load(skullstripped)
+            bmask = new_img_like(mask, mask.get_data() > 0)
+            resampled_mask = resample_to_img(bmask, img, 'nearest')
+            masked_image = new_img_like(img, img.get_data() * resampled_mask.get_data())
```
effigies: Mentioned this to Chris, but for anybody else who's reviewing, just note that the choice of binarizing before resampling will be slightly more prone to declaring voxels "not brain" than after. This is because, at the boundary, an interpolated value will always be `> 0`, while a nearest neighbor in a resampled binary map may be 0.

Just in case anybody wants to object.

fmriprep/workflows/anatomical.py
```diff
@@ -74,10 +76,130 @@ def t1w_preprocessing(name='t1w_preprocessing', settings=None):
...
+            masked_image.to_filename(bm_auto)
+
+            copyfile(bm_auto, bm, copy=True, use_hardlink=True)
```
effigies: Hardlinks should be fine, right? Seems a waste to copy.

fmriprep/workflows/anatomical.py
```diff
@@ -74,10 +76,130 @@ def t1w_preprocessing(name='t1w_preprocessing', settings=None):
...
+            import os
+            return os.path.join(dirname, basename)
+
+        subject_dir = pe.Node(
+            niu.Function(
+                function=joiner,
+                input_names=['dirname', 'basename'],
+                output_names=['path']),
+            name='SubjectDir',
+            run_without_submitting=True)
```
effigies: Is there a better way to get the full subject directory? It's 11 lines for `os.path.join`.

fmriprep/workflows/anatomical.py
```diff
@@ -74,10 +76,130 @@ def t1w_preprocessing(name='t1w_preprocessing', settings=None):
...
+
+        subject_out = pe.Node(
+            nio.DataSink(
+                base_directory=settings['output_dir'],
+                container='derivatives'),
+            name='FSSubjOut',
+            run_without_submitting=True)
```
effigies: Used `DataSink` instead of `DerivativesDataSink` because the latter doesn't like directories.

fmriprep/workflows/anatomical.py
```diff
@@ -74,10 +76,130 @@ def t1w_preprocessing(name='t1w_preprocessing', settings=None):
...
+                        img.header.get_zooms() == t1w_ref.header.get_zooms())):
+                    t1w_outs.append(t1w_outs)
+
+            autorecon1_flags = [default_flags]
+            reconall_flags = [default_flags]
+            if hires:
+                autorecon1_flags.append('-hires')
+                reconall_flags.append('-hires')
+            return (t1w_outs, ' '.join(autorecon1_flags),
+                    ' '.join(reconall_flags))
```
chrisfilo: Doesn't `-hires` need to be entered directly before a corresponding `-i` with a T1 volume for it to work?

fmriprep/workflows/anatomical.py
chrisfilo: Brain masks are so small and hard links are not supported on all systems. Will this throw an exception if hard links are not supported?

fmriprep/workflows/anatomical.py
effigies: Not that I can tell from reading the Csh source, it looks like it's set globally. Do you have a source suggesting this?

effigies: Though, it does look like it should come with an [experts file](https://surfer.nmr.mgh.harvard.edu/fswiki/SubmillimeterRecon), which we should probably produce. I guess just go with their `mris_inflate -n 15`?

fmriprep/workflows/anatomical.py
effigies: Yeah, the fallback behavior would be to copy.

fmriprep/workflows/anatomical.py
chrisfilo: Did you check in the logs that the right resolution is used internally? 
BTW it actually runs without an experts file (at least in my experience).

effigies: Yeah, it does run without the experts file.

Not sure what to look for in the log, but (from a currently running test):

```Python
>>> nib.load('/local-scratch/cjmarkie/12650555/workflow_enumerator/sub-387/t1w_preprocessing/Reconstruction/sub-387/mri/T1.mgz').header.get_zooms()
(0.89840001, 0.89840001, 0.89840001)
```

effigies: Pinged the FreeSurfer list about `-hires` best practices. My guess is that 15 max iterations should work for anything -- if something works well with a max of 10, it should stop by 10, but we should give smaller voxels a chance. 760d9de has the update I'd make if we want to go with that strategy. That seems like it's probably a good idea, since we'll want to support people with smaller voxels than we currently have access to.

requirements.txt
```diff
@@ -1 +1,2 @@
--e git+https://github.com/poldracklab/niworkflows.git@1c00a2490b7c5c9fd2c5d9afa7b6861d4ccecd08#egg=niworkflows
+-e git+https://github.com/effigies/nipype.git@a5fe6bb#egg=nipype
+-e git+https://github.com/effigies/niworkflows.git@7051b39#egg=niworkflows
```
effigies: Is it okay to have these pins go to my user? I figure we should probably have these issues resolved by the time we make another release, but in the short term, these are both needed.

Nipype pin: nipype/master version string is `0.13.0-dev` < `0.13.0-rc1` in `info.py`. This pin just updates to `rc2`.
Niworkflows pin: The nipype pin includes nipy/nipype#1784, which was also not in `0.13.0-rc1`. This makes niworkflows compatible with the above pin.

The simplest resolution for both is to have another release candidate in nipype, and let both depend on `>=0.13.0-rc2`.

chrisfilo:
We could also temporarily remove nipype version check from fmriprep and
niworkflwos (in info.py) and reinstate it whenever rc2 gets released.

docs/workflows.rst
```diff
@@ -47,6 +49,14 @@ warp to the MNI space.
 
     Animation showing T1 to MNI normalization (ANTs)
 
+Additionally, FreeSurfer surfaces are reconstructed from T1-weighted structural
+image(s), using the ANTs-extracted brain mask.
+This feature may be disabled with the ``--no-freesurfer`` flag.
+
+.. figure:: _static/reconall.svg
+
+    Surface reconstruction (FreeSurfer)
```
effigies: Can someone try building the docs on their machine? This one is cropped, unlike the others. It may be a consequence of the new figure generation. (The missing `:scale:` line doesn't make any difference for me.)

On that topic, we should probably update the sample report. What subject were you running it on? [`ds000005_R2.0.1`](https://openfmri.org/dataset/ds000005/)`/sub-01`? I'll run it with ANTS + FreeSurfer today and copy the new report.

chrisfilo:
I can give you access to readthedocs so you could build docs there. I need
your username though.

As for the example report it does not quite matter which subject you use. I
would consider using something with higher resolution because it more
impressive, and now with compression the reports should not be too big.

effigies:
Readthedocs user: markiewicz

chrisfilo:
done!

effigies:
Build a [`tmp_doc`](https://fmriprep.readthedocs.io/en/tmp_doc/workflows.html) version for testing. It looked the same as on my local machine. Updated the SVG using inkscape (Document Properties -> Resize page to content).

I believe the missing "height" and "width" properties were the problem. Did we have to get rid of them to make full-width scaling work?

chrisfilo:
I believe that was the case. Have you tried specifying the height and width
when linking the image in the docs (in the HTML tag)?

oesteban:
Yes, if I'm not wrong (@rwblair will correct me if so), you need to specify the ```viewBox``` property, and remove ```height``` and ```width```

effigies:
Manually entering `:height:` and `:width:` directives and leaving the SVG unchanged fixes the width, but doesn't rescale the image properly.
screenshot from 2017-02-13 12-54-24

It may be that for documentation purposes, we're just going to have to manually edit these figures. I'll look into an inkscape CLI invocation that will simplify the process as much as possible.

effigies:
So this is nearing final state.

I'm inclined to leave the documentation in its [current state](https://fmriprep.readthedocs.io/en/tmp_doc/workflows.html) rather than try to fix SVG issues across-the-board here. Let me know if there's any more you want added to the docs, though.

I'm re-running a high-res test on a Beast subject (387) on Sherlock. The only remaining issue here is whether we want to increase the `mris_inflate -n` default. [This response to my question to the FS list](https://mail.nmr.mgh.harvard.edu/pipermail//freesurfer/2017-February/050091.html) suggests that a number like 50 gives reasonable flexibility while not adding computation time (when it's not needed). 1c5c5ba is my proposal to use 50, and I'm running a separate test on Sherlock so we can compare results/runtime. These tests will also verify that `fsaverage` gets properly copied into `derivatives/freesurfer`.

requirements.txt
chrisfilo: Could we instead temporarily remove '>=0.13.0-rc1' from info (in both niworkflows and fmriprep) and switch those to two lines in requirements.txt to point to main niworkflows and nipype repositories? It will make it more straightforward for others to build on top.

effigies: Oh, right. That's still unresolved. See @oesteban's comments on poldracklab/niworkflows#117, where I started to do that. I'm still unsure exactly how to proceed here.

effigies:
WRT the expert file, I can't actually make it work. I can create it (and [verify it exists](6b3fbd7)), but when `Reconstruction2` runs, it can't be found. I assume it's getting deleted, but I couldn't find any obvious reason for it in the `nipype.interfaces.utility.Function` hierarchy.

If we want to do this, I can try more to figure out what's going on, but it's probably not important to this PR.

chrisfilo:
Agreed

On Feb 13, 2017 6:22 PM, "Chris Markiewicz" <notifications@github.com>
wrote:

WRT the expert file, I can't actually make it work. I can create it (and verify
it exists <http://6b3fbd7>), but when Reconstruction2 runs, it can't be
found. I assume it's getting deleted, but I couldn't find any obvious
reason for it in the nipype.interfaces.utility.Function hierarchy.

If we want to do this, I can try more to figure out what's going on, but
it's probably not important to this PR.

—

<https://github.com/poldracklab/fmriprep/pull/347#issuecomment-279588295>,
or mute the thread

effigies:
Derivatives properly copying. Rebased against `master`, waiting on tests.

Any more comments?

fmriprep/workflows/anatomical.py
```diff
@@ -72,10 +74,131 @@ def t1w_preprocessing(name='t1w_preprocessing', settings=None):
...
+
+        subject_out = pe.Node(
+            nio.DataSink(
+                base_directory=settings['output_dir'],
+                container='derivatives'),
+            name='FSSubjOut',
+            run_without_submitting=True)
+        # Copy fsaverage subject into derivatives/freesurfer
+        setattr(subject_out.inputs, 'freesurfer.@fsaverage',
+                '/opt/freesurfer/subjects/fsaverage')
```
effigies: Working on `bbregister`/`mri_vol2surf`, I'm realizing we don't want this inside `t1w_preprocessing` -- or really anything below `workflow_enumerator` -- as then it would be re-run for every subject.

Does it make sense to make something like:

```Python
class BIDSFreeSurferDir(BaseInterface):
    inputs = (freesurfer_home, derivatives_dir)
    outputs = subjects_dir

         # Create subjects directory
         # Copy fsaverage
```

And then pass `subjects_dir` through to workflows that need access to it? We would then be `recon-all`'ing directly in `derivatives/freesurfer/sub-{subj_id}/`, which seems okay to me.

WDYT, @chrisfilo? This can wait, but I think we'll have to deal with it in the next PR (where bbregister will need to know about the common subjects directory to do the registration) if we don't, now.

effigies:
Okay, so I've tested the following cases on Sherlock:

1) New derivatives directory - copies `fsaverage`, creates new subject
2) Existing `fsaverage`, new subject - leaves `fsaverage` untouched, creates new subject
3) Existing `fsaverage` and subject - leaves `fsaverage` untouched, runs `ReconAll` without error (also not doing anything)
4) Existing `fsaverage` and subject, deleted nipype scratch directory - leaves `fsaverage` untouched, runs `ReconAll` without error (overwriting brainmasks)

Pushed a commit to address this last case and not overwrite brainmasks. I believe we agreed that existing recon directories should be treated as "correct", and if someone wants to start over, they can delete the recon directory.

chrisfilo:
+1

effigies:
Alright. No errors, and existing brainmasks were left alone in case 4. This is ready for final review and merge.

fmriprep/workflows/anatomical.py
```diff
@@ -74,10 +76,130 @@ def t1w_preprocessing(name='t1w_preprocessing', settings=None):
...
+            name='ReconConfig',
+            run_without_submitting=True)
+        recon_config.inputs.default_flags = '-noskullstrip'
+
+        def bidsinfo(in_file):
+            from fmriprep.interfaces.bids import BIDS_NAME
+            match = BIDS_NAME.search(in_file)
+            params = match.groupdict() if match is not None else {}
+            return tuple(map(params.get, ['subject_id', 'ses_id', 'task_id',
+                                          'acq_id', 'rec_id', 'run_id']))
```
chrisfilo: This could be useful for pybids. I've opened an issue https://github.com/INCF/pybids/issues/31

chrisfilo:
Excellent work!

PR #349, opened 2017-01-28
Reorder Docker commands to reduce build time

effigies:
This change should reduce the time needed to build a new docker image. The premise is that the most expensive, most stable actions should be first, to take advantage of cache. FreeSurfer release is out, and takes longer than AFNI/FSL, so it's moved forward. Git/graphviz are moved toward the end, as these are cheap, and make a natural place to add new apt packages without forcing a re-fetch of FreeSurfer/AFNI/FSL/ANTs/etc.

I haven't really measured the other parts, so if there are some other ordering changes people want made, let me know.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,33 +1,12 @@
 # Use Ubuntu 16.04 LTS
 FROM ubuntu:xenial-20161213
 
-# Installing ubuntu packages (FSL, AFNI, git)
+# Prepare environment
 RUN apt-get update && \
     apt-get install -y --no-install-recommends curl bzip2 ca-certificates xvfb && \
     curl -sSL http://neuro.debian.net/lists/xenial.us-ca.full >> /etc/apt/sources.list.d/neurodebian.sources.list && \
     apt-key adv --recv-keys --keyserver hkp://pgp.mit.edu:80 0xA5D32F012649A5A9 && \
-    apt-get update && \
-    apt-get install -y --no-install-recommends \
-                    fsl-core=5.0.9-1~nd+1+nd16.04+1 \
-                    git=1:2.7.4-0ubuntu1 \
-                    afni=16.2.07~dfsg.1-2~nd16.04+1 \
-                    graphviz=2.38.0-12ubuntu2 && \
-    apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
-
-ENV FSLDIR=/usr/share/fsl/5.0
-ENV FSLOUTPUTTYPE=NIFTI_GZ
-ENV PATH=/usr/lib/fsl/5.0:$PATH
-ENV FSLMULTIFILEQUIT=TRUE
-ENV POSSUMDIR=/usr/share/fsl/5.0
-ENV LD_LIBRARY_PATH=/usr/lib/fsl/5.0:$LD_LIBRARY_PATH
-ENV FSLTCLSH=/usr/bin/tclsh
-ENV FSLWISH=/usr/bin/wish
-ENV FSLOUTPUTTYPE=NIFTI_GZ
-ENV AFNI_MODELPATH=/usr/lib/afni/models
-ENV AFNI_IMSAVE_WARNINGS=NO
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -64,6 +44,26 @@ ENV MNI_PERL5LIB /opt/freesurfer/mni/lib/perl5/5.8.5
 ENV PATH /opt/freesurfer/bin:/opt/freesurfer/fsfast/bin:/opt/freesurfer/tktools:/opt/freesurfer/mni/bin:$PATH
 RUN echo "cHJpbnRmICJrcnp5c3p0b2YuZ29yZ29sZXdza2lAZ21haWwuY29tXG41MTcyXG4gKkN2dW12RVYzelRmZ1xuRlM1Si8yYzFhZ2c0RVxuIiA+IC9vcHQvZnJlZXN1cmZlci9saWNlbnNlLnR4dAo=" | base64 -d | sh
 
+# Installing Neurodebian packages (FSL, AFNI, git)
+RUN apt-get install -y --no-install-recommends \
+                    fsl-core=5.0.9-1~nd+1+nd16.04+1 \
+                    afni=16.2.07~dfsg.1-2~nd16.04+1
+
+ENV FSLDIR=/usr/share/fsl/5.0
```
oesteban: All these ENV statements would be a bit quicker and cached together with a single command.

```
ENV FSLDIR=/usr/share/fsl/5.0 \
        FSLOUTPUTTYPE=NIFTI_GZ
```

I think this is a good opportunity to fix this one

Dockerfile
```diff
@@ -50,6 +29,7 @@ ENV OS Linux
 ENV FS_OVERRIDE 0
```
oesteban: These ENV instructions could be wrapped together in a single ENV (see comment later on)

effigies:
Let me know what you think of this change. Instead of setting all variables at once, I reduced to a few blocks, so that we could use existing variables to define new ones, when it seemed to make sense.

Dockerfile
```diff
@@ -45,25 +24,44 @@ RUN curl -sSL https://surfer.nmr.mgh.harvard.edu/pub/dist/freesurfer/6.0.0/frees
...
-ENV FREESURFER_HOME /opt/freesurfer
-ENV FSFAST_HOME /opt/freesurfer/fsfast
-ENV MINC_BIN_DIR /opt/freesurfer/mni/bin
-ENV MINC_LIB_DIR /opt/freesurfer/mni/lib
-ENV MNI_DATAPATH /opt/freesurfer/mni/data
-ENV FMRI_ANALYSIS_DIR /opt/freesurfer/fsfast
-ENV PERL5LIB /opt/freesurfer/mni/lib/perl5/5.8.5
-ENV MNI_PERL5LIB /opt/freesurfer/mni/lib/perl5/5.8.5
-ENV PATH /opt/freesurfer/bin:/opt/freesurfer/fsfast/bin:/opt/freesurfer/tktools:/opt/freesurfer/mni/bin:$PATH
+ENV FSL_DIR=/usr/share/fsl/5.0 \
```
oesteban: I see that FSL_DIR (and maybe others) are set twice, is this necessary?. Other than that, LGTM!

oesteban:
The split in blocks makes total sense :). The only detail remaining is that some variables seem to be set twice (or maybe only the FSL_DIR).

Dockerfile
effigies: `FSL_DIR` and `FSLDIR` are not the same. I think the former is FreeSurfer's variable, and the latter FSL's. Annoying, but that's how my machine is set up, too.

oesteban: Gotcha

oesteban:
I'll merge when circle finishes

oesteban:
Seems like a ```\``` is missing:

```
Step 5 : FREESURFER_HOME=/OPT/FREESURFER 
Unknown instruction: FREESURFER_HOME=/OPT/FREESURFER
```

effigies:
Good catch. Fixed and rebased.

PR #350, opened 2017-01-30
Flake8 cleanup workflows/anatomical

effigies:
Mostly line length and block spacing issues. These keep popping up when I style check before committing in #347, but I don't really want to infect that PR with these cleanups.

If people don't like these changes, can we put in a tox.ini or something to tell flake8 what style issues to report?

### DIFF
--- a/fmriprep/workflows/anatomical.py
+++ b/fmriprep/workflows/anatomical.py
@@ -11,8 +11,6 @@ Originally coded by Craig Moodie. Refactored by the CRN Developers.
 import os.path as op
 
 from nipype.interfaces import ants
-from nipype.interfaces import fsl
-from nipype.interfaces import io as nio
 from nipype.interfaces import utility as niu
 from nipype.pipeline import engine as pe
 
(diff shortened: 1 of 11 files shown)


rwblair:
LGTM. Maybe we bump minimum line length? I know its against the holy scripture for public code... Pylint  only complains at 100 characters. But it also has some interesting ideas about what good looking code is that are not in pep8.

oesteban:
I'm totally open to whatever standard you guys want to stick with. My only issue is that 79 characters is quite reduced, I would advocate for 99.

effigies:
99 works. Added a `tox.ini`, and ran through the rest of the workflows directory to get a sense of how much changes.

Mostly finding some blank line counts and unused imports. A few undefined names, which I left alone, not knowing what they were supposed to be.

oesteban:
Is this ready to merge?

effigies:
Yup.



Sent on my Samsung Galaxy S® 5

PR #357, opened 2017-02-13
Make BIDSDataGrabber great (python<3) again

oesteban:
Right now, the DictStrAny does not take in unicode strings.
This PR patches that.


oesteban:
Change issue title to turn this PR into clickbait for reviewers :P

chrisfilo:
There is a million maybe million and a half reasons to merge this pull request.

oesteban:
We're discussing in slack if we maybe should drop python 2

PR #362, opened 2017-02-17
[RTM] Replace FLIRT-BBR with FreeSurfer bbregister when FS preprocessing is enabled

### LINKED ISSUE #64: Replace BBR with bbregister

effigies:
Closes #64


chrisfilo:
It would be important to make sure that the transformation matrix produced by bbregister moves the image to the original T1 space not to the FS conformed T1.

effigies:
@oesteban Could you have a look at https://circleci.com/gh/effigies/fmriprep/102?

[This node](https://github.com/poldracklab/fmriprep/blob/18d3c701d33cb983e162772e7a300d1ed366c267/fmriprep/workflows/fieldmap/phase_diff_and_magnitudes.py#L88-L90) is getting mis-connected, it looks like.

```
170221-15:17:46,817 workflow INFO:

	 Executing: DemeanFmap ID: 36

170221-15:17:46,869 workflow INFO:

	 Executing node DemeanFmap in dir: /scratch/workflow_enumerator/100185/phase_diff_and_magnitudes/DemeanFmap

  File "/root/src/nipype/nipype/pipeline/plugins/base.py", line 249, in run

    result=result))

  File "/root/src/nipype/nipype/pipeline/plugins/base.py", line 300, in _clean_queue

    raise RuntimeError("".join(result['traceback']))

RuntimeError: Traceback (most recent call last):

  File "/root/src/nipype/nipype/pipeline/plugins/multiproc.py", line 52, in run_node

  File "/root/src/nipype/nipype/pipeline/engine/nodes.py", line 366, in run

    self._run_interface()

  File "/root/src/nipype/nipype/pipeline/engine/nodes.py", line 476, in _run_interface

  File "/root/src/nipype/nipype/pipeline/engine/nodes.py", line 607, in _run_command

  File "/root/src/nipype/nipype/interfaces/base.py", line 1085, in run

    runtime = self._run_wrapper(runtime)

  File "/root/src/nipype/nipype/interfaces/base.py", line 1033, in _run_wrapper

  File "/root/src/nipype/nipype/interfaces/utility.py", line 498, in _run_interface

    out = function_handle(**args)

  File "<string>", line 16, in demean_image

NameError: name 'NUMPY_MMAP' is not defined

Interface Function failed to run. 
```

This shouldn't have been touched by the changes here. But I suppose it's possible that it's something on the nipype end with your recent PRs that we haven't adjusted for? (I'm pinning a newer nipype.)

chrisfilo:
BTW - I think you can switch to main branch on both niworkflows and nipype now.

effigies:
I have a niworkflows branch (no PR yet) with changes needed for this bit. When I'm sure I don't need more, I'll put a PR there.

fmriprep/workflows/epi.py
```diff
@@ -138,6 +138,17 @@ def ref_epi_t1_registration(reportlet_suffix, inv_ds_suffix, name='ref_epi_t1_re
 
     explicit_mask_epi = pe.Node(fsl.ApplyMask(), name="explicit_mask_epi")
 
+    if settings['freesurfer']:
+        bbregister = pe.Node(
+            BBRegisterRPT(
+                contrast_type='t2',
+                init='header',
+                registered_file=True,
+                out_fsl_file=True,
```
effigies: Running BBRegisterRPT, I'm running into the issue that `registered_file` and `out_reg_file` are undefined in the outputs, but not in the command (which [uses `_list_outputs`](https://github.com/nipy/nipype/blob/86cd8ffb6dae1362a07ed8e9c79d583db87780eb/nipype/interfaces/freesurfer/preprocess.py#L1043-L1051) to build the command). Neither exists in the working directory. Weirdly, `out_fsl_file` is defined and exists, so it doesn't seem like it should be an mixin/MRO issue.

```RestructuredText
Node: sub-387 (ref_epi_t1_registration (bbregister (registration)
=================================================================

 Hierarchy : workflow_enumerator.sub-387.ref_epi_t1_registration.bbregister
 Exec ID : bbregister.a2

Original Inputs
---------------

* args : <undefined>
* compress_report : auto
* contrast_type : t2
* environ : {'SUBJECTS_DIR': '/opt/freesurfer/subjects'}
* epi_mask : <undefined>
* generate_report : True
* ignore_exception : False
* init : header
* init_reg_file : <undefined>
* intermediate_file : <undefined>
* out_fsl_file : True
* out_reg_file : <undefined>
* out_report : report.svg
* reg_frame : <undefined>
* reg_middle_frame : <undefined>
* registered_file : True
* source_file : /scratch/workflow_enumerator/sub-387/ref_epi_t1_registration/_epi_..data..sub-387..func..sub-387_task-Beast_run-3_bold.nii.gz/explicit_mask_epi/sub-387_task-Beast_run-3_bold_mcf.nii.gz_mean_reg_masked.nii.gz
* spm_nifti : <undefined>
* subject_id : sub-387
* subjects_dir : /out/freesurfer
* terminal_output : allatonce

Execution Inputs
----------------

* args : <undefined>
* compress_report : auto
* contrast_type : t2
* environ : {'SUBJECTS_DIR': '/out/freesurfer'}
* epi_mask : <undefined>
* generate_report : True
* ignore_exception : False
* init : header
* init_reg_file : <undefined>
* intermediate_file : <undefined>
* out_fsl_file : True
* out_reg_file : <undefined>
* out_report : report.svg
* reg_frame : <undefined>
* reg_middle_frame : <undefined>
* registered_file : True
* source_file : /scratch/workflow_enumerator/sub-387/ref_epi_t1_registration/_epi_..data..sub-387..func..sub-387_task-Beast_run-3_bold.nii.gz/bbregister/sub-387_task-Beast_run-3_bold_mcf.nii.gz_mean_reg_masked.nii.gz
* spm_nifti : <undefined>
* subject_id : sub-387
* subjects_dir : /out/freesurfer
* terminal_output : allatonce

Execution Outputs
-----------------

* min_cost_file : <undefined>
* out_fsl_file : /scratch/workflow_enumerator/sub-387/ref_epi_t1_registration/_epi_..data..sub-387..func..sub-387_task-Beast_run-3_bold.nii.gz/bbregister/sub-387_task-Beast_run-3_bold_mcf.nii.gz_mean_reg_masked_bbreg_sub-387.mat
* out_reg_file : <undefined>
* out_report : /scratch/workflow_enumerator/sub-387/ref_epi_t1_registration/_epi_..data..sub-387..func..sub-387_task-Beast_run-3_bold.nii.gz/bbregister/report.svg
* registered_file : <undefined>

Runtime info
------------

* command : bbregister --t2 --init-header --fslmat /scratch/workflow_enumerator/sub-387/ref_epi_t1_registration/_epi_..data..sub-387..func..sub-387_task-Beast_run-3_bold.nii.gz/bbregister/sub-387_task-Beast_run-3_bold_mcf.nii.gz_mean_reg_masked_bbreg_sub-387.mat --reg /scratch/workflow_enumerator/sub-387/ref_epi_t1_registration/_epi_..data..sub-387..func..sub-387_task-Beast_run-3_bold.nii.gz/bbregister/sub-387_task-Beast_run-3_bold_mcf.nii.gz_mean_reg_masked_bbreg_sub-387.dat --o /scratch/workflow_enumerator/sub-387/ref_epi_t1_registration/_epi_..data..sub-387..func..sub-387_task-Beast_run-3_bold.nii.gz/bbregister/sub-387_task-Beast_run-3_bold_mcf.nii.gz_mean_reg_masked_bbreg.nii.gz --mov /scratch/workflow_enumerator/sub-387/ref_epi_t1_registration/_epi_..data..sub-387..func..sub-387_task-Beast_run-3_bold.nii.gz/bbregister/sub-387_task-Beast_run-3_bold_mcf.nii.gz_mean_reg_masked.nii.gz --s sub-387
* duration : 67.60915
* hostname : 9c7c0a3e6622

Terminal output
~~~~~~~~~~~~~~~
```

@chrisfilo Anything obvious stick out to you?

effigies: Oh. `source_file` changed.

Was: `[...]/_epi_..data..sub-387..func..sub-387_task-Beast_run-3_bold.nii.gz/explicit_mask_epi/sub-387_task-Beast_run-3_bold_mcf.nii.gz_mean_reg_masked.nii.gz`
Now: `[...]/_epi_..data..sub-387..func..sub-387_task-Beast_run-3_bold.nii.gz/bbregister/sub-387_task-Beast_run-3_bold_mcf.nii.gz_mean_reg_masked.nii.gz`

Looks like a nipype issue. I'll try to figure this out.

chrisfilo:
Probably has something to do with `copyfile`.

fmriprep/workflows/epi.py
effigies: Ugh. `remove_unnecessary_outputs` is `True` by default.

It does seem like if I say "Make this file", it should at least throw a warning if it's unused, but I suppose setting a config variable is easy enough.

effigies:
Yeah, it did, and wasn't at all the problem. Well, I guess I understand more of nipype's internals now. :1st_place_medal:

fmriprep/workflows/epi.py
chrisfilo: We should definitely make reportlets work independently off what `remove_unnecessary_outputs` is set to.

As far as I remember with DEBUG logging you get information which files are removed.

effigies: Reportlets work fine (mainly because we hook them up to an output sink). It's the outputs like the registered file (which I wanted to visually compare to the FLIRT-registered file) that get dropped if there isn't something else grabbing them.

Unfortunately, `needed_outputs` is an attribute of `Node`, not `BaseInterface`, so there's no simple way for an interface to mark the outputs that it uses as do-not-erase.

And yeah, the DEBUG logging is how I realized they were at least being created.

effigies:
Okay. We're definitely getting different outputs, and it appears that `bbregister` is registering to the 256x256x256 FreeSurfer space.

Visually, the output images align.

sub-387_T1w.nii.gz
------------------
```
Dimensions: 256 x 256 x 186
Voxel sizes: 0.8984 x 0.8984 x 0.9
Affine: [[   0.            0.           -0.89999998   78.08100128]
         [   0.           -0.89840001    0.          131.02000427]
         [  -0.89840001    0.            0.          110.31900024]
         [   0.            0.            0.            1.        ]]
```
sub-387_task-Beast_run-1_bold.nii.gz
------------------------------------
```
Dimensions: 100 x 100 x 64 x 650
Voxel sizes: 2.2 x 2.2 x 2.2 x 0.68
Affine: [[ -2.20000005e+00   0.00000000e+00   1.99987669e-03   1.00387001e+02]
         [  7.35627196e-04  -1.97774065e+00   9.63940501e-01   8.25609970e+01]
         [ -1.50989520e-03  -9.63606596e-01  -1.97757792e+00   1.03964500e+02]
         [  0.00000000e+00   0.00000000e+00   0.00000000e+00   1.00000000e+00]]
```
sub-387_task-Beast_run-1_bold_mcf.nii.gz_mean_reg_masked_flirt.nii.gz
---------------------------------------------------------------------
```
Dimensions: 186 x 256 x 256
Voxel sizes: 0.9 x 0.8984 x 0.8984
Affine: [[   0.89999998    0.            0.          -88.41899109]
         [   0.            0.89840001    0.          -98.0719986 ]
         [   0.            0.            0.89840001 -118.77300262]
         [   0.            0.            0.            1.        ]]
```

sub-387_task-Beast_run-1_bold_mcf.nii.gz_mean_reg_masked_bbreg.nii.gz
---------------------------------------------------------------------
```
Dimensions: 256 x 256 x 256
Voxel sizes: 0.8984 x 0.8984 x 0.8984
Affine: [[  -0.89840013    0.            0.          109.3762207 ]
         [   0.            0.            0.89840001  -98.97039795]
         [   0.           -0.89840001    0.          110.31900024]
         [   0.            0.            0.            1.        ]]
```

sub-387_task-Beast_run-1_bold_mcf.nii.gz_mean_reg_masked_flirt.mat
------------------------------------------------------------------
```
Registration: [[  9.99822613e-01   2.18406143e-03  -1.87130549e-02  -2.01251828e+01]
               [  1.02471905e-02  -8.96531511e-01   4.42861559e-01   1.78893223e+02]
               [ -1.58096043e-02  -4.42974773e-01  -8.96394706e-01   2.26519183e+02]
               [  0.00000000e+00   0.00000000e+00   0.00000000e+00   1.00000000e+00]]
```

sub-387_task-Beast_run-1_bold_mcf.nii.gz_mean_reg_masked_bbreg_sub-387.mat
--------------------------------------------------------------------------
```
Registration: [[  9.99783580e-01   4.73640000e-04  -2.07878600e-02   1.14698811e+01]
               [  1.83094900e-02   4.41616650e-01   8.96942620e-01   2.37868500e+00]
               [  9.54438000e-03  -8.97202850e-01   4.41665860e-01   1.80107742e+02]
               [  0.00000000e+00   0.00000000e+00   0.00000000e+00   1.00000012e+00]]
```

chrisfilo:
It might be good to pick @satra brains on this or peruse the workflows in
nipype. I have a feeling this might have been dealt with in the past.

effigies:
So 7e56d04 implements the FreeSurfer-conformed volume -> original T1w volume transformation. However, now that I'm trying to think through it, I'm confusing myself.

Basic idea: `t1mgz_affine = nb.load('T1.mgz').affine :: ijk -> LIA` (in FreeSurfer-conformed volume), and `t1w_affine =nb.load('t1w.nii.gz').affine :: ijk -> IPL` (for example; in raw MGZ space). Hence, `pinv(t1mgz_affine).dot(t1w_affine) :: LIA -> IPL`... except why should the `ijk`s correspond?

@satra, if you get a second, could you have a look and let me know if this is what you were describing? If anybody else sees something obvious, let me know.

satra:
```
fsl bbr -> fsl1.mat
fs bbr -> fsl2.mat 
```

fsl1.mat -> reference is regular T1.nii
fsl2.mat -> reference is freesurfer space (likely same as T1.mgz, but i'm never 100% sure till i try it out)

so technically if you know the mapping between T1.nii and freesurfer space, you should be able to go one way or the other. of the top of my head i was thinking that this would be the same as the mapping between fsl1.mat and fsl2.mat i.e., pinv(fsl1.mat).dot(fsl2.mat) or vice verse depending on which transform you want.

t1w_affine maps from ijk_t1w to xyz 
t1mgz_affine maps from ijk_mgz to xyz

it's the xyz that's common to both, which is what allows you to relate them.

effigies:
I think what you describe -- mapping between the two structural spaces using the functional registrations -- assumes that FSL BBR and FS bbregister are going to produce the same alignment? If that were the case, we should just choose the one that puts us in the space we want, in this case FSL.

I'd rather avoid the functional step altogether and use the structural affines, and given that FreeSurfer is doing nothing but resampling to 256^3 isometric voxels, this shouldn't be fancy. If what we're after is `ijk_mgz` -> `ijk_t1w`, I think we should be able to just turn this around and say `t1mgz_affine.dot(pinv(t1w_affine))`.

I might still be confused, but I feel like this make more sense.

satra:
@effigies - i wasn't saying you should map between functional spaces - since you don't really want to run both registrations. just that the mapping between those two spaces should roughly equal the structural transform as a sanity check.

however, each registration has it's own source of error so simply doing a structural transform is not going to be equivalent to running the other registration.

the structural transform is still the safest thing to do.

effigies:
Wondering if this is causing an issue: [Image+centering.pdf](https://github.com/poldracklab/fmriprep/files/828282/Image.centering.pdf)

The location of the center defined by the affine in the raw images seems to be very different, and the center in the `flt_bbr`'ed a `bbregister`ed images are also different. Is this normal/something that shouldn't affect the affine correction?

effigies:
So this has gone on for quite a while, and I want to try to give a status update, to clear my head and get people's opinions on how best to proceed. I'm testing on ds000005 (give or take some 0's) to avoid LR flip issues in the uh2 datasets.

We have been using the following FLIRT sequence (actual nipype filenames simplified):

```Shell
$ flirt -in $EPI -ref $T1 -out init.nii.gz -omat init.mat -dof 6
$ flirt -in $EPI -ref $T1 -omat flirt.mat -searchcost bbr -dof 6 -init init.mat \
    -schedule /usr/share/fsl/5.0/etc/flirtsch/bbr.sch -wmseg $WMMASK
```

Here `T1` refers to the ANTs-generated brainmask, not the original T1 in the BIDS folder.

Yielding flirt.mat:

```
 0.9997169368   -0.008484848857  -0.02222782656  -15.35086442
-0.006684317239  0.7964569286    -0.6046583701    70.34123553 
 0.02283393991   0.6046357384     0.7961747509    15.49617409
 0  0  0  1
```
We're replacing this with:

```Shell
$ recon-all -s $SUBJ -i $ORIG_T1 -all
$ bbregister --s $SUBJ --mov $EPI --init-coreg --t2 --reg bbreg.dat --fslmat fsl.mat
$ tkregister2 --no-edit --mov $EPI --reg bbreg.dat --targ $T1 --fslregout adjusted.mat
```

This yields fsl.mat (output of bbregister):

```
 0.99972117   -0.01106763   -0.02086867   33.85606003
-0.02328384   -0.61062598   -0.79157710  240.28797913
-0.00398207    0.79184198   -0.61071330   70.66278839
 0.00000000    0.00000000    0.00000000    1.00000000
```

And adjusted.mat (output of tkregister2):
```
-0.99972129    0.01106763    0.02086867  173.14398193
-0.02328384   -0.61062604   -0.79157716  240.28800964
-0.00398206    0.79184198   -0.61071336   70.66275787
-0.00000000   -0.00000000   -0.00000000    1.00000012 
```

These aren't the same as `flirt.mat`, but then, the affine matrices of the outputs of `flirt` and `bbregister` are, respectively:

```
   1.            0.            0.          -81.
   0.            1.33333302    0.         -133.
   0.            0.            1.33333302 -129.
   0.            0.            0.            1.
```

```
  -0.99999994    0.            0.          127.00000763
   0.            0.            0.99999994 -133.
   0.           -0.99999994    0.          126.99995422
   0.            0.            0.            1.
```

Aligning these on Freeview shows basically identical images:

flirt

bbr

So it may be that I've been making too much of differing afffines and transforms, and this is a perfectly acceptable place to be; the aligned functionals are scaled and centered slightly differently, but we have transforms back into the original T1 space. Maybe the best way to test is to determine whether or not the downstream workflow is unable to cope with these differences. What's most likely to get distorted by a messed up

Alternately, what's the best way to directly test the transforms? `flirt -in bbregistered.nii.gz -ref $T1 -init adjusted.mat -applyxfm -o out.nii.gz` produces those weirdly flipped images:

applyxfm
 

@chrisfilo @oesteban

effigies:
As discussed, I did:

```Shell
flirt -in bbregistered.nii.gz -ref $T1 -init adjusted.mat -applyxfm -o out.nii.gz
```

The non-shorthand command was:

```Shell
flirt -in /scratch/workflow_enumerator/01/ref_epi_t1_registration/_epi_..data..sub-01..func..sub-01_task-mixedgamblestask_run-01_bold.nii.gz/explicit_mask_epi/sub-01_task-mixedgamblestask_run-01_bold_mcf.nii.gz_mean_reg_masked.nii.gz \
      -ref /scratch/workflow_enumerator/01/t1w_preprocessing/ANTsBrainExtraction/Ants_T1_Brain_Extraction/highres001_BrainExtractionBrain.nii.gz \
      -init /scratch/workflow_enumerator/01/ref_epi_t1_registration/_epi_..data..sub-01..func..sub-01_task-mixedgamblestask_run-01_bold.nii.gz/tkreg2/epi2t1w.mat \
      -applyxfm -o /scratch/test.nii.gz
```

Checking `test.nii.gz` with Freeview:

applyxfm-mean_epi_tkreg

Verifying the `applyxfm` command with the `flt_bbr` output matrix:

```Shell
flirt -in /scratch/workflow_enumerator/01/ref_epi_t1_registration/_epi_..data..sub-01..func..sub-01_task-mixedgamblestask_run-01_bold.nii.gz/explicit_mask_epi/sub-01_task-mixedgamblestask_run-01_bold_mcf.nii.gz_mean_reg_masked.nii.gz \
      -ref /scratch/workflow_enumerator/01/t1w_preprocessing/ANTsBrainExtraction/Ants_T1_Brain_Extraction/highres001_BrainExtractionBrain.nii.gz \
      -init /scratch/workflow_enumerator/01/ref_epi_t1_registration/_epi_..data..sub-01..func..sub-01_task-mixedgamblestask_run-01_bold.nii.gz/flt_bbr/sub-01_task-mixedgamblestask_run-01_bold_mcf.nii.gz_mean_reg_masked_flirt.mat \
      -applyxfm -o /scratch/test.nii.gz
```

applyxfm-mean_epi_fltbbr

So we're definitely not working...

chrisfilo:
1. What if you set the ref as the FS conformed T1?
2. How does the original T1 and the FS conformed T1 relate?
4. Can you create an affine going from original to FS conformed T1?

effigies:
1. You get the right orientations, but shifted to the right.

applyxfm-mean_epi_tkreg_tomgz

2. Resampled to isotropic voxels, re-centered to anterior commissure.
3. That was the goal of the whole `pinv(t1w_affine).dot(t1mgz_affine)` thing, which doesn't really work:

applyxfm-mean_epi_transformed

I think that's because they set their `(0, 0, 0)` points in different places, so there's no common `xyz` space to work to. Which of course presents the question of how Freeview and mricron can show these files aligned in the first place.

I have tried doing ANTs registration from one to the other, but that seems to give affines following yet another convention. I suppose I can try using `flirt`, but even given that, I guess we're stuck at the problem where we have all these affines and no clear way to relate them.

chrisfilo:
Hmm... If you use ANTs to coregister original T1 to FS conformed T1
restricting transformation only to translations and scaling you should get
a good approximation to how such matrix should look in theory and thus
could be estimated de novo without having to performed coregistration. Or
am I missing something?

Alterantively, we can give up on trying to figure this out and output the
conformed T1 as the T1w_preproc (and adjust FAST to use it instead of the
original T1).

>

satra:
@effigies - there is a magical matrix that you may be missing:

```
M = np.array([[-1, 0, 0, 128],
              [ 0, 0, 1,-128],
              [ 0,-1, 0, 128],
              [ 0, 0, 0,   1]],dtype=float)
```
try applying this it essentially relates freesurfer surface space and conformed volume space.

effigies:
@satra Tried updating the transform generated by tkregister via `M.dot(<tkreg_mat>)` (and `<tkreg_mat>.dot(M)`) and then ran `flirt -applyxfm`. Might have fixed the axes, but shifted the center by quite a lot in both cases. Tried setting the fourth column to `[0; 0; 0; 1]`, and that just zeroed the files. Did I misunderstand what you were suggesting, or was this not the magic we needed?

@chrisfilo Not sure if you're missing something, but I definitely am. I'll try to figure out ANTs in the morning.

I did try `mri_vol2vol` (`mri/orig/001.mgz` is the original T1 just converted to `mgz`):

```Shell
$ mri_vol2vol --mov /out/freesurfer/sub-01/mri/T1.mgz \
              --targ /out/freesurfer/sub-01/mri/orig/001.mgz \
              --regheader --o /scratch/reorig.mgz --keep-precision
movvol /out/freesurfer/sub-01/mri/T1.mgz
targvol /out/freesurfer/sub-01/mri/orig/001.mgz
outvol /scratch/reorig.mgz
invert 0
tal    0
talres 2
regheader 1
noresample 0
interp  trilinear (1)
precision  uchar (0)
Gdiag_no  -1
Synth      0
SynthSeed  1490402579

Computing registration based on scanner-to-scanner

Final tkRAS-to-tkRAS Matrix is:
-1.00000   0.00000   0.00000  -0.00002;
 0.00000   0.00000  -1.00000  -0.00003;
 0.00000   1.00000   0.00000  -0.00001;
 0.00000   0.00000   0.00000   1.00000;


Vox2Vox Matrix is:
-1.00000   0.00000   0.00000   208.00002;
 0.00000   0.00000  -1.33333   255.99997;
 0.00000   1.33333   0.00000  -0.00001;
 0.00000   0.00000   0.00000   1.00000;

Resampling
Output registration matrix is identity

mri_vol2vol done
```

In the reverse:

```Shell
Final tkRAS-to-tkRAS Matrix is:
-1.00000   0.00000   0.00000  -0.00002;
 0.00000   0.00000   1.00000  -0.00002;
 0.00000  -1.00000   0.00000  -0.00004;
 0.00000   0.00000   0.00000   1.00000;


Vox2Vox Matrix is:
-1.00000   0.00000   0.00000   207.99997;
 0.00000   0.00000   0.75000   0.00002;
 0.00000  -0.75000   0.00000   191.99997;
 0.00000   0.00000   0.00000   1.00000;
```

So looks like I have three new matrices to try in the morning.

satra:
@effigies - i don't think that magic matrix is completely correct, but may still be relevant. 

could you put the following files in a shared folder somewhere: raw T1, orig.mgz, flirt.mat, bbregister.mat, meanEPI (i'll take a quick look)

effigies:
@satra https://qnl-linux1.bu.edu/~effigies/registration.tar.gz

satra:
thanks @effigies. 

so the info you need is this (it would be good to know how doug computes the mat file below or the steps to conform a file - it can likely be replaced with a few lines of nibabel+numpy code):

`tkregister2 --mov T1.mgz --targ sub-01_T1w.nii.gz --regheader --reg junk --fslregout freesurfer2subT1.mat --noedit`

then do: 
`np.loadtxt('freesurfer2subT1.mat').dot(np.loadtxt('bbregister.mat')).savetxt('newregforflirt.mat')`

then apply with flirt with the ref being sub-01_T1w.nii.gz

and correspondingly if you do 

`np.linalg.inv(np.loadtxt('freesurfer2subT1.mat')).dot(np.loadtxt('flirtbbr.mat')).savetxt('newregforflirt.mat')`

and use T1.mgz as the ref, you should get to the conformed space.

effigies:
@satra Thank you! This was exactly what I needed.

In the above case, we get a `T1.mgz -> T1w.nii.gz` transform of:

```
   1.00000012    0.00000000    0.00000000  -49.00001144 
   0.00000000    0.00000000    1.00000012    0.00000763 
   0.00000000   -1.00000012    0.00000000  256.00000000 
   0.00000000    0.00000000    0.00000000    1.00000012 
```

An `EPI -> T1.mgz` transform of:

```
   0.99971902   -0.01112530   -0.02093013   33.86444092 
  -0.02336754   -0.61066228   -0.79154629  240.31718445 
  -0.00397511    0.79181224   -0.61075121   70.68538666 
   0.00000000    0.00000000    0.00000000    1.00000000 
```

Which we combine to get `EPI -> T1w.nii.gz`:

```
0.999719139966 -0.011125301335 -0.0209301325116 -15.1355664563
-0.00397511047701 0.791812335017 -0.61075128329 70.6854027722
0.0233675428041 0.610662353279 0.791546384986 15.6827867119
0 0 0 1
```

Compare to the FLIRT-BBR generated:

```
0.9997169368  -0.008484848857  -0.02222782656  -15.35086442  
-0.006684317239  0.7964569286  -0.6046583701  70.34123553  
0.02283393991  0.6046357384  0.7961747509  15.49617409  
0  0  0  1  
```

Freeview confirms:

applyxfm-mean_epi_satrad

effigies:
Latest commit updates the language on the reports. Ready for review, pending CI build to verify non-FreeSurfer runs produce unchanged reports.

effigies:
Docs updated. I decided to go with the approach of noting when FreeSurfer adds to or modifies a step in the base workflow, and then describing all surface-based processing in its own section. I suspect that will be smoother when fieldmaps go in, and it allows the FreeSurfer derivatives to be described without cluttering up the main derivatives section.

An alternative approach would be to include FreeSurfer stuff by default, and have a section describing the removals/replacements if `--no-freesurfer` is used.

chrisfilo:
there seems to be something wrong with the MNI preproc files:

this is the preproc bold in the MNI space overlaid on top of the atlas
image

this is the proproc bold in the T1w space overlaid on top of the T1
image

This is the normalized T1w overlaid on the MNI
image

ds005 sub-01

I'm gonna check if the same thing happens in master.

chrisfilo:
The same thing when using master with FLIRT BBR

image

PR #363, opened 2017-02-17
[RTM] Pin new nipype - updates ReconAll response to existing subjects

effigies:
(Minor) unexpected behavior: `stats` and `labels` directories were getting updated by repeat runs of ReconAll.

Once nipy/nipype#1824 is merged, will update pin and mark [RTM].


effigies:
Rebased, but should wait on Circle. If it errors, we'll know that there's an upstream bug that needs dealing with.

effigies:
Should be good to go.

PR #370, opened 2017-02-22
HMC refactor

### LINKED ISSUE #267: MCFLIRT is introducing slow frequency confound
image

https://fsl.fmrib.ox.ac.uk/fsl/fslwiki/MCFLIRT
> As part of the initial 8mm search, an identity transformation is assumed between the middle volume and the adjacent volume. The transformation found in this first search is then used as the estimate for the transformation between the middle volume and the volume beyond the adjacent one. This scheme should lead to much faster optimization and greater accuracy for the majority of studies where subject motion is minimal. In the pathalogical cases, this scheme does not penalise the quality of the final correction. 
> 
> If mean registration is used, the curent motion correction parameters are applied to the time-series, the volumes are averaged to create a new template image and the same 3-stage correction is carried out using this new mean image as a template.

chrisfilo:
- Closes #267 by performing interpolation using ANTs
- Outputs EPI in T1w space instead of the EPI space
- Use float32 when applying transforms (turns out they were calculated using float32 anyway)

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -34,8 +34,9 @@ dependencies:
 test:
   override:
     - docker run -ti --rm --entrypoint="python" poldracklab/fmriprep:latest -m unittest discover test
-    - docker run -ti --rm -v $HOME/nipype.cfg:/root/.nipype/nipype.cfg:ro -v $HOME/data:/data:ro -v $HOME/ds054/scratch:/scratch -v $HOME/ds054/out:/out poldracklab/fmriprep:latest /data/ds054 /out/ participant --no-freesurfer --debug -w /scratch:
-        timeout: 4800
+    # Disabling until fieldmaps refactor is ready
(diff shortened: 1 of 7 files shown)


chrisfilo:
I'm waiting for @rwblair to run it through the paces and confirm that it indeed fixes #267.

rwblair:
On ds054 style datasets we error out with the following:
```
Module ConfoundDiscoverer has no input called inputnode.t1_transform
Module ConfoundDiscoverer has no input called inputnode.reference_image
```
Those two inputs are referenced in base.py, and are not defined in confounds.py

chrisfilo:
Yeah - the refactor, breaks ds054 (that's why the tests were
commented out). We agreed with @oesteban to not try to adapt the ds054 and
leave it to his fieldmap branch.

Did you check that this PR indeed removes the trend in DVARS?

rwblair:
Its currently being rerun on sub-313 from the beast dataset, will review its output in the morning.

PR #376, opened 2017-02-28
[RTM] Accept T2w images in recon-all

effigies:
Starting to track progress on T2w incorporation.

Follow-up to #347.

### DIFF
--- a/fmriprep/workflows/anatomical.py
+++ b/fmriprep/workflows/anatomical.py
@@ -90,7 +90,9 @@ def t1w_preprocessing(name='t1w_preprocessing', settings=None):
             t1w_list = filename_to_list(t1w_list)
             t2w_list = filename_to_list(t2w_list) if isdefined(t2w_list) else []
             t1w_ref = nib.load(t1w_list[0])
-            hires = round(max(t1w_ref.header.get_zooms()), 1) < 1
+            # Use high resolution preprocessing if voxel size < 1.0mm
+            # Tolerance of 0.05mm requires that rounds down to 0.9mm or lower
+            hires = max(t1w_ref.header.get_zooms()) < 1 - 0.05
             t1w_outs = [t1w_list.pop(0)]
             for t1w in t1w_list:
                 img = nib.load(t1w)
--- a/fmriprep/workflows/anatomical.py
+++ b/fmriprep/workflows/anatomical.py
@@ -81,7 +81,7 @@ def t1w_preprocessing(name='t1w_preprocessing', settings=None):
     if settings['freesurfer']:
         nthreads = settings['nthreads']
 
-        def detect_inputs(t1w_list, t2w_list=[], default_flags=''):
+        def detect_inputs(t1w_list, t2w_list=[]):
             import os
             from nipype.interfaces.base import isdefined
             from nipype.utils.filemanip import filename_to_list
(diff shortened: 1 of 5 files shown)


effigies:
I've run a test on sub-s483.

Nipype log excerpts:
```
170301-02:39:34,493 workflow INFO:
	 Executing node Reconstruction in dir: /scratch/workflow_enumerator/s483/t1w_preprocessing/Reconstruction
170301-02:39:34,498 workflow INFO:
	 Running: recon-all -autorecon1 -i /data/sub-s483/ses-1/anat/sub-s483_T1w.nii.gz -i /data/sub-s483/ses-2/anat/sub-s483_T1w.nii.gz -T2 /data/sub-s483/ses-1/anat/sub-s483_T2w.nii.gz -noskullstrip -hires -openmp 8 -parallel -subjid sub-s483 -sd /out/freesurfer
```

```
170301-07:04:35,101 workflow INFO:
	 Executing node Reconstruction2 in dir: /scratch/workflow_enumerator/s483/t1w_preprocessing/Reconstruction2
170301-07:04:35,167 interface INFO:
	 resume recon-all : recon-all -all -noskullstrip -hires -openmp 8 -parallel -subjid sub-s483 -sd /out/freesurfer -T2pial -nomotioncor -notalairach -nonuintensitycor -nonormalization
```

`-T2` flag is being added correctly to `recon-all -autorecon1`, and `-T2pial` to `recon-all -all`.

Waiting to finish to verify that `-T2pial` runs correctly, but this should be in good shape.

I will note that for hi-res images the inflated surface looks a bit pointy in places, so we may want to reconsider adjusting `mris_inflate -n` in an expert file. I'm pretty sure the issue I was having with that before is that the output file needed to be marked as `needed` to avoid being deleted by `remove_unnecessary_outputs`.

chrisfilo:
It's freesurfer doing some form of skull stripping on the t2 you had to
take care off?

effigies:
It uses `brainmask.mgz`, which we're injecting from ANTs. See https://surfer.nmr.mgh.harvard.edu/fswiki/ReconAllTableStableV6.0#line-116.

effigies:
T2 coregisters fine. I want to try adjusting `mris_inflate`, because the inflated surface (with and without T2) is kind of pointy in places.

effigies:
Upping `mris_inflate -n` definitely makes the inflated surface much smoother, so I'm including that.

No `expert`:
no_expert

`expert`:
expert

This PR is waiting on nipy/nipype#1857. The `expert` option for `ReconAll` is needed to mark the `expert.opts` file as not to be deleted.

Going to do a clean run on Sherlock and report total time.

chrisfilo:
Interesting. Is there an impact on how reconstructed surfaces look like in
our HTML reports?

effigies:
So reports uses `?h.white` and `?h.pial`, and I don't *think* those are affected by `mris_inflate`. I believe `?h.sphere` is affected, which is used for subject-subject surface registration.

effigies:
So `?h.white` *does* depend on `?h.inflated` in the smoothing stages, as do `?h.sphere`. So this isn't purely aesthetic wrt the `?h.inflated` surfaces.

effigies:
I set this as RTM because it works. If you want to wait on Sherlock build, that's fine. (If I can't get Sherlock to work, I'll run on my laptop overnight and give you an 8-core estimate.)

oesteban:
We also should wait for https://github.com/nipy/nipype/pull/1857 to be merged, right?

effigies:
Oh right. Yes, somehow I forgot that. I'll mark it [WIP] again. Once a final pin is set, I'll set it back to RTM.

effigies:
`SLURM Job_id=13250524 Name=sub-s483 Ended, Run time 09:38:39, COMPLETED, ExitCode 0`

fmriprep/workflows/anatomical.py
```diff
@@ -81,35 +81,38 @@ def t1w_preprocessing(name='t1w_preprocessing', settings=None):
...
+            import os
+            from nipype.interfaces.base import isdefined
             from nipype.utils.filemanip import filename_to_list
+            from nipype.interfaces.traits_extension import Undefined
             import nibabel as nib
             t1w_list = filename_to_list(t1w_list)
+            t2w_list = filename_to_list(t2w_list) if isdefined(t2w_list) else []
             t1w_ref = nib.load(t1w_list[0])
-            hires = max(t1w_ref.header.get_zooms()) < 1
+            hires = round(max(t1w_ref.header.get_zooms()), 1) < 1
```
effigies: Woops. Lost the thread in the rebase. Recap:

@oesteban suggested we set an explicit threshold, rather than round.

```Python
hires = np.all(np.array(t1w_ref.header.get_zooms()) - 1 < -tol)
```

Where `tol` here `= 0.05`.

If we're going to do that, I'd prefer to stick with the `max` formulation, and keep constants on the right-hand-side, e.g.:

```Python
hires = max(t1w_ref.header.get_zooms()) < 1 - 0.05
```

I'll do whatever the consensus is, but I think the `round(max(zooms), 1) < 1` and `max(zooms) < 1 - tol` are the most readable options. Opinions, @chrisfilo?

chrisfilo:
I like the explicit threshold. Just make sure you cast everything to float!

effigies:
I don't see where casting is required... even if zooms are `int` (Nifti zooms are [`'f4'`](https://github.com/nipy/nibabel/blob/ec4567fb09b4472c5a4bb9a13dbcc9eb0a63d875/nibabel/analyze.py#L119), anyway), numerical comparisons aren't going to be affected by an explicit cast. Am I missing something?

chrisfilo:
LGTM

effigies:
Another run, following the nipype PR: `Run time 09:22:15`

oesteban:
LGTM too

chrisfilo:
+1

Issue #259, opened 2016-12-09
GM/WM outline is yellow instead of red for some ds005 subjects


chrisfilo:
![image](https://cloud.githubusercontent.com/assets/238759/21041115/e1e70b88-bd9d-11e6-899d-6f3a41bd8c46.png)
Definitely a minor issue. Mysterious non-the-less

chrisfilo:
Another example note that the middle row has a different color than the rest:
![image](https://cloud.githubusercontent.com/assets/238759/21414362/b9ed84ac-c7cc-11e6-8ea9-c7c0c525b259.png)

oesteban:
This issue should be in niworkflows :)

chrisfilo:
It might be an intersection between the two packages - we need to make sure
the mask fed to the interface is correct.

PR #378, opened 2017-03-01
[WIP] update nipype and niworkflows versions

chrisfilo:


Dockerfile
```diff
@@ -89,25 +89,23 @@ RUN apt-get install -y nodejs
...
-RUN conda install -y mkl=2017.0.1 \
-                     numpy=1.11.2 \
+RUN conda install -y mkl=2017.0.1 mkl-service &&  \
+    conda install -y numpy=1.12.0 \
                      scipy=0.18.1 \
-                     scikit-learn=0.17.1 \
-                     matplotlib=1.5.3 \
-                     pandas=0.19.0 \
+                     scikit-learn=0.18.1 \
+                     matplotlib=2.0.0 \
```
effigies: Main effect here seems to be the white matter boundary for EPI registration is now lime green.

chrisfilo: I think this is related to https://github.com/poldracklab/fmriprep/issues/259

PR #381, opened 2017-03-02
[RTM] Don't use editable flags.

chrisfilo:
It turns out that "-e" makes packages appear in sys.path in front of whatever is in PYTHONPATH making it impossible to override installs in singualrity.

This will change installation path and require changes to #317


effigies:
Ah, that explains why I wasn't able to override in singularity. I'm not super familiar with other consequences here. @rwblair? Will this screw up docker patching?

oesteban:
We'll need to modify the mount point for docker patching, or provide hooks (softlink the fmriprep folder under sites-packages to somewhere or from somewhere easy to remember)

Issue #268, opened 2016-12-11
README in `derivatives` folder


shoshber:
According to BIDS spec: http://bids.neuroimaging.io/bids_spec1.0.0-rc4.pdf

rwblair:
A ​README​file should be found at the root of the “s​ourcedata”​or the “​derivatives”​folder (or both). This file should describe the nature of the raw data or the derived data. In the case of the existence of a “​derivatives”​folder, we recommend to include details about the software stack and settings used to generate the results. Non-imaging objects to improve reproducibility are encouraged (scripts, setting files, etc.).

Wonder what the best way of dumping all our software used and versions to a file should be? Maybe we should just have a link to the commit that is used on github.

oesteban:
Superseded by latest BEP advances.

PR #391, opened 2017-03-14
[RTM] Surface sample

### LINKED ISSUE #355: Surface-mapped functional time series
Following on #347, we're going to map functional time series onto surfaces. In FreeSurfer, the standard way to do this is through `preproc-sess`, which is kind of like `recon-all` for the functional bits. This is just starting some notes before there's any code to comment on.

I've initiated nipy/nipype#1805 to get this going. This will make sampling to the subject-native and template spaces (or any subject in the subjects directroy) straightforward.

On this end, we'll need to populate an FSFAST directory structure (unless this should be in Nipype?). Since images are already motion-corrected, we can probably skip that step. BBR is part of the preproc-sess pipeline, so we can use that, unless we don't want to force people to go through the FSFAST process for that. Are we going to want to do any slice-time correction or smoothing?

effigies:
Building on #362, this uses the registration produced by `bbregister` to sample volumes to the subject-native and `fsaverage` surfaces.

Looks like this will be fairly simple, but `mri_vol2surf` is pretty flexible in its sampling strategy. For each vertex on the surface, there is a normal vector from the white matter surface to the pial surface, and you can choose a point along that vector (in mm or as a fraction of the cortical thickness from 0-1), the average value along a range (start, stop, step; again in mm or fraction) or the maximum value on a range.

As a starter, I've set it to sample the midpoint of the normal vectors.

Looks like this'll be the thing that finally makes me learn to use `JoinNode`.

Closes #355.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,6 +3,7 @@ Next release
 
 * [ENH] Generate GrayWhite, Pial, MidThickness and inflated surfaces (#398)
 * [ENH] Memory and performance improvements for calculating the EPI reference (#436)
+* [ENH] Sample functional series to subject and ``fsaverage`` surfaces (#391)
 
 0.3.2 (7th of April 2017)
 =========================
--- a/docs/workflows.rst
+++ b/docs/workflows.rst
@@ -259,6 +259,29 @@ It also maps the T1w-based mask to MNI space.
 Transforms are concatenated and applied all at once, with one interpolation (Lanczos)
 step, so as little information is lost as possible.
 
+EPI sampled to FreeSurfer surfaces
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+:mod:`fmriprep.workflows.epi.epi_surf_sample`
+
+.. workflow::
+    :graph2use: colored
+    :simple_form: yes
+
+    from fmriprep.workflows.epi import epi_surf_sample
+    wf = epi_surf_sample("test",
+                         settings={'output_dir': '.',
+                                   'skip_native': False,
+                                   })
+
+If FreeSurfer processing is enabled, the motion-corrected functional series is sampled to the
+surface by averaging across the cortical ribbon.
+Specifically, at each vertex, the segment normal to the white-matter surface, extending to the pial
+surface, is sampled at 6 intervals and averaged.
+
+Surfaces are generated for the "subject native" surface, as well as transformed to the
+``fsaverage`` template space.
+All surface outputs are in GIFTI format.
+
 Confounds estimation
 ~~~~~~~~~~~~~~~~~~~~
 :mod:`fmriprep.workflows.confounds.discover_wf`
@@ -326,6 +349,8 @@ Derivatives related to EPI files are in the ``func`` subfolder:
 - ``*bold_confounds.tsv`` A tab-separated value file with one column per calculated confound and one row per timepoint/volume
 - ``*bold_space-T1w_preproc.nii.gz`` Motion-corrected (using MCFLIRT for estimation and ANTs for interpolation) EPI file in T1w space
 - ``*bold_space-MNI152NLin2009cAsym_preproc.nii.gz`` Same as above, but in MNI space
+- ``*bold_space-fsnative.[LR].func.gii`` Motion-corrected EPI file sampled to subject's "native" FreeSurfer surfaces
+- ``*bold_space-fsaverage.[LR].func.gii`` Same as above, but in FreeSurfer ``fsaverage`` template space
 
 
 FreeSurfer Derivatives
--- a/fmriprep/workflows/base.py
+++ b/fmriprep/workflows/base.py
@@ -70,7 +70,6 @@ def base_workflow_generator(subject_id, task_id, settings):
         raise Exception("No T1w images found for participant {}. "
                         "All workflows require T1w images.".format(subject_id))
 
-
     if all((subject_data['fmap'] != [],
             subject_data['sbref'] != [],
             "fieldmaps" not in settings['ignore'])):
--- a/fmriprep/workflows/epi.py
+++ b/fmriprep/workflows/epi.py
@@ -542,6 +542,7 @@ def epi_surf_sample(name='SurfaceSample', settings=None):
     sampler = pe.MapNode(
         fs.SampleToSurface(sampling_method='average', sampling_range=(0, 1, 0.2),
                            sampling_units='frac', reg_header=True,
+                           interp_method='trilinear', cortex_mask=True,
                            out_type='gii'),
         iterfield=['source_file', 'target_subject'],
         iterables=('hemi', ['lh', 'rh']),
--- a/fmriprep/workflows/epi.py
+++ b/fmriprep/workflows/epi.py
@@ -495,16 +495,30 @@ def epi_sbref_registration(settings, name='EPI_SBrefRegistration'):
 
 
 def epi_surf_sample(name='SurfaceSample', settings=None):
+    """ Sample functional images to FreeSurfer surfaces
+
+    For each vertex, the cortical ribbon is sampled at six points (spaced 20% of thickness apart)
+    and averaged.
+
+    Outputs are in GIFTI format.
+
+    Settings used:
+        skip_native : sample only to fsaverage space, skipping subject native surface
+        output_dir : directory to save derivatives to
+    """
     workflow = pe.Workflow(name=name)
     inputnode = pe.Node(
-        niu.IdentityInterface(fields=[
-            'source_file',
-            'reg_file',
-            'subjects_dir',
-            'name_source',
-        ]), name='inputnode')
+        niu.IdentityInterface(fields=['source_file', 'reg_file', 'subjects_dir', 'name_source']),
+        name='inputnode')
 
     def select_targets(reg_file, template, skip_native):
+        """ Select targets based on a provided template and whether to generate outputs in
+        native space.
+
+        Source target is defined from registration file to preclude mismatches
+
+        Returns targets (FreeSurfer subject names) and spaces (FMRIPREP output space names)
+        """
         targets = [template]
         spaces = [template]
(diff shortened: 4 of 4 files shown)


effigies:
So we have generated (in #362) a FreeSurfer `.dat` that goes from the mean EPI (outputted by [`explicit_mask_epi`](https://github.com/poldracklab/fmriprep/blob/ea45a6e/fmriprep/workflows/epi.py#L113)) to FreeSurfer-conformed `T1.mgz` space.

Am I correct in reading that we never actually save motion-corrected EPIs in native space? From what I can see, transforms are accumulated and include a template normalization prior to application. If this is the case, what would be the best point to insert into existing workflows (ideally general enough to plug into both `basic_wf` and `basic_fmap_sbref_wf`)?

effigies:
Proposed derivatives target (not yet coded):

```
fmriprep/
    func/
        lh.sub-<subject_label>_task-<task_label>_run-<run_index>_bold_space-fsaverage_preproc.gii
        lh.sub-<subject_label>_task-<task_label>_run-<run_index>_bold_space-fsnative_preproc.gii
        rh.sub-<subject_label>_task-<task_label>_run-<run_index>_bold_space-fsaverage_preproc.gii
        rh.sub-<subject_label>_task-<task_label>_run-<run_index>_bold_space-fsnative_preproc.gii
```

Question is mainly how to describe the space. `fsaverage` seems straightforward. `fsnative` would be an indicator that the subject is `sub-<subject_label>`. Should this also reflect the sampling technique (point, average, max)? And should that be in space or a new key/value?

And as to sampling technique, FreeSurfer defaults to `nearest-neighbor` interpolation, though also provides trilinear.

We can save the sampling/interpolation parameters as JSON. It might make more sense than trying to encode these concepts in the filename.

chrisfilo:
The file naming scheme looks good to me. In terms of space naming I wonder if we should be more explicit about the resolution of the space we are using - aren't there multiple versions of fsaverage with different number of vertices? I don't know what is the current naming convention in HCP/FreeSurfer world for those.

I would prefer using trilinear interpolation instead of nearest neighbours.

The details of interpolation and samplint technique should be explained in the documentation (and hopefully one day in README - see #268), but I would leave them out of the filename unless we want to provide multiple versions (but I don't see the need for that now).

BTW I checked and FSL's film_gls can take gifti files: https://github.com/Washington-University/Pipelines/blob/master/TaskfMRIAnalysis/scripts/TaskfMRILevel1.v1.0.sh#L156, but I'm not sure if FreeSurfers mri_glmfit can https://surfer.nmr.mgh.harvard.edu/fswiki/mri_glmfit

effigies:
It would probably be good to figure out the HCP/FS naming conventions for README or parameter recordings, but I think that at the moment  the "which fsaverage" question is answered: "The fsaverage in the `derivatives/freesurfer/fsaverage` directory."

As an aside, I do think it would be good to be able to choose an alternative template space, though, but it should be also in the `derivatives/freesurfer` directory, and I'm writing with the assumption that we'll at some point want to make a non-`fsaverage` template.

So wrt GIFTI, `mri_glmfit` can't take it directly... it needs surfaces concatenated into a "volume"-based file, and provides `mris_preproc` for this purpose. I'll try to pin down whether that can take GIFTI, but I have a strong suspicion it can, since `mri_vol2surf` and `mri_surf2surf` work facilely with .gii files.

chrisfilo:
BTW I'm looking through the standard: https://www.nitrc.org/frs/download.php/2871/GIFTI_Surface_Format.pdf

and it seems we should follow the:

```
fmriprep/
    func/
        sub-<subject_label>_task-<task_label>_run-<run_index>_bold_space-fsaverage_preproc.L.time..gii
        sub-<subject_label>_task-<task_label>_run-<run_index>_bold_space-fsnative_preproc.R.time.gii
        sub-<subject_label>_task-<task_label>_run-<run_index>_bold_space-fsaverage_preproc.L.time.gii
        sub-<subject_label>_task-<task_label>_run-<run_index>_bold_space-fsnative_preproc.R.time.gii
```
convention.

It's also worth checking if Connectome workbench can read those files and does not complain about any metadata (such as Intent is set to NIFTI_INTENT_TIME_SERIES, AnatomicalStructurePrimary is set to CortexLeft and CortexRight, TimeStep is set to correct TR )

effigies:
Marked [RTM] for review. Preparing to run ds114 on Sherlock to get FreeSurfer reconstructions to test that a surface-based GLM gives sensible outputs.

chrisfilo:
Could you mention the new derivatives and a short description of the sampling strategy in the docs?

effigies:
Was just working on that.

chrisfilo:
This LGTM assuming the ds114 models fits look reasonable.

effigies:
At least for `film_gls`, it looks like we'd need subsampled surfaces and functionals to keep it from taking a very long time: https://www.jiscmail.ac.uk/cgi-bin/webadmin?A2=ind1404&L=FSL&F=&S=&P=345205 (Also, nipype needs updating for `film_gls` in surface mode.)

I'm going to test with FreeSurfer's GLM.

chrisfilo:
Should we output the surface timeseries on a subsampled mesh?

effigies:
I think it makes sense to go ahead and continue producing at least the `fsaverage` datasets as we are, because a FreeSurfer-based GLM would typically be in that space that. But I'm not sure how much value there is in producing `fsnative` time series on the high-res mesh.

Maybe what makes the most sense is to have functionals in three spaces:

* `fsaverage` for FreeSurfer analyses
* Sub-sampled `fsaverage` and `fsnative` for `film_gls`-based analyses

So we'd need to output the sub-sampled GIFTI meshes. Possibly in place of the meshes we're currently producing?

chrisfilo:
Is there an advantage of using full resolution `fsaverage`? Why is it ok to use it with freesurfer but not with `film_gls`?

effigies:
The advantage of using full-resolution `fsaverage` is that I don't have to figure out how to make my data work with the `fsaverage` space. It just works like usual, if I'm doing analyses in FreeSurfer.

I assume the difference is that FreeSurfer was designed to work with high-res meshes, while surface analysis was hacked into `film_gls`? Or possibly FSL users are used to faster analyses from working in the volume. I'm not really clear on how much time the FSL thing should take. I expect a FreeSurfer analysis to take a few hours.

chrisfilo:
So this really boils down to the question: "If I want to use an atlas, parcellation or segmentation do I need to downsample it first?". For our volume outputs (both native and MNI) the answer is "yes". This is because we downsample the outputs to the native resolution to save space. From a consistency point of view it would make sense to do the same for surfaces (both native and fsaverage).

Does freesurfer GLM even care about the space (other making sure it's consistent across participants)? How much space would we save by downsampling?

BTW did we figure out if freesurfer glm can take gifti as input?

chrisfilo:
This could be usful:
https://brainder.org/2016/05/31/downsampling-decimating-a-brain-surface/

effigies:
In that case, I think we also need to distribute the down-sampled fsaverage surface so people can visualize results using FS tools. If it's a non-deterministic process (I'd suspect it is), we might want to piggy-back on HCP's work so that functionals sampled in FMRIPREP and HCP are comparable.

effigies:
I'm working on figuring out how to shoehorn externally preprocessed data into a FreeSurfer GLM, now.

chrisfilo:
Good point about visualisation - I forgot that surface viewers in contrast to volume ones do not do resampling on the fly. Not sure what is the best compromise between space saving and ease of use. Have you looked into the Conte69 and 32k_fs_LR spaces?

effigies:
Not in depth. 32k_fs_LR was on my radar, but I'll also have a look at Conte69.

effigies:
FSFAST-preprocessed EPI, mapped to `fsaverage` surface:

fsfast_func_surf

FMRIPREP-preprocessed EPI, mapped to `fsaverage surface:

fmriprep_func_surf

These two images have the exact same color scaling, so we're clearly getting something very close.

I realized that using the `bbregister`-made registration file was double-dipping, since we apply it to our functional data. In FSFAST land, the volumes are not moved; the registration is calculated and then applied at the time of sampling.

Getting closer to a working GLM.

chrisfilo:
Reply from a user:

> Regarding FMRIPREP: gifti format would be perfect, and I would suggest making a few resolution options available. We tended to go with fsaverage5, but the HCP 32k surface could also be quite useful...

effigies:
These are level 1 GLM results for `ds000114/sub-01/ses-test`.

FSFAST (from BIDS source functionals):
level1_fsfast

FMRIPREP surfaces -> FSFAST:
level1_fmriprep

So I think we're in pretty good shape here. Not only are we 

I want to write up some notes on what I actually had to do to get this to run, because it was non-trivial. I started with [Doug's advice](https://mail.nmr.mgh.harvard.edu/pipermail/freesurfer/2017-April/051270.html), but had to also create a global mean and produce an AFNI-ish motion parameter file and hack `selxavg3-sess` to be a little flexible in that regard. (These aren't the notes. They're just off the top of my head.)

Talked with @chrisfilo and will be emailing the list to get feedback on what would be useful output to them in terms of formats and "stage" of preprocessing. I can also, probably in a separate thread, discuss the issues I had with injecting pre-computed surface datasets into FSFAST. It would be nice if their pipeline were a little more modular; though I suppose only people trying to go from FMRIPREP to FSFAST would think so.

Remaining work to close this issue:

- [ ] Find out preferred sampling surfaces (should this be an option, similar to `--skip-native`, *e.g.* `--conte69`? )
- [ ] Double check data types in GIFTI files... Was getting some noisy output from `wb_view` and `film_gls`, so it could be the FreeSurfer-generated GIFTIs need a little cleaning here, too.

effigies:
`wb_view` was complaining about an absence of `AnatomicalStructurePrimary` field in the metadata, so the latest commit cleans that up. Re-ran `ds000114/sub-01/ses-test`, and the resulting file produces no complaints in `wb_view`.

PR #396, opened 2017-03-18
[RTM] Switch to composite transforms.

chrisfilo:
As indicated in https://github.com/poldracklab/fmriprep/pull/362#issuecomment-287490389 we have been applying affine and warp transforms in the wrong order causing slight misalignment. To avoid such problems in the future I propose to switching to composite transforms.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -7,6 +7,7 @@ What's new
 0.3.0 (upcoming)
 ================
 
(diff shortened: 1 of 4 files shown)


chrisfilo:
I have to say I'm surprised how little impact this bug has had on the outputs.

effigies:
The code looks reasonable. I'll run this one when I get a minute. In the meantime, do you have images for this, like the ones you posted on the Slack channel?

PR #408, opened 2017-03-24
FIX: Reference fs_transform to the reoriented T1, not original

effigies:
This may be the issue with Beast. The most likely reason some datasets were working is that `nibabel.as_closest_canonical` didn't change anything (or changed little enough).

### DIFF
--- a/fmriprep/workflows/anatomical.py
+++ b/fmriprep/workflows/anatomical.py
@@ -278,7 +278,7 @@ def t1w_preprocessing(name='t1w_preprocessing', settings=None):
             (injector, reconall, [('subjects_dir', 'subjects_dir'),
                                   ('subject_id', 'subject_id')]),
             (recon_config, reconall, [('use_T2', 'use_T2')]),
-            (t1wmrg, fs_transform, [('out_avg', 'target_image')]),
+            (arw, fs_transform, [('out_file', 'target_image')]),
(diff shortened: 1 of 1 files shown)


effigies:
Beast sub-387 (fixed)
------
new_reg_target

ds000005 sub-01 (unchanged)
-------
ds000005_check

chrisfilo:
There is no L/R flip 
image

I drew the ROI on the original data on the right hand side

effigies:
So yes, it was the conformation step that I was failing to account for, and some subjects simply didn't have a changed conformation.

Although this seems to have resolved the issue, as discussed on Slack, an alternative approach would be to replace steps 0 and 1 (merge T1s and conform) with FreeSurfer's autorecon1, and move the entire FMRIPREP pipeline into the FreeSurfer-conformed space.

This has the advantage of eliminating the need to transform between the nibabel-conformed space and the FreeSurfer-conformed space. The biggest disadvantage (if it is one) is that FreeSurfer reslices images that are not 1mm-isotropic.

chrisfilo:
If this is the easiest way to align surfaces in FreeView and Connectome
Workbench I would opt for that in the short term. Seems less error prone. I
would leave it to a separate PR though.

Let's merge this one and mint a new bugfix release.

PR #410, opened 2017-03-25
[RTM] Preliminary PR to gradually merge fieldmaps

oesteban:
Mostly, sorts out documentation.


fmriprep/info.py
```diff
@@ -38,7 +38,8 @@
     'pybids>=0.1.0',
     'nitime',
     'niworkflows',
-    'nipype'
+    'nipype',
+    'statsmodels'
```
chrisfilo: It doesn't seem you ever use this module.

oesteban: I'll add it in the next iteration. But I can add it back then.

docs/workflows.rst
```diff
@@ -95,7 +95,7 @@ EPIMNITransformation
     :scale: 100%
 
 The EPIMNITransformation sub-workflow uses the transform from
-`ref_epi_t1_registration`_ and a T1-to-MNI transform from `t1w_preprocessing`_ to
+`EPIMeanNormalization`_ and a T1-to-MNI transform from `t1w_preprocessing`_ to
```
chrisfilo: Let's carefully inspect the docs that will get built on Circle. I expect this line to be problematic since `EPIMeanNormalization` does not exist.

chrisfilo:
I have to wait for the docs to built on Circle to give you more feedback.

docs/workflows.rst
```diff
@@ -229,3 +229,12 @@ A FreeSurfer subjects directory is created in ``<output dir>/freesurfer``.
 
 A copy of the ``fsaverage`` subject distributed with the running version of
 FreeSurfer is copied into this subjects directory.
+
+
+Susceptibility Distortion Correction (SDC)
+==========================================
```
chrisfilo: I am skeptical about including SDC documentation without actually fixing SDC. Here's my reasoning: imagine we will merge this now and next week (before you will have a chance to submit a PR fixing SDC) we find a bug and have to release 0.3.2. It will include this documentation misleading users about the inclusion of SDC functionality (when in reality it will still remain broken and unusable).

chrisfilo:
The new command line presentation is a bit funky:
image

and sometimes confusing:
image

https://133-53608443-gh.circle-artifacts.com/0/home/ubuntu/_build_html/usage.html

oesteban:
Just pushed some changes documenting many arguments that didn't have a `help` metadata. Also reorganized the groups to improve organization and readability.

As regards the version funkyness, I'm afraid that is the way the sphinx extension takes it. Since it is a special argument, it shows that ugly `==SUPRESS==`. I can open an issue in their repo asking for a better message (something like: `<current fmriprep version>`) but I doubt it is worth the effort.

I would encourage using this sphinx extension because:
- We don't need to manually execute the help command when building documentation (a real failure point)
- We don't maintain the arguments description anymore: if argparse evolve, they will add the improvements.

oesteban:
Working on some refinements, don't merge yet.

oesteban:
Should be fine now

fmriprep/cli/run.py
```diff
@@ -109,7 +125,7 @@ def create_workflow(opts):
         'ignore': opts.ignore,
         'skip_native': opts.skip_native,
         'freesurfer': opts.freesurfer,
-        'reportlets_dir': op.join(op.abspath(opts.work_dir), 'reportlets'),
+        'reportlets_dir': op.join(op.abspath(opts.work_dir), 'reportlets')
```
effigies: Is this an intentional change?

I typically like having the comma there, so that adding a new line only puts one line changed in the diff, not two.

fmriprep/cli/run.py
```diff
@@ -20,75 +17,94 @@
...
     # IMPORTANT: they must go directly with the parser object
-    parser.add_argument('bids_dir', action='store', default=os.getcwd())
+    parser.add_argument('bids_dir', action='store',
+                        help='the root folder of a BIDS valid dataset (sub-XXXXX folders should '
+                             'be found at the top level in this folder).')
     parser.add_argument('output_dir', action='store',
-                        default=op.join(os.getcwd(), 'out'))
-    parser.add_argument('analysis_level', choices=['participant'])
+                        help='the output path for the outcomes of preprocessing and visual '
+                             'reports')
```
effigies: Maybe we should recommend that it be a `derivatives/` folder?

fmriprep/cli/run.py
```diff
@@ -20,75 +17,94 @@
...
-    g_input.add_argument('-s', '--session-id', action='store', default='single_session')
-    g_input.add_argument('-r', '--run-id', action='store', default='single_run')
-    g_input.add_argument('--task-id', help='limit the analysis only ot one task', action='store')
-    g_input.add_argument('-d', '--data-type', action='store', choices=['anat', 'func'])
-    g_input.add_argument('--debug', action='store_true', default=False,
+    parser.add_argument('-v', '--version', action='version', version=verstr)
+
+    g_bids = parser.add_argument_group('Options for filtering BIDS queries')
+    g_bids.add_argument('--participant_label', action='store', nargs='+',
+                        help='one or more participant identifiers (the sub- prefix can be '
```
effigies: s/can/should/

I don't think we remove `sub-` and some things assume it's not there. (Unless you've added sanitization and I haven't gotten there yet.)

chrisfilo: I never use `sub-` and things seem to work.

oesteban: Adding `sub-` prunning...

fmriprep/cli/run.py
oesteban: Makes sense

fmriprep/cli/run.py
oesteban: I leave this for the documentation, no problem on my end.

fmriprep/cli/run.py
oesteban: here https://github.com/oesteban/fmriprep/blob/c177f6969b955b401a77e3a42674d99b278a182f/fmriprep/cli/run.py#L176

oesteban: @chrisfilo I think Chris was referring to that: it probably failed before if you tried with `sub-`

oesteban:
thumbs up?

chrisfilo:
Is there a way to disable displaying the default values? Right now they are
a bit confusing and would require cleaning up before merging.

oesteban:
You got it. Just added an option to the sphinx directive.

oesteban:
https://145-53608443-gh.circle-artifacts.com/0/home/ubuntu/_build_html/usage.html

PR #411, opened 2017-03-25
[ENH/WIP] Fieldmap estimation workflows

oesteban:
~~~Merge after #410~~~

(EDITED)

CHANGES:

  - Refactor `fmriprep.workflows.fieldmap` (fieldmap-related workflows), that now have the following modules:
    - `phdiff` -- estimates the fieldmap from the phase-difference image of two GRE acquisitions (*traditional* siemens fieldmaps).
    - `fmap` -- estimates the fieldmap from spiral sequences (like in the BEAST).
    - `pepolar` -- PE-POLARity images (also known as TOPUP).
    - `unwarp` -- gets the outcomes of any of the three previous modules and applies it to the input image.
- Extends both workflows (base, and sbrefs) to estimate fieldmaps in case they are found and not ignored.
- As a pilot feature, it corrects ONLY the EPI images after head motion correction, and appends the deformation field to the list of transforms used in resampling.
- This PR involves many changes to the utilities found in `fmriprep.interfaces`. Particularly, adds a new `fmriprep.interfaces.nilearn` that could be sent to nipype.
- Updates the documentation accordingly.


chrisfilo:
To reiterate what we discussed in person. The first iteration of fieldmap support should:
- not depend on or require sbrefs
- not modify the motion correction procedure (apply the same warp field to all volumes)
- support spiral fieldmaps (GE style)

We agreed tit will be easier to tackle sbrefs, unwarping before motion correction and other types of fieldmaps in separate subsequent PRs. Please correct me if I didn't get something right.

oesteban:
That's correct

oesteban:
Ok, seems like this is flowing. Final steps involve checking that the connections make sense (order of transforms, etc), adding the reportlets to the final report, and making sure the derivatives are written in the right place.

chrisfilo:
Cool! Let me know what should I have a look at it. Also - could you share the reports from the beast tests using this branch somewhere?

chrisfilo:
BTW if you are adding conditional fieldmap support to `basic_wf` I am happy to completely drop (remove) `basic_fmap_sbref_wf` if it makes things easier for you. It does not work anyway in master and we can bring back sbrefs in a new PR (in a new more thought through way).

Also if you could avoid including interfaces and function in this PR that you don't use it would make code review easier (for example I don't see `RASReorient` being used anywhere).

Glad to see things are progressing!

oesteban:
Yes, I'm cherry picking files from the fieldmaps branch, so some extra code is coming along with the minimal code necessary. I'll have a review for all this code.

Particularly, that interface should be replaced by the Conform interface in nipype, if I manage to unlock it (I was stuck with the affines to generate the desired output orientation last time I had a chance to take on it).

Ref. SBRefs, I think it does not hurt having the code there. I don't think we should implement the fieldmaps with SBRefs in this PR (to keep the focus as we planned), but right after I can start with that.

chrisfilo:
Just to clarify - what, do you mean by "I'll have a review for all this code". Are you insisting on keeping the unused Interfaces and function in this PR?

oesteban:
On the contrary. When it is finished, I'll check that all the unused interfaces are removed.

fmriprep/workflows/fieldmap/fmap.py
```diff
@@ -0,0 +1,90 @@
...
+                      name='FieldmapFuse')
+
+    # de-gradient the fields ("bias/illumination artifact")
+    n4 = pe.Node(ants.N4BiasFieldCorrection(dimension=3), name='MagnitudeBias')
+    cphdr = pe.Node(CopyHeader(), name='FixHDR')
+    bet = pe.Node(BETRPT(generate_report=True, frac=0.6, mask=True),
+                  name='MagnitudeBET')
+    fmapenh = pe.Node(FieldEnhance(
+        # despike_threshold=1.0, mask_erode=1),
+        despike=False, njobs=settings.get('ants_nthreads', 4)), name='FieldmapMassage')
```
chrisfilo: Please set the interface level num_threads so MultiProc would be aware of it.

fmriprep/workflows/base.py
```diff
@@ -237,41 +250,72 @@ def basic_wf(subject_data, settings, name='fMRI_prep'):
...
+                                ('outputnode.fmap_mask', 'inputnode.fmap_mask')]),
+            (unwarp, epi_2_t1, [('outputnode.out_reference', 'inputnode.ref_epi'),
+                                ('outputnode.out_warps', 'inputnode.hmc_xforms'),
+                                ('outputnode.out_mask', 'inputnode.ref_epi_mask')]),
+            (unwarp, epi_mni_trans_wf, [('outputnode.out_warps', 'inputnode.hmc_xforms'),
+                                        ('outputnode.out_mask', 'inputnode.epi_mask')])
+        ])
+
+        # Report on EPI correction
+        epireport = epi_preproc_report(settings=settings)
```
chrisfilo: This seems like a short workflow that is newer reused. All other reportlet datasinks were defined in the main workflow directly - shouldn't we do the same for consistency?

oesteban: I will reuse this one for the sbref, when we extend the correction to them.

chrisfilo:
- [ ] plot the unwarped EPI mask instead of the original one (and make sure there is no funny business going on)
- [ ] apply the mode of the Jacobian to correct for intensity changes (BTW check how HCP is dealing with this)
- [ ] fix s495 coregistration
- [ ] s548 is overcorrecting (low priority)
- [ ] set DOF for both FLIRT and bbregister as well
- [ ] turn fieldmaps off by default

oesteban:
I'm trying to address all these remaining issues:

- set DOF for both FLIRT and bbregister: turns out it was set for FLIRT only, because bbregister forces 6 dof (even though nipype has the `dof` option implemented, may not be documented?). The code is there, but commented out, until we decide what to do.

- the `SDC_unwarp` workflow now has an `out_jacobian` output that can be used for the modulation of drop-out. I'm not sure (yet) if we should use the inverse of this.

- I'm re-running on uh2 to check on s495 and s548.

- I've checked the mask and seems OK. So I should replace the current skullstripping report by the corresponding one with the corrected EPI and the mask?

chrisfilo:
On Mon, Apr 10, 2017 at 5:18 PM, Oscar Esteban <notifications@github.com>
wrote:

> I'm trying to address all these remaining issues:
>
>    -
>
>    set DOF for both FLIRT and bbregister: turns out it was set for FLIRT
>    only, because bbregister forces 6 dof (even though nipype has the dof
>    option implemented, may not be documented?). The code is there, but
>    commented out, until we decide what to do.
>
> Let's use that option even though its not well documented.

>
>    -
>
>    the SDC_unwarp workflow now has an out_jacobian output that can be
>    used for the modulation of drop-out. I'm not sure (yet) if we should use
>    the inverse of this.
>
> We can leave it for another PR.

>
>    -
>
>    I'm re-running on uh2 to check on s495 and s548.
>    -
>
>    I've checked the mask and seems OK. So I should replace the current
>    skullstripping report by the corresponding one with the corrected EPI and
>    the mask?
>
> This makes more sense since this is the mask we are using. Is the pre
uwarping brain mask used for anything?

oesteban:
Right now, I produce a mask that is the union of the brain mask from the magnitude image (supposedly anatomically correct) and the mask that nilearn obtains on the uncorrected data.

I will change this (so it uses the mask nilearn computes after correction). The idea of making the union of the two masks is adding to the brain mask regions that were not totally corrected.

This is the mask we should use after susceptibility correction and on.

So, should I replace the skull-stripping report?. Instead, the corrected reference with the mask contour would be plotted. WDYT?

chrisfilo:
Yes - since this is the final mask that will be used for EPI-T1
coregistration as well as saved in derivatives it makes sense to plot it in
the reports.

PR #415, opened 2017-03-28
[RTM] slice timing correction

### LINKED ISSUE #27: Slice timing correction
- only when metadata are available
- consider using http://nipype.readthedocs.io/en/latest/interfaces/generated/nipype.interfaces.nipy.preprocess.html#spacetimerealigner (simultaneous realignment and slice time correction)

chrisfilo:
* Adds slice time correction if metadata preset (can be disabled, closes #27)
* Refactors the basic workflow to allow for different workflows per each bold file depending on metadata
* Removes redundant avscale (it was producing the same results as mcflirts' .par file)

This implementation uses a two step approach using AFNI. There were big discrepancies in estimated motion between nipy and FSL so I decided to avoid 4dRealign until this was figured out.


fmriprep/workflows/base.py
```diff
@@ -217,60 +219,118 @@ def basic_wf(subject_data, settings, name='fMRI_prep'):
...
+                 [('subjects_dir', 'inputnode.subjects_dir')]),
+                (t1w_pre, bold_pre,
+                 [('outputnode.subject_id', 'inputnode.subject_id'),
+                  ('outputnode.fs_2_t1_transform', 'inputnode.fs_2_t1_transform')]),
+            ])
+
+    return workflow
+
+
+def bold_preprocessing(name, metadata, settings):
```
oesteban: Can we deprecate the `fmriprep.workflows.epi` module in favor of `fmriprep.workflows.bold` module which would contain this `bold_preprocessing` workflow?

oesteban: BTW, I like the idea of splitting the preprocessing bit very much

fmriprep/workflows/confounds.py
```diff
@@ -39,7 +38,7 @@ def discover_wf(settings, name="ConfoundDiscoverer"):
     dvars.interface.estimated_memory_gb = settings[
                                               "biggest_epi_file_size_gb"] * 3
     # Frame displacement
-    frame_displace = pe.Node(confounds.FramewiseDisplacement(parameter_source="FSL"),
+    frame_displace = pe.Node(confounds.FramewiseDisplacement(parameter_source="SPM"),
```
oesteban: Just to confirm: this parameter source is compatible with ITK matrices right?

oesteban: Actually FSL? Are we sure angles are in radians?

chrisfilo: This interface is never fed anything that used to be ITK matrices - so the answer is no.

"FSL" and "SPM" differ only in order of parameters - it should not influence the FD calculations, but since we are conforming all motion parameters to SPM style I did this for consistency.

chrisfilo:
I'm still planning on adding some comments and do some refactoring before this is ready for review.

I also need to verify everything is ok with the motion parameters.

chrisfilo:
The motion parameters are roughly similar, but not identical. Not sure how much difference I should expect judging how different this procedure is. This is FD

image

fmriprep/workflows/epi.py
```diff
@@ -248,7 +374,7 @@ def apply_fs_transform(fs_2_t1_transform, bbreg_transform):
             (inputnode, ds_t1w, [(('name_source', _first), 'source_file')]),
             (inputnode, ds_t1w_mask,
              [(('name_source', _first), 'source_file')]),
-            (merge, ds_t1w, [('merged_file', 'in_file')]),
+            (outputnode, ds_t1w, [('epi_t1', 'in_file')]),
```
effigies: This seems to be a null change. Why use `outputnode` as the source? Just to move the `DerivativesDataSink` to the bottom of the graph?

chrisfilo:
It's a leftover from previous attempt where depending on the metadata I was
using split or merged volumes. Therefore using outputnode allowed me to
make this connection consistent across the two options without a need for a
conditional statement. Even though it's a left over it might be useful in
the future and does not change the results.

On Thu, Mar 30, 2017 at 11:37 AM, Chris Markiewicz <notifications@github.com

effigies:
Fair.

Actually, I guess one more question: Is there a good report we can add to verify STC function?

chrisfilo:
I wish there was! I don't know how to visualize this correction - that's
why it's so controversial.

On Thu, Mar 30, 2017 at 11:42 AM, Chris Markiewicz <notifications@github.com

fmriprep/info.py
```diff
@@ -38,7 +38,8 @@
     'pybids>=0.1.0',
     'nitime',
     'niworkflows',
-    'nipype'
+    'nipype',
+    'nipy'
```
chrisfilo: Done.

PR #421, opened 2017-04-01
[RTM] HMC refactor around non-steady state

chrisfilo:
* Non-steady state detector was added to HMC workflow
* Reference image estimator for HMC was added:
  * If non-steady state is detected it will be used as a reference (they tend to have better tissue contrast)
  * Otherwise a median of 20 motion corrected volumes will be used as a reference
* EPI skull mask reportlet was brought back

### DIFF
--- a/docs/workflows.rst
+++ b/docs/workflows.rst
@@ -175,7 +175,7 @@ TR. Slice time correction can be disabled with ``--ignore slicetiming`` command
 line argument.
 
 FSL MCFLIRT is used to estimate motion
-transformations using a automatically estimated reference scan. If T1-saturation effects
+transformations using an automatically estimated reference scan. If T1-saturation effects
 ("dummy scans" or non-steady state volumes) are detected they are used as reference due to
 their superior tissue contrast. Otherwise a median of motion corrected subset of volumes is used.
 
--- a/docs/workflows.rst
+++ b/docs/workflows.rst
@@ -142,8 +142,8 @@ Preprocessing of BOLD files is split into multiple sub-workflows decribed below.
 
 .. epi_hmc :
 
-Head-motion and slice time correction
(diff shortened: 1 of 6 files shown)


docs/workflows.rst
```diff
@@ -165,14 +165,21 @@ Head-motion and slice time correction
...
+Slice time correction is performed
 using AFNI 3dTShift. All slices are realigned in time to the middle of each
 TR. Slice time correction can be disabled with ``--ignore slicetiming`` command
-line argument. FSL MCFLIRT is used to estimate motion
-transformations and ANTs is used to apply them using Lanczos interpolation. Nilearn
-is used to perform skullstripping of the mean EPI image.
+line argument.
+
+FSL MCFLIRT is used to estimate motion
+transformations using a automatically estimated reference scan. If T1-saturation effects
```
oesteban: an automatically

fmriprep/interfaces/utils.py
```diff
@@ -10,11 +10,75 @@
...
+        if n_volumes_to_discard == 0:
+            if in_nii.shape[-1] > 40:
+                slice = in_nii.get_data()[:, :, :, 20:40]
+                slice_fname = os.path.abspath("slice.nii.gz")
+                nb.Nifti1Image(slice, in_nii.affine,
+                               in_nii.header).to_filename(slice_fname)
+            else:
+                slice_fname = self.inputs.in_file
+
+            res = Volreg(in_file=slice_fname, args='-Fourier -twopass', zpad=4,
```
oesteban: Why not mcflirt here again? (just a question, I assume 3dvolreg works better here)

chrisfilo: I trust AFNIs interpolation more.

PR #422, opened 2017-04-01
ENH: Permit disabling sub-mm FS reconstruction

### LINKED ISSUE #420: Allow users to turn off -hires mode even if submillimiter data is avalable
https://neurostars.org/t/fmriprep-error-interface-reconallrpt-failed-to-run/331/9

effigies:
Closes #420.

### DIFF
--- a/fmriprep/cli/run.py
+++ b/fmriprep/cli/run.py
@@ -86,6 +86,8 @@ def get_parser():
     g_fs = parser.add_argument_group('Specific options for FreeSurfer preprocessing')
     g_fs.add_argument('--no-freesurfer', action='store_false', dest='freesurfer',
                       help='disable FreeSurfer preprocessing')
+    g_fs.add_argument('--no-submm-recon', action='store_false', dest='hires',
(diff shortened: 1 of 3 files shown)


fmriprep/cli/run.py
```diff
@@ -86,6 +86,8 @@ def get_parser():
     g_fs = parser.add_argument_group('Specific options for FreeSurfer preprocessing')
     g_fs.add_argument('--no-freesurfer', action='store_false', dest='freesurfer',
                       help='disable FreeSurfer preprocessing')
+    g_fs.add_argument('--no-submm-recon', action='store_false', dest='hires',
+                      help='disable sub-millimeter (hires) reconstruction')
```
chrisfilo: Does this default to storing `True`?

effigies: [Yes](https://docs.python.org/3/library/argparse.html#action). See e.g. the "--no-freesurfer" flag above. If it would be better to be explicit, I can also set defaults, similar to `g_ants.set_defaults(skull_strip_ants=True)`.

PR #423, opened 2017-04-02
[RTM] minor bbregister option tweak

chrisfilo:


chrisfilo:
Depends on https://github.com/nipy/nipype/pull/1929

chrisfilo:
I've tested this on a couple of datasets. It seems to help with ds000114 without making any visible negative impact on the beast.

chrisfilo:
FS6 is already a hard dependency - `-hires` does not work in 5.3 AFAIK. I would prefer to avoid adding a try catch for partial support of 5.3.

effigies:
True. Maybe a note in the [External Dependencies](https://github.com/poldracklab/fmriprep/blob/master/docs/installation.rst#external-dependencies) section of `installation.rst`, then?

chrisfilo:
https://github.com/poldracklab/fmriprep/pull/424

Issue #426, opened 2017-04-03
[bug] --reports-only argument fails to run


erramuzpe:
```
DATA_DIR="/home/asier/Desktop/fmriprep/ds000030"
OUTPUT_DIR="/home/asier/Desktop/fmriprep/output"
WORK_DIR="/home/asier/Desktop/fmriprep/scratch"

docker run -ti  \
	-v $DATA_DIR:/data:ro \
	-v $OUTPUT_DIR:/output \
	-v $WORK_DIR:/work \
	-w /work \
	poldracklab/fmriprep:latest \
	/data /output participant  \
	-w /work --no-freesurfer --skip-native --no-skull-strip-ants
```
This works great! :+1: 
```
docker run -ti --rm  \
	-v $DATA_DIR:/data:ro \
	-v $OUTPUT_DIR:/output \
	-w $WORK_DIR:/work \
	poldracklab/fmriprep:latest \
	/data /output participant \
	-w /work --reports-only 
```
`
TypeError: run_reports() missing 3 required positional arguments: 'out_dir', 'subject_label', and 'run_uuid'`

[1 argument](https://github.com/poldracklab/fmriprep/blob/master/fmriprep/cli/run.py#L150-L152) is given to run_reports but [run_reports](https://github.com/poldracklab/fmriprep/blob/master/fmriprep/viz/reports.py#L169) takes 4 arguments. 

I can try to solve this

effigies:
Thanks for the bug report. I've had this on my to do list for a couple weeks. I'll check out your PR and see what I can do to help.


-------- Original message --------

PR #427, opened 2017-04-03
[WIP] Solving --reports-only issue #426

erramuzpe:
I reorganized a bit the code:

 - if --reports-only: just run reports, as if preproc_wf had been already run. 

 - else: normal preproc_wf. 

wip. help.


fmriprep/cli/run.py
```diff
@@ -183,9 +183,9 @@ def create_workflow(opts):
     report_errors = 0
     if opts.reports_only:
         for subject_label in subject_list:
-            run_reports(settings['reportlets_dir'],
-                        settings['output_dir'],
-                        subject_label, run_uuid=run_uuid)
+            report_errors += run_reports(settings['reportlets_dir'],
+                                         settings['output_dir'],
+                                         subject_label, run_uuid=run_uuid)
```
erramuzpe: some kind of checker of `report_errors` ?

erramuzpe: line 212, inside `else`

erramuzpe:
I'm trying to run docker as explained [here](http://fmriprep.readthedocs.io/en/stable/contributors.html):

```
DATA_DIR="/home/asier/Desktop/fmriprep/ds000030"
OUTPUT_DIR="/home/asier/Desktop/fmriprep/output"
WORK_DIR="/home/asier/Desktop/fmriprep/scratch"

docker run -ti --rm  \
	-v $DATA_DIR:/data:ro \
	-v $OUTPUT_DIR:/output \
	-v $HOME/projects/fmriprep/fmriprep:/usr/local/miniconda/lib/python3.6/site-packages/fmriprep:ro \
	-w $WORK_DIR:/work \
	poldracklab/fmriprep:latest \
	/data /output participant   \
	-w /work --reports-only 
```

and I get the following error: 

```
  File "/usr/local/miniconda/bin/fmriprep", line 11, in <module>
    load_entry_point('fmriprep==0.3.1', 'console_scripts', 'fmriprep')()
  File "/usr/local/miniconda/lib/python3.6/site-packages/setuptools-27.2.0-py3.6.egg/pkg_resources/__init__.py", line 565, in load_entry_point
  File "/usr/local/miniconda/lib/python3.6/site-packages/setuptools-27.2.0-py3.6.egg/pkg_resources/__init__.py", line 2598, in load_entry_point
  File "/usr/local/miniconda/lib/python3.6/site-packages/setuptools-27.2.0-py3.6.egg/pkg_resources/__init__.py", line 2258, in load
  File "/usr/local/miniconda/lib/python3.6/site-packages/setuptools-27.2.0-py3.6.egg/pkg_resources/__init__.py", line 2264, in resolve
ModuleNotFoundError: No module named 'fmriprep.run_workflow'
```

I've checked into the container with `docker run -ti --entrypoint /bin/bash -v $HOME/projects/fmriprep/fmriprep:/usr/local/miniconda/lib/python3.6/site-packages/fmriprep:ro poldracklab/fmriprep:latest ` and the volume mounts correctly with my changes. ?¿

effigies:
This last issue is because we [recently](https://github.com/poldracklab/fmriprep/pull/410) updated the run environment. Could you go ahead and rebuild your docker image? From the source directory:

```Shell
docker build -t fmriprep:issue427 .
```

And then when you run, replace `poldracklab/fmriprep:latest` with `fmriprep:issue427`.

erramuzpe:
Thanks  @effigies for the fast reply!
so...
Rebuilt (1 hour)
Called docker again. Avoided previous error, report generated (empty) and:
```
/usr/local/miniconda/lib/python3.6/site-packages/nipype/workflows/dmri/mrtrix/group_connectivity.py:16: 
UserWarning: cmp not installed
  warnings.warn('cmp not installed')
```

effigies:
In the future the rebuilding the docker image shouldn't take so long, but thanks for going through the process. That warning isn't really relevant to us. It would be nice to suppress, as that's caused others confusion, too.

Looking at your `docker` command, you do have one error: `-w $WORK_DIR:/work` should be `-v $WORK_DIR:/work`. It looks to me like you're not actually saving your work directory, so FMRIPREP doesn't have anything to build from. You'll want to re-run from scratch once to build out the working directory, and then you should be able to do `--reports-only` (assuming it's been fixed).

If it hasn't been fixed, let us know, and I can make a priority of digging deeper into this. I appreciate your work so far. But in any event, you can always rebuild your reports by re-running without `--reports-only`. As long as nothing's changed in your inputs or working directory, it shouldn't take very long to check and regenerate your reports.

erramuzpe:
Oh, yes, you are right, I've been copy/pasting commands and this last docker call was wrong... the correct call is:

```
docker run -ti --rm  \
	-v $DATA_DIR:/data:ro \
	-v $OUTPUT_DIR:/output \
	-v $HOME/projects/fmriprep/fmriprep:/usr/local/miniconda/lib/python3.6/site-packages/fmriprep:ro \
	-v $WORK_DIR:/work \
	-w /work \
	fmriprep:issue427 \
	/data /output participant   \
	-w /work --reports-only 
```
I had all the (...)/scratch/workflow_enumerator/sub-001/BIDSDatasource(etc.) from a previous correct call. With the one above (corrected) the reports are not empty anymore :smile: Looks like working.

erramuzpe:
I guess tests have to be runned inside the container, right?
**13 passed, 1 skipped in 32.96 seconds**

fmriprep/cli/run.py
```diff
@@ -147,9 +147,7 @@ def create_workflow(opts):
     make_folder(settings['output_dir'])
     make_folder(settings['work_dir'])
 
-    if opts.reports_only:
-        run_reports(settings['output_dir'])
-        sys.exit()
+
```
effigies: This is now three lines between `make_folder(settings['work_dir'])` and `# nipype plugin configuration`. Can you reduce it to one?

fmriprep/cli/run.py
```diff
@@ -181,6 +179,14 @@ def create_workflow(opts):
         subject_list = [sub[4:] if sub.startswith('sub-') else sub for sub in subject_list]
 
     logger.info('Subject list: %s', ', '.join(subject_list))
+    
+    report_errors = 0
+    if opts.reports_only:
+        for subject_label in subject_list:
+            report_errors += run_reports(settings['reportlets_dir'],
+                                         settings['output_dir'],
+                                         subject_label, run_uuid=run_uuid)
```
effigies: Since we're not using `report_errors` before exiting in this case want to just make this:

```
if opts.reports_only:
    for subject_label in subject_list:
        run_reports(...)
    sys.exit()
```

And move `report_errors = 0` back to its previous location?

Note that I'm moving `sys.exit()` back one indentation level. As it is, I think you'll only regenerate the report for the first subject.

effigies:
Yes, it's generally best to run tests in the container. You can check out [circle.yml](https://github.com/poldracklab/fmriprep/blob/master/circle.yml#L43) for the tests we usually run.

erramuzpe:
thanks for the comments!

why is that circleCI is not running in this PR? Do you guys control it privately? :thought_balloon:

erramuzpe:
Changes work for me. 

I preprocessed just `--participant_label sub-001`, so I called again for just sub-001. Doing the general call (without --participant_label), generates empty reports for non-preprocessed subjects.

effigies:
Perhaps we should disable report building for subjects that have not been previously run, but I think in the short term this is fine. I think I've figured out the PR building, so I sent you a patch that should trigger one on this branch, once you merge it.

erramuzpe:
great!

effigies:
This LGTM. Any objections to merging, @chrisfilo?

chrisfilo:
Thanks Asier!

Issue #425, opened 2017-04-02
Preexisting expert options


chrisfilo:
When rerunning a hires dataset:

```
ERROR: there is a pre-existing expert options file and
you have specified an expert options file on the command-line.
If you want to use the file specified on the command-line,
you must run with -xopts-overwrite, or you may edit the 
existing file /root/src/fmriprep/work/workflow_enumerator/sid000080/t1w_preprocessing/Reconstruction/expert.opts (and not use the -expert option).

Standard error:

Return code: 1
Interface ReconAll failed to run.
```

https://neurostars.org/t/fmriprep-error-interface-reconallrpt-failed-to-run/331/12?u=chrisgorgolewski

effigies:
This is at the weird convergence of nipype and `recon-all`'s resume functions, and kind of highlights the problem that is FreeSurfer using the output directory as the state directory, and us permitting people to use externally-prepared FreeSurfer subject directories. The short story is that expert options can by default only be defined once, since they're then saved in the FreeSurfer state, so that running `recon-all` in the future keeps the same parameters. We currently set them in the `-autorecon1` phase. Which is a problem if that phase breaks and needs to be restarted, because we can't distinguish that situation from an externally-prepared (if partial) subject directory.

I think the best thing to do here is to split our FreeSurfer pipeline into one extra stage, separating T1/T2 population (`recon-all -i ... -T2 ...`) from `recon-all -autorecon1`, and include the expert options in the initial step. This has the advantage that no attempt to set expert options occurs if the directory already exists, and it would be very difficult for this first step to fail and need to be re-run.

There is the weird corner case of someone running the population step manually, for whatever reason, so we perform `-autorecon1` with `-hires`, but never get a chance to specify the expert options file. But maybe this is just overthinking it. We do say in the documentation that we will perform missing steps in the `recon-all` pipeline. We need to make a note of `-hires` processing in the docs, and maybe also note that letting FMRIPREP run a partial reconstruction can produce unexpected results?

The other corner case is somebody starts recon-all with hires processing, and wants to switch to `--no-submm-recon`, but there I think the answer is clearly "delete the subject directory and start over".

mike21499:
Hi there

I get an error regarding the recon-all. ReconallRPT failed to run. Tested two subjects with version 0.4.2. Attached is the traceback file.
[recon_traceback.txt](https://github.com/poldracklab/fmriprep/files/985896/recon_traceback.txt)


[...]

PR #429, opened 2017-04-03
[WIP] Separate loading of structural images from autorecon1

effigies:
Solution proposed in #425.

Untested at the moment, but would appreciate comments on whether this seems like a good move.

### DIFF
--- a/fmriprep/workflows/anatomical.py
+++ b/fmriprep/workflows/anatomical.py
@@ -16,6 +16,7 @@ from nipype.interfaces import ants
 from nipype.interfaces import freesurfer
 from nipype.interfaces import utility as niu
 from nipype.interfaces import io as nio
+from nipype.interfaces.traits_extension import Undefined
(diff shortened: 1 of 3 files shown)


fmriprep/workflows/anatomical.py
```diff
@@ -129,13 +129,21 @@ def bidsinfo(in_file):
             name='BIDSInfo',
             run_without_submitting=True)
 
+        pre_recon = pe.Node(
+            freesurfer.ReconAll(
+                directive='autorecon1',
+                flags='-nomotioncor -notalairach -nonuintensitycor '
+                      '-nonormalization -noskullstrip',
+            ),
+            name='PreRecon')
```
effigies: There's no good way to run `recon-all` in nipype without trying to run a part of it, so this just sets it to `autorecon1` and disables all of the sub-stages. Could push a new PR to nipype to make `directive` undefined by default.

PR #434, opened 2017-04-06
DOC: Add contribution thanks in changelog

effigies:
Let me know if you'd rather have a contributors file, and if you think we should credit specific PRs to listed contributors.

Also update the copyright to 2017 and to include me.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,8 @@
 Next release
 ============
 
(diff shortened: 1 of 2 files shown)


chrisfilo:
We should have a zenodo file. I'll set it up after this release. We also need to appropriatelly credit all of the methods used in FMRIPREP (calls for a duecredit coding sprint).

PR #436, opened 2017-04-09
[RTM] use dataobj instead of get_data

chrisfilo:
Avoids loading all of the big multiband images.

### DIFF
--- a/fmriprep/interfaces/utils.py
+++ b/fmriprep/interfaces/utils.py
@@ -44,7 +44,8 @@ class EstimateReferenceImage(SimpleInterface):
 
     def _run_interface(self, runtime):
         in_nii = nb.load(self.inputs.in_file)
-        global_signal = in_nii.get_data()[:, :, :, :50].mean(axis=0).mean(
+        data_slice = in_nii.dataobj[:, :, :, :50]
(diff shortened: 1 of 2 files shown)


fmriprep/interfaces/utils.py
```diff
@@ -44,7 +44,8 @@ class EstimateReferenceImage(SimpleInterface):
 
     def _run_interface(self, runtime):
         in_nii = nb.load(self.inputs.in_file)
-        global_signal = in_nii.get_data()[:, :, :, :50].mean(axis=0).mean(
+        data_slice = in_nii.dataobj[:, :, :, :50]
+        global_signal = data_slice.mean(axis=0).mean(
             axis=0).mean(axis=0)
```
effigies: Could reduce the number of operations a little:
```Python
global_signal = data_slice.reshape(-1, 50).mean(axis=0)
```

Test:
```Python
import numpy as np
x = np.random.rand(10, 9, 11, 20)
np.allclose(x.reshape(990, 20).mean(axis=0), x.mean(axis=0).mean(axis=0).mean(axis=0))
```

chrisfilo: This would break if the file had fewer than 50 timepoints, because reshape needs exact length and indexind does not.

PR #440, opened 2017-04-10
[RTM] Remove "sub-" prefix

chrisfilo:
BIDS Apps expect labels without prefixes.


effigies:
I definitely find `--participant_label 387` a bit opaque, and I think seeing the `sub-` prefix was actually informative, so maybe we should describe what `participant_label` refers to.

chrisfilo:
Good point - where do you think we should add this description?

PR #446, opened 2017-04-17
[RTM] fix the order of reports, remove silent try catch

chrisfilo:


effigies:
This looks fine to me, though the fact that this loop is populating a dictionary suggests that the order shouldn't matter at all.

@rwblair was there any particular reason for reverse-ordering elements?

chrisfilo:
From @rwblair  "Not positive why it was done that way. If going through the elements fixes the issue there shouldn't be any problem with the changes that I can see."

PR #447, opened 2017-04-17
[RTM] RF: Switch from --skip-native to --output-space=...

### LINKED ISSUE #445: Replace --skip-native with --output-spaces
Multiple choice with possible values
- T1w
- MNI152NLin2009cAsym (on by default)
- fsaverage 
- fsaverage6
- fsaverage5 (on by default unless --no-freesurfer)

effigies:
* [x] Use output spaces to select fsaverage subjects to copy to derivatives
* [x] Update T1w and surface space logic in workflows/epi

To do:

* [x] Permit disabling functionals resampled to MNI

Closes #445.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -56,7 +56,8 @@ ENV PERL5LIB=$MINC_LIB_DIR/perl5/5.8.5 \
 RUN echo "cHJpbnRmICJrcnp5c3p0b2YuZ29yZ29sZXdza2lAZ21haWwuY29tXG41MTcyXG4gKkN2dW12RVYzelRmZ1xuRlM1Si8yYzFhZ2c0RVxuIiA+IC9vcHQvZnJlZXN1cmZlci9saWNlbnNlLnR4dAo=" | base64 -d | sh
 
 # Installing Neurodebian packages (FSL, AFNI, git)
-RUN apt-get install -y --no-install-recommends \
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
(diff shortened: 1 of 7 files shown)


effigies:
Functional mapped to `fsaverage5` (compare to https://github.com/poldracklab/fmriprep/pull/391#issuecomment-292623823)

fsaverage5_func

Dockerfile
```diff
@@ -49,7 +47,8 @@ ENV PERL5LIB=$MINC_LIB_DIR/perl5/5.8.5 \
 RUN echo "cHJpbnRmICJrcnp5c3p0b2YuZ29yZ29sZXdza2lAZ21haWwuY29tXG41MTcyXG4gKkN2dW12RVYzelRmZ1xuRlM1Si8yYzFhZ2c0RVxuIiA+IC9vcHQvZnJlZXN1cmZlci9saWNlbnNlLnR4dAo=" | base64 -d | sh
 
 # Installing Neurodebian packages (FSL, AFNI, git)
-RUN apt-get install -y --no-install-recommends \
+RUN apt-get update && \
```
effigies: These add slightly to build time, but if `apt-get install` is called too long after `apt-get update` (such as in a partial image rebuild), it can fail.

effigies:
Status on this: MNI, T1w and various surface resamplings are all toggle-able. Currently no way to turn them all off (actually `--no-freesurfer --output-space=fsaverage` would do it), but I assume that's not actually desirable anyway.

chrisfilo:
Awesome work. Could you add a changelog entry and update the "Derivatives" section of the documentation?

PR #458, opened 2017-04-20
[RTM] Refactor working directory folder names

### LINKED ISSUE #417: Node object and label consistency
We should make sure all node objects have the same names as their labels.

For example:

Instead of
```
t1wmrg = pe.Node(IntraModalMerge(), name='MergeT1s')
```

we should do:
```
t1wmrg = pe.Node(IntraModalMerge(), name='t1wmrg')
```

It will make debugging easier.

chrisfilo:
Closes #417


effigies:
If we're going to do this, can we also move to more meaningful variable names? The node names have sometimes been very nice for clarifying what a component was meant to do.

fmriprep/workflows/anatomical.py
```diff
@@ -45,26 +45,26 @@ def t1w_preprocessing(settings, name='t1w_preprocessing'):
...
 
     # 2. T1 Bias Field Correction
     # Bias field correction is handled in skull strip workflows.
 
     # 3. Skull-stripping
-    asw = skullstrip_wf()
+    asw = skullstrip_wf(name='asw')
     if settings.get('skull_strip_ants', False):
-        asw = skullstrip_ants(settings=settings)
+        asw = skullstrip_ants(name='asw', settings=settings)
```
effigies: Giving `skullstrip_wf` and `skullstrip_ants` the name `'asw'` means if I ever re-run with/without ANTs, the `asw` directory is going to get clobbered. Perhaps in these swappable workflows, we could be slightly less rigid without losing anything?

```Python
skullstrip = skullstrip_wf(name='skullstrip_wf')
if settings.get('skull_strip_ants', False):
    skullstrip = skullstrip_ants(name='skullstrip_ants', settings=settings)
```

We keep the same prefix as the variable, the workflow name matches the workflow-generating function (which is where we'll want to look, if we're matching directories with variables) and there's no conflation of the two workflows in the working directory.

fmriprep/workflows/anatomical.py
```diff
@@ -74,43 +74,44 @@ def t1w_preprocessing(settings, name='t1w_preprocessing'):
...
+        name='t1_2_mni'
     )
     # should not be necesssary but does not hurt - make sure the multiproc
     # scheduler knows the resource limits
     t1_2_mni.interface.num_threads = settings['ants_nthreads']
 
     # 6. FreeSurfer reconstruction
     if settings['freesurfer']:
-        surface_recon = surface_reconstruction(settings=settings)
+        surface_recon = surface_reconstruction(name='surface_recon',
```
effigies: Again, here I think it's better to give workflows the name of the generator function, not the variable.

chrisfilo:
I agree that many of the variables should be renamed to improve clarity, but I would prefer to leave this to separate (potentially many small) PRs.

effigies:
Sounds good. Any thoughts on the workflow name issue?

fmriprep/workflows/anatomical.py
chrisfilo: Yeah I was wondering about this as well - what's the best way to do this. My reasoning was that:
- you are interested in the instance of a particular workflow not the name of the function (which is helpful when distinguishing between multiple instances of the same workflow)
- this would be consistent with the Node naming scheme

However, in FMRIPREP we almost never reuse functions for creating workflows. Therefore in all cases "instance == function" and naming workflow after the function actually makes it faster to find relevant code.

effigies: What about:

* The variable name should be a substring or superstring of the function name
* The workflow name should be the function name if only one is made, the variable name if multiple

Then we have either:

```Python
surface_recon = surface_reconstruction(name='surface_reconstruction')
```
or
```Python
surface_recon1 = surface_reconstruction(name='surface_recon1')
surface_recon2 = surface_reconstruction(name='surface_recon2')
```

chrisfilo: That sounds good, but wouldn't this mean that

`s = surface_reconstruction(name='surface_reconstruction')`

would be allowed? We might also need some recommendations on function naming schemes.

effigies: Right, we wouldn't want to accept arbitrary sub/superstrings. If we want to set a few guidelines, here are some:

* Remove `_wf` from the function name before applying other rules
* For substrings, the final word of the function name can be abbreviated or removed
  * Substrings can have a number or `_desc` appended
* Superstrings must be the function name followed by a number or `_desc`

Where `_desc` is an underscore followed by a one word (or abbreviation) description.

Looking at what we've got and proposed names that fit these rules:

* `skullstrip_wf`/`skullstrip_ants` => `skullstrip`
* `surface_reconstruction` => `surface_recon`
* `epi_hmc` => `epi_hmc`
* `ref_epi_t1_registration` => `ref_epi_t1_reg`
* `discover_wf` => `discover`
* `sdc_unwarp` => `sdc_unwarp`
* `fieldmap_unwarping_report` => `fieldmap_unwarping_report`
* `epi_mni_transformation` => `epi_mni`
* `epi_surf_sample` => `epi_surf`
* `cleanup_edge_pipeline` => `cleanup_edge`
* `phdiff_workflow` => `phdiff`

Not perfect, but it doesn't seem too far off.

-----

For function naming schemes, I think a reasonable rule would be for workflow-returning functions to end in `_wf` or `_workflow`. (Slight preference for `_wf`.)

I don't think we have functions that return nodes.

What about functions that are intended to be wrapped with the `Function` interface? Having a name scheme might make things easier. Supposing we ended all such functions with `_fn`:

```Python
def gifti_name_fn(arg1, arg2):
    ...
    return ret

gifti_name = pe.Node(niu.Function(function=gifti_name_fn, ...), name='gifti_name')
```

I don't really like `_fn`, since it isn't obviously for wrappable functions, but something like that could keep the namespace clean.

chrisfilo: Some thoughts:
1) Let's keep a naming convention for Function functions for another PR
2) We should probably enforce that all Workflow constructors have `_wf` suffix
3) It might be actually beneficial to add `_wf` to the Workflow name - it makes it easier to distinguish Workflow folder from a Node folder in the working directory
4) Also - are functions the most intuitive way of creating subworkflows. Maybe those should just be classes inheriting from Workflow?
4) The scheme you outlined has a lot of rules which will make it hard to follow for newcomers. If we get rid of the sub/super strings things could become easier.

effigies: 1. Sounds good
2. Seems reasonable
3. Then wouldn't the variable have the same name as the generator? `x_wf = x_wf(name="x_wf")`
4. Do you have an example of a sub-class workflows? I think I've only ever seen function-generated, so it'd need to be a pretty compelling improvement to break with standard usage.
5. It is a number of rules. I'd think of them more as heuristics, nothing hard and fast. It just tries to keep an orthographic link between the workflow name and the variable name, while not ending up with ridiculously long names or running into namespace collisions (see 3).

effigies: What about this rule:

* All workflow-generating functions end with `_wf`.
* For workflow `x_wf`, variables are `xN = x_wf(name="xN_wf")`
  * If there's only one: `N = ''`
  * Else: `N = <number>` or `N = '_desc'`

chrisfilo: What about `init_` prefix and `_wf` suffix? for example

`xN_wf = init_x_wf(name="xN_wf")`

(with N playing the same role as you described)

effigies: That works for me. I can make a PR against this branch real quick...

effigies:
This'll be the same test as the one that just passed. Any last changes before merge?

PR #461, opened 2017-04-20
Update nipype dependency in niworkflows and fmriprep

effigies:


effigies:
Closing in favor of #462.

PR #465, opened 2017-04-21
CI: Pull latest image before rebuilding

effigies:
Because we rebuild from scratch every time, every docker image is unrelated, even though there may be no difference in the docker file. The primary goal of this PR is to reduce redundant downloads for users of the `poldracklab/fmriprep` docker images on new releases.

It has a futher advantage of nearly halving the image creation stage on Circle:

### Before

no_docker_pull

### After

with_docker_pull

The failure on saving the image is a little weird. I'll see if that keeps happening.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -33,6 +33,7 @@ dependencies:
     - if [[ -e $HOME/docker/image.tar ]]; then docker load -i $HOME/docker/image.tar; fi
     - sed -i -E "s/(__version__ = )'[A-Za-z0-9.-]+'/\1'$CIRCLE_TAG'/" fmriprep/info.py
     - sed -i -E "s/(__version__ = )'[A-Za-z0-9.-]+'/\1'$CIRCLE_TAG'/" wrapper/fmriprep_docker.py
+    - docker pull poldracklab/fmriprep:latest || true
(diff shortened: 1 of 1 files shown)


effigies:
Okay, that seems to be working. Is there any reason this is a bad idea?

There is the issue that if a major change happens near the top of the Docker file, the time savings are going to be negated and have the additional cost of the pull; under those conditions, it would probably be a good policy to comment out the line until the next release updates `:latest`.

apt packages will get frozen for a while, though that might be a feature, rather than a bug.

And it may still be worth rebuilding from scratch every so often. That could be done by pushing a clean-built `poldracklab/fmriprep:latest` just before a release, so the release image is based on the fresh build.

effigies:
Tests passing. Thoughts?

chrisfilo:
So in principle this should speed up builds and make them less dependant on say MGH servers?
If this works we could turn off saving and restoring the docker image using .tar (to speed things up a little more) - right?

effigies:
Right. This should only hit MGH if either docker hub goes down or the Dockerfile has changed to the extent of needing it.

I don't want to kill the cache just yet. I'm curious if after a couple rounds through, this will actually make the cache work; the pull might give the needed context to the layers that are saved in the `.tar` file.

effigies:
Merging. If we don't see a speed-up after a couple more PRs, we can remove the caching.

PR #469, opened 2017-04-23
[RTM] Refactor settings

### LINKED ISSUE #418: Explicit arguments instead of dictionaries
Often we are passing `settings` dictionaries to workflow creation functions. This makes debugging hard, because it's not clear which fields of this dictionary are mandatory. I propose we use explicit names.

For example instead of:

```
wf = t1w_preprocessing(settings={'ants_nthreads': 1, 'nthreads': 1, 'freesurfer': True, 'reportlets_dir': '.', 'output_dir': '.'})
```

```
wf = t1w_preprocessing(ants_nthreads=1, nthreads=1, freesurfer=True, reportlets_dir='.', output_dir='.'})
```

In addition, this will make coding easier because IDEs understand local variables but do not understand dictionary keys.

### LINKED ISSUE #442: Set defaults only once
On many occasions, we set defaults in the run script:

```
g_ants.add_argument('--skull-strip-ants', dest="skull_strip_ants", action='store_true',
                        help='use ANTs-based skull-stripping (default, slow))')
```
https://github.com/poldracklab/fmriprep/blob/a6e941ec3e51ce7d5f90d7ef50357d9a39ee1e59/fmriprep/cli/run.py#L79
(in this case the default is implicit)

and then we set the default again down the execution stream
```
    if settings.get('skull_strip_ants', False):
        asw = skullstrip_ants(settings=settings)
```
https://github.com/poldracklab/fmriprep/blob/57572ed2dc414f0fd175d2eff4268981066597ed/fmriprep/workflows/anatomical.py#L59

This is a dangerous practice in which the two defaults can accidentally get out of sync without any clear error. 

If #418 will be implemented the `settings.get()` will go away, but even then we should avoid default argument values where the default is already set in the run script.

chrisfilo:
- Removes settings dictionary in favour of explicit function arguments. Closes #418
- All defaults are set in one place - run.py. Closes #442
- Memory restrictions are now set per each run independently.
- Minor workflow name tweaking.

### DIFF
--- a/docs/workflows.rst
+++ b/docs/workflows.rst
@@ -150,7 +150,7 @@ BOLD preprocessing
     :simple_form: yes
 
     from fmriprep.workflows.epi import init_func_preproc_wf
-    wf = init_func_preproc_wf("test_file",
+    wf = init_func_preproc_wf('/completely/made/up/path/sub-01_task-nback_bold.nii.gz',
                               ants_nthreads=1,
                               ignore=[],
                               freesurfer=True,
--- a/docs/workflows.rst
+++ b/docs/workflows.rst
@@ -16,24 +16,25 @@ slice-timing information and no fieldmap acquisitions):
     :graph2use: orig
     :simple_form: yes
 
-    from fmriprep.workflows.base import init_basic_wf
-    wf = init_basic_wf(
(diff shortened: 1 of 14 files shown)


effigies:
LGTM.

Issue #456, opened 2017-04-19
Workflow interface: input / output / reports


effigies:
By giving each workflow an `inputnode`, `outputnode` and `reportnode`, we can reduce the clutter within each workflow where sinking/derivatives are put on visual parity with processing nodes.

Reportlets can then be passed to an interface/workflow that constructs the full report, isolated from the processing/reportlet construction.

Similarly, it would be nice (though possibly not feasible) to reduce output sinks to at most one per sub-workflow, and ideally fewer, grabbing outputs from `sub_wf.outputnode`.

chrisfilo:
Could you explain a bit more what role the `reportnode` would play?

Please keep in mind that with the current architecture we need one datasink per reportlet, because we want as many reportlets in a partial report (when something fails). If we connect multiple reportles to one data sink and one of the interfaces creating reportlets fails none of them will be saved and all of them will be missin in the final report.

effigies:
I might be missing something. Why is a datasink needed, as opposed to an interface that grabs the reportlets from their original locations in the working directory?

But to be (hopefully) a little clearer, the reportnode would collect and give named ports for each of the reportlets that a workflow generates. If the reportlet is undefined, we can skip over it, as we do now. The only thing that needs a sink here is the final report.

If this sounds like too much of a pain, NBD. I just find the reportlet derivatives to be very cluttering when I'm trying to understand a new workflow (or part of a workflow).

chrisfilo:
On Wed, Apr 19, 2017 at 11:13 AM, Chris Markiewicz <notifications@github.com
> wrote:

> I might be missing something. Why is a datasink needed, as opposed to an
> interface that grabs the reportlets from their original locations in the
> working directory?
>
We could do that, but this means we will heavily depend on the structure of
the working directory not changing over time.

> But to be (hopefully) a little clearer, the reportnode would collect and
> give named ports for each of the reportlets that a workflow generates. If
> the reportlet is undefined, we can skip over it, as we do now. The only
> thing that needs a sink here is the final report.
>
Ok - get it.

> If this sounds like too much of a pain, NBD. I just find the reportlet
> derivatives to be very cluttering when I'm trying to understand a new
> workflow (or part of a workflow).
>
You mean the reportlet datasinks?

effigies:
> We could do that, but this means we will heavily depend on the structure of
the working directory not changing over time.

Will it? Since the full reportlet path is what's passed to the report-constructor, it seems like this should be fairly robust.

[...]

PR #471, opened 2017-04-24
RF: Split reports/derivatives from anatomical workflow

effigies:
This is the work discussed in #456. Looking at it, it's actually kind of large, so I want to make it a separate PR before starting in on the functional workflows.

I also updated some names. The overall pattern I took was `<space>_<desc>`, which seemed roughly how things were being done. So `arw` (which I think meant anatomical reoriented... w) is now `t1_conform`, since it's essentially a conformation step. `bmask_mni` is `mni_mask`, to mirror `t1_mask`.

I think for the functional workflows, it makes sense to do a derivatives and reports node at the just-under-top level again, even though it's a bigger workflow overall. That will help keep the internal sub-workflows more modular and focused on processing than book-keeping.


effigies:
Hierarchical workflow graph following a successful run (ds000114/sub-01/ses-01):
refactored_workflow

effigies:
@chrisfilo What tests or further updates do you want to see before this goes in?

A few things I'm thinking of:

* [ ] Rethink a couple names in `surface_recon_wf`
  * [ ] `injector -> skull_strip_extern` or similar
  * [ ] `gifticonv -> fs_2_gii` or similar
* [ ] (Maybe) push all derivatives into `outputnode` first, and from there to `anat_derivatives_wf`

(I'll work on `func_preproc_wf` separately, just to keep the diffs more manageable.)

chrisfilo:
This already look like a nice improvement in readability - I cannot think of anything in this workflow that desperately needs renaming.

effigies:
Docs: https://326-78558509-gh.circle-artifacts.com/0/home/ubuntu/docs/workflows.html

PR #472, opened 2017-04-24
[RTM] CI: Remove non-working docker cache

effigies:
We don't seem to be getting any speed-up from maintaining a docker cache. Every pull still grabs a 2.7GB layer; if the cache were actually loading, we should only see the pip builds and on get pulled and rebuilt each time.

Assuming I'm right and there's no speed up granted by the cache, we should be seeing a pull on the order of 5-6 minutes and the same for the build step.

This will end up saving ~4 minutes per build.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,7 +1,4 @@
 machine:
-  pre:
-    - sudo curl -L -o /usr/bin/docker 'https://s3-external-1.amazonaws.com/circle-downloads/docker-1.9.1-circleci'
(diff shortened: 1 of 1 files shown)


chrisfilo:
Since we are not using cache we can also drop the `pre` section that was installing a supposedly cache enabled docker version https://github.com/poldracklab/fmriprep/pull/472/files#diff-29944324a3cbf9f4bd0162dfe3975d88R2

effigies:
Updated. Good to go?

PR #473, opened 2017-04-24
[WIP] ENH: Apply 9 DOF alignment in T1 merge

### LINKED ISSUE #439: T1 merge is failing on s497

effigies:
Moves to using merged and conformed T1 as basis for FreeSurfer reconstruction, rather than raw T1w images.

Fixes #439.


oesteban:
This is interesting: which of the two T1s is the reference?

While the first 6 parameters are equivalent between registration directions (i.e. which T1 is the reference), the scaling parameters (the 3 between 6 and 9) are not. Maybe we need a subject-template space at the midpoint of all transforms.

effigies:
So we're not actually using 9-DOF, we moved to `mri_robust_template` in #481. There the reference is the first T1 as passed by the BIDS grabber (which I believe will always be `ses-1` or the lexically "smallest" session). It does use one additional DOF to allow intensity scaling.

#481 put in machinery that I think shouldn't be too hard to adapt to an unbiased template (see discussion).

PR #476, opened 2017-04-24
[RTM] CI: Test FreeSurfer workflows with pre-reconstructed subject

### LINKED ISSUE #432: Incorporate precomputed freesurfer outputs into CircleCI tests
FreeSurfer is our default flow now. We should start testing it on a regular basis.

effigies:
Closes #432

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -2,6 +2,7 @@ machine:
   environment:
     OSF_PROJECT: "https://files.osf.io/v1/resources/fvuh8/providers/osfstorage"
     DS005_URL: "${OSF_PROJECT}/57f32a429ad5a101f977eb75"
+    DS005_FS_URL: "${OSF_PROJECT}/58fe59eb594d900250960180"
     DS054_URL: "${OSF_PROJECT}/57f32c22594d9001ef91bf9e"
   services:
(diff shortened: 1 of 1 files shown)


effigies:
Set [RTM] because this is working and can be merged as is. Adds about 20 minutes onto test time, so we can count on about half of that being `mris_expand` (see #477) and I guess the rest is bbregister and sampling to the `fsaverage5` surface.

Main thing I see here is to reduce the functionals to a single truncated run (40 time points?). Should I modify `ds005_downsampled.tar.gz` or create a new variant? And should we do the same thing for ds054 to help reduce test time there, too?

cc @oesteban

chrisfilo:
I think it makes sense to have two but no more (there are currently 3) to test iterating over runs. Great work!

effigies:
I checked the down-sampled datasets. ds054 only has two runs, with 5 time points each. ds005 EPIs have 16 time points, so I just removed the third run and uploaded a new tarball.

Re-running with a cleared cache to get the new ds005. I think it might be worth updating this to use two parallel builds. We can run the FreeSurfer test on one build and everything else on the other.

Issue #337, opened 2017-01-13
Add option to us 2mm voxels size for the MNI transformation


chrisfilo:
We currently keep the same voxel size of the MNI outputs as the original inputs. This avoids upsampling and saves disk space.

However, images in MNI space that do not have the canonical MNI template resolution and dimensions can cause problems with some legacy software (for example FSLView). It would be good for compatibility reasons to provide an option to disable resolution preserving feature and use cannonical 2mm resolution.

Suggested by @poldrack.

feilong:
Does FSL in the Docker/Singularity image contains standard brains?  I didn't find the data directory of FSL.

chrisfilo:
Indeed the atlases are not included to save space.

However, to add an option of forcing 2mm voxel size (instead of the voxel size of the inputs as it is done right now) you can modify this Interface https://github.com/poldracklab/fmriprep/blob/cecf99dd25a25a840673c900b3f905616bd35666/fmriprep/interfaces/images.py#L42

Most importantly it would be interesting to learn why are you interested in this feature. Using 2mm voxel size when your original data has lower resolution will waste disk space without much of an advantage.

feilong:
One data set that I'm working on was collected from different sites and has different voxel sizes.  It would be helpful to resample data to the same resolution (not necessarily upsampling to 2mm).

chrisfilo:
Ah I see - let me know if you need help implementing this. Opening a pull request with "[WIP]" (as in Work in Progress) could be a good starting point.

feilong:
I just sent a pull request that allows to use an external image as the resampling reference.

Alternatively we could specify a resolution for resampling and create the reference using the interface mentioned above, but the created 2-mm reference would be slightly different from the one provided by FSL in terms of shape and affine matrix.

chrisfilo:
I think there are two issues we need to disentangle here:

- Selection of the normalization target template. We are currently using `MNI152NLin2009cAsym` while SPM and FSL are using an older `MNI152NLin6Asym` (more details [here](http://www.lead-dbs.org/?p=1241)). Those templates are very similar, but not the same and will produce different results. If you want your results to be in the exactly same space as FSL we need to add support for `MNI152NLin6Asym` space.

[...]

PR #480, opened 2017-04-25
[WIP?] Add mni_ref parameter that allows external reference image for resampling

feilong:
An attempt to solve the issue:
https://github.com/poldracklab/fmriprep/issues/337

### DIFF
--- a/fmriprep/cli/run.py
+++ b/fmriprep/cli/run.py
@@ -92,6 +92,11 @@ def get_parser():
         choices=['MNI152NLin2009cAsym'], default='MNI152NLin2009cAsym',
         help='volume template space (default: MNI152NLin2009cAsym)')
 
+    g_conf.add_argument(
+        '--mni-ref', required=False, action='store', default=None,
+        help='Reference image for resampling BOLD files to MNI152NLin2009cAsym space. '
+             'It determines the grid (i.e., affine matrix) of the output files if provided.')
+
     #  ANTs options
     g_ants = parser.add_argument_group('Specific options for ANTs registrations')
     g_ants.add_argument('--skull-strip-ants', dest="skull_strip_ants", action='store_true',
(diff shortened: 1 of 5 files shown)


effigies:
Thanks for this. I leave it to @chrisfilo as to whether the exact method for specifying this within the workflow is appropriate, but could you also update the `fmriprep-docker` script, so this can be easily used inside docker?

Options for pulling in files or directories from outside are defined in [`g_wrap`](https://github.com/poldracklab/fmriprep/blob/34fe8d2b0dda060081fc444359cb0df62c27ab0a/wrapper/fmriprep_docker.py#L220-L224), and are then [mapped](https://github.com/poldracklab/fmriprep/blob/34fe8d2b0dda060081fc444359cb0df62c27ab0a/wrapper/fmriprep_docker.py#L313-L315).

Something like:

```Python
g_wrap.add_argument('--mni-ref', required=False, action='store', type=os.path.abspath)

...

if opts.mni_ref:
    target = '/imports/' + os.path.basename(opts.mni_ref)
    command.extend(['-v', ':'.join((opts.mni_ref, target, 'ro'))])
    unknown_args.extend(['--mni-ref', target])
```

If this turns out to be too much of a pain, don't worry, and I'll figure it out once this is merged.

feilong:
Thanks @effigies, I updated (copy/paste) the file as you said.  I'm not familiar with Docker, so I'll leave the rest to the experts.

chrisfilo:
As we discussed in https://github.com/poldracklab/fmriprep/issues/337#issuecomment-300376362 it would make sense to rename this to `--output-grid-reference`. Could you also rebase/merge master?

feilong:
Sure, will do tomorrow.

feilong:
I rebased my commits and renamed the parameter. I tested it locally and the outputs look fine.

fmriprep/cli/run.py
```diff
@@ -92,6 +92,11 @@ def get_parser():
         choices=['MNI152NLin2009cAsym'], default='MNI152NLin2009cAsym',
         help='volume template space (default: MNI152NLin2009cAsym)')
 
+    g_conf.add_argument(
+        '--output-grid-reference', required=False, action='store', default=None,
+        help='Grid reference image for resampling BOLD files to volume template space. '
+             'It determines the grid (i.e., affine matrix) of the output files if provided.')
```
chrisfilo: I would mention that this "determines the field of view and resolution of the output images, but is not used in normalization".

Issue #310, opened 2017-01-10
Support for multisession data with multiple T1s


chrisfilo:
In case the dataset includes multiple sessions **and** each session includes at least one T1w image the pipeline should default to coregistering EPI to the T1w image from the corresponding session.

Alternatively an **unbiased** subject specific template (combination of all T1w images from a given subject) should be created and set as the coregistration target.

Currently the pipeline defaults to combining all T1w images (even across sessions) by coregistering all of them to the first T1 found. This is acceptable for combining multiple T1w images from a single session, but results in a bias towards the first T1w and thus is suboptimal in context of longitudinal studies.

Creating unbiased structural templates is necessary for both the volume as well as surface analysis. Former can be achieved using ANTs or AFNI - latter is supported by the FreeSurfer longitudinal stream.

fliem:
I am planning to implement an [unbiased template approach](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3389460/).
Roughly this means:
- creating an unbiased template from the T1w images
- registering from template-T1w to MNI space
- within each session: registering from session-EPI to session-T1w, from session-T1w to template-T1, and combining with base-T1w-to-MNI-T1w-transformation

The template could be create with 
- the longitudinal freesurfer stream (e.g., [Tracula](https://surfer.nmr.mgh.harvard.edu/fswiki/Tracula) uses that for longitudinal registration of DWI data)
and/or
- ANTs

@chrisfilo and others: do you have experience with either of those approaches?

chrisfilo:
This sounds good to me. I only have small amount of experience with
freesurfer longitudinal processing. Since there are more users in volume
based processing I would prefer at least starting with ants. Thanks!

fliem:
I’m not 100 percent sure, but I think Tracula only uses volumes as templates and basing the fmri workflow on the dmri workflow probably would be straightforward.

As far as ATNs goes, it seems there is 
- [buildtemplateparallel.sh](https://github.com/stnava/ANTs/blob/master/Scripts/buildtemplateparallel.sh) which has a (legacy) [nipype interface](http://nipy.org/nipype/interfaces/generated/nipype.interfaces.ants.legacy.html#buildtemplateparallel) and
- more recently: [antsMultivariateTemplateConstruction2.sh](https://github.com/stnava/ANTs/blob/master/Scripts/antsMultivariateTemplateConstruction2.sh) which has no nipype interface

Also there are two nipype workflows
[...]

PR #481, opened 2017-04-25
[RTM] ENH: Use mri_robust_template for T1 merge

effigies:
We've found that FLIRT alignment can sometimes fail on T1 images. Although a 9-DOF affine registration resolves the issue, it appears to be an artifact problem, rather than a scaling issue that would require the extra 3 degrees of freedom.

This PR uses FreeSurfer's `mri_robust_template` to generate a structural reference image.

`mri_robust_template` throws an error on single volumes, so I subclass `freesurfer.RobustTemplate` to copy the input file in that case. (I use a hack similar to that in `freesurfer.ReconAll` to fiddle with the `cmdline`.)

Closes: #439, #473

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,8 @@
+Next release
+============
+
+* [ENH] Use robust template generation for multiple T1w images (#481)
+
 0.4.1 (20th of April 2017)
 ==========================
 
--- a/fmriprep/interfaces/__init__.py
+++ b/fmriprep/interfaces/__init__.py
@@ -4,4 +4,5 @@
 # vi: set ft=python sts=4 ts=4 sw=4 et:
 from fmriprep.interfaces.bids import ReadSidecarJSON, DerivativesDataSink, \
     BIDSDataGrabber, BIDSFreeSurferDir
-from fmriprep.interfaces.images import IntraModalMerge, CopyHeader, StructuralReference
+from fmriprep.interfaces.images import IntraModalMerge, CopyHeader
+from fmriprep.interfaces.freesurfer import StructuralReference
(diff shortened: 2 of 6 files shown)


effigies:
This is running fine for me. Do we still want to pass the pre-merged T1 to FreeSurfer? On the one hand there's no scaling issue to correct before passing it in; on the other, it's what FS will do immediately, anyway, so no need to duplicate work.

Leaving it as is also puts the original T1s into the FreeSurfer directory, which is either good (for provenance sake) or bad (for taking up space), depending on your priority.

I can go either way.

chrisfilo:
I would lean towards feeding one merged image.

chrisfilo:
BTW - can we get a changelog entry? It would be good that now the templates are unbiased. This also closes #310

effigies:
When you say unbiased, you mean that we're no longer aligning to the first image before taking the average? Because we're still doing that. The FreeSurfer invocation is `--inittp 1 --fixtp`, which means "use first frame as the 'init' time point and align to that time point", so I kept on with this.

I can play with dropping those to see what effect that has, but I think maybe that makes sense as a separate PR. This fixes #439 in the short term while keeping as close as possible to existing behavior.

chrisfilo:
Ok then this does not close #310. Lets leave it to another PR.

effigies:
The new merge stripped original filename info from the pipeline. This should resolve that, though another option would be to pull the `bids_info` node out of `surface_recon_wf` and just pass in the `subject_id`. But I figured there might be something else that counts on this name getting propagated through properly, so I went with this option.

effigies:
Re-ran s497 from scratch overnight. Outputs LGTM. Moving to `interfaces.freesurfer` to be consistent with #484. Will merge on pass.

PR #482, opened 2017-04-26
[RTM] RF: Split/consolidate functional derivatives and reports

effigies:


effigies:
So this is a lot more challenging a cleanup than anatomical, simply because there are so many sub-workflows. I've taken the strategy of moving reports and derivatives into outputnodes of the individual sub-workflows, then copying derivatives to a `func_preproc_wf.outputnode` before passing to `func_derivatives_wf.inputnode`, and sending reports directly to `func_reports_wf.inputnode`.

This was to be consistent with `anat_preproc_wf`, but it may be that this isn't the best strategy here. Here's a figure with output-spaces `T1w MNI[...] fsaverage5`:

graph0

Documentation workflows will be up for viewing when Circle tests finish. With the relatively small number of output nodes per work node, I don't think this will make much of a visual improvement.

effigies:
Here's the execution graph for the FreeSurfer-enabled ds005 test:



And [the docs](https://359-78558509-gh.circle-artifacts.com/0/home/ubuntu/docs/workflows.html).

effigies:
I guess one additional point: I've left the fieldmap report workflow alone.

fmriprep/workflows/epi.py
```diff
@@ -73,53 +73,56 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
...
+
+    func_reports_wf = init_func_reports_wf(reportlets_dir=reportlets_dir,
+                                           freesurfer=freesurfer)
+
+    func_derivatives_wf = init_func_derivatives_wf(output_dir=output_dir,
+                                                   output_spaces=output_spaces,
+                                                   freesurfer=freesurfer)
+
+    workflow.connect([
+        (inputnode, func_reports_wf, [(('epi', _first), 'inputnode.source_file')]),
```
chrisfilo: Why is `_first` necessary here?

effigies: Any derivatives/reports used [this construct](https://github.com/poldracklab/fmriprep/pull/482/files#diff-4c51aa6368169bddb20f09d1f776098dL440), so I just copied it. Looking at it, I think it's unnecessary because `'epi'` is always a single path, so I'll go ahead and remove it.

effigies:
Didn't break anything by removing the `_first` calls. Merging...

PR #484, opened 2017-04-26
[RTM] RF: Skip mris_expand on pre-existing midthickness files

### LINKED ISSUE #477: Skip mris_expand on pre-existing graymid/midthickness files
With #476, I'm remembering that `mris_expand` gets called every time the scratch directory gets wiped out, even though the output (midthickness files) exists. This is a roughly ~12min step. I'm not sure what the best way to skip this would be.

One option would be: Add a graymid/midthickness output to `nipype.interfaces.io.FreeSurferSource`. Then subclass `MRIsExpand` to take that field as an input. If it exists, change the command to `cp $GRAYMID $HEMI.midthickness`.

In general I don't love that we're writing into a subject directory, but for reasons explored in https://github.com/poldracklab/fmriprep/pull/398#issuecomment-291497871 it's a least-bad option. But especially if somebody provides a pre-computed midthickness file, I think it's better not to replace it, and the above proposal overwrites it, but with a copy of itself, so it's pretty tame.

Any other thoughts?

effigies:
Just to get this testing, while waiting on upstream. Same style of command-line hack as in #481.

Here I use explicit copying, rather than `copyfile`, since I don't want the datasink to cause an error by overwriting a file with itself. (This might not actually be a problem; I'm not sure.)

Closes #477.

### DIFF
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -28,5 +28,5 @@ dependencies:
     - svgutils
     - nitime
     - nilearn
-    - git+https://github.com/nipy/nipype.git@dc68922bf9ac1dd910c3ca9b5b8c0dbb093f4691#egg=nipype
+    - git+https://github.com/nipy/nipype.git@ccd0527d65c63b22b86b8e7df0d3103e4f4fe743#egg=nipype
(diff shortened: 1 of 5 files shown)


effigies:
Previous [Circle build](https://circleci.com/gh/effigies/fmriprep/351) reduced build time to 1h10m. Plan to merge in the morning if there are no comments.

PR #486, opened 2017-04-26
[RTM] ENH: Check for error during memory check

### LINKED ISSUE #485: Issues using docker container
I just installed the docker container on Ubuntu 16.04LTS on VirtualBox and I get the following two errors:

Issue 1:
Got permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Get http://%2Fvar%2Frun%2Fdocker.sock/v1.27/images/json?filters=%7B%22reference%22%3A%7B%22poldracklab%2Ffmriprep%3A0.4.1%22%3Atrue%7D%7D: dial unix /var/run/docker.sock: connect: permission denied

Issue 2:
Image 'poldracklab/fmriprep:0.4.1' is missing
Would you like to download? [Y/n] Y
  File "/home/vooz/.local/bin/fmriprep-docker", line 11, in <module>
    sys.exit(main())
  File "/home/vooz/.local/lib/python2.7/site-packages/fmriprep_docker.py", line 282, in main
    mem_total = check_memory(opts.image)
  File "/home/vooz/.local/lib/python2.7/site-packages/fmriprep_docker.py", line 122, in check_memory
    if line.startswith(b'Mem:')][0]
IndexError: list index out of range

effigies:
Should give a more helpful error message than in #485.

Any thoughts on testing this in Circle? I guess something like `sudo -u unprivileged_user fmriprep-docker ...`?

Closes #485.

### DIFF
--- a/wrapper/fmriprep_docker.py
+++ b/wrapper/fmriprep_docker.py
@@ -117,6 +117,9 @@ def check_memory(image):
                           image, '-m'],
                          stdout=subprocess.PIPE,
                          stderr=subprocess.DEVNULL)
+    if ret.returncode:
+        return -1
+
     mem = [line.decode().split()[1]
            for line in ret.stdout.splitlines()
            if line.startswith(b'Mem:')][0]
@@ -280,6 +283,10 @@ def main():
 
     # Warn on low memory allocation
     mem_total = check_memory(opts.image)
+    if mem_total == -1:
+        print('Could not detect memory capacity of Docker container.\n'
+              'Do you have permission to run docker?')
+        return 1
     if mem_total < 8000:
         print('Warning: <8GB of RAM is available within your Docker '
               'environment.\nSome parts of fMRIprep may fail to complete.')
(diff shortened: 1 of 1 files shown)


effigies:
In addition to this, it would be good to give some indication that the image is being downloaded. We could let `stderr` pass to the console in `check_memory` or add a `pull` step, which could either just directly output to the terminal or give some visual indication such as `Downloading......DONE`.

chrisfilo:
As discussed it would be good to pass the std(out/err) of docker from the wrapper to the console.

effigies:
Stderr is being passed to the console. Stdout is being parsed. The stdout is the output of the command being run, stderr that of docker.



Sent on my Samsung Galaxy S® 5


chrisfilo:
Just to clarify - if someone runs the wrapper without ever pulling the fmriprep image before they will see the docker output describing progress of downloading different layers?

effigies:
Yes. It's not the pretty progress bars that Docker gives you, but it gives output. (You can come see what it looks like, if you want.)

Issue #491, opened 2017-04-27
Handle IsRunning files in recon-all


effigies:
recon-all creates `IsRunning` files to indicate that it is currently running, and blocks other instances of recon-all from running. Some crashes aren't recovered from gracefully enough for recon-all to remove the file, resulting in a need to manually remove the file.

We have a few options:

1) Leave it to the user to remove; `recon-all` gives a reasonably thorough error message
2) Always remove; assume that nobody will ever run two instances of fmriprep on the same subject
3) Sometimes remove, based on a flag that indicates the user knows what they're doing?

I'm open to suggestions. I'm a little inclined to 1 or 3.

effigies:
@ianEisenberg As someone both using fmriprep and running into this issue, do you have any thoughts on the best way to resolve this?

IanEisenberg:
Is a 4th option not possible - recognize when such a crash has happened and remove the "IsRunning" file under those conditions?

chrisfilo:
Wouldn't this mean that if someone accidentally runs two instances of
FMRIPREP on the same subject (at the same time) two instances of freesurfer
will start running in parallel leading to all sort of unforeseen problems?

effigies:
@IanEisenberg We'd need a pretty reliable signature of a crash. There are a few different crash cases I can immediately think of:

* recon-all crashes, but fmriprep keeps running normally
  * Here we can catch the error and make the cleanup... But need to be careful *not* to, if the error was "Another recon-all instance may be running." This would be doable, but I suspect non-trivial.

* fmriprep crashes, interrupting recon-all
  * We *may* be able to catch this, depending on the nature of the crash. If it's a crash where we get a chance to clean up after ourselves, then we're in the same place as above, but a bit worse off.

* Our session is killed by the cluster manager or admin
  * Not clear how to identify this case in the first place. Especially when running on a cluster, the other copy (if it's running) is probably on a different machine.

@chrisfilo Yeah, that's the big thing we need to avoid, and why I'm a little inclined toward 3.

effigies:
One option would be to check the mtime on `recon-all.log`. If difference > N hours (for some sensible N), you can go ahead and delete.

A deeply hacky option on similar lines follows:

[...]

PR #492, opened 2017-04-27
[RTM] FIX: Check for expert options file

### LINKED ISSUE #425: Preexisting expert options

effigies:
Incorporates nipy/nipype#1966, which should resolve our self-created problems with expert options files.

@ianEisenberg If you have any datasets that are throwing the expert options error, could you try running this branch?

Closes #429 
Closes #425

### DIFF
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -28,5 +28,5 @@ dependencies:
     - svgutils
     - nitime
     - nilearn
-    - git+https://github.com/nipy/nipype.git@ccd0527d65c63b22b86b8e7df0d3103e4f4fe743#egg=nipype
+    - git+https://github.com/nipy/nipype.git@5a88840c31836d393cf8fea28de0777e4d4922c3#egg=nipype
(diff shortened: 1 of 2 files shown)


effigies:
I'll try testing this by running a new subject and cancelling before it gets very far. Might not get around to it until morning.

effigies:
In addition the the `expert-options` file, I did also get an `IsRunning` that needed to be deleted. I'm okay with updating the pin and merging this when nipy/nipype#1978 goes in, but we should make some decisions on #491.

PR #493, opened 2017-04-28
[RTM] Adding support for pepolar "fieldmaps"

### LINKED ISSUE #218: Implement movement parameters for ApplyTOPUP from mcflirt
Per conversation with Jesper Andersson

> Date:    Mon, 12 Sep 2016 12:45:40 +0000
> From:    Jesper Andersson <jesper.andersson@NDCN.OX.AC.UK>
> Subject: Re: Convert FUGUE fieldmaps to TOPUP fieldcoef
> 
> Dear Oscar,
> 
> I'd like to raise this question again, hoping that Jesper could jump in and enlighten me with some leads. This was the problem:
> 
> 1. I am building a workflow that can estimate fieldmaps using TOPUP when the input dataset is appropriate for that estimation and FUGUE when it comes with a phase difference map.
> 
> 2. I want to plug the estimated fieldmap (regardless the tool used for generation) into ApplyTOPUP for all cases. I don't want to use FUGUE at all for this.
> 
> 3. I have successfully created the workflow that estimates the fieldmap with FUGUE, converts it to Hz units, and with several tricks, I generate the corresponding fieldcoeff map. I also manipulate the headers so that they reproduce the headers you get for the fieldcoeff nifti you get with TOPUP.
> 
> 4. I use a fake movpar file with all zeros.
> 
> 5. All this is apparently working fine.
> 
> 6. Now, I want to use a real movpar file, generated with MCFLIRT but rotations are apparently exaggerated, translations do not look that bad.
> 
> These are my questions:
> 
> Q1. Do steps 1-5 look right? Am I missing something even though ApplyTOPUP seems to work fine?
> 
> This all looks good, though I can of course not say anything with certainty about for example your generation of the fieldcoeff map (which sounds like hard work). Why is it you want to use Applytopup instead of FUGUE? Would it not make sense to have these cases (dual-echotime fieldmap vs opposing PE volumes) as separate threads through the process (given that they have anyway diverged in how to estimate the field)?
> 
> Also, if these are diffusion data, why not use EDDY and skip the applytopup/fugue step altogether? The latest patch of eddy allows the input of a dual-echotime fieldmap instead of a topup-field.
> 
> Q2. Is the movpar file in the same format and conventions (order of columns, units, center of rotation) as the par file generated by MCFLIRT with save plots?
> 
> No, unfortunately they are different in every aspect. If you have the transformation matrices from flirt you can convert these to topup/eddy movement-parameters using the TOPUP::Matrix2MovePar global function that you can find in topup_file_io.h in the TOPUP project.
> 
> Jesper

The particular piece of code to use or reimplement is:
```
NEWMAT::ColumnVector Matrix2MovePar(const NEWMAT::Matrix&           M,
				    const NEWIMAGE::volume<float>&  vol)
{
  NEWMAT::ColumnVector mp(6); mp = 0.0;
  NEWMAT::ColumnVector rot(3);

  MISCMATHS::rotmat2euler(rot,M);
  mp.Rows(4,6) = rot;

  NEWMAT::Matrix MM = MovePar2Matrix(mp,vol);
  mp.Rows(1,3) = M.SubMatrix(1,3,4,4) - MM.SubMatrix(1,3,4,4);

  return(mp);  
}
```

MovPar2Matrix:
```
NEWMAT::Matrix MovePar2Matrix(const NEWMAT::ColumnVector&     mp,
                              const NEWIMAGE::volume<float>&  vol) 
{
  if (mp.Nrows() != 6) throw TopupFileIOException("MovePar2Matrix: mp must have 6 elements");

  NEWMAT::ColumnVector tmp(6);
  tmp(1) = mp(4); tmp(2) = mp(5); tmp(3) = mp(6);
  tmp(4) = mp(1); tmp(5) = mp(2); tmp(6) = mp(3);
  
  NEWMAT::ColumnVector cntr(3);
  cntr(1) = ((vol.xsize()-1)*vol.xdim())/2.0;
  cntr(2) = ((vol.ysize()-1)*vol.ydim())/2.0;
  cntr(3) = ((vol.zsize()-1)*vol.zdim())/2.0;

  NEWMAT::Matrix mat(4,4);
  MISCMATHS::construct_rotmat_euler(tmp,6,mat,cntr);

  return(mat);
}
```

rotmat2euler:
```

  int rotmat2euler(ColumnVector& angles, const Matrix& rotmat)
    {
      // uses the convention that R = Rx.Ry.Rz
      Tracer tr("rotmat2euler");
      float cz, sz, cy, sy, cx, sx;
      cy = std::sqrt(Sqr(rotmat(1,1)) + Sqr(rotmat(1,2)));
      if (cy < 1e-4) {
	//cerr << "Cos y is too small - Gimbal lock condition..." << endl;
	cx = rotmat(2,2);
	sx = -rotmat(3,2);
	sy = -rotmat(1,3);
	angles(1) = atan2(sx,cx);
	angles(2) = atan2(sy,(float)0.0);
	angles(3) = 0.0;
      } else {
	// choose by convention that cy > 0
	// get the same rotation if: sy stays same & all other values swap sign
	cz = rotmat(1,1)/cy;
	sz = rotmat(1,2)/cy;
	cx = rotmat(3,3)/cy;
	sx = rotmat(2,3)/cy;
	sy = -rotmat(1,3);
	//atan2(sin,cos)  (defined as atan2(y,x))
	angles(1) = atan2(sx,cx);
	angles(2) = atan2(sy,cy);
	angles(3) = atan2(sz,cz);
      }
      return 0;
    }
```

construct_rotmat_euler:
```
  int construct_rotmat_euler(const ColumnVector& params, int n, Matrix& aff,
			     const ColumnVector& centre)
    {
      Tracer tr("construct_rotmat_euler");
      ColumnVector angl(3);
      Matrix newaff(4,4);
      aff=IdentityMatrix(4);

      if (n<=0) return 0;
      // order of parameters is 3 rotation + 3 translation
      // angles are in radians 
      //  order of parameters is (Rx,Ry,Rz) and R = Rx.Ry.Rz
      angl=0.0;
      angl(1)=params(1);
      make_rot(angl,centre,newaff);
      aff = aff * newaff;
      if (n==1) return 0;

      angl=0.0;
      angl(2)=params(2);
      make_rot(angl,centre,newaff);
      aff = aff * newaff; 
      if (n==2) return 0;

      angl=0.0;
      angl(3)=params(3);
      make_rot(angl,centre,newaff);
      aff = aff * newaff;
      if (n==3) return 0;

      aff(1,4)+=params(4);
      if (n==4) return 0;
      aff(2,4)+=params(5);
      if (n==5) return 0;
      aff(3,4)+=params(6);
      if (n==6) return 0;

      return 1;
    } 
```

chrisfilo:
Closes #218. Comments:

- I've tried TOPUP but 3dQwarp gave visually better results. We might revisit this after consulting with FSL folks.
- I've tested in on the Beast (total of three iterations) and it works very well.
- I had to restrict the displacement to the phase encoding direction (which is not what other AFNI users usually do) - otherwise I was seeing distortion induced by the correction (visible in alignment of ventricles or shape of the corpus callosum).
- This method is sensitive to skull stripping. I had to change it from nilearn to BET. I've updated documentation and modified other subworkflows that performed skullstripping of the same type of data.
- I also tested this on LR/RL data - namely example HCP subject. It works, but the correction is not as impressive as the example in the HCP NeuroImage paper. More work might be required here.


fmriprep/workflows/fieldmap/unwarp.py
```diff
@@ -215,10 +221,270 @@ def init_sdc_unwarp_wf(reportlets_dir, ants_nthreads, fmap_bspline,
...
+    prepare_epi_opposite_wf = init_prepare_epi_wf(ants_nthreads=ants_nthreads,
+                                                  name="prepare_epi_opposite_wf")
+    prepare_epi_opposite_wf.inputs.inputnode.fmaps = usable_fieldmaps_opposite_pe
+
+    qwarp = pe.Node(afni.QwarpPlusMinus(pblur=[0.05, 0.05],
+                                        blur=[-1, -1],
+                                        noweight=True,
+                                        minpatch=9,
+                                        nopadWARP=True,
+                                        environ={'OMP_NUM_THREADS': str(ants_nthreads)},
```
chrisfilo: We should rename `ants_nthreads` to `omp_nthreads` - since both ANTs and AFNI use the same library similar limitations occur. As for freesurfer we will need to investigate more what is the optimal number of threads. I've asked on the mailing list and created and issue https://github.com/poldracklab/fmriprep/issues/499

fmriprep/workflows/fieldmap/unwarp.py
```diff
@@ -215,10 +221,270 @@ def init_sdc_unwarp_wf(reportlets_dir, ants_nthreads, fmap_bspline,
...
+
+    workflow.connect([
+        (inputnode, explicit_mask_epi, [('in_reference', 'in_file'),
+                                        ('in_mask', 'mask_file')]),
+        (inputnode, cphdr_warp, [('in_reference', 'hdr_file')]),
+        (qwarp, cphdr_warp, [('source_warp', 'in_file')]),
+        (cphdr_warp, to_ants, [('out_file', 'in_file')]),
+        (to_ants, unwarp_reference, [('out', 'transforms')]),
+        (inputnode, unwarp_reference, [('in_reference', 'reference_image')]),
+        (inputnode, unwarp_reference, [('in_reference', 'input_image')]),
```
chrisfilo: Yes that is the intention.

fmriprep/workflows/fieldmap/unwarp.py
effigies: That seems reasonable. FreeSurfer uses OMP_THREADS as well, so that makes sense. The only issue is that if you use `-parallel`, you might end up using `2x nthreads` when it parallelizes by hemisphere. (But then you spend a ton of time actually using 1 thread total even though nipype is blocking on the assumption that you're using `nthreads`.)

fmriprep/workflows/fieldmap/unwarp.py
```diff
@@ -215,10 +221,270 @@ def init_sdc_unwarp_wf(reportlets_dir, ants_nthreads, fmap_bspline,
...
+
+    merge = pe.Node(
+        StructuralReference(auto_detect_sensitivity=True,
+                            initial_timepoint=1,
+                            fixed_timepoint=True,  # Align to first image
+                            intensity_scaling=True,
+                            # 7-DOF (rigid + intensity)
+                            no_iteration=True,
+                            subsample_threshold=200),
+        name='merge')
```
effigies: Regarding this: `mri_robust_template` takes either 1 4D image or a series of 3D images.

https://github.com/freesurfer/freesurfer/blob/5367eec84621b331271e97f4e1f1502c3f6d0d87/mri_robust_register/MultiRegistration.cpp#L143-L298

chrisfilo: Thanks for the confirmation. I'm loving the fact freesurfer is in github!

fmriprep/workflows/fieldmap/unwarp.py
chrisfilo: Let's revisit this in another PR. Here's my email inquiry https://mail.nmr.mgh.harvard.edu/pipermail//freesurfer/2017-May/051705.html

effigies: Oh sure. This was hugely off-topic.

fmriprep/workflows/epi.py
```diff
@@ -161,47 +162,59 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
...
-        fmap_estimator_wf = init_fmap_estimator_wf(fmap_bids=fmaps,
-                                                   reportlets_dir=reportlets_dir,
+        # In case there are multiple fieldmaps prefer EPI
+        fmaps.sort(key=lambda fmap: {'epi': 0, 'fieldmap': 1, 'phasediff': 2}[fmap['type']])
+        fmap = fmaps[0]
+
+        LOGGER.info('Fieldmap estimation: type "%s" found', fmap['type'])
+
+        if fmap['type'] == 'epi':
+            epi_fmaps = [fmap['epi'] for fmap in fmaps if fmap['type']=='epi']
```
effigies: Spaces around `==`.

fmriprep/workflows/fieldmap/unwarp.py
```diff
@@ -24,14 +24,20 @@
 from nipype.pipeline import engine as pe
 from nipype.interfaces import utility as niu
 from nipype.interfaces import fsl
+from nipype.interfaces import afni
+from nipype.interfaces import freesurfer as fs
```
effigies: I don't think you're using `fs.*` in this file.

chrisfilo:
Let me know if there is anything else you would like me to change.

fmriprep/workflows/fieldmap/unwarp.py
```diff
@@ -24,14 +24,20 @@
...
-from niworkflows.interfaces.masks import ComputeEPIMask
+from niworkflows.interfaces.masks import BETRPT
 
 from fmriprep.interfaces import itk
 from fmriprep.interfaces import ReadSidecarJSON
 from fmriprep.interfaces.bids import DerivativesDataSink
 
+from nipype.interfaces import ants
+from fmriprep.interfaces import CopyHeader
+from fmriprep.interfaces.freesurfer import StructuralReference
```
effigies: `StructuralReference` is also imported into `fmriprep.interfaces`, if you want to combine these lines.

fmriprep/workflows/fieldmap/unwarp.py
```diff
@@ -215,10 +221,268 @@ def init_sdc_unwarp_wf(reportlets_dir, ants_nthreads, fmap_bspline,
 
     return workflow
 
+
+def init_pepolar_unwarp_wf(fmaps, bold_file, ants_nthreads, layout=None,
+                           fmaps_pes=None, bold_file_pe=None,
+                           name="pepolar_unwarp_wf"):
+    """
+    This workflow takes in a set of EPI files with opposite phase encoding 
+    direction than the target file and calculates a displacements field
```
effigies: "displacements field" shows up twice and "displacement field" twice. Is one of these a typo, or are they two different things?

fmriprep/workflows/fieldmap/unwarp.py
```diff
@@ -215,10 +221,268 @@ def init_sdc_unwarp_wf(reportlets_dir, ants_nthreads, fmap_bspline,
...
+    direction than the target file and calculates a displacements field 
+    (in other words, an ANTs-compatible warp file).
+    
+    This procedure works if there is only one '_epi' file is present
+    (as long as it has the opposite phase encoding direction to the target 
+    file). The target file will be used to estimate the field distorton.
+    However, if there is another '_epi' file present with a matching
+    phase encoding direction to the target it will be used instead.
+    
+    Currently different phase encoding dimension between target file and the
```
effigies: Comma after "Currently". And maybe "in the" instead of "between".

fmriprep/workflows/fieldmap/unwarp.py
```diff
@@ -215,10 +221,268 @@ def init_sdc_unwarp_wf(reportlets_dir, ants_nthreads, fmap_bspline,
...
+    (as long as it has the opposite phase encoding direction to the target 
+    file). The target file will be used to estimate the field distorton.
+    However, if there is another '_epi' file present with a matching
+    phase encoding direction to the target it will be used instead.
+    
+    Currently different phase encoding dimension between target file and the 
+    '_epi' file(s) (for example 'i' and 'j') is not supported.
+    
+    The warp field correcting for the distortions is estimated using AFNI's 
+    3dQwarp with displacement estimation limited to the target file phase
```
effigies: Comma after "3dQwarp".

fmriprep/workflows/fieldmap/unwarp.py
```diff
@@ -215,10 +221,268 @@ def init_sdc_unwarp_wf(reportlets_dir, ants_nthreads, fmap_bspline,
...
+
+        out_reference
+            the ``in_reference`` after unwarping
+        out_warp
+            the corresponding :abbr:`DFM (displacements field map)` compatible with
+            ANTs
+        out_mask
+            mask of the unwarped input file
+        out_mask_report
+            reportled for the skullstripping
```
effigies: reportlet

fmriprep/workflows/fieldmap/unwarp.py
```diff
@@ -215,10 +221,268 @@ def init_sdc_unwarp_wf(reportlets_dir, ants_nthreads, fmap_bspline,
...
+            if len(fmap_pe) != len(bold_file_pe):
+                add_list = usable_fieldmaps_opposite_pe
+            else:
+                add_list = usable_fieldmaps_matching_pe
+            add_list.append(fmap)
+
+    if len(usable_fieldmaps_opposite_pe) == 0:
+        raise Exception("None of the discovered fieldmaps has the right "
+                        "phase encoding direction. Possibly a problem with"
+                        "metadata. If not rerun with '--ignore fieldmaps' to"
```
effigies: Comma after "If not". Add a space at the end of each of the preceding two strings.

CHANGES.rst
```diff
@@ -3,6 +3,8 @@ Next release
 
 * [ENH] Use robust template generation for multiple T1w images (#481)
 * [ENH] Anatomical MNI outputs respect ``--output-space`` selection (#490)
+* [ENH] Added support for distortion correction using opposite phase encoding direction (#493)
```
effigies: "opposite phase encoding direction field maps"? Or "EPI images"?

fmriprep/workflows/epi.py
```diff
@@ -56,19 +57,19 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
...
+                }]
     else:
         metadata = layout.get_metadata(bold_file)
         # Find fieldmaps. Options: (phase1|phase2|phasediff|epi|fieldmap)
-        fmaps = layout.get_fieldmap(bold_file) if 'fieldmaps' not in ignore else {}
+        fmaps = layout.get_fieldmap(bold_file, return_list=True) if 'fieldmaps' not in ignore else []
 
     # TODO: To be removed (supported fieldmaps):
-    if not fmaps.get('type') in ['phasediff', 'fieldmap']:
+    if not set([fmap['type'] for fmap in fmaps]).intersection(['phasediff', 'fieldmap', 'epi']):
```
oesteban: Are `fmap` and `fmap['type']` always defined? Otherwise we might expect None has no Attribute and AttributeErrors here.

fmriprep/workflows/epi.py
```diff
@@ -161,47 +162,59 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
...
-                                                   reportlets_dir=reportlets_dir,
+        # In case there are multiple fieldmaps prefer EPI
+        fmaps.sort(key=lambda fmap: {'epi': 0, 'fieldmap': 1, 'phasediff': 2}[fmap['type']])
+        fmap = fmaps[0]
+
+        LOGGER.info('Fieldmap estimation: type "%s" found', fmap['type'])
+
+        if fmap['type'] == 'epi':
+            epi_fmaps = [fmap['epi'] for fmap in fmaps if fmap['type']=='epi']
+            sdc_unwarp_wf = init_pepolar_unwarp_wf(fmaps=epi_fmaps,
```
oesteban: I would imagine `init_pepolar_unwar_wf` does not estimate the fieldmap only, it also corrects for distortion right?

I think that's perfect for now, but we should aim at the original design: one thing is the estimation and a different issue is applying that estimation to in fact correct for distortions.

chrisfilo: There was no advantage in splitting those steps. This has been explained in the documentation.

fmriprep/workflows/epi.py
chrisfilo: pybids always returns fmap type. fmap is part of a for loop

effigies:
LGTM.

PR #494, opened 2017-04-28
[RTM] ENH: Validate graymid inputs from FreeSurferSource

effigies:
`FreeSurferSource` returns an `OutputMultiPath`, so the `graymid` input should follow suit. One particular possibility is that a pre-computed subject directory may have `?h.graymid` in it and we create `?h.midthickness` (copies of `?h.graymid`, since the introduction of `MakeMidthickness`).

A second run-through would receive both, which would cause an error. This change accepts either as input, preferring `?h.midthickness` when available.

It also rejects files from the wrong hemisphere, which means this now can be safely used from a `FreeSurferSource` that did not specify hemisphere.

@chrisfilo Possibility of pre-computed subject dirs containing this file is what convinced me to do this. (I'm test running fmriprep on my PhD data, which contains `?h.graymid` surfaces.)

### DIFF
--- a/fmriprep/interfaces/freesurfer.py
+++ b/fmriprep/interfaces/freesurfer.py
@@ -9,7 +9,7 @@ FreeSurfer tools interfaces
 """
 from __future__ import print_function, division, absolute_import, unicode_literals
 
-from nipype.interfaces.base import isdefined, File
+from nipype.interfaces.base import isdefined, InputMultiPath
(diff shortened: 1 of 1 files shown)


chrisfilo:
That makes a lot of sense. There will be manu usecases with preexisting freesurfer directories.

Testing FMRIPREP on your PhD data is a great idea!

PR #501, opened 2017-05-02
[RTM] CI: Fix typo in .circle/tests.sh

effigies:
Currently `ds054/scratch` isn't being saved to artifacts as it is supposed to be. This checks whether it's a permissions issue.


effigies:
This should go in after the build is fixed by #503.

PR #504, opened 2017-05-03
[RTM] fixing header issues in HMC

### LINKED ISSUE #497: Output EPI voxel size slightly different from input
It happened to me occasionally in some runs of some subjects.  The input voxel size was always `(3.0, 3.0, 3.0)` (as shown by `nib.load(filename).header.get_zooms()[:3]`), but the output voxel size became `(2.9999998, 3.0, 3.0)`.  An annoying side effect of this discrepancy is the output grid of that run became `(66, 77, 65)` instead of `(65, 77, 65)`, and the two grids were half a voxel apart and I couldn't simply delete a slice to make them match.

As I tracked down this issue, the voxel size of the generated reference image was already different, and influenced all following steps.  It seems related to [`EstimateReferenceImage`](https://github.com/poldracklab/niworkflows/blob/master/niworkflows/interfaces/registration.py#L261).

The filename of my singularity image was `poldracklab_fmriprep_latest-2017-04-08-0600a0f79358.img`.

chrisfilo:
* Fix header issue caused by bias field correction. Closes #497 
* Use reference image with bias field instead of after correction when doing motion correction. The reasoning is that we should match the bias field between the reference image and the 4D bold volume (which is not corrected for bias).


chrisfilo:
Waiting for @feilong to confirm this fix works.

fmriprep/workflows/epi.py
```diff
@@ -289,12 +289,16 @@ def normalize_motion_func(in_file, format):
...
         (inputnode, gen_ref, [('epi', 'in_file')]),
         (gen_ref, inu, [('ref_image', 'input_image')]),
-        (inu, hmc, [('output_image', 'ref_file')]),
-        (inu, skullstrip_epi, [('output_image', 'in_file')]),
-        (inu, outputnode, [('output_image', 'ref_image')]),
+        (inu, cphdr, [('output_image', 'in_file')]),
+        (gen_ref, cphdr, [('ref_image', 'hdr_file')]),
+        (gen_ref, hmc, [('ref_image', 'ref_file')]),
+        (cphdr, skullstrip_epi, [('out_file', 'in_file')]),
+        (cphdr, outputnode, [('out_file', 'ref_image')]),
```
effigies: I'm trying to understand this a bit better. `cphdr.out_file` is just the bias-field corrected image with the original header. We're sure that this is safe -- i.e. that a different header on `inu.output_image` doesn't reflect a change in the data?

chrisfilo: That is the assumption. the problem is that `N4BiasFieldCorrection` strips/modies the affines in the header.

effigies: Really? Do they give a reason, or is this a bug we should report? And is this the primary use for the `CopyHeader` interface? If so, maybe we should make a reusable workflow that is simply `N4BiasFieldCorrection` + `CopyHeader` so that it's obvious why we're doing this.

chrisfilo: We should report it and yes - we use it enough to create a wee workflow for this.

feilong:
Last commit (91161f7f1a8058c2de31897a26524fd84255303a) gave me correct headers and resolutions for all my subjects.
With the most recent commit (f4bf657361a320514fd7dfd503d2799967b21173) the problems are back again.

fmriprep/workflows/epi.py
```diff
@@ -27,12 +27,13 @@
 
 from fmriprep.interfaces import DerivativesDataSink
 
-from fmriprep.interfaces.images import GenerateSamplingReference
+from fmriprep.interfaces.images import GenerateSamplingReference, CopyHeader
```
effigies: You can drop the `CopyHeader` import.

feilong:
Commit 51bab778107fad6f2935c3b90ba5ea96965f8cb7 works well on my data.

chrisfilo:
That's great news - just to confirm did your test include upgrading niworkflows (which is part of this PR)?

feilong:
I got some problems when I first tried to use the latest commit, so I upgraded niworkflows, nipype, and pybids to versions shown in `requirements.txt`

chrisfilo:
Awesome! Thank you so much for testing this. I feel much better knowing we have resolved this issue!

feilong:
You are most welcome!  Thanks for fixing this issue!

PR #506, opened 2017-05-03
[RTM] RF: Split recon-all into coarse chunks to improve resource use estimates

effigies:
This PR eschews the `-parallel` flag in FreeSurfer, preferring the directives that are volume or hemisphere specific. Thus, we run:

* `-autorecon1` - Single thread (`-parallel` has no effect, but `-openmp` does in places)
* Inject externally skull-stripped image
* `-autorecon2-volonly` - Single thread
* `-autorecon-hemi {lh,rh}` - Replace `-parallel` with nipype parallelism
* `-all` - Run stats and any other post-hemisphere-specific processing

Also takes advantage of #494 to get rid of the separate `get_surfaces` node.

### DIFF
--- a/fmriprep/workflows/anatomical.py
+++ b/fmriprep/workflows/anatomical.py
@@ -356,12 +356,13 @@ def init_surface_recon_wf(omp_nthreads, hires, name='surface_recon_wf'):
     reconhemis.interface.num_threads = omp_nthreads
     reconhemis.inputs.hemi = ['lh', 'rh']
 
-    reconall = pe.Node(
+    autorecon3 = pe.Node(
         ReconAllRPT(
+            directive='autorecon3',
             openmp=omp_nthreads,
             generate_report=True),
-        name='reconall')
-    reconall.interface.num_threads = omp_nthreads
+        name='autorecon3')
+    autorecon3.interface.num_threads = omp_nthreads
 
     fs_transform = pe.Node(
         fs.Tkregister2(fsl_out='freesurfer2subT1.mat', reg_header=True),
@@ -450,13 +451,13 @@ def init_surface_recon_wf(omp_nthreads, hires, name='surface_recon_wf'):
                                           ('subject_id', 'subject_id')]),
         (autorecon2, reconhemis, [('subjects_dir', 'subjects_dir'),
                                   ('subject_id', 'subject_id')]),
-        (reconhemis, reconall, [(('subjects_dir', _dedup), 'subjects_dir'),
-                                (('subject_id', _dedup), 'subject_id')]),
-        (reconall, save_midthickness, [('subjects_dir', 'base_directory'),
-                                       ('subject_id', 'container')]),
-        (reconall, outputnode, [('subjects_dir', 'subjects_dir'),
-                                ('subject_id', 'subject_id'),
-                                ('out_report', 'out_report')]),
+        (reconhemis, autorecon3, [(('subjects_dir', _dedup), 'subjects_dir'),
+                                  (('subject_id', _dedup), 'subject_id')]),
+        (autorecon3, save_midthickness, [('subjects_dir', 'base_directory'),
+                                         ('subject_id', 'container')]),
+        (autorecon3, outputnode, [('subjects_dir', 'subjects_dir'),
+                                  ('subject_id', 'subject_id'),
+                                  ('out_report', 'out_report')]),
         # Reconstruction phases
         (inputnode, autorecon1, [('t1w', 'T1_files')]),
         (recon_config, autorecon1, [('t2w', 'T2_file'),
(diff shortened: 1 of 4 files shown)


effigies:
Quick update: `-autorecon-hemi` pipelines have a couple statistics steps that presume both sides were already done. Nothing depends on them, so I just let them be run in the `-autorecon3` step.

Re-running from scratch to verify.

fmriprep/workflows/anatomical.py
```diff
@@ -341,32 +340,46 @@ def inject_skullstripped(subjects_dir, subject_id, skullstripped):
...
+
+    autorecon_surfs = pe.MapNode(
+        fs.ReconAll(
+            directive='autorecon-hemi',
+            flags=['-noparcstats', '-noparcstats2', '-noparcstats3'],
+            openmp=omp_nthreads),
+        iterfield='hemi',
+        name='autorecon_surfs')
+    autorecon_surfs.interface.num_threads = omp_nthreads
+    autorecon_surfs.inputs.hemi = ['lh', 'rh']
```
effigies: On a 6-core machine, I'm getting both 5-thread jobs in this mapnode running simultaneously. @oesteban I think you're more familiar with the resource manager than I am. Do you see something wrong here?

chrisfilo: doesn't omp_nthreahds default to nthreads-1?

effigies: Yes. But we shouldn't be able to run two 5-thread jobs on a 6-thread queue... I think?

chrisfilo: Yes that is correct. I wonder if this is because MultiProc is not handling MapNodes correctly.

effigies: Yeah, I think the possibilities are:

1) MapNodes are handled incorrectly by MultiProc
2) MapNodes need to have `num_threads` specified differently from other nodes
3) I've screwed something subtle up here

chrisfilo: Might be worth opening an issue at nipy/nipype and pinging Cameron.

effigies:
Race conditions involving `mris_anatomical_stats` are now avoided by removing the `-parcstats*` and `-balabels` steps from `-autorecon-hemi` and letting them run in `-autorecon3`.

Switched to an iterable to avoid the MapNode issue (nipy/nipype#2016). (May revert: nipy/nipype#2017)



Can verify that one node runs if `omp-nthreads > nthreads / 2`, and both if `omp-nthreads == nthreads / 2`.

chrisfilo:
This looks great - let me know when it will be ready for review.

effigies:
Well, I thought this was ready to go, and was just running one last full recon... Turns out that there are more cross-hemisphere dependencies (that only show up when you run them in sequence). I think we'll need one extra node to handle these dependencies. I'm splitting the post-skullstrip recon bit into its own workflow just to keep the logic separate.

This will be easier once nipy/nipype#2017 goes in, since `MapNode`s handle syncing naturally without `JoinNode`s.

effigies:
@chrisfilo This is ready for review. Let's wait on nipype 0.13.1 to merge. If either nipy/nipype#2019 or nipy/nipype#2020 go in, I'll be able to make a couple edits that make it more concise.

Relatedly, with the nipype release(s), we may want to revisit our dependency strategy, as we discussed at one point in January/February.

effigies:
Final commit is for the sake of comparing the hierarchical graph to the simple graph, to see which we'd rather have in the docs.

effigies:
Here are our doc options:

Simple:



Colored:

chrisfilo:
LGTM have you noticed any influence on performance?

effigies:
I've really only tested it on a machine without enough cores to get workflow-level parallelism, to make sure that it doesn't depend on these per-hemisphere stages moving in lockstep. I can try to get timing info from a 16 core machine tonight.

Any preference on the graph for the docs? I have one more update to make to the docs before this is fully ready.

chrisfilo:
I prefer the simpler one (but not strongly).

effigies:
On a 16-core machine, running identical commands, I see effectively no difference in total fmriprep runtime between 0.4.3 and this branch, both right around 6 hours total.

I have start and end times for the entire pipeline instances. I forgot to save detailed logs, so my terminal history is all I have. I can verify that we did get parallel execution in both cases, but in the 0.4.3 branch, that's because we were using the `-parallel` flag, so we were using 2n threads even though we only claimed n.

chrisfilo:
Thanks for looking into this. If you could sync with Master so the tests will start passing I'm happy to merge.

effigies:
Woops. Rebased this with a docs-only on the end. Full tests running now.

PR #507, opened 2017-05-04
[RTM] Add run_without_submitting to simple functions, datasinks

effigies:
While resuming a mostly-completed workflow, I noticed a number of simple nodes being scheduled, running in <2 seconds, and thus causing the scheduler to wait unnecessarily.

I don't know if this will even buy more than a minute speedup for a large, multi-subject run, so consider this a super low-priority review.


chrisfilo:
I wonder if this would make sense for niu.Merge MapNodes.

effigies:
Probably. Most of the `niu` nodes, really, except for `Function` nodes that do real processing. (My rule was: If it reads and writes a file bigger than an affine matrix, leave it alone.) I didn't think to look for `Merge`.

PR #509, opened 2017-05-04
CI: Add a flag to commits to instruct Circle only to build docs

effigies:
Assuming this works, we'll be able to tag commits that don't need to run real code.

### DIFF
--- a/.circle/tests.sh
+++ b/.circle/tests.sh
@@ -15,6 +15,20 @@ if [ "${CIRCLE_NODE_TOTAL:-}" != "2" ]; then
   exit 1
 fi
 
+# Only run docs if DOCSONLY or "docs only" (or similar) is in the commit message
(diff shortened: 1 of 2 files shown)


chrisfilo:
Super handy!

effigies:
Well that went quickly. 15min build time, 10 of that `docker pull` + `docker build`.

Any commit message that contains (case-insensitive) `docs only` `docs_only` or `docsonly` will trigger this condition. If we'd prefer, I can change it to some other set of triggers.

oesteban:
Have we tried a commit that builds everything? I'm implementing the same idea to mriqc and it seems to identify all commits as DOCSONLY.

chrisfilo:
We looked at CircleCI together @oesteban and everything looks good.

oesteban:
Sorry, I misread the circle.yml file.

PR #510, opened 2017-05-05
RF: Add init_n4bias_wf to run N4 bias correction

effigies:
As discussed in #504, this wraps N4BiasFieldCorrection into a small workflow that copies the original header to the bias-corrected image.


oesteban:
This looks good to me. Do you think it'd be worth adding to nipype? at the interface level itself. @chrisfilo ?

effigies:
I'm good with moving it to niworkflows (would it be useful in MRIQC?) or nipype, or with keeping it here for now.

If we don't specifically need it for MRIQC, let's maybe merge it here, propose it in nipype, and then switch to nipype's version once it's accepted upstream?

oesteban:
Sure, this should be included in fmriprep ASAP. I meant that in the long term this is something every nipype user may want.

PR #522, opened 2017-05-16
[RTM] Accommodate Sphinx >=1.6 warning behavior

### LINKED ISSUE #521: Circle builds failing due to old AFNI
Looks like Sphinx has a [new release out today](https://pypi.python.org/pypi/Sphinx/1.6.1), so we're getting nipype warnings raised to Sphinx warnings, which are treated as errors.

```
AFNI is outdated, detected version Debian-16.2.07~dfsg.1-2~nd16.04+1 and AFNI_17.1.05 is available.
```

effigies:
Closes #521

### DIFF
--- a/fmriprep/__init__.py
+++ b/fmriprep/__init__.py
@@ -23,3 +23,24 @@ from .info import (
     __description__,
     __longdesc__
 )
+
+import warnings
+
+# cmp is not used by fmriprep, so ignore nipype-generated warnings
+warnings.filterwarnings('ignore', r'cmp not installed')
+
+# Monkey-patch to ignore AFNI upgrade warnings
+from nipype.interfaces.afni import Info
+
+_old_version = Info.version
+def _new_version():
+    from nipype import logging
+    iflogger = logging.getLogger('interface')
+    level = iflogger.getEffectiveLevel()
+    iflogger.setLevel('ERROR')
+    v = _old_version()
+    iflogger.setLevel(level)
+    if v is None:
+        iflogger.warn('afni_vcheck executable not found')
+    return v
+Info.version = staticmethod(_new_version)
--- a/fmriprep/info.py
+++ b/fmriprep/info.py
@@ -57,7 +57,7 @@ TESTS_REQUIRES = [
 ]
 
 EXTRA_REQUIRES = {
-    'doc': ['sphinx<1.6.1', 'pydotplus', 'sphinx_rtd_theme', 'sphinx-argparse'],
+    'doc': ['sphinx>=1.5.3', 'pydotplus', 'sphinx_rtd_theme', 'sphinx-argparse'],
     'tests': TESTS_REQUIRES,
     'duecredit': ['duecredit']
 }
--- a/fmriprep/workflows/epi.py
+++ b/fmriprep/workflows/epi.py
@@ -54,7 +54,7 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
     # For doc building purposes
     if layout is None or bold_file == 'bold_preprocesing':
 
-        LOGGER.warning('No valid layout: building empty workflow.')
+        LOGGER.info('No valid layout: building empty workflow.')
         metadata = {"RepetitionTime": 2.0,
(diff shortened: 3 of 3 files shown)


chrisfilo:
Suppressing all warnings is a bit dangerous. Could we do something more nuanced using filtering? https://docs.python.org/3.1/library/warnings.html#the-warnings-filter

effigies:
I'll look into it.

effigies:
Logger warnings are homogeneous, as far as I can see. I can switch to reducing warnings in `docs/conf.py`, rather than in the program in general.

chrisfilo:
It seem that you can filter warning using regular expression on the text of the warning (unless I am misreading the documentation). This would allow us to suppress warning that we know can be ignored (for example the AFNI version).

effigies:
Gotcha. Pushed a few filters. Also included some that I saw in the docs output, even though they don't seem to be causing sphinx to throw up. We've had users mention the cmp warning a couple times, I think.

Let me know what you think.

effigies:
Well, looks like ignored warnings can still be caught by Sphinx. Want to keep the fitlers, but restore the `setLevel`s from 0815732?

chrisfilo:
Does the filtering of warnings by content work at all? I cannot see since all the CircleCI builds are docs only.

effigies:
Amended the commit so the full build will run.

chrisfilo:
The warnings seem to still be there when running pipelines:

```
170516-21:29:49,255 interface WARNING:

	 AFNI is outdated, detected version Debian-16.2.07~dfsg.1-2~nd16.04+1 and AFNI_17.1.05 is available.
```

effigies:
I don't think we can suppress logged warnings without changing the log level.

However, search for "Fieldmap estimation:" in the output of (container 0) builds [440](https://circleci.com/gh/effigies/fmriprep/440) and [428](https://circleci.com/gh/effigies/fmriprep/428). In 440, the `cmp not installed` warning is suppressed from the output.

chrisfilo:
I see - so such suppression would require changes at nipype level right?

effigies:
I'm not sure what they'd do differently. They use the standard logger to log messages of different [criticality](https://docs.python.org/3/library/logging.html#logging-levels).

If our goal is to suppress the AFNI message, we could monkey-patch `interfaces.afni.Info` to not check for new AFNI. This might be a good idea anyway, so that running FMRIPREP doesn't have to try to make an outgoing HTTP connection every time an AFNI interface that checks version information is run.

Or we could try to convince Neurodebian to upgrade their AFNI packages.

chrisfilo:
Would it be hard to monkey patch this or should we do it in another PR?

effigies:
Proposed a monkey patch. We'll see how it goes.

effigies:
Monkey patch seems to work. matplotlib warnings aren't suppressed in doc building, and only show up in doc building, so not sure those are worth filtering.

cmp warning does show up in container 0 runtime tests, and this does filter those. So I'll remove the AFNI and matplotlib filters, but keep cmp. Sound reasonable?

docs/conf.py
```diff
@@ -15,6 +15,11 @@
 import os
 import sys
 import subprocess
+from nipype import logging
+
+# Disable informational interface warnings
+logging.getLogger('interface').setLevel('ERROR')
+logging.getLogger('workflow').setLevel('ERROR')
```
chrisfilo: If monkey patching works this should not be necessary anymore right?

effigies: True. Let me drop those. Do we need a changelog entry?

chrisfilo:
I think we can skip it for this one.

On Wed, May 17, 2017 at 10:25 AM, Chris Markiewicz <notifications@github.com

effigies:
Oh right. We're getting [this warning](https://github.com/poldracklab/fmriprep/blob/master/fmriprep/workflows/epi.py#L57) now. That at least was being suppressed by setting the workflow logger level to 'ERROR' in `docs/conf.py`. Should I revert to that, or do you have another preferred solution?

chrisfilo:
I would prefer not to do a blanket warning suppression. Changing the level of this one log message to INFO seems less error prone.

PR #523, opened 2017-05-16
[RTM] move missing values to the beginning of the confounds file

chrisfilo:
It's more intuitive in context of using those for nuissance regression.


effigies:
This is beyond my pandas experience. I'm willing to take your word that this does what you say, though.

chrisfilo:
Let me add a changelog entry

effigies:
Resolved conflicts in changelog. I can't cancel the resulting build for you, but there's no need for it to run to merge this.

PR #524, opened 2017-05-16
[RTM] fix the version of sphinx (for now)

chrisfilo:
This would allow us to evaluate other PRs until #522 will be finalized.


effigies:
Sure, if #522 isn't ready to go when the current tests finish.

oesteban:
I think we need sphinx>=1.4  for the nipype extension. But I guess that then the pinning in nipype would apply here and sphinx will be >=1.4

Issue #334, opened 2017-01-13
Temporal ICA for artifact detection.


rwblair:
FSL FIX is a tool that can do this but it requires a model to be trained.

chrisfilo:
Another option that does not require training is ICA-AROMA: https://github.com/rhr-pruim/ICA-AROMA

The easiest way is to apply it at the very end - after resampling to MNI space.

oesteban:
Closed here I guess #539

PR #539, opened 2017-05-27
WIP: Adding ICA_AROMA functionality

jdkent:
This is my attempt at adding ICA_AROMA functionality to fmriprep. Almost 100% certain it doesn't work as is.

I have ICA_AROMA completing it's own transforms since it appears the template space could be something something different than ICA AROMA is expecting, in addition, ICA_AROMA requires a affine fsl matrix and fnirt fsl warp to pass in so ANTs is off the table (afaik). 

I may continue editing/working on this, but first I wanted feedback if this is a worthwhile endeavour or if it's better to hi-jack the original code from ICA-AROMA to make it flexibly fit in the fmriprep workflows (as was done with compcor).

Thanks!
James


chrisfilo:
Hey. This is a great contribution. In fact we planned to do this - see https://github.com/poldracklab/fmriprep/issues/334#issuecomment-283877729 so I am really happy to see your contribution! I prefer reusing rather than reinventing so using the Nipype interfaces as you are doing is preferable to hijacking the code.

A few things I would suggest:
- Do we need to perform the normalization again using FNIRT or maybe we can just reused the normalized times generated using ANTs (outputs of `epi_mni_trans_wf`)? It would save some time.

- We should make sure the smoothing is performed with the same algorithm/parameters as in the original paper to match the training set.

Things worth considering, but potentially a separate PR:

- It would be great to add visualization of the good/bad components to the report. I wrote a generic melodic reportlet generator that could be reused here: https://github.com/poldracklab/niworkflows/pull/134. This would require running melodic outside of aroma, but is should support this use case.

- Not sure we need to make users chose between aggressive and non-agressive aggressors. We could output both sets in the confounds.tsv file as a default. There should be a way to generate both sets without much computational overhead. It would avoid users having to rerun all of fmriprep if they change their minds.

Once again - thanks for this PR!

jdkent:
Thanks you for your comments! Would the template: MNI152NLin2009cAsym always have to be specified to run ICA_AROMA, since it looks like it's currently not mandatory.

 

- I can see about reporting the good/bad components, but I will at least run melodic separately to generate the report.

- aggressive v. non-aggressive. This is completed via fsl_regfilt, where the non-aggressive denoising is completed through partial regression (e.g. variance shared with the other components is kept in the data) versus full regression (e.g. the shared variance is removed from the data). So I would want to return the coefficients from the partial regression, I think. fsl_regfilt does not return the coefficients sadly, so I would need to calculate that myself, perhaps with [this?](http://scikit-learn.org/stable/modules/generated/sklearn.cross_decomposition.PLSRegression.html)

- It would be easier just to pass in the epi_mni and forgo the transforms, is this less recommended? I guess I'm less sure about whether the ouputs are organized so we can pass one affine and one warp if there are fieldmaps in the processing stream. 

- I will look into the published smoothing method

chrisfilo:
On Sat, May 27, 2017 at 5:33 PM, James Kent <notifications@github.com>
wrote:

> Would the template: MNI152NLin2009cAsym always have to be specified to run
> ICA_AROMA, since it looks like it's currently not mandatory.
>
Currently ANTs normalization is only performed if "template" is specified
as one of output spaces (which is the default). I would change this to "or
ICA AROMA is required".

>
>    -
>
>    I can see about reporting the good/bad components, but I will at least
>    run melodic separately to generate the report.
>
> Great!

>
>    -
>
>    aggressive v. non-aggressive. This is completed via fsl_regfilt, where
>    the non-aggressive denoising is completed through partial regression (e.g.
>    variance shared with the other components is kept in the data) versus full
>    regression (e.g. the shared variance is removed from the data). So I would
>    want to return the coefficients from the partial regression, I think.
>    fsl_regfilt does not return the coefficients sadly, so I would need to
>    calculate that myself, perhaps with this?
>    <http://scikit-learn.org/stable/modules/generated/sklearn.cross_decomposition.PLSRegression.html>
>
> I don't think we need to run fsl_regfilt for this. The aggressive
regressors would be just the timecourses of the "bad" ICs. The
non-agressive regressors would be the timecourses of "bad" ICs after
regressing out timecourses of "good" ICs. The main difference is that the
non-agressive regressors do not share any variance with timecourses of
"good" ICs. This should be pretty straightforward to do, but we can leave
it to a separate PR.

>
>    -
>
>    It would be easier just to pass in the epi_mni and forgo the
>    transforms, is this less recommended? I guess I'm less sure about whether
>    the ouputs are organized so we can pass one affine and one warp if there
>    are fieldmaps in the processing stream.
>
> I think ICA AROMA has a mode where you do not pass any transformations and
the inputs are just assumed to be in the MNI space. We should use that.

>
>    -
>
>    I will look into the published smoothing method
>
> Thanks!

jdkent:
I'm still not done with this. Things I still need to work on:

- ica aroma partial regression: I've given it some effort, but I can't replicate the results from fsl_regfilt for non-aggressive denoising. My attempt followed this line of reasoning.
linear regression: y_hat = mx +b, so the residual (or partial) after regression should be y - y_hat, right? I can send a link to some example data if what I did wasn't a simple mathematical error.

- I still need to work on the melodicRPT, I'll continue reading up on that to add that functionality.

- clean up the code and get formatting consistent. I still don't have a good sense of what should be a method, a node, and a workflow, and I saw in a tutorial that some simple mathematical operations (like number output from ImageStats) can be modified before being passed into another node. 

Just making an update to keep consistent with working on the code/learn the nipype framework.

chrisfilo:
On Wed, May 31, 2017 at 8:35 PM, James Kent <notifications@github.com>
wrote:

> I'm still not done with this. Things I still need to work on:
>
>    -
>
>    ica aroma partial regression: I've given it some effort, but I can't
>    replicate the results from fsl_regfilt for non-aggressive denoising. My
>    attempt followed this line of reasoning.
>    linear regression: y_hat = mx +b, so the residual (or partial) after
>    regression should be y - y_hat, right? I can send a link to some example
>    data if what I did wasn't a simple mathematical error.
>
> That sounds correct.  Are you including all "good" IC in the right hand
side of the equation? We can also leave it to another PR.

>
>    -
>
>    I still need to work on the melodicRPT, I'll continue reading up on
>    that to add that functionality.
>
> Let me know if you have any questions. Here's an example how this can be
used:
https://github.com/poldracklab/mriqc/blob/91368d14167267009bea81ab833c4a76dd919371/mriqc/workflows/functional.py#L167

jdkent:
RE: regression
here is the relevant bit of code:
```   
def calc_residuals(x,y):
        X = np.column_stack(x+[[1]*len(x[0])])
         beta_hat = np.linalg.lstsq(X,y)[0]
         y_hat = np.dot(X,beta_hat)
         residuals = y - y_hat
 
         return residuals
```
Where y is the "bad" motion IC, and x is a matrix of the good_ICs. Maybe will leave it to a separate PR if I can't figure it out soon. I asked on the [FSL listserve](https://www.jiscmail.ac.uk/cgi-bin/webadmin?A2=FSL;719d1657.1706) to possibly get an answer

chrisfilo:
Great move!

jdkent:
Okay, I'm getting closer to making this a reality, there are still several TODOs before it's ready for a formal review:

- I can't get the basic melodic reportlet to work, I get the following error:
```
170612-01:58:13,360 workflow INFO:
	 Executing node ds_ica_aroma_report in dir: /out/work2/fmriprep_wf/single_subject_controlGE159_wf/func_preproc_task_flanker_wf/func_reports_wf/ds_ica_aroma_report
170612-01:58:13,463 workflow ERROR:
	 ['Node ds_ica_aroma_report failed to run on host 6c638518fad3.']
170612-01:58:13,466 workflow INFO:
	 Saving crash info to /out/fmriprep/sub-controlGE159/log/20170612-015349_3ed5f63d-a4cb-4a3c-9244-2955be966222/crash-20170612-015813-root-ds_ica_aroma_report-8fc746fc-5653-40cb-99ba-ced522bd80e3.pklz
170612-01:58:13,467 workflow INFO:
  File "/usr/local/miniconda/lib/python3.6/site-packages/nipype/pipeline/plugins/multiproc.py", line 322, in _send_procs_to_workers
    self.procs[jobid].run()
  File "/usr/local/miniconda/lib/python3.6/site-packages/niworkflows/nipype/pipeline/engine/nodes.py", line 372, in run
    self._run_interface()
  File "/usr/local/miniconda/lib/python3.6/site-packages/niworkflows/nipype/pipeline/engine/nodes.py", line 482, in _run_interface
  File "/usr/local/miniconda/lib/python3.6/site-packages/niworkflows/nipype/pipeline/engine/nodes.py", line 613, in _run_command
  File "/usr/local/miniconda/lib/python3.6/site-packages/niworkflows/nipype/interfaces/base.py", line 1066, in run
    self._check_mandatory_inputs()
  File "/usr/local/miniconda/lib/python3.6/site-packages/niworkflows/nipype/interfaces/base.py", line 971, in _check_mandatory_inputs
    raise ValueError(msg)
ValueError: DerivativesDataSink requires a value for input 'in_file'. For a list of required inputs, see DerivativesDataSink.help()
```
^it looks like a simple error but I can't find the solution right now.
[Here is the link](https://drive.google.com/open?id=0B9eMc5Q-tEdSdkw3Q05BdlRMQWM) to the workflow graph made, in case I connected the nodes incorrectly. 

- Have ica_aroma output in the correct directory: right now all the output is going into a /tmp/ directory, but I think with nipy/nipype#2056, this issue should be fixed as soon as niworkflows uses the updated version of nipype.

- change the testing suite to reflect ica aroma functionality.

- I've changed the dockerfile to include fsl atlases, but I don't know if this is needed, or if I could make a symbolic link between the template image fmriprep already uses to the standard place ica aroma expects the MNI images.

Good news: within a docker container, I have it working (albiet the melodic report error above), but the confounds tsv file is generated with the expected output (the partial regression implemented here is not identical to fsl_regfilt's partial regression, but it's close)

chrisfilo:
Great progress! I had a look at the code, but unfortunately I cannot see what could be wrong with the way you use MELODICRPT. Things worth checking:

- Is the SVG file present in the working directory of the MELODICRPT node? You might need to set `remove_unnecessary_outputs` to false in nipype.cfg

- the only difference between how you are using MELODICRPT and how it it [used in mriqc](https://github.com/poldracklab/mriqc/blob/cdca267d5ccdd5ab92022f44b6aa2f929e33fdf0/mriqc/workflows/functional.py#L174) is setting the report_mask. This should not be necessary, but it is a good idea anyway since you already have a mask available and there is no need to calculate it again the the MELODICRPT.

fmriprep/workflows/confounds.py
```diff
@@ -7,21 +7,24 @@
...
 from niworkflows.interfaces.masks import ACompCorRPT, TCompCorRPT
+from niworkflows.interfaces import segmentation as nws
+
 
 from niworkflows.nipype.interfaces.nilearn import SignalExtraction
 from fmriprep.interfaces.utils import prepare_roi_from_probtissue
 
 
-def init_discover_wf(bold_file_size_gb, name="discover_wf"):
+def init_discover_wf(bold_file_size_gb, name="discover_wf",use_aroma=False):
```
chrisfilo: `use_aroma` should not have a default set here. This might be the core of the error you are getting.

fmriprep/workflows/confounds.py
```diff
@@ -243,3 +263,172 @@ def reverse_order(inlist):
...
+            [('out_dir','melodic_dir')]),
+        #geneerate tsvs from ICA_AROMA
+        (ica_aroma, ica_aroma_confound_node,
+            [('out_dir','ica_out_dir')]),
+        #output for processing and reporting
+        (ica_aroma_confound_node, outputnode,
+            [('aroma_confounds','aroma_confounds')]),
+            #TODO change melodic report to reflect noise and non-noise components
+        (melodic, outputnode,
+            [('out_report','ica_aroma_report')]),
```
effigies: `ica_aroma_report` => `out_report`

chrisfilo: Nice catch!

jdkent:
Thanks @chrisfilo and @effigies, you guys are rockstars! That fixed the report error I was getting in the workflow. Sorry that re-running takes forever, when nipy/nipype#2056 is implemented, debugging will be way faster (I think).

jdkent:
I think the code is pretty much ready for a review, only a couple items remain that I know of:

-  The melodic report may need some extra space on the report to load correctly (I think i'll leave the labeling of the components as good/bad to a separate PR; right now people can painstakingly reference the components in the confounds.tsv with the melodic report.)

- Still need to wait on the updated nipype to automatically set the output directory of ICA_AROMA

- misc: formatting code to be in line with PEP8 standards

- Add more documentation to places I may have forgotten.

fmriprep/workflows/epi.py
```diff
@@ -148,7 +149,11 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
             ('outputnode.acompcor_report', 'inputnode.acompcor_report'),
             ('outputnode.tcompcor_report', 'inputnode.tcompcor_report')]),
     ])
-
+    if use_aroma:
+        workflow.connect([
+            (discover_wf, func_reports_wf, [
+                ('outputnode.ica_aroma_report', 'inputnode.ica_aroma_report')]),
+        ])
```
effigies: This doesn't need to be in an `if` block. You can just add it to the `(discover_wf, func_reports_wf, ...)` tuple above.

fmriprep/workflows/epi.py
```diff
@@ -233,7 +238,12 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
                 ('outputnode.itk_epi_to_t1', 'inputnode.itk_epi_to_t1')]),
             (epi_mni_trans_wf, outputnode, [('outputnode.epi_mni', 'epi_mni'),
                                             ('outputnode.epi_mask_mni', 'epi_mask_mni')]),
-        ])
+                        ])
+        if use_aroma:
+            workflow.connect([
+                (epi_mni_trans_wf, discover_wf, [('outputnode.epi_mask_mni', 'inputnode.epi_mask_mni'),
+                                                 ('outputnode.epi_mni', 'inputnode.epi_mni')])
+                            ])
```
effigies: Similarly, this doesn't need its own `if` block. Unused connections are safe to make.

fmriprep/workflows/epi.py
```diff
@@ -728,6 +738,11 @@ def init_func_reports_wf(reportlets_dir, freesurfer, name='func_reports_wf'):
         DerivativesDataSink(base_directory=reportlets_dir,
                             suffix='tcompcor'),
         name='ds_tcompcor_report', run_without_submitting=True)
+    if use_aroma:
+        ds_ica_aroma_report = pe.Node(
+            DerivativesDataSink(base_directory=reportlets_dir,
+                                suffix='ica_aroma'),
+            name='ds_ica_aroma_report', run_without_submitting=True)
```
effigies: You can create the node unconditionally. The cost of creating the node and then not connecting it is miniscule, and avoiding unnecessary branching makes reasoning easier.

fmriprep/workflows/epi.py
```diff
@@ -738,7 +753,12 @@ def init_func_reports_wf(reportlets_dir, freesurfer, name='func_reports_wf'):
                                          ('acompcor_report', 'in_file')]),
         (inputnode, ds_tcompcor_report, [('source_file', 'source_file'),
                                          ('tcompcor_report', 'in_file')]),
-        ])
+                    ])
+    if use_aroma:
+        workflow.connect([
+            (inputnode, ds_ica_aroma_report, [('source_file', 'source_file'),
+                                              ('ica_aroma_report', 'in_file')]),
+                        ])
```
effigies: This `if` block should stay. By not making the connection, the node won't get run.

fmriprep/cli/run.py
```diff
@@ -154,6 +158,12 @@ def create_workflow(opts):
     if opts.debug:
         logger.setLevel(logging.DEBUG)
 
+    #ERROR check if use_aroma wa specified, but template was not
+    if opts.use_aroma is True and str(opts.template) != 'MNI152NLin2009cAsym':
+        #some error statement
+
+        logger.warn('ERROR: if use_aroma is set, the template must be set to MNI152NLin2009cAsym not %s' % opts.template)
```
chrisfilo: Shouldn't we break here. A warning alone could get lost.

docs/workflows.rst
```diff
@@ -314,15 +316,17 @@ Confounds estimation
...
     from fmriprep.workflows.confounds import init_discover_wf
     wf = init_discover_wf(name="discover_wf",
-                          bold_file_size_gb=3)
+                          use_aroma=False, bold_file_size_gb=3)
 
 Given a motion-corrected fMRI, a brain mask, MCFLIRT movement parameters and a
 segmentation, the `discover_wf` sub-workflow calculates potential
 confounds per volume.
+Optional: give a motion corrected fMRI in MNI space, and mask in MNI space,
+the `discover_wf` sub-workflow calculates potential motion related components
```
chrisfilo: Let's mention that this is ICA AROMA.

Dockerfile
```diff
@@ -59,6 +59,7 @@ RUN echo "cHJpbnRmICJrcnp5c3p0b2YuZ29yZ29sZXdza2lAZ21haWwuY29tXG41MTcyXG4gKkN2dW
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
                     fsl-core=5.0.9-1~nd+1+nd16.04+1 \
+                    fsl-atlases=5.0.7-2 \
```
chrisfilo: Do we still need this?

fmriprep/workflows/confounds.py
```diff
@@ -243,3 +262,201 @@ def reverse_order(inlist):
...
+        aggr_tsv = aroma_add_header_func(no_signal_arr, 'no_signal_aggr_', ['00'])
+        nonaggr_tsv = aroma_add_header_func(no_signal_arr, 'no_signal_nonaggr_', ['00'])
+        aroma_confounds = (aggr_tsv, nonaggr_tsv)
+        return aroma_confounds
+
+    # nonaggr denoising confounds
+    nonaggr_confounds = np.asarray([calc_residuals(good_ic_arr, y) for y in aggr_confounds])
+
+    # add one to motion_ic_indices to match melodic report.
+    aggr_tsv = aroma_add_header_func(aggr_confounds.T, 'aggr_comp_',
```
chrisfilo: All other columns use camel case. Could we do "AROMAAggrComp01" and "AROMANonAggrComp01"?

Dockerfile
jdkent: We may not need this, but we need some file named ['MNI152_T1_2mm_brain.nii.gz'](https://github.com/rhr-pruim/ICA-AROMA/blob/master/ICA_AROMA_functions.py#L146) in the ${FSLDIR}/data/standard directory. [Here](http://neuro.debian.net/pkgs/fsl-mni152-templates.html#binary-pkg-fsl-mni152-templates) is a smaller package from neurodebian that has what ICA_AROMA needs. Alternatively we could use a symlink to the MNI152NLin2009cAsym template that's named MNI152_T1_2mm_brain.nii.gz in the directory that ICA_AROMA expects. I guess I don't know how those images differ, so I don't know if that is a viable option.

chrisfilo: I see - this is needed to get the right reference grid! Let's try using `fsl-mni152-templates` to save some space.

effigies:
For the test case, I'd suggest enabling on ds005. ds054 is already taking a bit longer because of the fieldmaps.

jdkent:
1. I think (I'm positive) I added the test to circleci incorrectly

2. I've deleted the dependency of having to set the ICA_AROMA outdir, but fmriprep throwing an error, because niworkflows hasn't been updated to include nipy/nipype#2056. Is there a timeline for the next release of niworkflows, since this should not be merged until the next release of niworkflows (which contains the relevant version of nipype).

3. I think I (or someone with more knowledge) should change how the melodic report is printed. On my tests with noisy subjects that have many components, the melodic report is very tiny, and doesn't load correctly in the html file.

effigies:
To update niworkflows, you can somewhat follow its [circle.yml](https://github.com/poldracklab/niworkflows/blob/master/circle.yml#L18) to get working with submodules.

```Shell
git clone git@github.com:jdkent/niworkflows.git
cd niworkflows
git submodule init
git submodule update
git -C nipype fetch origin
git -C nipype checkout origin/master
git checkout -b update_pin
git add nipype
git commit -m 'PIN: Update nipype to include ICA_AROMA fixes'
git push -u origin update_pin
```

I demonstrated checking out `origin/master` for simplicity. You can reference another commit if it seems like a good idea.

Once this is merged into niworkflows, then we can [pin fmriprep to that commit](https://github.com/effigies/fmriprep/commit/44babfb9be22dfe220686ad237d1936c3fd037d8). When it's time to release fmriprep (or mriqc), we will release niworkflows and set the dependency to the new version.

jdkent:
@effigies, Thank you for walking me through the commands to do that. That did help me understand the process better. When I look at niworkflows though (before I submit a pull request), I see the pinned version of nipype already contains the fixes to ICA_AROMA, but the [requirements.txt ](https://github.com/poldracklab/fmriprep/blob/master/requirements.txt#L1) for fmriprep only selects the released versions of niworkflows (as far as I understand it). version 0.1.2 does not include the selected fix for ICA_AROMA. So I followed your instruction to pin fmriprep to the most recent niworkflows commit. Thanks again for walking me through this!

requirements.txt
```diff
@@ -1 +1 @@
-niworkflows>=0.1.2
+git+git://github.com/poldracklab/niworkflows@a2dc98c#egg=niworkflows
```
effigies: `git+https`

.circle/tests.sh
```diff
@@ -46,4 +46,8 @@ case ${CIRCLE_NODE_INDEX} in
     fmriprep-docker -i poldracklab/fmriprep:latest --config $HOME/nipype.cfg -w $HOME/ds005/scratch $HOME/data/ds005 $HOME/ds005/out participant --output-space fsaverage5 --debug --write-graph
```
effigies: Just add `--use_aroma` to this line. We don't need to add another container. If you're going to need outputs in MNI space, just remove the `--output-space fsaverage5` part.

You don't need to set the template; that's the default.

.circle/tests.sh
```diff
@@ -43,7 +43,7 @@ case ${CIRCLE_NODE_INDEX} in
     ;;
   1)
     fmriprep-docker -i poldracklab/fmriprep:latest --help
-    fmriprep-docker -i poldracklab/fmriprep:latest --config $HOME/nipype.cfg -w $HOME/ds005/scratch $HOME/data/ds005 $HOME/ds005/out participant --output-space fsaverage5 --debug --write-graph
+    fmriprep-docker -i poldracklab/fmriprep:latest --config $HOME/nipype.cfg -w $HOME/ds005/scratch $HOME/data/ds005 $HOME/ds005/out participant --template --debug --write-graph --use_aroma
```
effigies: Sorry, I was unclear. `--template` is for specifying a *different* template. This will think it's getting `--template "--debug"`. Just drop the `--template` flag altogether.

jdkent: okay, got it. :+1:

effigies:
Looks like the AROMA images aren't getting included in the report. 

https://888-53175327-gh.circle-artifacts.com/1/home/ubuntu/ds005/out/fmriprep/sub-01.html

jdkent:
Yeah, that's what happened in my tests too, I can spend some time learning how to make those reports, it's what I meant by my number 3 point in my last post. The melodic reports are large, so they probably deserve their own page. So do you think the images aren't being included at all in the report or are the images too large to fit in that space?

chrisfilo:
As long as we leave AROMA as optional and experimental feature I am happy to leave the reports for a separate PR. The visualization of the components without an indication how they were classified is of limited use anyway.

I had a look at the confounds.tsv created on Circle and there are only two extra regressors - one agressive and one non aggressive. The non aggressive is just a row of zeros. Is it because of the downsampled nature of the test data?

There is also a more general question what to do if no signal or no noise components were detected. You opted for a warning and dummy regressors. I feel this is too silent - I would prefer at least to not include dummy regressors in such cases. We can also consider raising an error - after all this is a sign that something is wrong. What do others think?

effigies:
If we're not going to fix the reports in this PR, the blank images should be removed from the report file.

I agree a row of zeros isn't great to include in the confounds file. Raising an error seems like a reasonable way to handle the issue.

fmriprep/workflows/confounds.py
```diff
@@ -243,3 +262,197 @@ def reverse_order(inlist):
     if isinstance(inlist, list):
         inlist.reverse()
     return inlist
+
+
+def get_ica_confounds(ica_out_dir):
```
chrisfilo: Hi @maartenmennes! We are trying to add ICA AROMA to FMRIPREP, but without explicitly denoising the data - only estimating noise regressors and giving them to the users to include in their design. Could you have a quick look at how we calculate residuals for non aggressive components? Much obliged!

fmriprep/workflows/confounds.py
```diff
@@ -243,3 +262,197 @@ def reverse_order(inlist):
...
+
+        df = pd.DataFrame(np_arr, columns=[str(col_base) + str(index) for index in comp_nums])
+        df.to_csv(str(col_base) + "ica_confounds.tsv",
+                  sep="\t" if PY3 else '\t'.encode(), index=None)
+
+        return os.path.abspath(str(col_base) + "ica_confounds.tsv")
+
+    # Partial regression (close to, but not identical to fsl_regfilt)
+    def calc_residuals(x, y):
+        X = np.column_stack(x + [[0] * len(x[0])])
```
effigies: This seems to be equivalent to `X = x`. Say `x` has shape `(m, n)`. This gives us:

```Python
X = np.column_stack(x + [[0] * n])
  = np.column_stack(x + np.zeros((1, n)))
  = np.column_stack(x)
  = x
```

Am I missing something?

jdkent: You're right, the intended purpose of this was to concatenate a column of zeros to the coefficient matrix, but this does not do that, and furthermore, I believe I should be [appending a column of ones](https://docs.scipy.org/doc/numpy/reference/generated/numpy.linalg.lstsq.html) to the matrix instead of a column of zeros (sorry, my matrix algebra is lacking). If my thought process is correct, the command should like this:
```
X = np.concatenate((x, np.ones((x.shape[0], 1))), axis=1)
```
Thus X would look something like:
```
[ 2.34, 4.23, 1],
[ 6.23, 1.33, 1],
[ 2.67, 9.21, 1],
...
[ 3.56, 5.34, 1]
```
That fits better with the numpy example provided in the link.

effigies: That makes more sense to me. Adding a column of ones calculates a mean regressor (the `c` in their example, so your residuals will be de-meaned. Which is probably what you want, but I just want to be explicit about this, in case that's not actually what you want, here.

jdkent:
RE: what to do if there are no signal/noise components. I agree to raise a runtime error if there are no signal components in the data as that signals something is wrong with the data, but I don't know what are the implications if there are no noise components in the data, [ICA-AROMA](https://github.com/rhr-pruim/ICA-AROMA/blob/master/ICA_AROMA_functions.py#L547-552) doesn't treat no noise components as an error, but doesn't really leave a clue on how to represent the data in a tsv.

maartenmennes:
Hello, great to hear about adding ICA_AROMA to fmriprep! Seems like you've come a long way already! I don't think I have enough knowledge of the fmriprep code to comment on its correctness, but I did want to add a couple of remarks that might be useful.

- Getting the residuals should indeed be a straightforward regression. The one thing to keep in mind is how you treat the mean. If you remove this by demeaning the data or by adding a mean regressor, it is a good idea to add the mean back to the residuals afterwards. Demeaned data might screw up some calculations afterwards.

- In AROMA, as you've probably noticed, noise component classification happens in standard space, the denoising regression in native space. This fits with our (and fsl's) philosophy to keep data in native space as long as possible, only bringing results to standard space. That said, I see no reason to not use an ANTs calculated transformation to bring the data into standard space. The spatial features used (i.e., edge mask and csf mask) are not so fine-grained that they would be hugely affected by using a different registration. In fact, we do have the option built in to only use linear registration files.

- When applying ICA_AROMA we always investigate the number of ICA components derived, and the number of components flagged as noise, as low/high values in either of these could signal some weird stuff in the data. Of course this nicely fits with: 'ALWAYS LOOK AT YOUR DATA' ;)

chrisfilo:
Thanks for your comments @maartenmennes. Since we are not denoising the data (only producing the regressors) I think we should opt for ones that are demeaned (which is what I think is happening now).

@jdkent I cannot think of a situation when no components classified as noise is the correct input, but I might be lacking experience with AROMO. For the time being I would go for an error in such case and see what different use cases people will come back to us with.

.circle/tests.sh
```diff
@@ -43,7 +43,7 @@ case ${CIRCLE_NODE_INDEX} in
     ;;
   1)
     fmriprep-docker -i poldracklab/fmriprep:latest --help
-    fmriprep-docker -i poldracklab/fmriprep:latest --config $HOME/nipype.cfg -w $HOME/ds005/scratch $HOME/data/ds005 $HOME/ds005/out participant --output-space fsaverage5 --debug --write-graph
+    fmriprep-docker -i poldracklab/fmriprep:latest --config $HOME/nipype.cfg -w $HOME/ds005/scratch $HOME/data/ds005 $HOME/ds005/out participant --debug --write-graph --use_aroma
```
effigies: Consider:

```Shell
fmriprep-docker -i poldracklab/fmriprep:latest --config $HOME/nipype.cfg -w $HOME/ds005/scratch $HOME/data/ds005 $HOME/ds005/out participant --debug
fmriprep-docker -i poldracklab/fmriprep:latest --config $HOME/nipype.cfg -w $HOME/ds005/scratch $HOME/data/ds005 $HOME/ds005/out participant --debug --write-graph --use_aroma || true
```

This will run the majority of the pipeline, verifying that  it runs without error. Then it will run with `--use_aroma`, reusing earlier results, but not failing when components aren't found.

chrisfilo:
But this will silently pass for any other errors now right?

What about additional flag something along the lines of `--ignore-missing-noise-error`?

effigies:
Only other errors in ICA, but yes. I like the CLI flag option better than mine (sorry @jdkent for prompting you to change that before discussion completed). I'd say it should have "(for testing only)" or similar in the description. And I'd group it with `--debug`.

jdkent:
@chrisfilo, @effigies Would it be best to treat the flag option as I did with `--use_aroma`? As in, pass the variable on through the workflow until it reaches its destination in confounds?

effigies:
@jdkent Yes.

jdkent:
I added a flag called `--ignore-aroma-denoising-errors`, but I found the melodic code needed to be changed as well. Passing a mask into melodic makes a big difference in ICA AROMA classification, and I thought I added it in before when @chrisfilo suggested, but I only added the report mask, not the actual mask to use in melodic. This fix results in the classifier giving fewer noise components. However, my result with using Non-aggressive denoising is still not identical to ICA-AROMA's fsl_regfilt calculation. The code I use to check is:
```
tail -n +2 'nonaggr_tsv' > 'no_header_nonaggr_tsv'
Text2Vest 'no_header_nonaggr_tsv'  'nonaggr_des'
fsl_glm -i 'smoothed_image' -o 'denoised_glm' -d  'nonaggr_des' --out_res='fmriprep_denoised_data'
 ```
The difference between calculations amounts to more than what I would expect from a rounding error.

P.S. Thank you all for helping me with this, I have learned a lot by going through this process.

**Comment by chrisfilo** [comment]
Thank you for doing this!

Could you share correlation plots between the regressors provided by
fsl_glm and those calculated by you? This could help us see what is going
on.

**Comment by jdkent** [comment]
Sorry, I wasn't clear. I'm trying to compare the denoised result from fsl_regfilt as implemented by ICA_AROMA to the regressors derived from ICA_AROMA in the script passed into fsl_glm. I'm trying to get an identical between the two methods. I can make the results identical with aggressive denoising, but I cannot do the same with the non-aggressive method. 

So the ICA_AROMA method would be something like:
`fsl_regfilt -i smooth_file -o denoised_file -f classified_motion_ICs.txt -d melodix_mix -a`

Then using the non-aggressive derived in fmriprep, I would do this:
`fsl_glm -i smooth_file -o glm_file -d AROMANonAggr.mat -out_res=denoised_file`

I've tried adding the demean option in fsl_glm and add the mean back in after, but I can't get the residual image from fsl_glm to match the output from fsl_regfilt.

[Here is a link](https://drive.google.com/drive/folders/0B9eMc5Q-tEdSQWg5VGc5ZXVTMEk?usp=sharing) to ica_aroma_wf for a subject. 

Unless I misunderstood what you are asking, I don't think fsl_regfilt outputs regressors as ICA_AROMA uses it.

**Comment by chrisfilo** [comment]
Hmm - I'm not quite sure what is going on then. I tried looking through the  source of fsl_regfilt https://surfer.nmr.mgh.harvard.edu/pub/dist/freesurfer/tutorial_packages/OSX/fsl_501/src/melodic/fsl_regfilt.cc but the regression is don in a funky way.

Maybe @maartenmennes could give us a hint. Are we doing the orthogonalization in a wrong way?

**Comment by jdkent** [comment]
To help facilitate my understanding and for those that can help, I've made an [ipython notebook](https://drive.google.com/drive/folders/0B9eMc5Q-tEdSbklYZkM3djRiMFk?usp=sharing) to explain the steps I'm taking to perform multiple linear regression with the output from ICA-AROMA. I'll start here to assume there is some mistake in the code I wrote, before I move on to thinking there is something wrong in the way I called fsl_glm.

**Comment by jdkent** [comment]
[additional insight:](https://www.jiscmail.ac.uk/cgi-bin/webadmin?A2=ind1307&L=fsl&D=0&P=288785) I may need to load the entire image into a numpy array and then do the regression described in the link except return the residuals (This description should match what's in the source code @chrisfilo posted).  I'll keep you posted as I try to emulate this method.

**Comment by effigies** [comment]
Sounds good. Let us know if there's anything we can do. I've just pushed out a new release today that includes the PR that's been taking up a lot of my energy, so I can direct some more attention to this PR.

**Comment by jdkent** [comment]
I submitted [another question ](https://www.jiscmail.ac.uk/cgi-bin/webadmin?A2=fsl;308d72cd.1706) to the fsl listserve (the first one didn't get a reply). Since Dr. Christian F. Beckmann is the author of the fsl_regfilt code, perhaps I should directly contact him?

I'm having trouble converting his denoising partial regression to a regression equation that would output  residual noise components. 

Conceptually, I want to remove any shared variance in the noise components. Previously I was only giving the shared variance to the "signal" melodic components, but it looks like there is additional shared variance between the noise components and the image data. To include the data, I first regress out all the melodic components and use the residual image data as a regressor to soak up shared variance between itself and the noise components.

**Comment by jdkent** [comment]
Yay! [Dr. Christian F. Beckmann replied](https://www.jiscmail.ac.uk/cgi-bin/webadmin?A2=ind1706&L=fsl&F=&S=&X=0E95A4350CBB09A848&Y=james-kent%40uiowa.edu&P=204617) to my question. Now I need to figure out how to implement the equation he gave me in numpy/scipy:

> If X=(X_1|X_2) is the melodic mixing matrix where we want to remove effects in X_2 only then the denoised data is given by
> Y_clean = ( Id - X_2 * (0 | Id_2) pinv(X)) * Y
> Where Y is the original data and Id is the Identity matrix (t x t) and Id_2 is the identity matrix of size n2, where n2 is the number of nice components. 

This may take some work, but will be a good learning opportunity
Given his answer I believe  to accomplish our goal we need to solve the equation
pinv(( Id - X_2 * (0 | Id_2) pinv(X)) * Y)

I understand most the operations, but I am unfamiliar with the (0 | id_2) notation (specifically the vertical bar) 
That's my first question.

**Comment by chrisfilo** [comment]
I think in this context this means concatenation.

**Comment by jdkent** [comment]
Alright, I've given my best faith effort, and I wasn't able to emulate the equation. Moreover, even if I was, based on Dr. Beckmann's response, the matrix would have as many columns as there are voxels of interest in the dataset (if I'm interpreting the equation correctly). A tsv of that size would be unruly and potentially confusing to users.

Here's my suggestion: pass the ICA_AROMA classification file and the melodic mix to the subject derivative directory and let the user run fsl_regfilt on their data if they want non-aggressive denoising applied to their data. The drawback to this being more clutter in the directory. Less-so of a drawback is coming up with another naming scheme for this unique type of confound that is not a voxelwise regressor, but doesn't quite fit in with the confound tsv. 


My Naming Proposal: 
<BIDS_subject_info>_aroma_noise_classification.csv (for the file that contains a comma separated list of components suggested to be related to noise)
<BIDS_subject_info>_aroma_melodic_mix.tsv (the full melodic mixing matrix)

Then document furiously to let users know how to implement non-aggressive de-noising. 

Thoughts?

**Comment by effigies** [comment]
Okay. Let me try to summarize the situation. As I think I've said, I've never used MELODIC or AROMA; hopefully I understand roughly what's going on, but I recognize what follows is likely to be laughably simplistic. Please correct me if it's also wrong.

1) MELODIC finds IC time series, which I assume are global signals, independent of space. Some of these get labeled noise, some get labelled signal.
2) Aggressive de-noising simply regresses out all noise ICs from every voxel.
3) Non-aggressive does some kind of per-voxel modification of the ICs, so we can't actually generate these regressors in such a way as to apply in a GLM anyway.

If that's basically right, then I think your strategy seems reasonable.

I believe the BIDS preference is always for TSV, not CSV, so I'd use TSV for both of your files. We should also avoid `_` in the descriptive parts of the filenames, since that's used as a field delimiter in BIDS. How about:

* `aroma_noise_classification.csv` -> `AROMAnoiseICs.tsv`
* `aroma_melodic_mix.txv` -> `MELODICmix.tsv`

Or something like that? Since the MELODIC output is actually not specific to AROMA post-processing, I don't think we need to label it with both AROMA and MELODIC. I'm open to alternative capitalization schemes.

**Comment by jdkent** [comment]
Looks like you got it right for points 1 and 2, and I'm not sure if I have it right for point 3, but I'll try explaining what I think I know more.
RE: 3, The response I got from the [fsl listserv](https://www.jiscmail.ac.uk/cgi-bin/webadmin?A2=ind1706&L=fsl&D=0&X=0E95A4350CBB09A848&Y=james-kent%40uiowa.edu&P=204617) suggests it's possible to make a non-aggressive denoising glm matrix, but I'm not mathematically inclined enough to recreate the process. Additionally, the pseudo-inverse of Y_clean (where Y_clean is the image data) is what we need to pass into the glm. It may be possible to simplify the matrix, but the direct solution appears to be each voxel would get a column. (if my understanding is correct, haha).

> I believe the BIDS preference is always for TSV, not CSV, 

I also believe the preference is for tsv, but this is another special case. I include the noise IC classification as a csv because fsl_regfilt does not accept tab separated values with it's "-f" flag. However, fsl_regfilt does not tell you it can't read the .tsv, instead it only filters the first index/number in the list. Making the AROMAnoiseICs.tsv could lead to misuse on the user end if they try to pass the tsv directly to fsl_regfilt without knowing that quirk. ugh, that is annoying.

**Comment by effigies** [comment]
Alright. I think in any case, it's reasonable to go ahead in this PR and not produce the non-aggressive regressors, and just add documentation. If future investigation shows we can do it for users without producing a column per voxel, that can be another PR. (Incidentally, if it would end up being a column per-voxel, it probably makes more sense to output it as a `.nii.gz` time series so that the spatial information is directly encoded.)

Thanks for the context on the CSV issue. That makes sense and let's definitely go ahead with your plan, there.

**Comment by jdkent** [comment]
I'm not entirely sure what the errors for the ci/circle mean for me, I've tested it on my data (however, there were not edge cases where no signal components were detected), so I am unsure where I should look to fix the error.

**Comment by jdkent** [comment]
Now I'm out of ideas on how to pass the tests, it looks like running ICA_AROMA.py from the container failed, perhaps due to some property of the data?

**Comment by jdkent** [comment]
@effigies @chrisfilo Question: what software do you use to preview the *rst files in docs? I tried [retest](https://apps.ubuntu.com/cat/applications/raring/retext/) for ubuntu, but my first impression is disappointing.

Again thank you for taking the time to help me through this journey, I believe the code is ready barring documentation formatting and any further suggestions you have.

**Comment by effigies** [comment]
```Shell
python setup.py build_sphinx
```

Generally run it in a docker container, because it's pretty touchy.

effigies:
And I'll have a thorough look through (probably Monday). Thanks so much for all the effort you've put into this.

fmriprep/workflows/confounds.py
```diff
@@ -243,3 +264,211 @@ def reverse_order(inlist):
...
+    # Return dummy list of ones if no noise compnents were found
+    if motion_ic_indices.size == 0:
+        if ignore_aroma_err:
+            LOGGER.warn('WARNING: No noise components were classified')
+            no_noise_arr = np.ones((melodic_mix_arr.shape[0], 1))
+            aggr_tsv = aroma_add_header_func(no_noise_arr, 'AROMANoNoiseAggr', ['00'])
+            aroma_confounds = aggr_tsv
+            return aroma_confounds, motion_ics_out, melodic_mix_out
+        else:
+            raise RuntimeError('ERROR: ICA-AROMA found no noise components!')
```
effigies: I vaguely remember there being some discussion earlier, but it's hard to find at this point.

Why are we returning a dummy vector as opposed to one with `size == 0`, or `None`?

fmriprep/workflows/confounds.py
```diff
@@ -243,3 +264,211 @@ def reverse_order(inlist):
...
+                                     generate_report=True),
+                      name="melodic")
+
+    # ica_aroma node
+    ica_aroma = pe.Node(aroma.ICA_AROMA(denoise_type='no'),
+                        name='ica_aroma')
+
+    # extract the confound ICs from the results
+    ica_aroma_confound_extraction = pe.Node(utility.Function(input_names=['ica_out_dir',
+                                                             'ignore_aroma_err'],
```
effigies: This can be done a little cleaner:

```Python
ica_aroma_confound_extraction = pe.Node(
    utility.Function(function=get_ica_confounds,
                     output_names=['aroma_confounds', 'aroma_noise_ics', 'melodic_mix']),
    name='ica_aroma_confound_extraction')
```

Other nodes can be cleaned up similarly. (I won't hold up merging on this, though. It's mostly an aesthetic thing.)

fmriprep/workflows/confounds.py
```diff
@@ -243,3 +264,211 @@ def reverse_order(inlist):
...
+        (ica_aroma, ica_aroma_confound_extraction,
+            [('out_dir', 'ica_out_dir')]),
+        # output for processing and reporting
+        (ica_aroma_confound_extraction, outputnode,
+            [('aroma_confounds', 'aroma_confounds'),
+             ('aroma_noise_ics', 'aroma_noise_ics'),
+             ('melodic_mix', 'melodic_mix')]),
+        # TODO change melodic report to reflect noise and non-noise components
+        (melodic, outputnode,
+            [('out_report', 'out_report')]),
```
effigies: A lot of these connections can be reduced to one or two lines. I find this extremely hard to read as is.

fmriprep/workflows/epi.py
```diff
@@ -40,8 +40,10 @@
 def init_func_preproc_wf(bold_file, ignore, freesurfer,
                          bold2t1w_dof, reportlets_dir,
                          output_spaces, template, output_dir, omp_nthreads,
-                         fmap_bspline, fmap_demean, use_syn, force_syn,
-                         debug, output_grid_ref, layout=None):
+                         fmap_bspline, fmap_demean, debug, output_grid_ref,
+                         use_aroma, ignore_aroma_err,
+                         use_syn, force_syn, layout=None):
```
effigies: Just to reduce the diff to a single addition, rather than 2-/4+.

```Python
fmap_bspline, fmap_demean, use_syn, force_syn,
use_aroma, ignore_aroma_err,
debug, output_grid_ref, layout=None):
```

fmriprep/workflows/epi.py
```diff
@@ -142,11 +150,11 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
                                  ('subject_id', 'inputnode.subject_id'),
                                  ('fs_2_t1_transform', 'inputnode.fs_2_t1_transform')
                                  ]),
-        (inputnode, discover_wf, [('t1_tpms', 'inputnode.t1_tpms')]),
+        (inputnode, discover_wf, [('t1_tpms', 'inputnode.t1_tpms'),
+                                  ]),
```
effigies: Let's revert this newline.

fmriprep/workflows/epi.py
```diff
@@ -157,8 +165,12 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
...
         (discover_wf, func_reports_wf, [
             ('outputnode.acompcor_report', 'inputnode.acompcor_report'),
-            ('outputnode.tcompcor_report', 'inputnode.tcompcor_report')]),
-    ])
+            ('outputnode.tcompcor_report', 'inputnode.tcompcor_report'),
+            ('outputnode.ica_aroma_report', 'inputnode.ica_aroma_report')]),
+        (discover_wf, func_derivatives_wf, [
+             ('outputnode.aroma_noise_ics', 'inputnode.aroma_noise_ics'),
+             ('outputnode.melodic_mix', 'inputnode.melodic_mix')]),
+        ])
```
effigies: Our current pattern is to pass anything that needs to go to derivatives to `outputnode`, and then pass the `outputnode` fields to `func_derivatives_wf.inputnode`. The reasoning is basically that if it's a derivative, it's something another workflow should reasonably expect access to.

fmriprep/workflows/epi.py
```diff
@@ -1034,7 +1054,12 @@ def init_func_reports_wf(reportlets_dir, freesurfer, use_syn, name='func_reports
                                          ('acompcor_report', 'in_file')]),
         (inputnode, ds_tcompcor_report, [('source_file', 'source_file'),
                                          ('tcompcor_report', 'in_file')]),
-        ])
+                    ])
+    if use_aroma:
+        workflow.connect([
+            (inputnode, ds_ica_aroma_report, [('source_file', 'source_file'),
+                                              ('ica_aroma_report', 'in_file')]),
+                        ])
```
effigies: The close brace should be aligned with the start of the tuple. e.g.

```Python
if x:
    wf.connect([
        (node1, node2, [(...)]),
        ])
```

docs/workflows.rst
```diff
@@ -320,15 +324,29 @@ Confounds estimation
...
 tCompCor, aCompCor, Framewise Displacement, 6 motion parameters and DVARS.
 
+Optional (--use_aroma): aggressive ICA confounds classified by ICA-AROMA.
+Non-aggressive confounds will not be given in the confounds tsv, but denoising can be completed
+by the user. A noise component classification csv and a melodic mix tsv will be passed
+into the derivatives directory. To get non-aggressively denoised data, use the command
+below as a guide:
+
+``fsl_regfilt -i <BIDS SUBJECT>_task-<TASK>_bold_space-<SPACE>_preproc.nii.gz
+-f $(cat <BIDS SUBJECT>_task-<TASK>_bold_AROMAnoiseICs.csv)
```
jdkent: correct, and that is the only way it can be read by [fsl_regfilt](https://surfer.nmr.mgh.harvard.edu/pub/dist/freesurfer/tutorial_packages/OSX/fsl_501/src/melodic/fsl_regfilt.cc):
```
Option<string> filter(string("-f,--filter"),string(""),
		string("filter out part of the regression model, e.g. -f \"1,2,3\" "),
		false, requires_argument);
```
Do you have another idea about formatting?

effigies: Nope, that's reasonable. Just making sure that the line was doing what I thought it was.

fmriprep/workflows/confounds.py
```diff
@@ -7,36 +7,46 @@
 Calculates frame displacement, segment regressors, global regressor, dvars, aCompCor, tCompCor
 '''
 from __future__ import print_function, division, absolute_import, unicode_literals
-
 from niworkflows.nipype.interfaces import utility
+from niworkflows.nipype.interfaces import fsl
 from niworkflows.nipype.algorithms import confounds
 from niworkflows.nipype.pipeline import engine as pe
+from niworkflows.nipype.interfaces.fsl import ICA_AROMA as aroma
```
jdkent: After messing with this, I realized I need to add ICA_AROMA to the __init__.py in [nipype](https://github.com/nipy/nipype/blob/master/nipype/interfaces/fsl/__init__.py#L7), I'll submit a pull request to fix this.

effigies: Oh, sorry. I didn't realize it was `aroma.ICA_AROMA` below. I was assuming it was just `pe.Node(aroma())` and didn't check. Let's keep this as it is.

fmriprep/workflows/confounds.py
```diff
@@ -243,3 +264,211 @@ def reverse_order(inlist):
...
+    # Pass in numpy array and column base name to generate headers
+    # modified from add_header_func
+    def aroma_add_header_func(np_arr, col_base, comp_nums):
+        import pandas as pd
+        from sys import version_info
+        PY3 = version_info[0] > 2
+
+        df = pd.DataFrame(np_arr, columns=[str(col_base) + str(index) for index in comp_nums])
+        df.to_csv(str(col_base) + "AROMAConfounds.tsv",
+                  sep="\t" if PY3 else '\t'.encode(), index=None)
```
jdkent: Honestly, this was copied from [add_header_func](https://github.com/poldracklab/fmriprep/blob/master/fmriprep/workflows/confounds.py#L145), I don't know who wrote the original function, so I don't know the original reasoning behind this.

fmriprep/workflows/confounds.py
jdkent: `None` would work, my reasoning for adding it back in was the conditional code would only be ran if the user also specified `--ignore-aroma-denoising-errors`,  and were actively looking for subjects that didn't have signal/noise components. A unique column name in confounds would make them stand out. However, if they ran it normally, fmriprep would error out and that would also be an easy way to find which subjects didn't have noise or signal components. 

I'll switch to `None`, I believe that is most reasonable and fits well with the rest of the code.

fmriprep/workflows/confounds.py
```diff
@@ -7,36 +7,46 @@
...
-        from motion-corrected fMRI ('inputnode.fmri_file').
+    from motion-corrected fMRI ('inputnode.fmri_file').
     Calculates DVARS from the fMRI and an EPI brain mask ('inputnode.epi_mask')
     Calculates frame displacement from MCFLIRT movement parameters ('inputnode.movpar_file')
     Calculates segment regressors and aCompCor
         from the fMRI and a white matter/gray matter/CSF segmentation ('inputnode.t1_seg'), after
         applying the transform to the images. Transforms should be fsl-formatted.
-
+    Optional:
+        Calculates motion related components from ICA_AROMA
```
jdkent: [That is how it is advertised](https://www.ncbi.nlm.nih.gov/pubmed/25770991), but I've heard from colleagues (and from my own data) some of the artifacts removed appear to be more physiological in origin.

fmriprep/workflows/confounds.py
effigies: Okay. We can leave it. I suspect at some point we'll purge `PY3` and similar references, since we don't support Python 2.

fmriprep/workflows/confounds.py
effigies: Sounds good. Can you do that for the other error condition, as well?

fmriprep/workflows/confounds.py
effigies: Maybe let's just shift to "noise components identified by ICA-AROMA", or similar. "Noise" is a reasonably broad characterization that would include physiological or motion-related components.

.circle/tests.sh
```diff
@@ -43,7 +43,7 @@ case ${CIRCLE_NODE_INDEX} in
         | tee $HOME/docs/builddocs.log
     cat $HOME/docs/builddocs.log && if grep -q "ERROR" $HOME/docs/builddocs.log; then false; else true; fi
     fmriprep-docker -i poldracklab/fmriprep:latest --help
-    fmriprep-docker -i poldracklab/fmriprep:latest --config $HOME/nipype.cfg -w $HOME/ds005/scratch $HOME/data/ds005 $HOME/ds005/out participant --output-space fsaverage5 --debug --write-graph
+    fmriprep-docker -i poldracklab/fmriprep:latest --config $HOME/nipype.cfg -w $HOME/ds005/scratch $HOME/data/ds005 $HOME/ds005/out participant --debug --write-graph --use_aroma --ignore-aroma-denoising-errors
```
effigies: Need to update `--use-aroma` here.

jdkent: oops, thanks!

effigies:
@jdkent Thanks so much for spending a month on this. This was a huge contribution.

chrisfilo:
Thank you! It's great to see this merged!

jdkent:
@effigies @chrisfilo This was an amazing learning experience for me and I can't emphasize enough how thankful I am for you two taking the time to guide me through this contribution. I plan to keep contributing to fmriprep and projects like it (hopefully I'll need less help and be able to help others as I learn more about nipype and python).

PR #542, opened 2017-05-30
[RTM] RF: Use N4BiasFieldCorrection(copy_header=True)

effigies:
Taking advantage of nipy/nipype#2034 to simplify workflows slightly.

### DIFF
--- a/fmriprep/workflows/anatomical.py
+++ b/fmriprep/workflows/anatomical.py
@@ -81,7 +81,7 @@ def init_anat_preproc_wf(skull_strip_ants, output_spaces, template, debug, frees
         RobustMNINormalizationRPT(
             generate_report=True,
             num_threads=omp_nthreads,
-            testing=debug,
+            flavor='testing' if debug else 'precise',
         ),
         name='t1_2_mni'
     )
--- a/fmriprep/workflows/fieldmap/phdiff.py
+++ b/fmriprep/workflows/fieldmap/phdiff.py
@@ -75,7 +75,7 @@ def init_phdiff_wf(reportlets_dir, name='phdiff_wf'):
     magmrg = pe.Node(IntraModalMerge(), name='magmrg')
 
     # de-gradient the fields ("bias/illumination artifact")
-    n4 = pe.Node(ants.N4BiasFieldCorrection(dimension=3), name='n4')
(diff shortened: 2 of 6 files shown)


fmriprep/workflows/fieldmap/phdiff.py
```diff
@@ -75,7 +75,7 @@ def _pick1st(inlist):
     magmrg = pe.Node(IntraModalMerge(), name='magmrg')
 
     # de-gradient the fields ("bias/illumination artifact")
-    n4 = pe.Node(ants.N4BiasFieldCorrection(dimension=3), name='n4')
+    n4 = pe.Node(ants.N4BiasFieldCorrection(dimension=3, copy_header=True), name='n4')
```
effigies: Note that this wasn't using `init_n4bias_wf` already. If there's a good reason for not copying the header, let me know and I'll revert this.

effigies:
This is failing a 4GB memory limit. ([Memory usage](https://457-78558509-gh.circle-artifacts.com/0/tmp/memory-usage.txt))

I'm not sure why this would newly be an issue. Could be the type of the bias-corrected image (which is now restored to that of the input T1w image). Should I re-run, or is there something obvious to fix up?

oesteban:
This is very weird. Additionally, the process crashing is antsRegistration, not the bias correction.

chrisfilo:
Let's try running first.

effigies:
Rerunning. Also, there's a workflow in niworkflows that should be updated along with this batch.

Blocking merge for poldracklab/niworkflows#173.

effigies:
That one worked. Updating pins and rebasing for a final test.

fmriprep/workflows/anatomical.py
```diff
@@ -81,7 +81,7 @@ def init_anat_preproc_wf(skull_strip_ants, output_spaces, template, debug, frees
         RobustMNINormalizationRPT(
             generate_report=True,
             num_threads=omp_nthreads,
-            testing=debug,
+            flavor='testing' if debug else 'precise',
```
effigies: @oesteban I believe this is the 0.0.7+ equivalent?

oesteban: yes

PR #544, opened 2017-06-02
[RTM] ENH: Nonlinear warp-based susceptibility distortion correct

### LINKED ISSUE #496: Susceptibility distortion correction without fieldmaps
Traditionally to perform susceptibility distortion correction (SDC) you need to have some sort of fieldmap data. However, there was some work showing that SDC can be performed without the need to acquire extra data.
- Galiphour et. al 2008 proposed using a field distortion average template http://ieeexplore.ieee.org/document/4649099/ Another template can be found here: http://neurovault.org/collections/2426/
- @juhuntenburg  investigated non-linear coregistration between EPI and T1w of the same subject http://pubman.mpdl.mpg.de/pubman/item/escidoc:2327525:5/component/escidoc:2327523/master_thesis_huntenburg_4686947.pdf
- Wang, @danjonpeterson et al. 2017 further improved and evaluated the non-linear coregistration technique http://journal.frontiersin.org/article/10.3389/fninf.2017.00017/full

It would be great if could add SDC to FMRIPREP that does not require fieldmaps. I would suggest starting, by implementing the approach from Wang et al. on a dataset with fieldmaps  (for example Beast or UH2) and see how that compares. I suspect (judging from @juhuntenburg work) that some amount of regularization and use of priors will be required. Here are some ideas to try:

- use 12 DOF (affine) boundary based coregistration as the initial transformation
- use a field distortion template as the initial transformation
- limit the non-linear coregistration only to the dimension of the phase encoding direction
- create a mask of usually distorted areas from the distortion template an limit the coregistration using it

effigies:
Closes #496


effigies:
For the [distortion atlas](http://neurovault.org/collections/2426/), the authors provide an unspecified template (I'll call the *distortion template*), I would guess a symmetric MNI template. My current strategy for getting a distortion atlas in the EPI ref space is:

* Find an affine transformation from the distortion template to the template in use by FMRIPREP
* Apply (distortion template -> template) + inv(T1 -> template) + inv(EPI ref -> T1) to distortion atlas

If we don't think an affine transformation is sufficient to go between templates, then I think it would make more sense to calculate a transform from the T1 to the distortion template and invert that, since we're going to go through the trouble of a nonlinear transform anyway.

Alternately, we can pre-calculate a (distortion template -> template) warp for each template we support.

WDYT?

chrisfilo:
I think affine is perfectly fine for this purpuse. BTW the template they
used (defining the space) is this one http://neurovault.org/images/43982/
AFAIK.

effigies:
Right, they used that template, but it's not clear if it's a standard template or if they created it from their collection of subjects. Not sure that it makes a difference for our purposes; just noting the situation.

effigies:
How do I get the phase-encoding direction from the JSON file associated with a BOLD image? Just make sure I have the original filename and replace `nii.gz` with `json`? Or does pybids have some relevant magic?

Another thing to consider: This atlas is only valid for anterior/posterior phase-encoding. In the event of an `i/i-` or `k/k-` image, we probably wouldn't want to use the atlas...

chrisfilo:
There are two things to consider:
- AFAIK distortions happen in the same place in the brain independently of the phase encoding direction. It only changes which way the tissue will be pulled/pushed. Thus I think we can use this atlas for both i/i- and j/j-
- most datasets without fieldmaps do not have any phase encoding information. In such cases it is worth assuming that the distortions happen in either i or j (because we already assume k is slice direction). There are two things we could do: 1) restrict ants to two direction - i and j (instead of one) 2) run ants twice once with i and once with j and see which gives better fit with bbregister

effigies:
1) I've searched a bit, and can't find any articles that discuss the effect of PE direction on where artifacts occur. Presumably there's some effect, even though the "epicenters" might be consistent. ANTs' smooth deformation constraints may save us here. In either event, we can start from that assumption.

2) I suppose I'd rather go with the second option in cases where we don't know; that way we're not using an effectively different algorithm when we don't know. In this case, we should verify on datasets where we do know PE direction that our heuristic (e.g. bbregister min cost) selects the true one.

oesteban:
Yes, theoretically the fieldmap is the same regardless the encoding direction. I don't think you'll find a scan with k/-k for PE since that's the B0 direction.

As regards the case of unknown PEs, as @effigies I'd say better going to option 2 (two registration processes). However, it shouldn't be that hard to look at the disagreement map of the T1 mask linearly mapped into EPI space and the EPI mask. Then project to X and Y. The PE axis should a bimodal distribution in the PE axis and monomodal in the FE axis.

effigies:
Current status: This performs the warping as expected and generates a report (currently SyN-registered reference EPI to the inverted-T1-in-EPI space).

Still in progress:

* [ ] PE-direction guessing (switching to MapNode, followed by BBR-based assessment, though making sure this works with FLIRT-BBR will be necessary)
* [ ] Atlas-based restriction of SyN warp

effigies:
I believe this is basically set. The main remaining issue is that `--use_syn` needs `t1_2_mni` to generate a transform, so if our `--output-space` doesn't include `template`, we'll throw an error. The hangup here is that, if we move to `--use-syn` by default: If I have fieldmaps and don't want MNI outputs, then I don't want to waste time with that conversion; if I don't have fieldmaps, I still want SyN to work, so we'll have to run that.

I'll need to think about the best way to manage this, but for now I think we should just warn if `'template'` is not in `output_spaces` and `use_syn` is on. That will minimize the logic that needs reworking when we settle on an approach.

effigies:
Looks like this is going to require upgrading to ANTs 2.2.0. Sorry for the delay. I'll work on upgrading existing ANTs nodes, and then return to this.

effigies:
This is just about ready. I'm going to work on some documentation today and plan to release 0.5.0 by tonight.

chrisfilo:
+1

effigies:
Added to the SDC docs, in the fieldmap estimation section.

It may make sense to move this workflow out of `workflows.epi` into `workflows.fieldmaps`, and manage its selection in a similar way to `fmap` and `phdiff`, but I don't want to hold this up further right now. That refactorization can come along with stabilizing, after we've had a chance to test the functionality more broadly.

PR #549, opened 2017-06-04
[RTM] Port to niworkflows.nipype

effigies:


effigies:
Pinning against niworkflows/master. Should be good to go if tests pass.

PR #556, opened 2017-06-13
Rolling back docker version

chrisfilo:


effigies:
Doesn't look like this is resolving the issue, but maybe the `:latest` image that it pulls needs to be created with the same version of Docker? If so, it could resolve after the next release. Hard to say. What was the reason for changing to this version?

chrisfilo:
We wanted compatibility with singularity (for automatic image size
assessment), but it did not work anyway. No benefits for using the never
version. We can try with this one and do a quick release fixing the
conformation issues.

On Tue, Jun 13, 2017 at 10:21 AM, Chris Markiewicz <notifications@github.com

effigies:
I'm okay with merging this without a full build... it'll get built after merging anyway.

PR #558, opened 2017-06-13
CLI: Add useful error message for missing subjects

effigies:


effigies:
Looks good. Do we need to add anything to the wrapper script? Barring a way to discover whether there's a failure to mount, I'm not sure what there is to add.

PR #560, opened 2017-06-16
[RTM] ENH: Add --anat-only workflow option

effigies:
This may be instant-reject, but it's not clear to me that there is an advantage to rejecting datasets without functional images. We already throw a warning in their absence.

(I created this branch a while ago to test the anat pipeline on a user's T1w image.)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,7 +1,7 @@
 Next release
 ============
 
-
+* [ENH] Add --anat-only workflow option (#560)
 
 0.4.6 (14th of June 2017)
 =========================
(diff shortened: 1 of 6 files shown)


chrisfilo:
My concern is that almost no one will intentionally run **FMRI**PREP on a dataset with no fMRI data (or mislabeled fMRI data). I'm afraid that this default behavior would lead to confusion.

effigies:
Would a CLI option (`--anat-only`) be acceptable? It is handy for testing the anatomical workflow. If not, it's only a couple lines to kill whenever I need to test.

chrisfilo:
Yup that would be much better.

On Jun 16, 2017 7:00 PM, "Chris Markiewicz" <notifications@github.com>
wrote:

Would a CLI option (--anat-only) be acceptable? It is handy for testing the
anatomical workflow. If not, it's only a couple lines to kill whenever I
need to test.

—
You are receiving this because you commented.

<https://github.com/poldracklab/fmriprep/pull/560#issuecomment-309186119>,
or mute the thread

PR #561, opened 2017-06-16
[RTM] Upgrade to ANTs 2.2.0

### LINKED ISSUE #516: Update ANTs to 2.2.0
Relevant changes:
- different masking in N4BiasFieldCorrection: https://github.com/stnava/ANTs/commit/6816c5a7ece6090febf90496dd5bd95cd48bb46e
- different metric in brain extraction: https://github.com/stnava/ANTs/commit/98644cd2caa1d30f0ef7cc804098111ee680bc39
- different registration init in brain extraction: https://github.com/stnava/ANTs/commit/10d8c4ccecc80ed2f90a6512e4984ec8e713e2e8

effigies:
Uses the Neurodebian build.

Will test on a couple subjects to see if any additional changes need to be made.

Closes #516.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -59,8 +59,7 @@ RUN echo "cHJpbnRmICJrcnp5c3p0b2YuZ29yZ29sZXdza2lAZ21haWwuY29tXG41MTcyXG4gKkN2dW
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
                     fsl-core=5.0.9-4~nd16.04+1 \
-                    afni=16.2.07~dfsg.1-5~nd16.04+1 \
-                    ants=2.2.0-1~nd16.04+1
+                    afni=16.2.07~dfsg.1-5~nd16.04+1
 
 ENV FSLDIR=/usr/share/fsl/5.0 \
     FSLOUTPUTTYPE=NIFTI_GZ \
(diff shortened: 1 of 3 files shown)


effigies:
I've tested this on a UH2 and ds114 subject over the weekend. The brain masks look reasonable. Can't see obvious issues that might have resulted from N4 changes, but we generate reports explicitly of the output of that one.

chrisfilo:
Please add changelog entry and update docs.

effigies:
How should the version string be formatted? AFNI's is `version Debian-16.2.07`, and the previous ANTs was (I think incorrectly) `(version 2.1.0.Debian-Ubuntu_X64)`.

Also, NeuroDebian's version string is `2.1.0-gGIT-N`, so we should probably wait until they fix this before updating the Docker image.

effigies:
Related: https://lists.alioth.debian.org/pipermail/neurodebian-users/2017-June/001154.html

chrisfilo:
I think we'll need to use a workaround until this is fix. Don't want to get
stuck on this.

On Mon, Jun 19, 2017 at 10:45 AM, Chris Markiewicz <notifications@github.com

effigies:
Yarik said he was going to patch and rebuild, but in the meantime we can just use the Neurodocker package.

Dockerfile
```diff
@@ -72,15 +72,14 @@ ENV FSLDIR=/usr/share/fsl/5.0 \
...
-RUN mkdir -p /opt/ants && \
-    curl -sSL "https://github.com/stnava/ANTs/releases/download/v2.1.0/Linux_Ubuntu14.04.tar.bz2" \
-    | tar -xjC /opt/ants --strip-components 1
-
-ENV ANTSPATH /opt/ants
-ENV PATH $ANTSPATH:$PATH
+ENV PATH=/usr/lib/fsl/5.0:/usr/lib/afni/bin:$PATH
+
+# Installing ANTs 2.2.0 (NeuroDocker build)
+ENV ANTSPATH=/usr/lib/ants
```
effigies: Sticking with modified `ANTSPATH` to be consistent with NeuroDebian. Let me know if you'd rather keep `/opt/ants`. It's no big deal to me, either way.

chrisfilo: works for me

PR #566, opened 2017-06-17
FIX: update fsl-core from neurodebian

jdkent:
Neurodebian doesn't appear to have the [5.0.9-1](http://neuro.debian.net/pkgs/fsl-core.html#binary-pkg-fsl-core) fsl-core version anymore. This update brings the fsl-core version to the version available on neurodebian. I remember seeing a conversation about versioning with neurodebian, so maybe there is a more general solution available.


effigies:
Looks good. Thanks. I'll merge when tests pass.

PR #569, opened 2017-06-19
[RTM] ENH: Output T1-space tissue probability maps, MNI-space segmentations

### LINKED ISSUE #559: probtissue present only in MNI space dtissue only in T1w space
image
Both should be in both spaces.

effigies:
Closes #559


effigies:
Not quite right...

dtissue

effigies:
dtissue2

PR #577, opened 2017-06-29
ENH: Add cosine-based high-pass filter interface

### LINKED ISSUE #361: Perform simple denoising before estimating CompCor
CompCor can pick on trivial noise patterns that would be removed through other means. To avoid this and let CompCor focus on the non-trivial noise the following simple denoising steps should be performed prior to signal extraction:

- Regressing out non-steady state outliers (see #360)
- High-pass filter with 0.008 Hz (remove signals with slower frequencies than 0.008 Hz)

effigies:
Closes #361.

### DIFF
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -28,4 +28,4 @@ dependencies:
     - svgutils
     - nitime
     - nilearn
-    - niworkflows>=0.1.3
+    - git+https://github.com/poldracklab/niworkflows.git@2de9c245b11f5328743c73d3ec80a969a20c6e5c#egg=niworkflows
--- a/docs/workflows.rst
+++ b/docs/workflows.rst
@@ -323,9 +323,11 @@ Confounds estimation
     :simple_form: yes
 
     from fmriprep.workflows.confounds import init_discover_wf
-    wf = init_discover_wf(name="discover_wf",
-                          use_aroma=False, ignore_aroma_err=False,
-                          bold_file_size_gb=3)
+    wf = init_discover_wf(
+        name="discover_wf",
+        use_aroma=False, ignore_aroma_err=False, bold_file_size_gb=3,
+        metadata={"RepetitionTime": 2.0,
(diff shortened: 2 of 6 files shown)


chrisfilo:
Great to see this coming along. Some comments:
- We need this for both a and t compcor
- It would be good to avoid saving the bandpass filtered intermediate files
- We should output the cosines in the confounds TSV file

It might be worth considering building this filtering directly into the compcor interface.

fmriprep/interfaces/images.py
```diff
@@ -239,6 +241,49 @@ def _run_interface(self, runtime):
...
+    design_matrix = File(exists=True)
+    residuals = File(exists=True)
+
+
+class HighPassFilter(SimpleInterface):
+    input_spec = HighPassFilterInputSpec
+    output_spec = HighPassFilterOutputSpec
+
+    def _run_interface(self, runtime):
+        TR = 2.5
```
chrisfilo: hardcoded?

effigies: Temporary.

effigies:
Yeah, if we don't want to have a file of residuals for CompCor to work on, it's going to need to be directly in the interface.

effigies:
Ready for review. Merge depends on nipy/nipype#2107 and bumping niworkflows.

chrisfilo:
Not sure why the tests are failing, but otherwise looks ready to merge.

effigies:
The bug is when the BOLD series is too short to have a cosine regressor. @jdkent resolved this by returning `None`. I don't want to encode that into the nipype interface, so I'll try to figure out a better way to handle that on Wednesday.

effigies:
Anybody know how `mock` works? There's a failure that's [appearing](https://circleci.com/gh/effigies/fmriprep/605#tests/containers/1) in the tests. I can try tracking it down, but perhaps the issue would be obvious to someone who's used it before.

effigies:
The failing test is now passing.

Issue #260, opened 2016-12-09
normalizing/conforming input data


chrisfilo:
To make sure data fed to FMRIPREP from different datasets is looking as similar as possible we are currently reorienting anatomical images to RAS. However, there is more that can be done:
- make sure that the information in the header concerning data type and affine transformation (sform, qform) are correct; fix if possible
- apply the same procedure to all input data (_bold, _sbref, fieldmaps etc.)

Problems:
- If we change the orientation of input files we need to adjust the phase encoding direction as well.
- Applying those fixes to big 4D _bold files will create an (almost) identical copy and waste a lot of disk space

oesteban:
Possible duplicate of #581, #388, #433 ?

effigies:
Copying in part of https://github.com/poldracklab/fmriprep/issues/573#issuecomment-313799752:

> To address further issues raised in #260:
> 
> * If we change the orientation of input files we need to adjust the phase encoding direction as well.
> 
> NIfTI does have a phase_dim header entry. We could update the header, and use that later.
> 
> * Applying those fixes to big 4D _bold files will create an (almost) identical copy and waste a lot of disk space
>
> True, and adding the PE direction header entry will cause the same issue, if we don't already need to create a new file.
> 
> I haven't tried this out yet, but we could consider creating a filename.hdr.gz with a data offset matching the original header, and symlink to the original file with filename.img.gz. Assuming this works, we couldn't reorient, but we could update the headers without copying the data.

oesteban:
Superseded by new issue (link above)

PR #581, opened 2017-07-10
[RTM] ENH/FIX: Validate and minimally conform BOLD images

### LINKED ISSUE #573: BOLD skullstripping failure
https://openneuro.org/datasets/202020206473303030303137/versions/3030303031372d3030303031?app=FMRIPREP&version=5

effigies:
As discussed in #573, BOLD series with invalid headers (no valid qform/sform) can cause a number of issues, including failure of skull stripping. I've also found that bbregister will fail to run.

It was decided that, instead of rejecting these images, we allow nibabel to assume an LAS orientation and set a default sform, and note this prominently in the reports.

The current warning is:

-----

### WARNING - Invalid header

Input file does not have valid qform or sform matrix. A default, LAS-oriented affine has been constructed. A left-right flip may have occurred. Analyses of this dataset MAY BE INVALID.

-----

Related #260 

Closes #573


effigies:
Here's what it looks like on `ds000017/sub-3`:

sample

(The black outlines in the middle of the image are a quirk of my screenshot's "select area" tool.)

PR #583, opened 2017-07-10
ENH: minor typos + func change while going through the code

yarikoptic:


fmriprep/workflows/base.py
```diff
@@ -104,7 +104,7 @@ def init_single_subject_wf(subject_id, task_id, name,
                             "All workflows require BOLD images.".format(
                                 subject_id, task_id if task_id else '<all>'))
 
-        if subject_data['t1w'] == []:
+        if not subject_data['t1w']:
```
yarikoptic: FWIW (why): more pythonic and a good idiom for generic programming (avoiding explicit data typing)

PR #584, opened 2017-07-10
[RTM] Add summary reports

### LINKED ISSUE #574: Suggestion to add list of pipeline steps to report
A small suggestion:

I am trying to check that fmriprep has run correctly on a new dataset, and there are a couple of steps I can't verify-- for instance, I can't tell from the report or the output files whether slice-timing actually ran. It'd be useful to have a list of all pipeline steps (completed, skipped, or failed) included in the report file, whether there's a corresponding visualization or not. I imagine I can dig through the terminal output, but better to have it saved in the stand-alone report. Thanks!

### LINKED ISSUE #594: Links to "about" and "erros" are repeated
image

https://s3.amazonaws.com/openneuro.outputs/b2b76d1b40fe1196eb7db904aaf4bdf1/d7d6f355-ca15-4dc4-97bf-544085dc5188/fmriprep/sub-2.html

effigies:
First pass at adding summary reports to the anatomical and functional workflows.

Todo:

* [x] Generate reportlets
* [x] Include reportlets in report
* [x] Include FMRIPREP version in report

Should we be adding the version information at the top, or bottom (just before exceptions)? And anything besides the version string? (Our version string should be fairly informative, since we pin niworkflows/nipype.)

Closes #574
Closes #594

### DIFF
--- a/fmriprep/interfaces/reports.py
+++ b/fmriprep/interfaces/reports.py
@@ -61,6 +61,7 @@ class AnatomicalSummaryInputSpec(BaseInterfaceInputSpec):
     subjects_dir = Directory(desc='FreeSurfer subjects directory')
     subject_id = traits.Str(desc='FreeSurfer subject ID')
     output_spaces = traits.List(desc='Target spaces')
+    template = traits.Enum('MNI152NLin2009cAsym', desc='Template space')
 
 
 class AnatomicalSummaryOutputSpec(SummaryOutputSpec):
@@ -89,9 +90,13 @@ class AnatomicalSummary(SummaryInterface):
             else:
                 freesurfer_status = 'Run by FMRIPREP'
 
+        output_spaces = [self.inputs.template if space == 'template' else space
+                         for space in self.inputs.output_spaces
+                         if space[:9] in ('fsaverage', 'template')]
+
         return ANATOMICAL_TEMPLATE.format(n_t1s=len(self.inputs.t1w),
                                           freesurfer_status=freesurfer_status,
-                                          output_spaces=', '.join(self.inputs.output_spaces))
+                                          output_spaces=', '.join(output_spaces))
 
 
 class FunctionalSummaryInputSpec(BaseInterfaceInputSpec):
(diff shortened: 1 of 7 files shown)


effigies:
Note that in order to detect the FreeSurfer situation, the anatomical summary needs to be run before recon-all, so I added a dependency to enforce ordering.

effigies:
Last commit fixes a display issue introduced in #587, which causes reports to double up the "About" and "Errors" headers. Sorry about that. Once this tests clean, I'd like to bump 0.5.4 to fix the report.

effigies:
WRT the above comment, see the header of this sample report.

https://958-53175327-gh.circle-artifacts.com/0/home/ubuntu/ds054/out/fmriprep/sub-100185.html

effigies:
Fixed reports from latest test:

* https://638-78558509-gh.circle-artifacts.com/1/home/ubuntu/ds005/out/fmriprep/sub-01.html
* https://638-78558509-gh.circle-artifacts.com/0/home/ubuntu/ds054/out/fmriprep/sub-100185.html

effigies:
@chrisfilo Do you have time for a review? I'd like to push 0.5.4 with this reporting improvement.

chrisfilo:
1. Shouldn't AROMA be listed in the confounds for ds005? https://638-78558509-gh.circle-artifacts.com/1/home/ubuntu/ds005/out/fmriprep/sub-01.html
2. Could you update the example report in the docs?

Otherwise ready to merge.

effigies:
@chrisfilo 

> Shouldn't AROMA be listed in the confounds for ds005?

In this case, it isn't showing up because there were no confounds found. (The file had size 0.) This isn't a very likely case except in these test sets, but I can make sure it shows up as long as the file exists.

> Could you update the example report in the docs?

Will do.

effigies:
Oh, right ICA-AROMA returns `None`, rather than an empty file when no confounds are found. That should be a quick fix, but I'd rather get this out than hold this PR up for another round of tests for an edge case. I'll open an issue.

PR #586, opened 2017-07-11
FIX: Bad arg in pe_chooser Function interface

effigies:
This might be causing some of our problems when PE direction is unknown.


effigies:
To catch this, we could remove the PE direction from one of the BOLD series in the sample ds054. So we'd do fieldmaps + SyN for one series and 2x SyN for the other.

effigies:
Merging because this needs doing, but we should consider updating the testing dataset.

Issue #474, opened 2017-04-24
Coregister T2w images


effigies:
#439 raises the issue of cross-session scaling differences between T1w images, which introduces the possibility that T2w images may also need to be scaled, and not merely realigned to the T1w images.

Reading the `recon-all` log for uh2/sub-s483, the T2 image is aligned using `bbregister`, which is 6-dof by default.

It seems reasonable to do a 9-dof registration of T2w images, which would then be saved to `derivatives/fmriprep/<subject>/anat/<subject>_T2w_preproc.nii.gz` (or should it be `<subject>_T2w_space-T1w_preproc.nii.gz`?). This re-aligned T2 would also be passed to FreeSurfer.

Questions:

Does FLIRT work well with T2w->T1w coregistrations? `bbregister` has a `--t2` flag to indicate the expected contrast in the moving image; it's not clear from the docs how that should be done with FLIRT. Or should we go ahead and do `bbregister --9`? Which would mean we need to wait until after the first `white` surfaces are created (`autorecon2`) to load the T2.

I'm inclined toward using FLIRT to align all of the structurals at the very beginning, because we run the risk of getting too complicated with the `recon-all` pipeline.

chrisfilo:
Yes - let's try FLIRT or ANTs for this first.

oesteban:
For inter-modality registration I'd try ANTs first. I can take this one if you want :+1:

effigies:
If you like. It's a little moot at the moment, simply because we were wrong that we needed non-rigid transformations between structurals. But sure, if you want to do this.

PR #591, opened 2017-07-18
[RTM] ENH: Merge T1w images to unbiased template if N < 3 or --longitudinal is passed

### LINKED ISSUE #310: Support for multisession data with multiple T1s

effigies:
Not running against CircleCI at this point because we'll need to test outside of CI against datasets with multiple T1w images before we merge this. This is a placeholder PR until then.

I'd also like to go ahead and use `mri_robust_register` to align any T2w images (see #474).

Note that this could break alignment between result images from previous FMRIPREP versions. Should probably be put into release notes.

Closes #310


effigies:
This adds serious run time to the `t1_merge` node for ds031, which merges 9 T1w images (testing after rebasing on top of #601). We could make it an option (`--longitudinal`?) with a note that runtime may be increased.

Also, I think we can use OpenMP, so I'll give that a shot.

chrisfilo:
how serious?

effigies:
It's taken over half an hour. Before I think it took ~10min, but I'll run both versions tonight with less other things going on the system to get a difference.

Once I was sure it was a major slowdown, I killed it to try to play with OpenMP. Just to verify that it does actually parallelize.

effigies:
For ds031/sub-01 (9 images):

Alignment | `OMP_NUM_THREADS` | Time
--|--|--
Aligning to first image| | 16m
Aligning to first image | 7 | 5m
Unbiased template|| 1h46m
Unbiased template | 7 | 1h21m

effigies:
Re: ds031, it's unclear whether we'd get a better speed-up profile if the number of images matched the number of threads; it could be that the parallel sections are going from 9 to 2. But even if so, it looks like the serial component for constructing an unbiased template is ~74min, and maximum concurrency might bring us down to 77-78min.

Similarly, for a simple alignment, it looks like we have a serial section of ~2.5min with maximum concurrency reducing to 3.6-3.7min.

I would say it makes sense to parallelize to `min(num_images, omp_nthreads)`, which keeps small merges from eating up all the cores to little benefit, and to make unbiased templates optional. Given the low additional cost for a small number of input images, it could also make sense to make unbiased templates the default for fewer than N images (and I'll work out what a reasonable N is).

chrisfilo:
That's really long! I agree with your suggestions.

effigies:
Haha, wow. N=2. Going from 2 to 3 images increases from <2min to ~25min. I do not need to see the curve badly enough to keep going.

effigies:
It's the stochastic ICA-AROMA bug. If it pops up again, I'll merge #611 before running another round of tests.

PR #605, opened 2017-07-25
[RTM] ENH: Save text crashfiles in output directory

### LINKED ISSUE #240: Use output log out/log/run_workflow
We are writing a small log in ```out/log/run_workflow```.

Currently, the content of that file after running is just a list of subjects processed.

I would love to have a file like that with all the output from processing (preferably with .txt extension so that it is convenient to open with the browser from, i.e. circle's artifacts).

### LINKED ISSUE #595: Pipeline crashes, but no errors are reported in the HTML report
just like poldracklab/mriqc#434

effigies:
Fixes #595.

This creates a `derivatives/fmriprep/logs` directory and saves logs and text crashfiles there.

Some other minor cleanups I saw while in there. Probably most significant is only building the graph if we're going to run it or write the graph. This allows `--reports-only` without `--write-graph` to run more quickly.

Closes #240.

### DIFF
--- a/fmriprep/cli/run.py
+++ b/fmriprep/cli/run.py
@@ -244,45 +244,43 @@ def create_workflow(opts):
 
     logger.info('Subject list: %s', ', '.join(subject_list))
 
-    # Build main workflow and run
+    # Build main workflow
     reportlets_dir = op.join(op.abspath(opts.work_dir), 'reportlets')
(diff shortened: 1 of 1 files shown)


effigies:
Hmm. If this worked, shouldn't log files have shown up in the artifacts? @oesteban would you mind seeing if you see something obvious? Like, should I move the configuration update to earlier/later in the program?

PR #606, opened 2017-07-25
added zenodo entry

chrisfilo:
@jdkent, @craigmoodie, @feilong, @erramuzpe, @zenkavi, @ilkayisik, @yarikoptic please post your pen names, affiliations and ORCID if you would like to be included as coauthors in the Zenodo entry for FMRIPREP.

PR #603, opened 2017-07-25
[RTM] ENH: Add non-steady-state detection before CompCor

effigies:
Closes #361.

Because pin includes poldracklab/niworkflows#188:

Fixes #616
Closes #619

PR #602, opened 2017-07-25
[RTM] FIX: Ensure un-nested lists are not over-flattened

effigies:
Do we have a good dataset to test this? It showed up with blip-up/blip-down, presumably single volume field maps.

Closes #598.

PR #601, opened 2017-07-25
[RTM] FIX: Correctly resample T1w images in ConformSeries

effigies:
This PR stems from #599, which revealed that we've been conforming wrongly. The error was so small that the error only appears for highly anisotropic voxels in an oblique orientation.

In addition, it introduced the case where high- and low-quality T1w scans coexist in the same dataset, and the upsampling involved would ultimately degrade the quality of a template produced by averaging the images.

The following changes are introduced:

* Rescaling is applied to affine columns, as opposed to rows, which will correctly set the zooms in the resampled dataset
* A size (dimension) factor is recalculated to be less sensitive to large differences in dimension
* A `max_scale` factor (default: 3) discards images which require upsampling over a given ratio
* An HTML reportlet is created to describe the conformation step

Todo:

* [x] Integrate reportlet into report
* [x] Test on ds000114 sub-02 (See #538/#545)
* [x] Test on ds000114 sub-07/09 (See #562/#564)
* [x] Test on ds000031

Closes #599.

PR #611, opened 2017-07-26
ENH: Use updated ICA-AROMA

### LINKED ISSUE #578: Stochastic ICA-AROMA bug
- https://circleci.com/gh/effigies/fmriprep/589#tests/containers/1
- https://circleci.com/gh/poldracklab/fmriprep/929#tests/containers/1

I suspect this is related to either the low resolution or the small number of time points. It's the ICA-AROMA interface failing to run, so the solution might need to be in nipype.

effigies:
Closes #578

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,7 @@ With thanks to Yaroslav Halchenko and Ilkay Isik for contributions.
 * [FIX] Conformation failure in thick-slice, oblique T1w datasets (#601)
 * [FIX] Crash/report failure of phase-difference SDC pipeline (#602, #604)
 * [DOC] Save logs to output directory (#605)
+* [ENH] Upgrade to ICA-AROMA 0.4.1-beta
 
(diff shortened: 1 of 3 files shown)


effigies:
Would recommend holding off on merging this until we're ready to make a release.

I think we should be able to merge #601, #602, #603 (assuming these latest tests pass), #608 and #612 today or tomorrow. How do we feel about a 0.5.5/0.6.0 release, once these are in?

effigies:
Pre-release status:

 - #619 is a fix for an (admittedly rare) bug, so it would be good to get it in.
 - Got sign off from Satra on nipy/nipype#2122, so #603 is a couple merges away.
 - If we want to make a Zenodo release, we should get #606 in.
 - #608 doesn't seem to be quite ready to go for this round.

I think the next release should be 0.6.0, since we are upgrading ICA-AROMA.

Getting late here, so maybe aim to merge things over the weekend as tests pass and release Monday morning?

Issue #225, opened 2016-12-05
Revise memory constraints throughout workflow, add missing


chrisfilo:
...for bottleneck nodes that are expected to use a lot of memory (see #223)

oesteban:
I take this one. And make it broader, I'll do a full code-review.

oesteban:
Please check out https://github.com/poldracklab/fmriprep/pull/629#issuecomment-319528238

We should probably set the memory limits to all nodes.

PR #621, opened 2017-07-29
[ENH] Revise command line script

oesteban:
Part of #225.

In this PR the `run.py` script is revised:

- [x] Create a new module `fmriprep.utils.bids` to bring together those methods handling bids. This module is a potential candidate to be wrapped in niworkflows and use it with mriqc.
- [x] Some improvements to the logging, will work on this in a future.
- [x] Reorganized logical flow (which, btw was broken: it did not have much sense to run `--reports-only` and `--write-graph`.
- [x] Checking some flags brought earlier. We probably want to customize the parser to remove the clutter from run.py. For now we only have two checks, so no big deal.
- [x] Some linting and PEP8 fixes.

Added 07/31/2017:
- [x] Replaced tests under `test/utils/` and `test/interfaces/` with doctests, which are easier to understand, help with documentation and easier to maintain (since they go alongside the corresponding units under test).
- [x] Replaced python unittest with py.test, which was installed anyways.
- [x] Moved BIDS-examples downloads to niworkflows (see: https://github.com/poldracklab/niworkflows/pull/189)


fmriprep/cli/run.py
```diff
@@ -9,21 +9,34 @@
...
 import uuid
+import warnings
 from argparse import ArgumentParser
 from argparse import RawTextHelpFormatter
 from multiprocessing import cpu_count
 from time import strftime
+from fmriprep.info import __version__
 
-from niworkflows.nipype import config as ncfg
+logging.addLevelName(25, 'INFO')  # Add a new level between INFO and WARNING
```
effigies: If this is between INFO and WARNING, shouldn't it have a different name than INFO?

fmriprep/cli/run.py
```diff
@@ -212,78 +226,78 @@ def create_workflow(opts):
...
-        if not subject_list:
-            print('Could not find subjects in {}\n'
-                  'If you are using Docker for Mac or Docker for Windows, you '
-                  'may need to adjust your "File sharing" preferences.'.format(bids_dir))
-            sys.exit(1)
-    else:
-        subject_list = [sub[4:] if sub.startswith('sub-') else sub for sub in subject_list]
+    # Set up directories
+    output_dir = op.abspath(opts.output_dir)
+    log_dir = op.join(opts.output_dir, 'fmriprep', 'logs')
```
effigies: `output_dir`, not `opts.output_dir`, to make sure we're using an absolute path.

fmriprep/cli/run.py
```diff
@@ -212,78 +226,78 @@ def create_workflow(opts):
...
+    # Nipype config (logs and execution)
+    ncfg.update_config({
+        'logging': {'log_directory': log_dir, 'log_to_file': True},
+        'execution': {'crashdump_dir': log_dir, 'crashfile_format': 'txt'},
+    })
 
+    # Called with reports only
     if opts.reports_only:
-        if opts.write_graph:
-            fmriprep_wf.write_graph(graph2use="colored", format='svg', simple_form=True)
```
effigies: You don't ever want to get the graph without running?

fmriprep/utils/bids.py
```diff
@@ -0,0 +1,134 @@
...
+"""
+Utilities to handle BIDS inputs
+"""
+import os.path as op
+import warnings
+from glob import glob
+from copy import deepcopy
+from bids.grabbids import BIDSLayout
+
+INPUTS_SPEC = {'fieldmaps': [], 'func': [], 't1': [], 'sbref': [], 't2w': []}
```
effigies: This can be removed, I think.

fmriprep/utils/bids.py
```diff
@@ -0,0 +1,134 @@
...
+    Lists the participants under the BIDS root and checks that participants
+    designated with the participant_label argument exist in that folder.
+
+    Returns the list of participants to be finally processed.
+
+
+    """
+    bids_dir = op.abspath(bids_dir)
+    all_participants = [op.basename(subdir)[4:]
+                        for subdir in glob(op.join(bids_dir, 'sub-*'))]
```
effigies: Not a big deal to change, but `glob` seems a bit heavy for this job:

```Python
all_participants = [subdir[4:] for subdir in os.listdir(bids_dir) if subdir.startswith('sub-')]
```

Could also add an `op.isdir` check, if there's any concern there.

fmriprep/utils/bids.py
```diff
@@ -0,0 +1,134 @@
...
+    found_label = list(set(participant_label) & set(all_participants))
+    if not found_label:
+        raise RuntimeError('Could not find participants [{}] in folder '
+                           '"{}".'.format(', '.join(participant_label), bids_dir))
+
+    notfound_label = list(set(participant_label) - set(all_participants))
+    if notfound_label:
+        warnings.warn('Some participants were not found: {}'.format(
+            ', '.join(notfound_label)), RuntimeWarning)
+    return found_label
```
effigies: Do we want the outputs (and exception/warning messages) of this function to be sorted? Feel like if I'm looking through logs, it'll be easier to see an issue in a list of participant labels if they're sorted than some random order generated by `set` or `glob`.

fmriprep/utils/bids.py
```diff
@@ -0,0 +1,134 @@
...
+        query_kwargs.append([(key, x) for x in get_kwargs[key]])
+
+    query_kwargs = itertools.product(*query_kwargs)
+
+    for elem in query_kwargs:
+        epi_files = [x.filename for x
+                     in layout.get(**dict(dict(elem), **queries['epi']))]
+        if epi_files:
+            imaging_data['func'] += epi_files
+    '''
```
effigies: This is in the git history. I think we can safely get rid of a commented section.

fmriprep/utils/bids.py
```diff
@@ -0,0 +1,134 @@
...
+    fieldmap_files = [x.filename for x in layout.get(**queries['fmap'])]
+    imaging_data['fmap'] = fieldmap_files
+    t1_files = [x.filename for x in layout.get(**queries['t1w'])]
+    imaging_data['t1w'] = t1_files
+    sbref_files = [x.filename for x in layout.get(**queries['sbref'])]
+    imaging_data['sbref'] = sbref_files
+    epi_files = [x.filename for x in layout.get(**queries['epi'])]
+    imaging_data['func'] = epi_files
+    t2_files = [x.filename for x in layout.get(**queries['t2w'])]
+    imaging_data['t2w'] = t2_files
```
effigies: Could really just replace all of this with:

```Python
return {modality: [x.filename for x in layout.get(**query)]
        for modality, query in queries.items()}
```

effigies:
Also, are you wanting to merge this and #622 changes before 0.6.0?

oesteban:
Yes, thanks for your comments. I'll incorporate them tomorrow morning

fmriprep/cli/run.py
oesteban: We can call it IMPORTANT or so, but it is just the name that appears in the logs.

I made this because the default log level in python is WARNING, so those loggers not handled with nipype were writing very little information. With this extra level of detail we can have some INFO messages that are printed by default, and some INFO more verbose.

The alternative is give this level 25 a name like IMPORTANT. But I don't think there is any gain.

In the long term, we should aim to control all the logs (nipype included) in a centralized way. My impression is that INFO is too detailed in nipype, with this new level we will easily filter those out.

Also, an easier way to specify the verbosity would be nice. In MRIQC I started with `-v`, `-vv`, `-vvv`, etc. to increase the verbosity level and it works very well.

effigies: Ah, gotcha. I thought it would override the regular "INFO". This is fine.

And yeah, verbosity flags for better control would be nice.

fmriprep/cli/run.py
oesteban: Good catch

fmriprep/cli/run.py
oesteban: We can add another argument to implement that, but it wasn't working that way even before. To do that, you would also run the reports. There wasn't any way to "just generate the graph without running".

Since we have the sphinx extension to write all the graphs, I don't think writing it is a realistic use case. We can leave it there as an extra when you execute the pipeline but I don't have a strong feeling about needing a better implementation for this.

oesteban: One thing I meant is: `--reports-only` means you want just the reports, and it should be independent of `--write-graph`. If you really just want the graph, it wasn't clear that you also needed to say `--reports-only` in the former version. It was like a hidden feature.

effigies: It was just a way of making `--write-graph` continue to work when `--reports-only` was run. But yeah, I don't know it would be obvious if your goal was to show the execution graph that that combo would do the trick.

I'm okay with leaving it out. If we need it back, we could add `--dry-run` to switch off `workflow.run()`.

fmriprep/utils/bids.py
oesteban: Indeed, I really wanted to get rid of this.

oesteban: I'm also removing this section a bit before:
```
    if session:
        session_list = [session]
    else:
        session_list = layout.unique('session')
        if session_list == []:
            session_list = [None]

    if run:
        run_list = [run]
    else:
        run_list = layout.unique('run')
        if run_list == []:
            run_list = [None]
```
which is not used.

fmriprep/utils/bids.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+def collect_data(dataset, subject, task=None):
+    """
+    Uses grabbids to retrieve the input data
+    """
+    layout = BIDSLayout(dataset)
+    queries = {
+        'fmap': {'subject': subject, 'modality': 'fmap', 'extensions': ['nii', 'nii.gz']},
+        'bold': {'subject': subject, 'modality': 'func', 'type': 'bold',
```
oesteban: Please note I changed `'epi'` to `'bold'`. Some functional data are not acquired with EPI.

fmriprep/utils/bids.py
```diff
@@ -0,0 +1,78 @@
...
+            'may need to adjust your "File sharing" preferences.'.format(bids_dir))
+
+    # No --participant-label was set, return all
+    if participant_label is None or not participant_label:
+        return all_participants
+
+    # Drop sub- prefixes
+    participant_label = [sub[4:] if sub.startswith('sub-') else sub for sub in participant_label]
+    # Remove duplicates
+    participant_label = sorted(list(set(participant_label)))
```
effigies: sorted takes an iterable and returns a list. Can drop the list().

fmriprep/utils/bids.py
```diff
@@ -0,0 +1,78 @@
...
+    # Remove labels not found
+    found_label = sorted(list(set(participant_label) & set(all_participants)))
+    if not found_label:
+        raise RuntimeError('Could not find participants [{}] in folder '
+                           '"{}".'.format(', '.join(participant_label), bids_dir))
+
+    # Warn if some IDs were not found
+    notfound_label = list(set(participant_label) - set(all_participants))
```
effigies: sorted here, too.

fmriprep/utils/bids.py
```diff
@@ -0,0 +1,78 @@
...
+    # No --participant-label was set, return all
+    if participant_label is None or not participant_label:
+        return all_participants
+
+    # Drop sub- prefixes
+    participant_label = [sub[4:] if sub.startswith('sub-') else sub for sub in participant_label]
+    # Remove duplicates
+    participant_label = sorted(list(set(participant_label)))
+    # Remove labels not found
+    found_label = sorted(list(set(participant_label) & set(all_participants)))
```
effigies: Same as above.

effigies:
Does this look like a quick fix, or should we push these two off to the next release?

oesteban:
Sure, np

oesteban:
@effigies @chrisfilo it'd be great that we all explicitly agree on that transforming as many tests as we can to doctests is a good idea.

Doctests would also play better with @effigies' future idea of adding the code to the generation of documentation to also cover the API.

oesteban:
Thanks a lot :). I'm enjoying this housekeeping! :+1:

Issue #624, opened 2017-07-30
anat affine appears off for ds000114 between data and fmriprep output


satra:
any reason why the T1 brainmask output from fmriprep has different affines from original T1 of the dataset? is this because of the conform step?

```
$ mri_info ../sub-01/anat/sub-01_t1w_brainmask.nii.gz 
Volume information for ../sub-01/anat/sub-01_t1w_brainmask.nii.gz
          type: nii
    dimensions: 256 x 156 x 256
   voxel sizes: 1.000000, 1.299473, 1.000000
          type: FLOAT (3)
           fov: 256.000
           dof: 0
        xstart: -128.0, xend: 128.0
        ystart: -101.4, yend: 101.4
        zstart: -128.0, zend: 128.0
            TR: 0.00 msec, TE: 0.00 msec, TI: 0.00 msec, flip angle: 0.00 degrees
       nframes: 1
       PhEncDir: UNKNOWN
       FieldStrength: 0.000000
ras xform present
    xform info: x_r =   0.9992, y_r =  -0.0393, z_r =   0.0096, c_r =    -2.1763
              : x_a =   0.0395, y_a =   0.9989, z_a =  -0.0269, c_a =    22.5293
              : x_s =  -0.0085, y_s =   0.0272, z_s =   0.9996, c_s =    -8.5970
Orientation   : RAS
Primary Slice Direction: axial

voxel to ras transform:
                0.9992  -0.0510   0.0096  -127.3170
                0.0395   1.2980  -0.0269   -80.3350
               -0.0085   0.0354   0.9996  -138.2132
                0.0000   0.0000   0.0000     1.0000

voxel-to-ras determinant 1.29947

ras to voxel transform:
                0.9992   0.0395  -0.0085   129.2109
               -0.0302   0.7687   0.0209    60.7968
                0.0096  -0.0269   0.9996   137.2195
                0.0000   0.0000   0.0000     1.0000

$ mri_info ../../../sub-01/ses-test/anat/sub-01_ses-test_T1w.nii.gz 
Volume information for ../../../sub-01/ses-test/anat/sub-01_ses-test_T1w.nii.gz
          type: nii
    dimensions: 256 x 156 x 256
   voxel sizes: 1.000000, 1.299376, 1.000000
          type: FLOAT (3)
           fov: 256.000
           dof: 0
        xstart: -128.0, xend: 128.0
        ystart: -101.4, yend: 101.4
        zstart: -128.0, zend: 128.0
[...]

PR #628, opened 2017-07-31
[WIP] ENH: Generate conformation transforms

effigies:
Insomnia led to playing around with this. Includes an assert to verify that the transform actually does get you from one affine to another. Tested on ds031.

Looks like `mri_robust_template` only outputs [LTA transforms](https://surfer.nmr.mgh.harvard.edu/fswiki/FsTutorial/LtaFormat).

Assuming this is worthwhile, I'd propose the following derivative structure:

```
derivatives/
    fmriprep/
        anat/
            transforms/
                <orig_T1w_name>.mat
```

Related: #624


effigies:
Closed in favor of #726.

PR #629, opened 2017-07-31
[RTM] Code revision of functional workflows

oesteban:
(part of #225, depends on #621)

This PR revises the code of the functional workflows:

- [x] Use relative imports
- [x] Nomenclature: EPI replaced with BOLD
- [x] PEP8 fixes (mostly indentations)
- [x] Memory constrains
- [x] https://github.com/oesteban/fmriprep/pull/1


effigies:
@oesteban Note that I've merged master to get tests going.

oesteban:
Thanks, I think I still need to work on this one quite a bit

On Aug 1, 2017 07:24, "Chris Markiewicz" <notifications@github.com> wrote:

@oesteban <https://github.com/oesteban> Note that I've merged master to get
tests going.

—
<https://github.com/poldracklab/fmriprep/pull/629#issuecomment-319385613>,
or mute the thread

effigies:
Cool. Was starting a review. Do you want some comments now, or should I wait until you're ready?

oesteban:
Better waiting for a finer iteration :+1:

oesteban:
I just realized that we are not setting `estimated_memory_gb` in most of the nodes, and [nipype's default is 250MB](https://github.com/nipy/nipype/blob/94a11f09ff79b328f788e41f2ba5a3444d5c4e12/nipype/interfaces/base.py#L775)

For many cores execution (i.e. stampede2) it may be worth lowering that to a tiny amount (I've set 10MB) for nodes that we are sure take nothing. Also, in systems with very limited memory like Circle, we might see some speedups if there are false memory bottlenecks that the scheduler is "averting".

fmriprep/workflows/bold.py
```diff
@@ -84,18 +92,19 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
     # Build workflow
     workflow = pe.Workflow(name=name)
     inputnode = pe.Node(niu.IdentityInterface(
-        fields=['epi', 't1_preproc', 't1_brain', 't1_mask', 't1_seg', 't1_tpms',
+        fields=['bold', 't1_preproc', 't1_brain', 't1_mask', 't1_seg', 't1_tpms',
                 't1_2_mni_forward_transform', 't1_2_mni_reverse_transform',
                 'subjects_dir', 'subject_id', 'fs_2_t1_transform']),
         name='inputnode')
-    inputnode.inputs.epi = bold_file
+    inputnode.inputs.bold = bold_file
```
effigies: How about making `'bold'` `'bold_file'`? I think that will help make it clear that it's the `bold_file` that's passed to this function, and that `name_source` (which is what it's often passed to) is looking at the original BOLD file name.

fmriprep/workflows/bold.py
```diff
@@ -15,26 +19,30 @@
...
+from ..interfaces.nilearn import Merge
+from ..interfaces.reports import FunctionalSummary
+from ..workflows import confounds
+from ..workflows.fieldmap.unwarp import init_pepolar_unwarp_wf
+from ..workflows.util import (
+    init_enhance_and_skullstrip_bold_wf, init_skullstrip_bold_wf,
     init_bbreg_wf, init_fsl_bbr_wf)
 
+
+DEFAULT_MEMORY_MIN_GB = 0.01
```
effigies: Not for this PR, but this makes me wonder if we can make this a config option in nipype. It would be nice to set an app-level default.

effigies:
Sorry about the flurry of commits. And forgetting the [docs only] on the last one. Ah well.

PR #633, opened 2017-08-01
[ENH] Add test to check on fmriprep outputs

### LINKED ISSUE #150: Test that workflows produce the expected outputs
This depends on closing #68.

Generate a list of expected files (per workflow) that should exist in `derivatives/` and check that they were created by the workflow.

This would be a more programatic approach to ensure that we generate the outputs we described in #44.

oesteban:
Two new text-files under `.circle/data` contain the list of expected files generated after running the tests. When we change the derivatives specification or add/remove outputs, these files will require a manual update.

Closes #150


oesteban:
This is only waiting for tests to pass. Also, we may want to make sure that these are actually the outputs we are expecting:
- [DS005](https://github.com/oesteban/fmriprep/blob/d08a4bf8d1b0e30a3b70389166dd0cf5c0e2ce17/.circle/data/ds005_outputs.txt)
- [DS054](https://github.com/oesteban/fmriprep/blob/d08a4bf8d1b0e30a3b70389166dd0cf5c0e2ce17/.circle/data/ds054_outputs.txt)

If the outputs of FMRIPREP change, we'll need to update these two `$DATASET_outputs.txt` files.

.circle/data/ds005_outputs.txt
```diff
@@ -0,0 +1,808 @@
...
+ds005/out/freesurfer/sub-01/touch/rusage.mris_sphere.lh.dat
+ds005/out/freesurfer/sub-01/touch/rusage.mris_sphere.rh.dat
+ds005/out/freesurfer/sub-01/touch/segstats.touch
+ds005/out/freesurfer/sub-01/touch/talairach.touch
+ds005/out/freesurfer/sub-01/touch/wmaparc.stats.touch
+ds005/out/freesurfer/sub-01/touch/wmaparc.touch
+ds005/out/freesurfer/sub-01/touch/wmsegment.touch
+ds005/out/freesurfer/sub-01/trait_added
+ds005/out/freesurfer/sub-01/trait_added/_SaveMidthickness0
+ds005/out/freesurfer/sub-01/trait_added/_SaveMidthickness1
```
effigies: The `trait_added` things are weird...

oesteban: Good catch, opened #635

PR #637, opened 2017-08-02
[RTM] FIX: Add second reorient step, in case LIA image is created

### LINKED ISSUE #636: Merged anatomicals are in LIA orientation, not RAS
`mri_robust_template` does not output in RAS orientation.

effigies:
This PR adds a second `ConformSeries` step post-merge, which ensures that the final template is RAS oriented. This reduces to a `nibabel.as_closest_canonical`, and will not make a new copy (even in memory) if the input is already RAS.

Fixes #636

### DIFF
--- a/fmriprep/workflows/anatomical.py
+++ b/fmriprep/workflows/anatomical.py
@@ -68,6 +68,9 @@ def init_anat_preproc_wf(skull_strip_ants, output_spaces, template, debug, frees
                             ),
         name='t1_merge')
 
+    # 1.5 Reorient template to RAS, if needed (mri_robust_template sets LIA)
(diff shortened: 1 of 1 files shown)


effigies:
@chrisfilo WDYT? Now that this stage is three nodes (conform -> template -> reorient), it may also make sense to abstract into a sub-workflow, just to keep the main workflow clean.

Especially if we produce affine transforms to relate the input images to the outputs (#628), which after talking to Satra does seem like a good idea to me, I think it'll be necessary, but that's also a refactoring that can go in that PR.

chrisfilo:
I think extracting this is a good idea, but we should leave it until this patter will be used more than once.

As for transformation matrices I asked Satra if he really has a use for them.

effigies:
Sounds good. Merging.

Issue #615, opened 2017-07-27
Add full API to documentation


effigies:
Even if we don't have full documentation for every class yet, I think having a searchable API might be useful to some users, and it should be a pretty low cost effort with Sphinx. It will also help highlight which areas off the docs need attention.

For `fmriprep.interfaces`, may want to reuse nipype's population of inputs and outputs from the specs, though possibly with a little bit better formatting.

oesteban:
I was thinking of creating a sphinx extension for nipype. But it's just a wish, for this PR it makes sense just reusing nipype's current code.

PR #638, opened 2017-08-02
[RTM] DOC: Document workflows

effigies:
Beginning to address #615.

* [x] anatomical.py
* [x] base.py
* [x] bold.py (will wait on #629 to avoid excessive collision)
* [x] confounds.py
* [x] fieldmap/*.py (already done)
* [x] utils.py

### DIFF
--- a/docs/api/index.rst
+++ b/docs/api/index.rst
@@ -14,3 +14,6 @@ Workflows
 
 Interfaces
 ----------
+
+.. autoclass:: fmriprep.interfaces.StructuralReference
+.. autoclass:: fmriprep.interfaces.NormalizeSurf
--- a/docs/api/index.rst
+++ b/docs/api/index.rst
@@ -9,6 +9,7 @@ Workflows
 
 .. automodule:: fmriprep.workflows.anatomical
 .. automodule:: fmriprep.workflows.bold
+.. automodule:: fmriprep.workflows.util
 
 Interfaces
 ----------
--- a/docs/api/index.rst
+++ b/docs/api/index.rst
@@ -1,0 +1,14 @@
+.. include:: ../links.rst
+
+===
+API
+===
+
+Workflows
+---------
+
+.. automodule:: fmriprep.workflows.anatomical
+.. automodule:: fmriprep.workflows.bold
+
+Interfaces
+----------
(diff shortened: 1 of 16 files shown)


effigies:
Just a note: It would be good to add citations to workflow docstrings, where appropriate. If you see any that you immediately know the references for, a comment would be appreciated.

effigies:
With the docstrings (more-or-less) in place. How do we want to organize the API documentation?

Warning: I have extremely little experience with Sphinx. Strong opinions are welcome.

effigies:
I think all workflows are now documented. Interfaces are underway, and the rough structure of the API documentation is in place.

Also note that this exists: https://sphinx-autoapi.readthedocs.io/en/latest/

effigies:
https://743-78558509-gh.circle-artifacts.com/1/home/ubuntu/docs/api/index.html

fmriprep/workflows/anatomical.py
```diff
@@ -3,9 +3,25 @@
...
++++++++++++++++++++++
+
+``fmriprep`` uses FreeSurfer_ to reconstruct surfaces from T1w/T2w
+structural images.
+
+.. autofunction:: init_surface_recon_wf
+.. autofunction:: init_autorecon_resume_wf
+.. autofunction:: init_gifti_surface_wf
+
+  .. note ::
```
oesteban: This note appears just below the gifti_surface workflow and that's misleading. Please move it to `init_anat_preproc_wf` (or maybe we should just write some acknowledgment in the About FMRIPREP and remove these notes from here and there).

docs/api/index.rst
```diff
@@ -0,0 +1,19 @@
...
+
+.. automodule:: fmriprep.workflows.base
+.. automodule:: fmriprep.workflows.anatomical
+.. automodule:: fmriprep.workflows.bold
+.. automodule:: fmriprep.workflows.util
+
+Interfaces
+----------
+
+.. autoclass:: fmriprep.interfaces.StructuralReference
```
oesteban: What are the plans for interfaces? Maybe in a different PR?

effigies:
Can do interfaces in a separate pr. Put those in as examples, and one of them so that a link in gifti_surfaces_wf would resolve and not show up as an error in circle.

Can move it to its own file for now, so it resolves, but doesn't look as incomplete.

oesteban:
That makes sense, let's leave interfaces for another PR, but keeping those two where they are right now :+1:

effigies:
@oesteban Good to go?

PR #640, opened 2017-08-03
[RTM] Code revision - fieldmaps

oesteban:
#225 - depends on #629

### DIFF
--- a/fmriprep/interfaces/fmap.py
+++ b/fmriprep/interfaces/fmap.py
@@ -8,7 +8,6 @@ Interfaces to deal with the various types of fieldmap sources
 """
 from __future__ import print_function, division, absolute_import, unicode_literals
 
-import os
 from builtins import range
 import numpy as np
 import nibabel as nb
@@ -67,16 +66,16 @@ class FieldEnhance(SimpleInterface):
                 mask = sim.binary_erosion(
                     mask, struc,
                     iterations=self.inputs.mask_erode
-                    ).astype(np.uint8)  # pylint: disable=no-member
+                ).astype(np.uint8)  # pylint: disable=no-member
 
         self._results['out_file'] = fname_presuffix(
-            self.inputs.in_file, suffix='_enh', newpath=os.getcwd())
+            self.inputs.in_file, suffix='_enh', newpath=runtime.cwd)
         datanii = nb.Nifti1Image(data, fmap_nii.affine, fmap_nii.header)
 
         if self.inputs.unwrap:
             data = _unwrap(data, self.inputs.in_magnitude, mask)
             self._results['out_unwrapped'] = fname_presuffix(
-                self.inputs.in_file, suffix='_unwrap', newpath=os.getcwd())
+                self.inputs.in_file, suffix='_unwrap', newpath=runtime.cwd)
             nb.Nifti1Image(data, fmap_nii.affine, fmap_nii.header).to_filename(
                 self._results['out_unwrapped'])
 
--- a/fmriprep/interfaces/fmap.py
+++ b/fmriprep/interfaces/fmap.py
@@ -70,13 +70,13 @@ class FieldEnhance(SimpleInterface):
                     ).astype(np.uint8)  # pylint: disable=no-member
 
         self._results['out_file'] = fname_presuffix(
-            self.inputs.in_file, suffix='enh', newpath=os.getcwd())
+            self.inputs.in_file, suffix='_enh', newpath=os.getcwd())
         datanii = nb.Nifti1Image(data, fmap_nii.affine, fmap_nii.header)
 
         if self.inputs.unwrap:
             data = _unwrap(data, self.inputs.in_magnitude, mask)
             self._results['out_unwrapped'] = fname_presuffix(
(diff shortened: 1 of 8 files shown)


effigies:
@oesteban Can you merge/rebase master? Will make it easier to review.

fmriprep/interfaces/fmap.py
```diff
@@ -68,12 +69,14 @@ def _run_interface(self, runtime):
                     iterations=self.inputs.mask_erode
                     ).astype(np.uint8)  # pylint: disable=no-member
 
-        self._results['out_file'] = genfname(self.inputs.in_file, suffix='enh')
+        self._results['out_file'] = fname_presuffix(
+            self.inputs.in_file, suffix='enh', newpath=os.getcwd())
```
effigies: `genfname` put a `_` before suffices, so this should be `suffix='_enh'` to keep behavior unchanged.

And since we're in `_run_interface`, we can avoid the syscall with `runtime.cwd`. (I know that this isn't a necessary optimization since we're not bottlenecked by syscalls; it's just a pattern I've been using in my interfaces to use the information provided in `runtime` when possible.)

(Not going to tag all of the uses of `fname_presuffix`, but both comments apply to most.)

fmriprep/workflows/fieldmap/fmap.py
```diff
@@ -96,9 +91,11 @@ def init_fmap_wf(reportlets_dir, omp_nthreads, fmap_bspline, name='fmap_wf'):
 
     else:
         torads = pe.Node(niu.Function(output_names=['out_file', 'cutoff_hz'],
-                                      function=_torads), name='torads')
+                                      function=_torads), name='torads',
+                         run_without_submitting=True)
```
effigies: We're not using `run_without_submitting` on `Function` nodes at present. See #554.

fmriprep/workflows/fieldmap/fmap.py
```diff
@@ -127,12 +124,14 @@ def init_fmap_wf(reportlets_dir, omp_nthreads, fmap_bspline, name='fmap_wf'):
...
+    import os
     from math import pi
     import nibabel as nb
-    from fmriprep.utils.misc import genfname
+    from niworkflows.nipype.utils.filemanip import fname_presuffix
 
     if out_file is None:
-        out_file = genfname(in_file, suffix='rad')
+        out_file = fname_presuffix(
+            in_file, suffix='rad', newpath=os.getcwd())
```
effigies: Marking this one: `suffix='_rad'`, but `os.getcwd()` is necessary here.

fmriprep/workflows/fieldmap/phdiff.py
```diff
@@ -25,11 +25,10 @@
     cleanup_edge_pipeline
 from niworkflows.interfaces.masks import BETRPT
 
-from fmriprep.interfaces import ReadSidecarJSON, IntraModalMerge
-from fmriprep.interfaces.bids import DerivativesDataSink
+from ...interfaces import ReadSidecarJSON, IntraModalMerge, DerivativesDataSink
 
 
-def init_phdiff_wf(reportlets_dir, name='phdiff_wf'):
+def init_phdiff_wf(reportlets_dir, omp_nthreads=1, name='phdiff_wf'):
```
effigies: Why a default value for `omp_nthreads`?

fmriprep/workflows/fieldmap/fmap.py
oesteban: What do you mean by "`os.getcwd()` is necessary here"?

effigies: That it can't be replaced with runtime.cwd. (Obviously, but just a flag in case you were going through with a search and replace.)

oesteban: oh, so you meant to replace `os.getcwd()` by `runtime.cwd` when it is possible? I missed that. Please confirm, and I'll update accordingly.

effigies: Yes, that's what I was suggesting. As I [said](https://github.com/poldracklab/fmriprep/pull/640#discussion_r131379154), it's no big deal. I'm happy to merge without that change.

oesteban: Sorry again, I missed that out. Updating it now...

fmriprep/workflows/fieldmap/phdiff.py
```diff
@@ -25,11 +25,10 @@
     cleanup_edge_pipeline
 from niworkflows.interfaces.masks import BETRPT
 
-from fmriprep.interfaces import ReadSidecarJSON, IntraModalMerge
-from fmriprep.interfaces.bids import DerivativesDataSink
+from ...interfaces import ReadSidecarJSON, IntraModalMerge, DerivativesDataSink
 
 
-def init_phdiff_wf(reportlets_dir, name='phdiff_wf'):
+def init_phdiff_wf(reportlets_dir, omp_nthreads, name='phdiff_wf'):
```
effigies: The `omp_nthreads` also needs to be added in the docstring.

PR #647, opened 2017-08-05
[RTM] ENH: Subject-level summary report

### LINKED ISSUE #639: Show subject ID and some other details in report summary
Report the subject ID and number of tasks, and runs per task on the summary.

effigies:
This one's taking a bit more work than I expected. The reportlet is in place, but I'm not sure what the best way to integrate it into reports would be, since it will end up at the same level as "Anatomical" and "Functional", when all of the other reportlets are going under subheadings.

I've started fiddling with the template, but figured I'd put up what I've got at the moment.

Closes #639.

### DIFF
--- a/fmriprep/interfaces/reports.py
+++ b/fmriprep/interfaces/reports.py
@@ -10,6 +10,7 @@ Interfaces to generate reportlets
 """
 
 import os
+from collections import Counter
 from niworkflows.nipype.interfaces.base import (
     traits, TraitedSpec, BaseInterfaceInputSpec,
     File, Directory, InputMultiPath, isdefined)
@@ -17,14 +18,17 @@ from niworkflows.interfaces.base import SimpleInterface
 
 from niworkflows.nipype.interfaces import freesurfer as fs
 
-ANATOMICAL_TEMPLATE = """\t\t<h3 class="elem-title">Summary</h3>
-\t\t<ul class="elem-desc">
-\t\t\t<li>Structural images: {n_t1s:d}</li>
-\t\t\t<li>FreeSurfer reconstruction: {freesurfer_status}</li>
-\t\t\t<li>Output spaces: {output_spaces}</li>
-\t\t</ul>
-"""
+from .bids import BIDS_NAME
 
+SUBJECT_TEMPLATE = """\t<ul class="elem-desc">
+\t\t<li>Subject ID: {subject_id}</li>
+\t\t<li>Structural images: {n_t1s:d} T1-weighted {t2w}</li>
+\t\t<li>Functional series: {n_bold:d}</li>
+{tasks}
+\t\t<li>Resampling targets: {output_spaces}
+\t\t<li>FreeSurfer reconstruction: {freesurfer_status}</li>
+\t</ul>
+"""
 
 FUNCTIONAL_TEMPLATE = """\t\t<h3 class="elem-title">Summary</h3>
 \t\t<ul class="elem-desc">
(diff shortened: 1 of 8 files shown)


effigies:
(Also, wanted to push it before rebasing/merging master, after Oscar's recent PRs.)

fmriprep/viz/config.json
```diff
@@ -2,14 +2,20 @@
 "sub_reports":
 [
     {
-        "name": "Anatomical",
+        "name": "Summary",
         "elements":
         [
             {
                 "name": "anat/summary",
                 "file_pattern": "anat/.*_summary",
```
effigies: `DerivativesDataSink` makes some naming assumptions that don't make it automatic to add a new top-level namespace. I'm not sure that it's worth it, for a single-element entry.

fmriprep/workflows/base.py
```diff
@@ -133,8 +146,16 @@ def init_single_subject_wf(subject_id, task_id, name,
 
     workflow.connect([
         (inputnode, anat_preproc_wf, [('subjects_dir', 'inputnode.subjects_dir')]),
+        (bidssrc, bids_info, [(('t1w', fix_multi_T1w_source_name), 'in_file')]),
+        (bidssrc, summary, [('t1w', 't1w'),
+                            ('t2w', 't2w'),
+                            ('bold', 'bold')]),
```
effigies: This currently provides the name of all BOLD series, rather than just the ones in the current run.

This makes this node more stable when running different task IDs in different processes, but it does mean that if someone doesn't choose to run FMRIPREP on all of their data, then the summary may not match the rest of the outputs.

To make it show just the images in the current execution, we'd just set `summary.inputs.bold = subject_data['bold']`

effigies:
Test reports:

* https://691-78558509-gh.circle-artifacts.com/0/home/ubuntu/ds054/out/fmriprep/sub-100185.html
* https://691-78558509-gh.circle-artifacts.com/1/home/ubuntu/ds005/out/fmriprep/sub-01.html

fmriprep/interfaces/reports.py
```diff
@@ -125,3 +161,18 @@ def _generate_segment(self):
...
                                           confounds=', '.join(self.inputs.confounds))
+
+
+class AboutSummaryInputSpec(BaseInterfaceInputSpec):
+    version = Str(desc='FMRIPREP version')
+    command = Str(desc='FMRIPREP command')
+    # Date not included - update timestamp only if version or command changes
+
+
+class AboutSummary(SummaryInterface):
```
effigies: This interface captures the most recent version/command line used. The timestamp is only updated if either changes. Seem reasonable @chrisfilo?

chrisfilo: yup

fmriprep/viz/config.json
oesteban: I think we can live with this :).

Additionally, we want #639 to be closed as soon as possible, and modifying `DerivativesDataSink` (which is currently working) may break it.

effigies: Yeah. This comment turned into #649. No plans to change it in this PR.

effigies:
Updated reports:

* https://696-78558509-gh.circle-artifacts.com/0/home/ubuntu/ds054/out/fmriprep/sub-100185.html
* https://696-78558509-gh.circle-artifacts.com/1/home/ubuntu/ds005/out/fmriprep/sub-01.html

effigies:
@oesteban @chrisfilo Any final comments?

fmriprep/interfaces/reports.py
```diff
@@ -55,26 +68,30 @@ def _run_interface(self, runtime):
         return runtime
 
 
-class AnatomicalSummaryInputSpec(BaseInterfaceInputSpec):
+class SubjectSummaryInputSpec(BaseInterfaceInputSpec):
     t1w = InputMultiPath(File(exists=True), desc='T1w structural images')
+    t2w = InputMultiPath(File(exists=True), desc='T2w structural images')
     subjects_dir = Directory(desc='FreeSurfer subjects directory')
-    subject_id = traits.Str(desc='FreeSurfer subject ID')
+    subject_id = traits.Str(desc='Subject ID')
```
oesteban: Use `Str` from `nipype.interfaces.base` better than `traits.Str`? I think we try to shadow `traits.Str` but it is safer to use the earlier (and it is already imported in this file...)

effigies:
Fixed. I'll merge on passing tests.

PR #648, opened 2017-08-06
[RTM] Save non-aggressively denoised outputs for convenience.

chrisfilo:
The assumption is that most people that turn on the AROMA flag will want to do non-aggressive denoising - this PR makes it easier for them.

### DIFF
--- a/docs/workflows.rst
+++ b/docs/workflows.rst
@@ -357,11 +357,15 @@ tCompCor, aCompCor, Framewise Displacement, 6 motion parameters, DVARS, and, if
 the ``--use-aroma`` flag is enabled, the noise components identified by ICA-AROMA
 (those to be removed by the "aggressive" denoising strategy).
 
-*Note*: Confounds for performing *non*-aggressive denoising cannot be generated in FMRIPREP.
-If the ``--use-aroma`` flag is passed to FMRIPREP, the MELODIC mix and noise component indices will
-be generated, and non-aggressive denoising may be performed with ``fsl_regfilt``, *e.g.*::
+*Note*: Confounds for performing *non*-aggressive AROMA denoising cannot be generated in FMRIPREP.
+However, for conveninece FMRIPREP produces *non*-aggressive denoised 4D NIFTI files in the MNI space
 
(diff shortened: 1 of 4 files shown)


docs/workflows.rst
```diff
@@ -357,14 +357,18 @@ tCompCor, aCompCor, Framewise Displacement, 6 motion parameters, DVARS, and, if
 the ``--use-aroma`` flag is enabled, the noise components identified by ICA-AROMA
 (those to be removed by the "aggressive" denoising strategy).
 
-*Note*: Confounds for performing *non*-aggressive denoising cannot be generated in FMRIPREP.
-If the ``--use-aroma`` flag is passed to FMRIPREP, the MELODIC mix and noise component indices will
-be generated, and non-aggressive denoising may be performed with ``fsl_regfilt``, *e.g.*::
+*Note*: Confounds for performing *non*-aggressive AROMA denoising cannot be generated in FMRIPREP.
+However, for conveninece FMRIPREP produces *non*-aggressive denoised 4D NIFTI files in the MNI space
```
effigies: I think maybe we should revisit the first line here; since confounds for performing non-aggressive denoising are the same as aggressive (it's the denoising process that's at issue), we don't need to say that FMRIPREP is unable to generate them. We could explain that non-aggressive denoising is its own process, not something that can be done by including nuisance regressors in a GLM. Or just leave it out, e.g.:

```RST
For convenience, FMRIPREP produces *non*-aggressively denoised BOLD series in the MNI space
```

In any event, "convenience" is misspelled and [NIfTI](https://nifti.nimh.nih.gov/) should have a lowercase f.

docs/workflows.rst
```diff
@@ -357,14 +357,18 @@ tCompCor, aCompCor, Framewise Displacement, 6 motion parameters, DVARS, and, if
...
 
-    ``fsl_regfilt -i sub-<subject_label>_task-<task_id>_bold_space-<space>_preproc.nii.gz \
+``*bold_space-MNI152NLin2009cAsym_variant-smoothAROMAnonaggr_brainmask.nii.gz``
+
+Additionally the MELODIC mix and noise component indices will
```
effigies: Comma after "Additionally".

PR #651, opened 2017-08-08
[RTM] singularity related workaround

chrisfilo:
See https://github.com/singularityware/singularity/issues/571#issuecomment-320826366

PR #654, opened 2017-08-09
[FIX] Docker&conda interaction breaks singularity

oesteban:
Closes #651. Reference https://github.com/singularityware/singularity/issues/571

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -118,7 +118,7 @@ ENV PATH=/usr/local/miniconda/bin:$PATH \
     LC_ALL=C.UTF-8
 
 # Installing precomputed python packages
-RUN conda install -y mkl=2017.0.1 mkl-service &&  \
+RUN conda install -y mkl=2017.0.1 mkl-service;  sync &&\
     conda install -y numpy=1.12.0 \
                      scipy=0.18.1 \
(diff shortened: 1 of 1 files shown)


effigies:
Out of curiosity, why is a `sync` at the end of the command insufficient?

oesteban:
Because the lack of synchronization with the filesystem may happen at any point between conda commands. Actually, that `chmod +x /usr/local/miniconda/bin/*` was introduced for this very reason: we used to have a first conda install command to update the python version and a second call installing new packages. Without the chmod or the sync, this would fail with some frequency.

effigies:
Haha, computers are awful.

I really thought the VFS was supposed to keep things consistent, even if the storage medium isn't synced. Is this a Docker failure to implement a VFS layer correctly?

Anyway, full-speed ahead, if this resolves the issues you're having. (I can't verify because I haven't experienced the problem.)

oesteban:
I've seen these problems on my desktop using AUFS and on Circle (using BTRFS), so it must be some kind of "optimization" on conda that docker build does not like.

effigies:
@oesteban Is this okay to merge?

oesteban:
Yup

Issue #253, opened 2016-12-08
Consider "EPImean": using the mean across all tasks/runs to do EPI -> T1 coregistrations


chrisfilo:
This would help with cases such as rewardBeast sub-317 where half of the runs have bad distortion and signal dropout, but the other half are fine. Such cases are rare though.

effigies:
Does this still seem like a good idea, and, just to be clear, the mechanics of this would be to move to an (EPI -> mean EPI) FLIRT alignment + (mean EPI -> T1w)?

chrisfilo:
Correct.

effigies:
Thinking about what's involved in this:

- Single workflow that takes all BOLD series
- Each series needs to be SDC unwarped
- Resulting reference files then merged with `StructuralReference`, and (EPI -> mean EPI) transform saved
- Mean EPI aligned to T1w image

This will involve a fairly major refactor from the current, isolated processing of each BOLD series. One problem with this approach is that you can get different results based on whether your analyze all tasks or just a subset in the same process.

However, if we took an EPInorm approach, then there's a common space to which all EPIs are registered that doesn't depend on which subset of tasks have been specified.

WDYT?

oesteban:
+1 to EPInorm approach. And we have the paper to support the decision.

I would leave that for the future though, beyond 1.0.0.

effigies:
Opened an issue for EPInorm, if that's the route we decide to go.

chrisfilo:
@oscar - from what I know EPInorm has not been compared with syn based SDC. It actually might not provide any improvement.

@effigies indeed this would be the case, but I don't think it's a big issue

oesteban:
Duplicate of #450 - I'll copy one of @effigies' comments there because it is relevant.

PR #655, opened 2017-08-09
[RTM] RF: Separate BOLD reference generation from head-motion correction

effigies:
As discussed in #253, constructing a mean BOLD image would require a reference from each run to be created and merged. This PR takes a first step toward that end by separating reference generation from HMC.

It is set up so that a list of workflows could be easily generated from a list of BOLD images, but at the moment, it's simply moved to the start of `init_func_preproc_wf`.

### DIFF
--- a/fmriprep/workflows/bold.py
+++ b/fmriprep/workflows/bold.py
@@ -134,8 +134,7 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
         ]),
     ])
 
-    validate = pe.Node(ValidateImage(), name='validate', mem_gb=DEFAULT_MEMORY_MIN_GB,
-                       run_without_submitting=True)
+    bold_reference_wf = init_bold_reference_wf(omp_nthreads=omp_nthreads)
 
     # HMC on the BOLD
     bold_hmc_wf = init_bold_hmc_wf(name='bold_hmc_wf',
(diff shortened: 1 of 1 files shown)


fmriprep/workflows/bold.py
```diff
@@ -381,6 +384,49 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
     return workflow
 
 
+def init_bold_reference_wf(omp_nthreads, bold_file=None, name='bold_reference_wf'):
+    """Generate a reference BOLD image for a series
+
+    This image is the target of head-motion-correction, and the subject of distortion
+    correction and registration to T1w and template spaces.
```
oesteban: Are you planning on adding the workflow generation code within #638 or should we already require ourselves to write the graph whenever we add a new workflow? (even for small ones like this).

fmriprep/workflows/bold.py
```diff
@@ -381,6 +384,49 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
...
+
+    inputnode = pe.Node(niu.IdentityInterface(fields=['bold_file']), name='inputnode')
+    outputnode = pe.Node(
+        niu.IdentityInterface(fields=['bold_file', 'raw_ref_image', 'skip_vols', 'ref_image',
+                                      'ref_image_brain', 'bold_mask', 'bold_mask_report',
+                                      'validation_report']),
+        name='outputnode')
+
+    # Simplify manually setting input image
+    if bold_file is not None:
```
oesteban: :+1:

fmriprep/workflows/bold.py
```diff
@@ -389,12 +435,11 @@ def init_bold_hmc_wf(metadata, bold_file_size_gb, ignore,
...
     workflow = pe.Workflow(name=name)
-    inputnode = pe.Node(niu.IdentityInterface(fields=['bold_file']),
+    inputnode = pe.Node(niu.IdentityInterface(fields=['bold_file', 'raw_ref_image', 'skip_vols']),
                         name='inputnode')
-    outputnode = pe.Node(niu.IdentityInterface(
-        fields=['xforms', 'bold_hmc', 'bold_split', 'bold_mask', 'ref_image',
-                'ref_image_brain', 'movpar_file', 'n_volumes_to_discard',
-                'bold_mask_report']), name='outputnode')
+    outputnode = pe.Node(
+        niu.IdentityInterface(fields=['xforms', 'bold_hmc', 'bold_split', 'movpar_file']),
```
oesteban: This is a lot better now!

fmriprep/workflows/bold.py
effigies: The plan was to get the docstring right in whichever PR comes second, since there are going to be conflicts either way.

PR #662, opened 2017-08-11
[RTM] FIX: Enable inactive slice timing, fieldmap unwarping options

effigies:
Currently, head motion correction is being estimated on slice-timing corrected volumes, but applied to the original, uncorrected volumes.

This PR ensures the inputs to HMC and fslsplit are the same, with slice-timing correction applied to both or neither.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,7 @@
 Next release
 ============
-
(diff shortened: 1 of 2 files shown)


effigies:
Applying the SDC warp in the EPI->MNI transform was disabled. A second commit fixes this.

Issue #613, opened 2017-07-26
Parallelize T1w conformation step


effigies:
In the event we need to resample more than 1 or 2 T1w images (e.g. ds031), `t1_conform` can become a bottleneck, only utilizing one core as it runs `nilearn.image.resample_img` in series.

This process can be broken down into two steps:

1) Read all images, decide on target zooms/dimensions (requires full list)
2) Resample images (only one image, and target zooms/dimensions required)

Issue #562, opened 2017-06-16
ERROR: talairach_afd: Talairach Transform: transforms/talairach.xfm ***FAILED*** (p=0.0000, pval=0.0000 < threshold=0.0050


chrisfilo:
```
#--------------------------------------------
#@# Talairach Failure Detection Fri Jun 16 02:55:05 UTC 2017
/output/data/freesurfer/sub-09/mri

 talairach_afd -T 0.005 -xfm transforms/talairach.xfm 

ERROR: talairach_afd: Talairach Transform: transforms/talairach.xfm ***FAILED*** (p=0.0000, pval=0.0000 < threshold=0.0050)

Manual Talairach alignment may be necessary, or
include the -notal-check flag to skip this test,
making sure the -notal-check flag follows -all
or -autorecon1 in the command string.
See:

http://surfer.nmr.mgh.harvard.edu/fswiki/FsTutorial/Talairach


ERROR: Talairach failed!

Linux 23248aa48e5f 4.4.51-40.58.amzn1.x86_64 #1 SMP Tue Feb 28 21:57:17 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux

recon-all -s sub-09 exited with ERRORS at Fri Jun 16 02:55:05 UTC 2017

For more details, see the log file /output/data/freesurfer/sub-09/scripts/recon-all.log
To report a problem, see http://surfer.nmr.mgh.harvard.edu/fswiki/BugReporting


Standard error:

Return code: 1
Interface ReconAll failed to run. 
INFO: transform src into the like-volume: orig.mgz
170616-02:55:05,846 workflow INFO:
	 ***********************************
170616-02:55:05,846 workflow ERROR:
	 could not run node: fmriprep_wf.single_subject_09_wf.anat_preproc_wf.surface_recon_wf.autorecon1
170616-02:55:05,846 workflow INFO:
	 crashfile: /output/data/fmriprep/sub-09/log/20170615-235823_9a07d737-404e-4210-9dd0-68c2f053ebb7/crash-20170616-025505-root-autorecon1-8022ed41-454e-4a1e-862d-faefba145cd2.pklz
170616-02:55:05,846 workflow INFO:
	 ***********************************
```

ds114 subjects 07 and 09 only. This did not happen with 0.4.5. More info: http://openneuro.dev.sqm.io/datasets/ds000118/versions/00001?app=fmriprep&version=14&job=050a6e62-7f12-48cb-bf8d-8e3e76f739c4

[...]

Issue #538, opened 2017-05-27
ERROR: MultiRegistration::loadMovables: images have different voxel sizes


chrisfilo:
ds000114 sub-02
```
  	 Saving crash info to /output/fmriprep/sub-02/log/20170527-000852_28440498-3e03-4a91-80b9-cf0b460ff9a7/crash-20170527-001301-root-t1_merge-5f63d74f-82be-4e65-ac5b-45abc515530e.pklz
  Interface StructuralReference failed to run. 
  	 Traceback (most recent call last):
    File "/usr/local/miniconda/lib/python3.6/site-packages/nipype/pipeline/plugins/multiproc.py", line 52, in run_node
      result['result'] = node.run(updatehash=updatehash)
    File "/usr/local/miniconda/lib/python3.6/site-packages/nipype/pipeline/engine/nodes.py", line 366, in run
      self._run_interface()
  170527-00:13:01,925 workflow INFO:
      self._result = self._run_command(execute)
    File "/usr/local/miniconda/lib/python3.6/site-packages/nipype/pipeline/engine/nodes.py", line 607, in _run_command
      result = self._interface.run()
    File "/usr/local/miniconda/lib/python3.6/site-packages/nipype/interfaces/freesurfer/base.py", line 162, in run
      return super(FSCommand, self).run(**inputs)
    File "/usr/local/miniconda/lib/python3.6/site-packages/nipype/interfaces/base.py", line 1085, in run
      runtime = self._run_wrapper(runtime)
    File "/usr/local/miniconda/lib/python3.6/site-packages/nipype/interfaces/base.py", line 1728, in _run_wrapper
      runtime = self._run_interface(runtime)
    File "/usr/local/miniconda/lib/python3.6/site-packages/nipype/interfaces/base.py", line 1762, in _run_interface
      self.raise_exception(runtime)
    File "/usr/local/miniconda/lib/python3.6/site-packages/nipype/interfaces/base.py", line 1686, in raise_exception
      **runtime.dictcopy()))
  RuntimeError: Command:
  mri_robust_template --satit --fixtp --mov /snapshot/sub-02/ses-retest/anat/sub-02_ses-retest_T1w.nii.gz /snapshot/sub-02/ses-test/anat/sub-02_ses-test_T1w.nii.gz --inittp 1 --iscale --noit --template sub-02_ses-retest_T1w_template.nii.gz --subsample 200
  Standard output:
[...]

PR #666, opened 2017-08-14
[ENH] Parallelize conformation

anibalsolon:
Step `t1_conform` can become a bottleneck since image resampling is executed in series. However, it is possible to run it in parallel, for multiple T1w images.

This PR intends to parallelize this process by breaking it into two steps:
1) Decide the zoom and scale that images should take, based on all images
2) For each image, resample it to fit the new requirements

This process is executed before and after `t1_merge`, so a sub-workflow was created to wrap both steps.

This PR solves issue #613 .


effigies:
Hi, thanks for this. I think if we're going to do this at this point, we should probably simplify the `t1_reorient` step, rather than build a workflow. Most of the machinery is entirely skipped, because `t1_reorient` only ever gets one image as input. I'll have a look through this in a bit; I think we can manage this with pretty minimal changes to what you've added.

anibalsolon:
Hi @effigies ! Thank you for the feedback.

I believe it is possible to abstract the nb.as_closest_canonical as a Step, and put it before PruneExcessiveZoom, so the wf become:
- MapNode "ClosestCanonical" (maybe better naming)
- Node PruneExcessiveZoom
- MapNode ConformSeries

and the `t1_reorient` do not use the whole wf, but just Node ClosestCanonical.

effigies:
I think the `ClosestCanonical` node would be perfectly reasonable, but there's no need to use it before and after the template creation; we can leave `nb.as_closest_canonical` in the `PruneExcessiveZoom` and `ConformSeries` interfaces; it's a basically free operation in memory, so there's no need to save the result. We can just run it after `t1_template`.

As to the name, maybe "Reorient", "OrientRAS"? I'm not too picky here.

Also, WRT "PruneExcessiveZoom", that's only one part of what that step does (and actually the rarest, in my experience). The main purpose it serves is to calculate the template dimensions. "TemplateDimensions", "CalcTemplateSize", "TemplateParameters", or something along those lines would be better.

What do you think?

fmriprep/interfaces/images.py
```diff
@@ -205,71 +192,33 @@ def _run_interface(self, runtime):
...
-            if img is not orig:
-                img.to_filename(out_name)
-            else:
-                copyfile(fname, out_name, copy=True, use_hardlink=True)
-
-            out_names.append(out_name)
-
-        self._results['t1w_list'] = out_names
+        self._results['target_zooms'] = target_zooms.tolist()
+        self._results['target_shape'] = target_shape.astype(np.int64).tolist()
```
effigies: If we need 64 bits to describe the shape of an array, we're in trouble. This should already be an int, so I think you can skip the `astype`.

fmriprep/interfaces/images.py
```diff
@@ -279,6 +228,93 @@ def _run_interface(self, runtime):
...
+                               desc='Target shape information')
+    target_span = traits.List(traits.Float,
+                              desc='Target span information')
+
+
+class ConformSeriesOutputSpec(TraitedSpec):
+    t1w = File(exists=True, desc='conformed T1w image')
+
+
+class ConformSeries(SimpleInterface):
```
effigies: Since the series is being mapped over, perhaps "ConformImage" or possibly "Conform"?

fmriprep/interfaces/images.py
```diff
@@ -279,6 +228,93 @@ def _run_interface(self, runtime):
         return runtime
 
 
+class ConformSeriesInputSpec(BaseInterfaceInputSpec):
+    t1w = File(exists=True, mandatory=True,
```
effigies: The reasons for using `t1w_list` are now obviated, so I'd switch to `in_file`, to make this more generic.

fmriprep/interfaces/images.py
```diff
@@ -279,6 +228,93 @@ def _run_interface(self, runtime):
...
+               desc='input T1w image')
+
+    target_zooms = traits.List(traits.Float,
+                               desc='Target zoom information')
+    target_shape = traits.List(traits.Int,
+                               desc='Target shape information')
+    target_span = traits.List(traits.Float,
+                              desc='Target span information')
```
effigies: Again, I'd drop `target_span`.

fmriprep/interfaces/images.py
```diff
@@ -279,6 +228,93 @@ def _run_interface(self, runtime):
...
+
+
+class ConformSeriesOutputSpec(TraitedSpec):
+    t1w = File(exists=True, desc='conformed T1w image')
```
effigies: `out_file = File(...)`

fmriprep/interfaces/images.py
```diff
@@ -124,6 +124,27 @@ def _run_interface(self, runtime):
...
+class PruneExcessiveZoomInputSpec(BaseInterfaceInputSpec):
+    t1w_list = InputMultiPath(File(exists=True), mandatory=True,
+                              desc='input T1w images')
+    max_scale = traits.Float(3.0, usedefault=True,
+                             desc='Maximum scaling factor in images to accept')
+
+
+class PruneExcessiveZoomOutputSpec(TraitedSpec):
+    t1w_dropped_list = OutputMultiPath(exists=True, desc='invalid T1w images')
+    t1w_valid_list = OutputMultiPath(exists=True, desc='valid T1w images')
```
anibalsolon: Yes! Initially, it was to be used for a posterior reporting, but I ended adding reporting in this same interface. I'll remove it for now.

fmriprep/interfaces/images.py
```diff
@@ -124,6 +124,27 @@ def _run_interface(self, runtime):
...
+
+    target_zooms = traits.List(traits.Float,
+                               desc='Target zoom information')
```
anibalsolon: Perfect! I didn't know this format. Will use it now :)

fmriprep/interfaces/images.py
```diff
@@ -124,6 +124,27 @@ def _run_interface(self, runtime):
...
+    target_shape = traits.List(traits.Int,
+                               desc='Target shape information')
+    target_span = traits.List(traits.Float,
+                              desc='Target span information')
```
anibalsolon: I was in doubt to parametrize it or compute the dot product on the next step, so I will now compute it. :ok_hand:

fmriprep/interfaces/images.py
anibalsolon: Hehe indeed! I cast it because it was using floats, but I will double check it, just in case. If it is not possible to remove this casting, I will switch `np.int64` to `int`.

fmriprep/interfaces/images.py
anibalsolon: Will do

fmriprep/interfaces/images.py
anibalsolon: `Conform` sounds good!

effigies:
@anibalsolon No rush, but let us know if there's anything you need from us.

anibalsolon:
Hi @effigies, sorry for the delay!

Thank you for the detailed review. If there is something that I can improve, please let me know.

fmriprep/interfaces/images.py
```diff
@@ -134,28 +154,12 @@ def _run_interface(self, runtime):
...
-
-class ConformSeries(SimpleInterface):
-    """Conform a series of T1w images to enable merging.
-
-    Performs two basic functions:
 
-    #. Orient to RAS (left-right, posterior-anterior, inferior-superior)
-    #. Along each dimension, resample to minimum voxel size, maximum number of voxels
+class TemplateDimensions(SimpleInterface):
+    """Filter a series of T1w images based on the requirements for up-sampling.
```
effigies: Proposed replacement for this first line, focusing on the main goal of this interface:

```
Finds template target dimensions for a series of T1w images, filtering low-resolution images, if necessary.

Along each axis, the minimum voxel size (zoom) and the maximum number of voxels (shape) are found
across images.
```

fmriprep/interfaces/images.py
```diff
@@ -205,76 +191,154 @@ def _run_interface(self, runtime):
...
+
+        out_name = fname_presuffix(fname, suffix='_ras', newpath=runtime.cwd)
+
+        # Image may be reoriented
+        if reoriented is not orig_img:
+            reoriented.to_filename(out_name)
+        else:
+            copyfile(fname, out_name, copy=True, use_hardlink=True)
+
+        self._results['out_file'] = out_name
```
effigies: Let's not bother copying, just point to the original file:

```Python
    # Load image, orient as RAS
    fname = self.inputs.in_file
    orig_img = nb.load(fname)
    reoriented = nb.as_closest_canonical(orig_img)

    if reoriented is not orig_img:
        out_name = fname_presuffix(fname, suffix='_ras', newpath=runtime.cwd)
        reoriented.to_filename(out_name)
    else:
        out_name = fname

    self._results['out_file'] = out_name
```

fmriprep/interfaces/images.py
```diff
@@ -205,76 +191,154 @@ def _run_interface(self, runtime):
...
-        self._results['t1w_list'] = out_names
+        out_name = fname_presuffix(fname, suffix='_ras', newpath=runtime.cwd)
 
-        # Create report
-        segment = self._generate_segment(dropped_images, target_shape, target_zooms)
+        # Image may be reoriented, rescaled, and/or resized
+        if reoriented is not orig_img:
+            reoriented.to_filename(out_name)
+        else:
+            copyfile(fname, out_name, copy=True, use_hardlink=True)
```
effigies: Same thing here, wrt copying.

fmriprep/interfaces/images.py
```diff
@@ -124,7 +124,27 @@ def _run_interface(self, runtime):
         return runtime
 
 
-CONFORMSERIES_TEMPLATE = """\t\t<h3 class="elem-title">Anatomical Conformation</h3>
+class TemplateDimensionsInputSpec(BaseInterfaceInputSpec):
+    t1w_list = InputMultiPath(File(exists=True), mandatory=True,
+                              desc='input T1w images')
+
+    max_scale = traits.Float(3.0, usedefault=True,
+                             desc='Maximum scaling factor in images to accept')
```
anibalsolon: Nice! Do you know if it is possible to enforce this pattern of attr spacing via flake?

fmriprep/interfaces/images.py
anibalsolon: Great :ok_hand:

anibalsolon:
I tested on Poldrack's MyConnectome dataset, as suggested in issue #613 
http://myconnectome.org/wp/data-sharing/ **Sessions 13-24**

fmriprep/interfaces/images.py
effigies: Not that I know of.

fmriprep/interfaces/images.py
```diff
@@ -205,76 +191,148 @@ def _run_interface(self, runtime):
...
+    out_file = File(exists=True, desc='Conformed T1w image')
+
+
+class Conform(SimpleInterface):
+    """Conform a series of T1w images to enable merging.
+
+    Performs two basic functions:
+
+    #. Orient to RAS (left-right, posterior-anterior, inferior-superior)
+    #. Along each dimension, resample to minimum voxel size, maximum number of voxels
```
effigies: Since this is no longer calculating the zooms, shape, we can just say:

"resample to target zooms (voxel sizes) and shape (number of voxels)"

anibalsolon:
Awesome!

I cannot merge because I don't have write access.

If there are some datasets that I can help testing, just let me know.

chrisfilo:
I think @effigies meant merging "master" branch from the upstream (poldracklab/fmriprep) into your branch.

anibalsolon:
Ok! Merged with the upstream. Thanks @chrisfilo

effigies:
Blocking merge on following tests (will check as I finish):

* [ ] Test on ds000114 sub-02 (See #538/#545)
* [ ] Test on ds000114 sub-07/09 (See #562/#564)
* [ ] Test on ds000031

effigies:
Good to go. Thanks a lot!

PR #670, opened 2017-08-16
[RTM] FIX: Handle missing functional data in SubjectSummary node

### LINKED ISSUE #669: Handle missing functional data in SubjectSummary node
See discussion in #656.

effigies:
The issue is that `OutputMultiPath` considers an empty list `Undefined`, so we have to handle that case specifically.

Closes #669.

### DIFF
--- a/fmriprep/interfaces/bids.py
+++ b/fmriprep/interfaces/bids.py
@@ -99,19 +99,17 @@ class BIDSDataGrabber(SimpleInterface):
         bids_dict = self.inputs.subject_data
 
         self._results['out_dict'] = bids_dict
+        self._results.update(bids_dict)
 
-        self._results['t1w'] = bids_dict['t1w']
         if not bids_dict['t1w']:
             raise FileNotFoundError('No T1w images found for subject sub-{}'.format(
                 self.inputs.subject_id))
 
-        self._results['bold'] = bids_dict['bold']
         if self._require_funcs and not bids_dict['bold']:
             raise FileNotFoundError('No functional images found for subject sub-{}'.format(
                 self.inputs.subject_id))
 
         for imtype in ['bold', 't2w', 'fmap', 'sbref']:
-            self._results[imtype] = bids_dict[imtype]
             if not bids_dict[imtype]:
                 LOGGER.warn('No \'{}\' images found for sub-{}'.format(
                     imtype, self.inputs.subject_id))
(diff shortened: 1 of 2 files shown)


effigies:
@yidilozdemir Can you test this fix?

To test it, you'll need to build a new Docker image:

```
curl -sSL https://github.com/effigies/fmriprep/archive/fix/nofuncs.tar.gz | tar xz
cd fmriprep-fix-nofuncs
docker build -t fmriprep:test
```

You can then add `-i fmriprep:test` to your `fmriprep-docker` command.

yidilozdemir:
```
lab@lab-ThinkStation-E31:~/Desktop/freesurfer/fmriprep-fix-nofuncs$ docker build -t fmriprep:test
"docker build" requires exactly 1 argument(s).
See 'docker build --help'.

Usage:  docker build [OPTIONS] PATH | URL | -

```

effigies:
Sorry, there's supposed to be a `.` after the `fmriprep:test`. Updated the command above, for clarity.

fmriprep/interfaces/bids.py
```diff
@@ -99,19 +99,17 @@ def _run_interface(self, runtime):
...
                 self.inputs.subject_id))
 
-        self._results['bold'] = bids_dict['bold']
         if self._require_funcs and not bids_dict['bold']:
             raise FileNotFoundError('No functional images found for subject sub-{}'.format(
                 self.inputs.subject_id))
 
         for imtype in ['bold', 't2w', 'fmap', 'sbref']:
-            self._results[imtype] = bids_dict[imtype]
             if not bids_dict[imtype]:
```
oesteban: shouldn't this line be updated to:

```Python
if imtype not in bids_dict:
...
```

fmriprep/interfaces/reports.py
```diff
@@ -114,8 +114,9 @@ def _generate_segment(self):
             t2w_seg = '(+ {:d} T2-weighted)'.format(len(self.inputs.t2w))
 
         # Add list of tasks with number of runs
+        bold_series = self.inputs.bold if isdefined(self.inputs.bold) else []
```
oesteban: Can we set a default value of `[]` in the input `bold`?

fmriprep/interfaces/bids.py
effigies: No, these entries exist, they just might be empty.

fmriprep/interfaces/reports.py
effigies: No, `InputMultiPath` also considers `[]` to be `Undefined`. And even if it didn't, the fact that the `OutputMultiPath` is pushing out an `Undefined` means we can't expect this to be defined.

oesteban:
Ok!

oesteban:
Ok!

effigies:
It would still be good to get an independent test, or else we might crank out another release that causes you issues.

yidilozdemir:
@effigies I ran it and it doesn't give any errors.

PR #672, opened 2017-08-17
RF: Drop custom ApplyMask for fsl.ApplyMask

effigies:
This looks like an unnecessary interface, as [every other instance](https://github.com/poldracklab/fmriprep/search?utf8=%E2%9C%93&q=ApplyMask&type=) uses [fsl.ApplyMask](https://nipype.readthedocs.io/en/latest/interfaces/generated/interfaces.fsl/maths.html#applymask). Rather than add documentation, we can just drop it.

If there's any reason not to do this, feel free to close without discussion.


oesteban:
I can't recall why this interface was included. So I guess we can remove it.

PR #659, opened 2017-08-10
[RTM] ENH: Erode TPM ROIs in T1 space, resample to BOLD

effigies:
This PR significantly modifies the creation of CSF/WM masks and eroded BOLD masks to accommodate partial field-of-view (FoV) BOLD series. It does so by performing erosions in the higher-resolution T1 space, resampling to BOLD, and applying the BOLD mask to limit results to the BOLD FoV.

These last two steps are encapsulated in the new `_resample_and_mask` function, while the original `_tpm2roi` works in the T1 space.

Eroded BOLD
------------------
1) T1 skull-stripping mask (`t1_mask`) is eroded
2) `_resample_and_mask` ensures that the result is clipped by the BOLD mask, in case of partial FoV

fov_tcompcor_mask

Blue: T1 mask
Red: BOLD mask
White: Eroded mask

CSF/WM ROIs
--------------------
1) If T1 mask is eroded, apply to thresholded tissue probability map (TTPM)
2) Erode TTPM in T1 space
3) `_resample_and_mask` ensures that the result is clipped by the BOLD mask, in case of partial FoV

fov_csf_mask

Red: Eroded BOLD mask
White: CSF ROI mask

fov_wm_roi

Red: BOLD mask (not eroded)
White: WM ROI mask

Closes #652

PR #674, opened 2017-08-18
ENH: Disable AFNI skullstrip flag

effigies:
AFNI skull stripping does not play nicely with eroding the brainmask in T1 space.

This may be reverted after poldracklab/niworkflows#194 is resolved.

Follow-up to #659.


oesteban:
Because of the holes it leaves in the ventricles?

effigies:
Yes, as well as how deep it penetrates into fissures. The erosion for tCompCor results in empty masks.

oesteban:
The holes are easy. Fissures are hard. Let's go ahead with this temporarily.

Opened #676 as a remainder to revert this PR when we can.

Issue #679, opened 2017-08-21
nan for medial wall


satra:
for the gifti files mapped to fsaverage5, would it be possible to make the values on the medial wall nan ?

chrisfilo:
Theoretically it should be as straighforward as adding a function node
after gifti conversion that would use nibabel to replace zeros with NaNs.
There might be a more reliable way to detect what is the medial wall though.

effigies:
This shouldn't be hard. I'll re-check the FSFAST code to see how they do masking.

effigies:
We're already using the cortical mask so the medial wall is 0. So it would need to be an additional step, equivalent to the following:

```Python
import os.path as op
import numpy as np
import nibabel as nb

cortex = nb.freesurfer.read_label('derivatives/freesurfer/fsaverage5/label/lh.cortex.label')
func = nb.load('derivatives/fmriprep/func/sub-01_task-rest_bold_space-fsaverage5.L.func.gii')

medial = np.delete(np.arange(len(func.darrays[0].data)), cortex)

for darray in func.darrays:
    darray.data[medial] = np.nan

func.to_filename(...)
```

So for us to go ahead and do this, we'd want to be sure that making it `NaN` would play well with FSFAST, FSL and SPM. Do you know how that would go, @satra?

We could also make this an optional preprocessing step, e.g. with a `--medial-surface-nan` (or similar) flag.

satra:
@effigies - i'm happy with the optional flag. FSL definitely wouldn't like it, SPM wouldn't care, and FSFAST, i don't know. also, i would be fine with using lh.cortex as well. so whether you make this an explicit change is totally up to you guys.

effigies:
Okay. If you're fine using your own filter for now, we're trying to focus on bugs rather than adding new features and flags in the run-up to the 1.0 release, so I'm going to mark this as low-priority for now. (@chrisfilo if you think this should be prioritized, let me know.)

That said, outputting masks that make working with our outputs meaningfully easier would be worth the effort of squeezing in before 1.0, IMO. So if, e.g., it would be a useful input to your toolchain to have the cortex labels as GIFTI label files instead of FS label files, we could do something like that.

PR #687, opened 2017-08-31
enh: replace medial wall values with nans

mgxd:
#679

### DIFF
--- a/fmriprep/cli/run.py
+++ b/fmriprep/cli/run.py
@@ -126,9 +126,9 @@ def get_parser():
              'but is not used in normalization.')
     g_conf.add_argument(
         '--medial-surface-nan', required=False, action='store', default=False,
-        help='Replace medial wall values with NaNs on surface gifti files. Only '
+        help='Replace medial wall values with NaNs on functional GIFTI files. Only '
         'performed for gifti files mapped to an fsaverage')
-     
+
     # ICA_AROMA options
     g_aroma = parser.add_argument_group('Specific options for running ICA_AROMA')
     g_aroma.add_argument('--use-aroma', action='store_true', default=False,
--- a/fmriprep/cli/run.py
+++ b/fmriprep/cli/run.py
@@ -119,12 +119,16 @@ def get_parser():
         '--template', required=False, action='store',
         choices=['MNI152NLin2009cAsym'], default='MNI152NLin2009cAsym',
         help='volume template space (default: MNI152NLin2009cAsym)')
-
     g_conf.add_argument(
         '--output-grid-reference', required=False, action='store', default=None,
         help='Grid reference image for resampling BOLD files to volume template space. '
              'It determines the field of view and resolution of the output images, '
              'but is not used in normalization.')
+    g_conf.add_argument(
+        '--medial-surface-nan', required=False, action='store', default=False,
+        help='Replace medial wall values with NaNs on surface gifti files. Only '
+        'performed for gifti files mapped to an fsaverage')
+     
     # ICA_AROMA options
     g_aroma = parser.add_argument_group('Specific options for running ICA_AROMA')
     g_aroma.add_argument('--use-aroma', action='store_true', default=False,
@@ -292,6 +296,7 @@ def create_workflow(opts):
         freesurfer=opts.freesurfer,
         output_spaces=opts.output_space,
         template=opts.template,
+        medial_surface_nan=opts.medial_surface_nan,
         output_grid_ref=opts.output_grid_reference,
         hires=opts.hires,
         bold2t1w_dof=opts.bold2t1w_dof,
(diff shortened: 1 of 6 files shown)


fmriprep/cli/run.py
```diff
@@ -119,12 +119,16 @@ def get_parser():
...
-
     g_conf.add_argument(
         '--output-grid-reference', required=False, action='store', default=None,
         help='Grid reference image for resampling BOLD files to volume template space. '
              'It determines the field of view and resolution of the output images, '
              'but is not used in normalization.')
+    g_conf.add_argument(
+        '--medial-surface-nan', required=False, action='store', default=False,
+        help='Replace medial wall values with NaNs on surface gifti files. Only '
+        'performed for gifti files mapped to an fsaverage')
```
effigies: GIFTI is caps. And these are functional GIFTI files (see section 9.0 of the spec), not surface.

fmriprep/workflows/base.py
```diff
@@ -120,6 +121,8 @@ def init_fmriprep_wf(subject_list, task_id, run_uuid,
              - fsaverage (or other pre-existing FreeSurfer templates)
         template : str
             Name of template targeted by `'template'` output space
+        medial_surface_nan : bool
+            Replace medial wall values with NaNs on surface gifti files
```
effigies: "functional GIFTI"

fmriprep/workflows/bold.py
```diff
@@ -969,6 +973,9 @@ def init_bold_surf_wf(output_spaces, name='bold_surf_wf'):
             If the list contains ``fsnative``, images will be resampled to the
             individual subject's native surface
 
+        medial_surface_nan : bool
```
effigies: Remove empty line above.

fmriprep/workflows/bold.py
```diff
@@ -1017,6 +1024,36 @@ def select_target(subject_id, space):
...
+
+        import nibabel as nb
+        import numpy as np
+        import os
+    
+        fn = os.path.basename(in_file)
+        space = fn.split('.')[1]
+
+        if not space.startswith('fsaverage'):
+            return in_file
```
effigies: Is this simply because `fsnative` isn't a subject, or is it because you don't want to set medial walls to NaN in `fsnative`?

If the former, you can use the output of `select_target` to get the actual subject name.

fmriprep/workflows/bold.py
```diff
@@ -1017,6 +1024,36 @@ def select_target(subject_id, space):
...
+
+        cortex = nb.freesurfer.read_label(os.path.join(
+            subjects_dir, '{}'.format(space), 'label', '{}.cortex.label'.format(fn[:2])))
```
effigies: No need to interpolate `space`.

```Python
os.path.join(subjects_dir, space, 'label', '{}.cortex.label'.format(fn[:2]))
```

fmriprep/workflows/bold.py
mgxd: I actually wasn't sure whether anyone would want that output - but I've added it in, thanks.

fmriprep/workflows/bold.py
```diff
@@ -943,7 +947,7 @@ def init_bold_reg_wf(freesurfer, bold2t1w_dof, bold_file_size_gb,
     return workflow
 
 
-def init_bold_surf_wf(output_spaces, name='bold_surf_wf'):
+def init_bold_surf_wf(output_spaces, medial_surface_nan=False, name='bold_surf_wf'):
```
effigies: Let's not set this by default. And you're going to want to find the instances of `init_fmriprep_wf`, `init_single_subject_wf`, `init_func_preproc_wf` and `init_bold_surf_wf` in `docs/` and add `medial_surface_nan=False` to the invocation. (Those are workflows that are created to add graphs to the docs.)

fmriprep/cli/run.py
```diff
@@ -119,12 +119,16 @@ def get_parser():
...
-
     g_conf.add_argument(
         '--output-grid-reference', required=False, action='store', default=None,
         help='Grid reference image for resampling BOLD files to volume template space. '
              'It determines the field of view and resolution of the output images, '
              'but is not used in normalization.')
+    g_conf.add_argument(
+        '--medial-surface-nan', required=False, action='store', default=False,
+        help='Replace medial wall values with NaNs on functional GIFTI files. Only '
+        'performed for gifti files mapped to an fsaverage')
```
effigies: Second sentence no longer applies.

fmriprep/workflows/bold.py
mgxd: @effigies couldn't find `init_fmriprep_wf` but I think I hit them all?

PR #694, opened 2017-09-07
[REVIEW] Fallback to initial registration, if BBR fails

### LINKED ISSUE #681: Fall-back registration for BBR failures
Original report: https://neurostars.org/t/strange-epi-to-t1w-registration-result/696

Summary: In cases of low gray/white tissue contrast in BOLD series, the default, 9-dof boundary-based registration (BBR) to T1w space can result in noticeable failures. Using the initial transform (vanilla FLIRT for FSL, mri_coreg for bbregister) can reduce the distortion.

cc @chrisfilo @oesteban @jdkent

effigies:
The easiest way to read this PR is to separately review #741, and read https://github.com/effigies/fmriprep/compare/fsfix_common...bbr_fallback.

Closes #681.

### DIFF
--- a/fmriprep/workflows/bold.py
+++ b/fmriprep/workflows/bold.py
@@ -258,11 +258,13 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
                 'aroma_noise_ics', 'melodic_mix', 'nonaggr_denoised_file']),
         name='outputnode')
 
-    summary = pe.Node(FunctionalSummary(output_spaces=output_spaces,
-                                        pe_direction=bold_pe), name='summary',
-                      mem_gb=0.05)
-    summary.inputs.slice_timing = run_stc
-    summary.inputs.registration = 'bbregister' if freesurfer else 'FLIRT'
+    summary = pe.Node(
+        FunctionalSummary(output_spaces=output_spaces,
+                          slice_timing=run_stc,
+                          registration='FreeSurfer' if freesurfer else 'FSL',
+                          registration_dof=bold2t1w_dof,
+                          pe_direction=bold_pe),
+        name='summary', mem_gb=DEFAULT_MEMORY_MIN_GB, run_without_submitting=True)
 
     func_reports_wf = init_func_reports_wf(reportlets_dir=reportlets_dir,
                                            freesurfer=freesurfer,
@@ -348,6 +350,7 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
             ('outputnode.validation_report', 'inputnode.validation_report')]),
         (bold_reg_wf, func_reports_wf, [
             ('outputnode.out_report', 'inputnode.bold_reg_report'),
+            ('outputnode.fallback', 'inputnode.bold_reg_fallback'),
         ]),
         (bold_confounds_wf, outputnode, [
             ('outputnode.confounds_file', 'confounds'),
@@ -362,6 +365,7 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
             ('outputnode.tcompcor_report', 'inputnode.tcompcor_report'),
             ('outputnode.ica_aroma_report', 'inputnode.ica_aroma_report')]),
         (bold_confounds_wf, summary, [('outputnode.confounds_list', 'confounds')]),
+        (bold_reg_wf, summary, [('outputnode.fallback', 'fallback')]),
         (summary, func_reports_wf, [('out_report', 'inputnode.summary_report')]),
     ])
 
@@ -880,6 +884,8 @@ def init_bold_reg_wf(freesurfer, bold2t1w_dof, bold_file_size_gb, omp_nthreads,
             BOLD mask in T1 space
         out_report
             Reportlet visualizing quality of registration
+        fallback
+            Boolean indicating whether BBR was rejected (mri_coreg registration returned)
 
     """
     workflow = pe.Workflow(name=name)
@@ -894,7 +900,7 @@ def init_bold_reg_wf(freesurfer, bold2t1w_dof, bold_file_size_gb, omp_nthreads,
     outputnode = pe.Node(
         niu.IdentityInterface(fields=['itk_bold_to_t1', 'itk_t1_to_bold',
                                       'bold_t1', 'bold_mask_t1',
-                                      'out_report']),
+                                      'out_report', 'fallback']),
         name='outputnode'
     )
 
@@ -913,7 +919,8 @@ def init_bold_reg_wf(freesurfer, bold2t1w_dof, bold_file_size_gb, omp_nthreads,
             ('t1_brain', 'inputnode.t1_brain')]),
         (bbr_wf, outputnode, [('outputnode.itk_bold_to_t1', 'itk_bold_to_t1'),
                               ('outputnode.itk_t1_to_bold', 'itk_t1_to_bold'),
-                              ('outputnode.out_report', 'out_report')]),
+                              ('outputnode.out_report', 'out_report'),
+                              ('outputnode.fallback', 'fallback')]),
     ])
 
     gen_ref = pe.Node(GenerateSamplingReference(), name='gen_ref',
@@ -1495,8 +1502,8 @@ def init_func_reports_wf(reportlets_dir, freesurfer, use_aroma, use_syn, name='f
     inputnode = pe.Node(
         niu.IdentityInterface(
             fields=['source_file', 'summary_report', 'validation_report', 'bold_mask_report',
-                    'bold_reg_report', 'acompcor_report', 'tcompcor_report', 'syn_sdc_report',
-                    'ica_aroma_report']),
+                    'bold_reg_report', 'bold_reg_fallback', 'acompcor_report', 'tcompcor_report',
+                    'syn_sdc_report', 'ica_aroma_report']),
         name='inputnode')
 
     ds_summary_report = pe.Node(
@@ -1523,9 +1530,14 @@ def init_func_reports_wf(reportlets_dir, freesurfer, use_aroma, use_syn, name='f
         name='ds_syn_sdc_report', run_without_submitting=True,
         mem_gb=DEFAULT_MEMORY_MIN_GB)
 
+    def _bold_reg_suffix(fallback, freesurfer):
+        if fallback:
+            return 'coreg' if freesurfer else 'flirt'
+        else:
+            return 'bbr' if freesurfer else 'flt_bbr'
+
     ds_bold_reg_report = pe.Node(
-        DerivativesDataSink(base_directory=reportlets_dir,
-                            suffix='bbr' if freesurfer else 'flt_bbr'),
+        DerivativesDataSink(base_directory=reportlets_dir),
         name='ds_bold_reg_report', run_without_submitting=True,
         mem_gb=DEFAULT_MEMORY_MIN_GB)
 
@@ -1554,8 +1566,10 @@ def init_func_reports_wf(reportlets_dir, freesurfer, use_aroma, use_syn, name='f
                                            ('validation_report', 'in_file')]),
         (inputnode, ds_bold_mask_report, [('source_file', 'source_file'),
                                           ('bold_mask_report', 'in_file')]),
-        (inputnode, ds_bold_reg_report, [('source_file', 'source_file'),
-                                         ('bold_reg_report', 'in_file')]),
+        (inputnode, ds_bold_reg_report, [
+            ('source_file', 'source_file'),
+            ('bold_reg_report', 'in_file'),
+            (('bold_reg_fallback', _bold_reg_suffix, freesurfer), 'suffix')]),
         (inputnode, ds_acompcor_report, [('source_file', 'source_file'),
                                          ('acompcor_report', 'in_file')]),
         (inputnode, ds_tcompcor_report, [('source_file', 'source_file'),
--- a/fmriprep/workflows/bold.py
+++ b/fmriprep/workflows/bold.py
@@ -899,7 +899,7 @@ def init_bold_reg_wf(freesurfer, bold2t1w_dof, bold_file_size_gb, omp_nthreads,
     )
 
     if freesurfer:
-        bbr_wf = init_bbreg_wf(bold2t1w_dof)
+        bbr_wf = init_bbreg_wf(bold2t1w_dof, omp_nthreads=omp_nthreads)
     else:
         bbr_wf = init_fsl_bbr_wf(bold2t1w_dof)
 
--- a/fmriprep/workflows/bold.py
+++ b/fmriprep/workflows/bold.py
@@ -65,7 +65,7 @@ LOGGER = logging.getLogger('workflow')
 
 
 def init_func_preproc_wf(bold_file, ignore, freesurfer,
-                         bold2t1w_dof, reportlets_dir,
+                         use_bbr, bold2t1w_dof, reportlets_dir,
                          output_spaces, template, output_dir, omp_nthreads,
                          fmap_bspline, fmap_demean, use_syn, force_syn,
                          use_aroma, ignore_aroma_err, medial_surface_nan,
@@ -88,6 +88,7 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
                                   output_spaces=['T1w', 'fsnative',
                                                  'template', 'fsaverage5'],
                                   debug=False,
+                                  use_bbr=True,
                                   bold2t1w_dof=9,
                                   fmap_bspline=True,
                                   fmap_demean=True,
@@ -108,6 +109,9 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
         freesurfer : bool
             Enable FreeSurfer functional registration (bbregister) and resampling
             BOLD series to FreeSurfer surface meshes.
+        use_bbr : bool or None
+            Enable/disable boundary-based registration refinement.
+            If ``None``, test BBR result for distortion before accepting.
         bold2t1w_dof : 6, 9 or 12
             Degrees-of-freedom for BOLD-T1w registration
         reportlets_dir : str
@@ -308,6 +312,7 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
     # mean BOLD registration to T1w
     bold_reg_wf = init_bold_reg_wf(name='bold_reg_wf',
                                    freesurfer=freesurfer,
+                                   use_bbr=use_bbr,
                                    bold2t1w_dof=bold2t1w_dof,
                                    bold_file_size_gb=bold_file_size_gb,
                                    omp_nthreads=omp_nthreads,
@@ -800,7 +805,7 @@ def init_bold_hmc_wf(bold_file_size_gb, omp_nthreads, name='bold_hmc_wf'):
     return workflow
 
 
-def init_bold_reg_wf(freesurfer, bold2t1w_dof, bold_file_size_gb, omp_nthreads,
+def init_bold_reg_wf(freesurfer, use_bbr, bold2t1w_dof, bold_file_size_gb, omp_nthreads,
                      name='bold_reg_wf', use_compression=True,
                      use_fieldwarp=False):
     """
@@ -821,12 +826,16 @@ def init_bold_reg_wf(freesurfer, bold2t1w_dof, bold_file_size_gb, omp_nthreads,
         wf = init_bold_reg_wf(freesurfer=True,
                               bold_file_size_gb=3,
                               omp_nthreads=1,
+                              use_bbr=True,
                               bold2t1w_dof=9)
 
     Parameters
 
         freesurfer : bool
             Enable FreeSurfer functional registration (bbregister)
+        use_bbr : bool or None
+            Enable/disable boundary-based registration refinement.
+            If ``None``, test BBR result for distortion before accepting.
         bold2t1w_dof : 6, 9 or 12
             Degrees-of-freedom for BOLD-T1w registration
         bold_file_size_gb : float
@@ -905,9 +914,10 @@ def init_bold_reg_wf(freesurfer, bold2t1w_dof, bold_file_size_gb, omp_nthreads,
     )
 
     if freesurfer:
-        bbr_wf = init_bbreg_wf(bold2t1w_dof, omp_nthreads=omp_nthreads)
+        bbr_wf = init_bbreg_wf(use_bbr=use_bbr, bold2t1w_dof=bold2t1w_dof,
+                               omp_nthreads=omp_nthreads)
     else:
-        bbr_wf = init_fsl_bbr_wf(bold2t1w_dof)
+        bbr_wf = init_fsl_bbr_wf(use_bbr=use_bbr, bold2t1w_dof=bold2t1w_dof)
 
     workflow.connect([
         (inputnode, bbr_wf, [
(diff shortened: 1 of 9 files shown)


effigies:
This is ready for a review. It's working locally on @jdkent's dataset. Both FreeSurfer and FSL are falling back to `mri_coreg`/`FLIRT`.

The only thing left to do is fix cost parsing for non-BBR FLIRT so that we can compare registration performance to determine phase-encoding direction for SyN-SDC.

Depends: poldracklab/niworkflows#195.

fmriprep/workflows/bold.py
```diff
@@ -247,10 +247,12 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
                 'aroma_noise_ics', 'melodic_mix', 'nonaggr_denoised_file']),
         name='outputnode')
 
-    summary = pe.Node(FunctionalSummary(output_spaces=output_spaces), name='summary',
-                      mem_gb=0.05)
+    summary = pe.Node(
+        FunctionalSummary(output_spaces=output_spaces,
+                          registration='FreeSurfer' if freesurfer else 'FSL',
+                          registration_dof=bold2t1w_dof),
+        name='summary', mem_gb=DEFAULT_MEMORY_MIN_GB)
```
oesteban: `run_without_submit=True` perhaps?

fmriprep/workflows/util.py
```diff
@@ -232,15 +253,33 @@ def init_bbreg_wf(bold2t1w_dof, report, reregister=True, name='bbreg_wf'):
...
+
     _BBRegister = BBRegisterRPT if report else fs.BBRegister
     bbregister = pe.Node(
-        _BBRegister(dof=bold2t1w_dof, contrast_type='t2', init='coreg',
-                    registered_file=True, out_fsl_file=True),
+        _BBRegister(dof=bold2t1w_dof, contrast_type='t2', registered_file=True, out_fsl_file=True,
+                    init_cost_file=True),
         name='bbregister')
 
+    transforms = pe.Node(niu.Merge(2), name='transforms')
```
oesteban: All these new nodes should `run_without_submitting=True`?

fmriprep/workflows/util.py
```diff
@@ -17,12 +17,33 @@
...
+    comp = mat2.dot(np.linalg.pinv(mat1))
+    trans, rotation_matrix, scales, shears = decompose44(comp)
+
+    max_trans = np.max(np.abs(trans))
+    rot = mat2axangle(rotation_matrix)[1]
+    max_scale = np.max(np.abs(scales))
+
+    fallback = any((max_trans > 2, rot > np.pi / 36, max_scale > 1.1))
+
+    return 0 if fallback else 1
```
oesteban: why not keeping the boolean type here (and downstream)?

fmriprep/workflows/util.py
```diff
@@ -17,12 +17,33 @@
...
+    mat2 = np.loadtxt(test_mat)
+    # [[R'], [A'], [S'], [1]] = mat2.dot(mat1_inv.dot([[R], [A], [S], [1]]))
+    comp = mat2.dot(np.linalg.pinv(mat1))
+    trans, rotation_matrix, scales, shears = decompose44(comp)
+
+    max_trans = np.max(np.abs(trans))
+    rot = mat2axangle(rotation_matrix)[1]
+    max_scale = np.max(np.abs(scales))
+
+    fallback = any((max_trans > 2, rot > np.pi / 36, max_scale > 1.1))
```
oesteban: 2mm is not too small?

fmriprep/workflows/util.py
effigies: It was intended as an index, but I can reverse these. That probably would be cleaner.

fmriprep/workflows/util.py
effigies: That might be. What seems like a reasonable translation threshold? It seems like it shouldn't be more than a couple voxels, right? (I think I picked 2 as a reasonable "1 voxel" threshold, but it was admittedly a quick and dirty guess.)

fmriprep/workflows/util.py
```diff
@@ -17,12 +17,35 @@
...
+    comp = mat2.dot(np.linalg.pinv(mat1))
+    trans, rotation_matrix, scales, shears = decompose44(comp)
+
+    shift_thresh = 5         # mm
+    rot_thresh = np.pi / 36  # 5 degrees
+    scale_thresh = 1.1       # scale factor
+
+    shift_magnitude = np.sqrt(trans.dot(trans))  # 2-norm
+    rot = mat2axangle(rotation_matrix)[1]
+    max_scale = np.max(np.abs(scales))
```
effigies: Two changes here:

1) Use the 2-norm to get total shift, not just max in any direction
2) Increase shift threshold to 5mm

effigies:
The translation differences are large, and I believe it's a function of the [.mat file format](https://fsl.fmrib.ox.ac.uk/fsl/fslwiki/FLIRT/FAQ#What_are_FLIRT_schedule_files.3F).

> The FSL convention for transformation matrices uses an implicit centre of transformation - that is, a point that will be unmoved by that transformation, which is an arbitrary choice in general. This arbitrary centre of the transformation for FSL is at the mm origin (0,0,0) which is at the centre of the corner voxel of the image.

> When using the transformation parameters from FLIRT, there is an additional complication in that the parameters are calculated in a way that uses a different centre convention: the centre of mass of the volume. The effect of this is that each of the three matrices above end up with an adjustment in the fourth column (top three elements only) that represents a shift between the corner origin and the centre of mass, while the rest of the matrix (first three columns) is unaffected. Once that is done the matrices are multiplied together, as indicated above, and you get your final matrix.

It may be that we just need to drop the translation check altogether.

oesteban:
:+1: I was ok with checking only the scaling and so I remain. The center
being in a corner will also change the concept of angle, so those
parameters will not be "trustworthy" either.

effigies:
Right. I was just playing with converting to LTA, and got identical scaling parameters but quite different translation and rotation angles.

effigies:
Looking at a lot of these, I'm not sure that there's a clear scaling threshold. Some are mostly rotations/translations, and a small rotation with a very large translation can cause as much distortion as a small(ish) translation with a large rotation. One possibility could be to have a threshold on some function of rotation and translation...

I'm going to try converting to [LTA](https://surfer.nmr.mgh.harvard.edu/fswiki/FsTutorial/LtaFormat), which I believe uses the RAS=(0,0,0) origin, which should at least be no worse than FSL's corner origin, and may result in more obvious thresholds.

I think there's a possibility we'll need override options, no matter how we decide this, allowing users to force BBR or fallback when our heuristics fail.

effigies:
Just as an update on this, I'm definitely not getting the level of distortion @jdkent was in his [original post](https://neurostars.org/t/strange-epi-to-t1w-registration-result/696). For the most part, the bbregister results look reasonable, while the FLIRT+BBR results are being rejected.

@jdkent Do you think you could re-run that subject with the latest release? I'm a little concerned that I'm no longer replicating the issue and getting bad metrics to find cut-offs for.

jdkent:
I'll re-reun with the latest release and report back

effigies:
I think I'm converging on using the "norm" from ArtifactDetect, which is a kind of compromise between @chrisfilo's idea of using framewise displacement and my earlier approach of using a differential affine matrix. The norm approach took a series of motion parameters, constructed affines, and finds the norm of the displacement, which is the maximum displacement of the centers of a bounding cube, and thus takes into account rotations, translations and scales (and shears, but we're only using 9-dof transforms).

With nipy/nipype#2198, I separate the norm calculation from constructing affines from the motion parameters, allowing us to directly pass the affine transform matrices.

The affines in [LTA format](https://surfer.nmr.mgh.harvard.edu/fswiki/FsTutorial/LtaFormat) (RAS2RAS) seem to produce differences of norms that correspond intelligibly with distortion, with norm > 20 being a pretty reliable indicator of a bad BBR, and < 10 indicating good. The exact threshold isn't obvious, with some that I marked "accept" and "reject" falling on either side as I moved between 15 and 20. For now I've put 15, and I'll see what you all think.

I'll package up the spreadsheet and visualizations first thing in the morning.

Given that I still haven't found a clear threshold, I'm inclined to add `--force-bbr` and `--force-no-bbr` options.

jdkent:
I reran the subject with 
[FMRIPREP version: 1.0.0-rc5](https://www.dropbox.com/s/ow0d2b90it5odxm/sub-experimentalGE176.html?dl=0), I still have the distorted results on my end. Did I share the fieldmap or do you just have the bold and anat?

effigies:
I do have the fieldmaps.

fmriprep/workflows/util.py
```diff
@@ -180,11 +191,14 @@ def init_bbreg_wf(bold2t1w_dof, report, reregister=True, name='bbreg_wf'):
         :simple_form: yes
 
         from fmriprep.workflows.util import init_bbreg_wf
-        wf = init_bbreg_wf(bold2t1w_dof=9, report=False)
+        wf = init_bbreg_wf(use_bbr=True, bold2t1w_dof=9, report=False, omp_nthreads=1)
```
effigies: Using `use_bbr=True` here to keep the workflow cleaner; the selection-related nodes don't seem informative to me.

effigies:
@jdkent Do you think you can try this branch with the `--force-bbr`? I'm getting much more reasonable `bbregister` results, which may be a consequence of splitting `mri_coreg` and `bbregister` (though I tried to make it equivalent).

I think you should be able to use this branch with `fmriprep-docker -f fmriprep -i poldracklab/fmriprep:1.0.0-rc5` without worrying about rebuilding the docker image.

If you're still getting the same issue, then I think it's possible there's a difference in our data.

For reference, here's what I'm seeing using `--force-bbr`:

bbregister_not_so_bad

jdkent:
quick question: does ```fmriprep-docker``` circumvent the need to use a singularity container if I'm running the subject on a cluster (that doesn't allow docker becuase sudo)?

effigies:
Ah, no, I assumed you were using docker. For singularity, it's also pretty simple.

From [the docs](https://fmriprep.readthedocs.io/en/latest/contributors.html):

> Patching containers can be achieved in Singularity by using the PYTHONPATH variable:
> 
> ```Shell
> $ PYTHONPATH="$HOME/projects/fmriprep" singularity run fmriprep.img \
>      /scratch/dataset /scratch/out participant -w /out/work/
> ```

Making whatever modifcations are needed.

effigies:
@jdkent Any word?

jdkent:
@effigies Hi, sorry, I got swamped with other duties, and dealing with an improperly made singularity container. I should be able to run the subject tonight since I think I fixed the singularity container.

effigies:
No worries. Thanks for the update.

jdkent:
@effigies, I'm sorry about this, I'm working with our IT to diagnose some permission errors we having on our University's cluster, What was supposed to be a simple fix has turned into an ordeal, but I "should" have a workaround to be able to run the container.

jdkent:
[I'm still getting the same issue](https://www.dropbox.com/s/ow0d2b90it5odxm/sub-experimentalGE176.html?dl=0), from running a singularity container containing the bbr_fallback branch from your github. The only difference between our data that I can think of off the top of my head is the anatomicals have pydeface.py run on them for images I shared, I'll try running the same container with defaced anatomicals, unless you have additional insights on what could be different between our sets.

effigies:
@jdkent would you mind trying with a [singularity image I just made](https://drive.google.com/open?id=0B4dQbkKCyJ6uUWk2R3hsdkJ3Nk0)? I'm going to rerun `sub-experimentalGE176` with a fresh working directory, and I'd appreciate it if you did the same. This should help make absolutely sure that we're really actually running the same code and getting different results. (And remember to set `PYTHONPATH=""`.)

Also, just to compare the input data:

```Shell
$ find . -type f | sort | xargs md5sum   
dc5192bba52512aa089d5b4acd78094a  ./sub-experimentalGE176/ses-post/anat/sub-experimentalGE176_ses-post_T1w.nii.gz
3d1943acdfe6b1101e7a7b5662a31137  ./sub-experimentalGE176/ses-post/anat/sub-experimentalGE176_ses-post_T2w.nii.gz
08f8d1c9b7f06eefb53ca8716e20e19c  ./sub-experimentalGE176/ses-post/fmap/sub-experimentalGE176_ses-post_fieldmap.json
2391c135b4bceafc19b5c1e8f9b2fb35  ./sub-experimentalGE176/ses-post/fmap/sub-experimentalGE176_ses-post_fieldmap.nii.gz
9ae05ce59c816d48ebd46f82e53b13e5  ./sub-experimentalGE176/ses-post/fmap/sub-experimentalGE176_ses-post_magnitude.nii.gz
b96de2c4fae0289bf5c37e9d276894a8  ./sub-experimentalGE176/ses-post/func/sub-experimentalGE176_ses-post_task-flanker_bold.nii.gz
c370895a23665a48e796b3a066797513  ./sub-experimentalGE176/ses-post/func/sub-experimentalGE176_ses-post_task-flanker_events.tsv
47b13928a1f98cdcc5c95384073734f5  ./sub-experimentalGE176/ses-post/func/sub-experimentalGE176_ses-post_task-rest_bold.nii.gz
54a16394068dc3cc5305a0c404058801  ./sub-experimentalGE176/ses-pre/anat/sub-experimentalGE176_ses-pre_T1w.nii.gz
44dd3617f16ec32c984c444f4585a5c9  ./sub-experimentalGE176/ses-pre/anat/sub-experimentalGE176_ses-pre_T2w.nii.gz
5962f62c2e4e99a022c601c92ceaa6c3  ./sub-experimentalGE176/ses-pre/fmap/sub-experimentalGE176_ses-pre_fieldmap.json
bb4b7d4c9737b31ad38ec9b9de290d8d  ./sub-experimentalGE176/ses-pre/fmap/sub-experimentalGE176_ses-pre_fieldmap.nii.gz
e74d79e5d1393a87d5948f31c071a480  ./sub-experimentalGE176/ses-pre/fmap/sub-experimentalGE176_ses-pre_magnitude.nii.gz
d866bbc3232d95f0a9c0cfb70832a035  ./sub-experimentalGE176/ses-pre/func/sub-experimentalGE176_ses-pre_task-flanker_bold.nii.gz
a94594652c401a9a2f81538f5170c4ae  ./sub-experimentalGE176/ses-pre/func/sub-experimentalGE176_ses-pre_task-flanker_events.tsv
418616bee67a679188a2705c5eb852e0  ./sub-experimentalGE176/ses-pre/func/sub-experimentalGE176_ses-pre_task-rest_bold.nii.gz
3a0c44992250ed25307bd417be7b7a49  ./task-flanker_bold.json
1fe14f86e825507e1a1c221a18d54fe1  ./task-rest_bold.json
```

effigies:
Here's [a tarball](https://drive.google.com/file/d/0B4dQbkKCyJ6ucFFraEJTZjdYdUk/view?usp=sharing) with the latest report, along with the SVG reportlets for all runs (both BBR and fallback).

effigies:
Okay, I've managed to clean this up to a large extent, including reducing some of the extraneous commits that were more useful during testing.

One thing to note if reading individual merges is that, up until the final commit, the transform comparison function assumed FSL `.mat` input, so the `mri_coreg`/`bbregister` comparison would not have worked.

I'm going to rebuild this docker image and run some tests, but I think this is (again) ready for review.

jdkent:
sorry for the delay again, just ran the md5sum on the data
```
dc5192bba52512aa089d5b4acd78094a  ./sub-experimentalGE176/ses-post/anat/sub-experimentalGE176_ses-post_T1w.nii.gz
3d1943acdfe6b1101e7a7b5662a31137  ./sub-experimentalGE176/ses-post/anat/sub-experimentalGE176_ses-post_T2w.nii.gz
08f8d1c9b7f06eefb53ca8716e20e19c  ./sub-experimentalGE176/ses-post/fmap/sub-experimentalGE176_ses-post_fieldmap.json
2391c135b4bceafc19b5c1e8f9b2fb35  ./sub-experimentalGE176/ses-post/fmap/sub-experimentalGE176_ses-post_fieldmap.nii.gz
9ae05ce59c816d48ebd46f82e53b13e5  ./sub-experimentalGE176/ses-post/fmap/sub-experimentalGE176_ses-post_magnitude.nii.gz
b96de2c4fae0289bf5c37e9d276894a8  ./sub-experimentalGE176/ses-post/func/sub-experimentalGE176_ses-post_task-flanker_bold.nii.gz
c370895a23665a48e796b3a066797513  ./sub-experimentalGE176/ses-post/func/sub-experimentalGE176_ses-post_task-flanker_events.tsv
47b13928a1f98cdcc5c95384073734f5  ./sub-experimentalGE176/ses-post/func/sub-experimentalGE176_ses-post_task-rest_bold.nii.gz
54a16394068dc3cc5305a0c404058801  ./sub-experimentalGE176/ses-pre/anat/sub-experimentalGE176_ses-pre_T1w.nii.gz
44dd3617f16ec32c984c444f4585a5c9  ./sub-experimentalGE176/ses-pre/anat/sub-experimentalGE176_ses-pre_T2w.nii.gz
5962f62c2e4e99a022c601c92ceaa6c3  ./sub-experimentalGE176/ses-pre/fmap/sub-experimentalGE176_ses-pre_fieldmap.json
bb4b7d4c9737b31ad38ec9b9de290d8d  ./sub-experimentalGE176/ses-pre/fmap/sub-experimentalGE176_ses-pre_fieldmap.nii.gz
e74d79e5d1393a87d5948f31c071a480  ./sub-experimentalGE176/ses-pre/fmap/sub-experimentalGE176_ses-pre_magnitude.nii.gz
d866bbc3232d95f0a9c0cfb70832a035  ./sub-experimentalGE176/ses-pre/func/sub-experimentalGE176_ses-pre_task-flanker_bold.nii.gz
a94594652c401a9a2f81538f5170c4ae  ./sub-experimentalGE176/ses-pre/func/sub-experimentalGE176_ses-pre_task-flanker_events.tsv
418616bee67a679188a2705c5eb852e0  ./sub-experimentalGE176/ses-pre/func/sub-experimentalGE176_ses-pre_task-rest_bold.nii.gz
3a0c44992250ed25307bd417be7b7a49  ./task-flanker_bold.json
1fe14f86e825507e1a1c221a18d54fe1  ./task-rest_bold.json
```
I'll start running the singularity container you provided in a fresh working directory. with `PYTHONPATH=""`

oesteban:
@jdkent, @effigies what is the current status of this PR?

effigies:
This is ready to review. I still like the affine difference we're taking, and think the main thing that needs to be decided is if we want a more stringent threshold.

jdkent:
Alright, I ran the subject with [--force-bbr](https://www.dropbox.com/s/42f32brq88tosmr/bad_10092017_sub-experimentalGE176.html?dl=0) and [without the flag](https://www.dropbox.com/s/ow0d2b90it5odxm/sub-experimentalGE176.html?dl=0), and the results are much improved, just not identical to @effigies. I ran it with a clear PYTHONPATH and in different working directories, one thing that isn't clean is the fact that I'm using an existing freesurfer directory, and not re-rerunning recon-all. However, the function did work and improved my results, so I don't think I should be holding up this pull request any longer.

effigies:
@oesteban Rebased and ready for review.

effigies:
@jdkent Your "without the flag" one looks very similar to mine, so I'm glad we're in a similar place. How do you feel about that threshold, by the way? We can make it more likely to reject that first run. The more I look at these results, the more I think we might want to make the threshold a little more stringent...

Thanks for testing this out for us, by the way.

jdkent:
oh, I thought the bbregister result was rejected for all the registrations. Then yes, a more stringent threshold could catch that case and improve the registration. I just don't know how much optimization for my case would sacrifice the generalization of the threshold to other datasets.

fmriprep/workflows/util.py
```diff
@@ -18,14 +18,34 @@
...
+from niworkflows.interfaces.registration import FLIRTRPT, BBRegisterRPT, MRICoregRPT
 from niworkflows.interfaces.masks import SimpleShowMaskRPT
 
 from ..interfaces.images import extract_wm
 
 DEFAULT_MEMORY_MIN_GB = 0.01
 
 
+def compare_xforms(lta_list, norm_threshold=15):
+    import numpy as np
```
oesteban: Please add a docstring with a link to the original issue to get quickly to the discussion and examples of how the threshold plays out (e.g. 10 = very small affine changes accepted and 20 = large changes accepted). Something along these lines.

```Python
"""
Computes a *distance* between two transforms as the maximum overall displacement of 
the midpoints of the faces of a cube due to translation and rotation.

Parameters

  lta_list : list or tuple of str
      the two given affines in LTA format

  norm_threshold : int
      the upper bound limit to the distance of the second transform w.r.t. the first; the lower
      the smaller distances are accepted; a default of 15 was found empirically appropriate
      to decide whether BBR transforms should be rejected.

"""
```

fmriprep/workflows/util.py
```diff
@@ -233,13 +258,13 @@ def init_bbreg_wf(bold2t1w_dof, name='bbreg_wf'):
...
     outputnode = pe.Node(
-        niu.IdentityInterface(['itk_bold_to_t1', 'itk_t1_to_bold', 'out_report']),
+        niu.IdentityInterface(['itk_bold_to_t1', 'itk_t1_to_bold', 'out_report', 'fallback']),
         name='outputnode')
 
-    bbregister = pe.Node(
-        BBRegisterRPT(dof=bold2t1w_dof, contrast_type='t2', init='coreg',
-                      registered_file=True, out_lta_file=True, generate_report=True),
-        name='bbregister')
+    mri_coreg = pe.Node(
```
oesteban: Does MRICoreg generate a new T1w space with the appropriate xforms? I'll need to get this clear before I start with #745.

fmriprep/workflows/util.py
```diff
@@ -174,7 +194,7 @@ def init_skullstrip_bold_wf(name='skullstrip_bold_wf'):
     return workflow
 
 
-def init_bbreg_wf(bold2t1w_dof, name='bbreg_wf'):
+def init_bbreg_wf(use_bbr, bold2t1w_dof, omp_nthreads, name='bbreg_wf'):
```
oesteban: An additional paragraph describing the feature being added would be very valuable. Just a couple of sentences to describe what happens if `use_bbr=None`.

fmriprep/workflows/util.py
```diff
@@ -268,12 +290,73 @@ def init_bbreg_wf(bold2t1w_dof, name='bbreg_wf'):
...
+        (reports, select_report, [('out', 'inlist')]),
+        (compare_transforms, select_report, [('out', 'index')]),
+        (select_report, outputnode, [('out', 'out_report')]),
         ])
 
     return workflow
 
 
-def init_fsl_bbr_wf(bold2t1w_dof, name='fsl_bbr_wf'):
+def init_fsl_bbr_wf(use_bbr, bold2t1w_dof, name='fsl_bbr_wf'):
```
oesteban: Refer to previous workflow regarding `use_bbr`

fmriprep/workflows/util.py
oesteban: Added link to original issue.

fmriprep/workflows/util.py
effigies: No. `mri_coreg` only generates a transform matrix. If you want to apply the transform, you can use `mri_vol2vol` (see [MRICoregRPT](https://github.com/poldracklab/niworkflows/blob/c55052435622efdc57959702af51178a4d774fae/niworkflows/interfaces/registration.py#L231-L234) or `mri_convert`).

Issue #700, opened 2017-09-13
Add optional smoothing of the outputs


chrisfilo:
> So I’m thinking of an --smooth-fwhm argument that could take a list fwhm for smoothing. For example, if I specify --smooth-fwhm 0.0 6.0, the output will contain a smoothed data for activation analysis and a unsmoothed data for MVPA analysis (also with ica denoised version).

https://neurostars.org/t/preprocessing-for-atlas-based-connectivity-matrices/743/9

It might be also worth considering adaptive smoothing methods or SUSAN.

effigies:
From discussion with @satra, one useful option would be to enable smoothing of surface-sampled functional series, which would just be an option to `mri_vol2surf`.

tsalo:
Given the [documentation on new features](https://www.nipreps.org/community/features/#some-questions-will-always-be-asked-about-a-new-feature), is this something that should be implemented _within_ fMRIPrep or is it better kept as a separate step? I'm not very familiar with surface data, but it's pretty easy to smooth volumetric data, and there are a bunch of ways to do it (SUSAN vs. standard Gaussian being one method already listed).

PR #706, opened 2017-09-17
WIP: add smoothing option

jdkent:
In reference to #700 

This is my attempt at adding a smoothing option (using FSL's SUSAN) to fmriprep. Currently the smoothed derivative (if selected) is generated in addition to the unsmoothed derivative to account for the use case mentioned in @chrisfilo 's  post. However, I did not implement smoothing as an iterable field, so separate calls to fmriprep would have to called if the user wanted several different smoothing kernels.

My goal is to try to get this merged with fewer than 100 comments (Not that I don't enjoy our conversations)! (see #539)

link to my suggestion on how to name smoothed derivatives: https://docs.google.com/document/d/1qBNQimDx6CuvHjbDvuFyBIrf2WRFUOJ-u50canWjjaw/edit#heading=h.mqkmyp254xh6


jdkent:
I may need some help for the sphinx error I'm getting, I'm still not familiar with this testing suite.

jdkent:
RE: [OSError: Duplicate node name smooth_wf found.](https://circleci.com/gh/poldracklab/fmriprep/1097#tests/containers/1)

I thought multiple instances of the same class (class smooth_wf) could be instantiated without them conflicting, unless I'm misunderstanding this error. Is this a property of networkx?

chrisfilo:
Some things we need to consider in context of this PR:

1. Should we have a single `--smooth-fwhm` option that applies to both surface and volume?

I would say yes (this is for example what HCP is doing). It's ok if this PR only applies smoothing to volume as long as everything is documented and we plan to make a similar PR for surface.

2. If user specifies `--smooth-fwhm` should we even output the unsmoothed data?

I would say no. In most cases if people specified smoothing kernel they want to use it. Extra data would just eat up their disk space. Happy to discuss though.

3. If user specifies `--use-aroma` and `--smooth-fwhm 8` (which that is different from 6mm what AROMA classifier expects) should we output 6mm smoothed aroma non-aggressively denoised or 8mm smoothed aroma non-aggressively denoised or both?

I am a bit on the fence here, but I am leaning towards saving only 8mm smoothed aroma non-aggressively denoised variant to save space.

jdkent:
1. Should we have a single ```--smooth-fwhm``` option that applies to both surface and volume?

I agree with as far as what I know. I'm not too familiar with surfaces, so I don't know how independent the analysis would be apart from the volume data, and if users would want a different kernel.

2. If user specifies --smooth-fwhm should we even output the unsmoothed data?

My impression is that MVPA is becoming more popular, and as mentioned in the neurostar's post, the images are commonly unsmoothed, so if someone wanted to a more traditional task analysis in addition to MVPA, they would want both. So then the balance is between providing outputs in many possible configurations to try to adapt to the user's need, versus keeping it simple and less storage intensive. We could add another commandline option that specifies if the user wants both smoothed and unsmoothed, and keep the default at only smoothed output.

3. If user specifies --use-aroma and --smooth-fwhm 8 (which that is different from 6mm what AROMA classifier expects) should we output 6mm smoothed aroma non-aggressively denoised or 8mm smoothed aroma non-aggressively denoised or both?

I don't about this one, I don't have an intuition on how much the ICA-AROMA classifier relies on how smooth the data are. I'm nervous letting that smoothing parameter be modified to any kernel size, and have the user believe it's a valid result. My ideal would be to have the classification done at 6mm smoothing and then non-aggressively filter the noise components from the 8mm smoothed image, then we could only pass the 8mm smooth aroma denoised bold.

chrisfilo:
One potential solution to question no 2 is making  --smooth-fwhm take a list of values. The default would be "0", but user could specify "0 6" (output unsmooth and 6mm smoothed) or just "6".

effigies:
Hi @jdkent. Just checking in on what you think the status of this PR is. If you're waiting for a review, I can make some comments.

To chime in a bit on the discussion here:

* I think having `--smooth-fwhm` take multiple arguments is reasonable, and only outputting unsmoothed data if 0 is among the arguments (or no smoothing option is specified). Probably the simplest way to go about this would be to make the `SUSAN` node a `MapNode`, and make the `fwhm` parameter a list/tuple that defaults to `(0,)`. Create a smoothing workflow if there are any non-zero elements in `fwhm`, and pass it the sequence with the zero removed. The derivatives datasink can also be made a `MapNode` to make it simpler.

* With smoothing surface-sampled outputs, the question is whether to smooth the input volume or the output surface. I don't have a strong preference here. It might be good to get on the FreeSurfer mailing list to see if there's a preference. If the preference is to smooth the volume and then sample to the surface, is that using the standard `mri_vol2surf --fwhm`, or would sampling a SUSAN-smoothed volume be better?

* For ICA-AROMA, if the user wants 6mm outputs, it would be *ideal* not to perform that smoothing twice, but if that's not easily engineered, we can leave that as an open issue.

* I was unable to find your comment about derivative naming in the linked BEP; not sure if it got marked resolved. I think in the short term `variant-SUSAN<N>mm` is fine, but perhaps something like `smooth-SUSAN<N>mm` would be good? Smoothing is common enough to warrant a key, IMO. Further, right now your actual derivatives are printing `6.0` instead of `6`. You can use `'{:.0g}'.format(6.0)` to print `6`.

Thanks again for this PR. Let us know if you need anything from us.

jdkent:
Hi,
Sorry I was working on NiBetaSeries to see if I can basic working version before I could focus my attention on this again. This is not ready for a review yet, as I will be incorporating the comments from our conversation.

jdkent:
Okay, I've been meaning to bring up some discussion once I had some time. If we add smoothing, there was also talk of adding 3d-despike, and potentially other processing steps. I was wondering if we wanted an entire workflow (or even file) dedicated to doing this processing. Have it be named `process_bold_wf` or something like that, where we complete smoothing, 3d-despike, fsl_regfilt (with the components derived from ICA-AROMA), etc. We want to make sure that we apply smoothing at all smoothing kernels the user wants to both the volume and surface bold. Then potentially have fsl_regfilt run on the smoothed outputs.

effigies:
Just a quick comment on this: I believe the internal consensus has been that smoothing isn't properly a job for fMRIPrep, as it multiplies the number of full-sized BOLD series produced. By leaving it to later steps in the analysis pipeline, flexibility (e.g. trying out different smoothing algorithms, parameters) doesn't have to come at the cost of replication of the largest part of the derivatives.

effigies:
Just a quick comment on this: I believe the internal consensus has been that smoothing isn't properly a job for fMRIPrep, as it multiplies the number of full-sized BOLD series produced. By leaving it to later steps in the analysis pipeline, flexibility (e.g. trying out different smoothing algorithms, parameters) doesn't have to come at the cost of replication of the largest part of the derivatives.

jdkent:
If that's the consensus, then I can close this pull request.

Issue #701, opened 2017-09-13
Poor skullstripping/registration BOLD->T1w results


jdkent:
There was extra dura mater included with the brain in the parietal area, as well as registration issues as already noted in #694, sharing this subject as per @chrisfilo's request from [this conversation](https://neurostars.org/t/fmriprep-spatial-normalization/762)

[download zip](https://osf.io/4tqa7/download)

fmriprep was run using our cluster via the following command:
```
singularity run -H ${HOME}/singularity_home -B /Shared/vosslabhpc:/mnt \
/Shared/vosslabhpc/UniversalSoftware/SingularityContainers/poldracklab_fmriprep_latest-2017-08-12-1b62280f96f1.img \
/mnt/Projects/PACR-AD/Imaging/BIDS /mnt/Projects/PACR-AD/Imaging/BIDS/derivatives \
participant --participant_label controlGE140 \
-w /mnt/Projects/PACR-AD/Imaging/BIDS/derivatives/work/fmriprep/ \
--write-graph --mem_mb 35000 --omp-nthreads 10 --nthreads 16 --use-aroma --bold2t1w-dof 6 \
--output-space fsnative template \
--template MNI152NLin2009cAsym
```

Let me know if you need any additional information.

effigies:
Using the NKI template (see #729) had no discernible effect on the skull-stripping. And I tried manually running `mri_watershed` on the `T1.mgz` outputs in FreeSurfer, and found that it was extremely overzealous, removing most of prefrontal cortex.

I'm wondering if what we should do is permit users to edit the outputs and make a second pass, with `--fixed-skull-strip img.nii.gz`. I would basically make a check that the passed image has the same dimensions/affine, and re-run any downstream preprocessing, including patching the fixed skull-strip into FreeSurfer.

I would not by default re-run FreeSurfer.

effigies:
Hi @jdkent. I've had a few pokes at this, but I don't think that we're going to be able to move very quickly here.

I think the fastest way to resolve this would be to move to the [ANTs discussion list](https://sourceforge.net/p/advants/discussion/840261/), and provide the commandline in your `anat_preproc_wf/skullstrip_ants_wf/t1_skull_strip/command.txt` file. Note that the `T_template0_*.nii.gz` files are from an [RAS-oriented version of their OASIS template](https://osf.io/bxuga/).

In any event, if you manage to fix this before we can, please let us know how, and we'll try to incorporate these fixes into fmriprep.

[...]

Issue #431, opened 2017-04-03
Refine ANTs brain mask with FreeSurfer


chrisfilo:
Sometimes ANTs brainmasks fails
![image](https://cloud.githubusercontent.com/assets/238759/24636174/349e69c6-188d-11e7-9f81-6bfe0f3f51d9.png) 
ds114 subject 01

However we can use the refined brainmask obtained from the FreeSurfer reconstruction (like here: https://github.com/BIDS-Apps/nipypelines/blob/master/run.py#L401)

effigies:
We inject the ANTs brainmask before aparc occurs.

effigies:
Well, I suppose as long as it only ever fails by including too much and never too little, we can use it.

chrisfilo:
Yes - it would only help in cases where ANTs includes parts of dura/skull.

effigies:
So the idea is to continue to inject ANTs as a prior, and then let a binarized `aparc` be the final mask we use?

That should be pretty easy. I assume we don't want to turn around and use the updated mask to fix `recon-all`.

chrisfilo:
Yes that's the idea (without rerunning recon-all). The refined mask
could be used to improve MNI coregistration and FAST tissue segmentation
(although wether we should use FAST if we have freesurfer segmentation is a
nother issues...)

chrisfilo:
So I've tried this last night and the problem I run into is that the aparc based mask is too tight and needs to be dilated. However, when you dilate it (even when using one voxel) the result is worse (includes more non-brain) than the initial ANTs mask.

effigies:
What about setting the mask to `sum(dilated_aparc + ANTs) == 2`?

oesteban:
I'd bet @binarybottle's opinion on this will be very valuable to make a decision here. Arno, can we ask you what do you think about FreeSurfer's and ANTS' brain masks? And how to combine them?

binarybottle:
Thanks for asking!  Because I have also found FreeSurfer and ANTs to under- and over-segment as mentioned above, I wrote the following function to try and reconcile the two:
https://github.com/nipy/mindboggle/blob/master/mindboggle/guts/segment.py#L1660

The docstring says::

    Combine the voxels of one label from two files and overlay them
    on the second label combined from the two files.

    An example application is to combine two segmentation volumes,
    such as from FreeSurfer and ANTs, to obtain a single cortex=2
    and noncortex=3 segmentation file, by taking the union of cortex voxels
[...]

PR #729, opened 2017-09-29
[RTM] ENH: Enable brain extraction with NKI template

effigies:
**State of this PR**

This PR builds in the infrastructure for using alternative templates, and adds an option to use the NKI template instead of the default OASIS.

Something similar has been done in the @kochalka [branch](https://github.com/poldracklab/fmriprep/compare/master...kochalka:master), which suggests that the ability to customize skull-stripping templates is desired by at least some users.

That said, I don't recommend adding NKI as an option without specific user demand or a demonstration that it improves skull-stripping in some cases.

**Original text**

An attempt to address #701. There was [previous discussion](https://sourceforge.net/p/advants/discussion/840261/thread/3fc8296f/#fcad) of using the NKI template as an alternative to OASIS. In that case, the issue was wrap-around, so it's not clear that we're going to see any improvement here.

I was able to replicate the poor skull stripping locally. Incidentally, the FreeSurfer reconstruction looked great despite that, possibly due to T2 pial surface refinement.

Will report back on skull-strip quality in the morning.


chrisfilo:
If FreeSurfer masks looked good it might be to revisit #431.

PR #733, opened 2017-10-02
[RTM] ENH: Register FreeSurfer template to FMRIPREP template

effigies:
Addresses issue raised in https://github.com/poldracklab/fmriprep/pull/726#issuecomment-333183529.

Given this, I think we should output the registration transform, to give users the tools to move between FMRIPREP and FreeSurfer spaces. I'd recommend the LTA format, since that has more metadata about the source and target spaces; if users need to convert to other formats, they will presumably have access to the `LTAConvert` tool.

To do:

* [x] Ensure output surfaces are aligned with T1_preproc
* [x] Affine

### DIFF
--- a/.circle/data/ds005_outputs.txt
+++ b/.circle/data/ds005_outputs.txt
@@ -24,6 +24,7 @@ ds005/out/fmriprep/sub-01/anat/sub-01_T1w_space-MNI152NLin2009cAsym_class-WM_pro
 ds005/out/fmriprep/sub-01/anat/sub-01_T1w_space-MNI152NLin2009cAsym_dtissue.nii.gz
 ds005/out/fmriprep/sub-01/anat/sub-01_T1w_space-MNI152NLin2009cAsym_preproc.nii.gz
 ds005/out/fmriprep/sub-01/anat/sub-01_T1w_space-MNI152NLin2009cAsym_warp.h5
+ds005/out/fmriprep/sub-01/anat/sub-01_T1w_target-fsnative_affine.txt
 ds005/out/fmriprep/sub-01/func
 ds005/out/fmriprep/sub-01/func/sub-01_task-mixedgamblestask_run-01_bold_AROMAnoiseICs.csv
 ds005/out/fmriprep/sub-01/func/sub-01_task-mixedgamblestask_run-01_bold_confounds.tsv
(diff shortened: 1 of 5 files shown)


effigies:
Addressing this approach to [BEP 014](https://docs.google.com/document/d/11gCzXOPUbYyuQx8fErtMO9tnOKC3kTWiL9axWkkILNE/edit#), I believe what we'll want is:

```
fmriprep/sub-<label>/anat/sub-<label>_T1w_preproc_target-fsnative_affine.txt
```

In which case, we want the inverse of what I'm currently generating, which is easily constructed with

```Shell
lta_convert --inlta transform.lta --outitk transform.txt --invert
```

I think the most appropriate alternative would be to add a transform to the `freesurfer/sub-<label>/mri/transforms/` directory, and I feel we should make minimal changes there.

effigies:
To review this PR, separately review #741, and then look at https://github.com/effigies/fmriprep/compare/fsfix_common...fix/freesurfer_register.

.circle/data/ds005_outputs.txt
```diff
@@ -24,6 +24,7 @@ ds005/out/fmriprep/sub-01/anat/sub-01_T1w_space-MNI152NLin2009cAsym_class-WM_pro
 ds005/out/fmriprep/sub-01/anat/sub-01_T1w_space-MNI152NLin2009cAsym_dtissue.nii.gz
 ds005/out/fmriprep/sub-01/anat/sub-01_T1w_space-MNI152NLin2009cAsym_preproc.nii.gz
 ds005/out/fmriprep/sub-01/anat/sub-01_T1w_space-MNI152NLin2009cAsym_warp.h5
+ds005/out/fmriprep/sub-01/anat/sub-01_T1w_target-fsnative_affine.txt
```
oesteban: I thought the filename was `sub-01_T1w_preproc_target-fsnative_affine.txt` (missing `_preproc`)

fmriprep/interfaces/surf.py
```diff
@@ -218,3 +224,20 @@ def normalize_surfs(in_file):
             pointset.meta.data.insert(2, geom_type)
     img.to_filename(fname)
     return os.path.abspath(fname)
+
+
+def load_transform(fname):
```
oesteban: a minimal docstring here?

.circle/data/ds005_outputs.txt
effigies: I did change it to make it more like `ds005/out/fmriprep/sub-01/anat/sub-01_T1w_space-MNI152NLin2009cAsym_warp.h5`, which is a transform from `T1w` space to `MNI152NLin2009cAsym`. Speaking of which, @chrisfilo, reading the BEP, should that be a `target-MNI...` instead of `space-MNI...`?

chrisfilo: Correct.

PR #740, opened 2017-10-06
[RTM] RF: Assume phase-encoding direction is A-P unless specified L-R

### LINKED ISSUE #714: SyN-SDC guessed phase encoding incorrectly
ds000243 sub-042 task-rest_run-2

Correction was attempted in L/R but should've been in A/P.

effigies:
As discussed in #714, the simplest way to build in a prior of A-P phase-encoding is to always use A-P unless otherwise specified. When this produces a discernible error, the correct response would be to update the dataset metadata.

With this change, we can eliminate the cost function machinery from the BBR workflows, which should simplify #694, if this goes in first.

Closes #714.

### DIFF
--- a/fmriprep/interfaces/reports.py
+++ b/fmriprep/interfaces/reports.py
@@ -33,6 +33,7 @@ SUBJECT_TEMPLATE = """\t<ul class="elem-desc">
 
 FUNCTIONAL_TEMPLATE = """\t\t<h3 class="elem-title">Summary</h3>
 \t\t<ul class="elem-desc">
+\t\t\t<li>Detected phase-encoding (PE) direction: {pedir}</li>
 \t\t\t<li>Slice timing correction: {stc}</li>
 \t\t\t<li>Susceptibility distortion correction: {sdc}</li>
 \t\t\t<li>Registration: {registration}</li>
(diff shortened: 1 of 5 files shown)


effigies:
The first three commits are some basic refactoring common to #694 and #733.

fmriprep/interfaces/reports.py
```diff
@@ -33,6 +33,7 @@
 
 FUNCTIONAL_TEMPLATE = """\t\t<h3 class="elem-title">Summary</h3>
 \t\t<ul class="elem-desc">
+\t\t\t<li>Detected phase-encoding (PE) direction: {pedir}</li>
```
oesteban: Remove "Detected"? since now it is the default AP or manually set

fmriprep/workflows/bold.py
```diff
@@ -1395,22 +1363,13 @@ def init_nonlinear_sdc_wf(bold_file, layout, freesurfer, bold2t1w_dof,
                                 mem_gb=DEFAULT_MEMORY_MIN_GB)
     fixed_image_masks.inputs.in1 = 'NULL'
 
-    if layout is None:
-        bold_pe = None
-    else:
-        bold_pe = layout.get_metadata(bold_file).get("PhaseEncodingDirection")
-
     restrict_i = [[1, 0, 0], [1, 0, 0]]
```
oesteban:
```Python
restrict = [[int(bold_pe[0] == 'i'), int(bold_pe[0] == 'j'), 0]] * 2
```
And then use `restrict_deformation=restrict` down in L1371

oesteban:
One more thing: the sdc-syn method is one method to estimate the fieldmap. Could we move the code to its proper module under the fieldmap/ folder?

effigies:
Yeah, I can do that. Can do that here or in a separate PR. Up to you.

effigies:
Actually I'll go ahead an say: Let's do that in a separate PR. We can move a lot of logic that's currently in `bold.py` into a `fieldmap/`, which could take a little work and we want to get this in ASAP, unless I'm mistaken.

PR #750, opened 2017-10-10
ENH: Remove AFNI version check monkey patch

effigies:
The need for this was removed in nipy/nipype#2196.


oesteban:
Weird, I just manually merged this in changing the message to contain [skip ci] so that it does not trigger one more build. I think it is safe to close.

effigies:
When you click "Merge", you can change the commit message:

merge_skip_ci

oesteban:
Oh shoot. Forgot that. Thanks for the tip.

Issue #749, opened 2017-10-10
Singularity image built from poldracklab/fmriprep:latest requires additional ldconfig


justbennet:
In Issue 648 for mriqc, https://github.com/poldracklab/mriqc/issues/648#issuecomment-335242120, I reported that when I converted to Singularity from Docker, there were problems because libGL.so.1 is missing.  This seems also to be an issue with fmriprep.

I used the following to create the Singularity container, where Singularity was compiled from the development branch, pre-3.4 release.

fmriprep.def
```
Bootstrap: docker
From: poldracklab/fmriprep:latest

%post
chmod -R a+rX /usr/local/miniconda
chmod -R a+rX /niworkflows_data
```
I am running tests against the `ds008` dataset from OpenfMRI.org.  When it got to
```
Node: fmriprep_wf.single_subject_15_wf.anat_preproc_wf.surface_recon_wf.fs_transform
```
it crashed when the following error was returned to `niworkflows/nipype`.
```
RuntimeError: Command:
tkregister2 --fslregout freesurfer2subT1.mat --mov /out/freesurfer/sub-15/mri/T1.mgz --noedit --reg register.dat --regheader --targ /home/bennet/Projects/mpeckins/MRIPREP/work/fmriprep_wf/single_subject_15_wf/anat_preproc_wf/t1_merge/sub-15_T1w_template.nii.gz
Standard output:

Standard error:
/opt/freesurfer/tktools/tkregister2.bin: error while loading shared libraries: libGL.so.1: cannot open shared object file: No such file or directory

Return code: 127
Interface Tkregister2 failed to run. 
```
Using `singularity shell` in both containers, and running `dkpg -l`, searching for either GL or mesa fails in both containers

```
Singularity mriqc.simg:/> dpkg -l | grep GL
Singularity mriqc.simg:/> dpkg -l | grep mesa

Singularity fmriprep.simg:/> dpkg -l | grep GL
Singularity fmriprep.simg:/> dpkg -l | grep mesa
```
In both containers, the files appear to be there, though.
```
Singularity mriqc.simg:/> find /usr/lib/x86_64-linux-gnu -name libGL\*
/usr/lib/x86_64-linux-gnu/libGL.so.1
/usr/lib/x86_64-linux-gnu/libGLU.so.1
/usr/lib/x86_64-linux-gnu/libGLU.so.1.3.1
/usr/lib/x86_64-linux-gnu/libGLw.so.1
[...]

PR #757, opened 2017-10-12
[RTM] DOCKER: Add rX permissions to make life easier on Singularity users

effigies:
See https://github.com/poldracklab/fmriprep/issues/749#issuecomment-336170941

This should be merged shortly before the next release. Doing so before then will waste time during builds to no benefit, since users won't see the images built until release.


oesteban:
Why don't we release rc7 today?

effigies:
Sure.

effigies:
Want to update the nipype pin in niworkflows, and we can push out a release of that, first?

oesteban:
Oh yes, we should include the latest patch for the grabbled output from autorecon.

PR #785, opened 2017-10-23
[FIX] Reduce FoV of outputs in T1w space

### LINKED ISSUE #722: request for output-space option of functional data in T1w coordinates with original scanning resolution
This came from a question [asked on neurostars](https://neurostars.org/t/why-are-functional-data-files-output-by-fmriprep-in-t1w-space-so-much-larger/796/2). I'm running some participant-specific analyses. An extra output-space option at the resolution of the original BOLD image would be helpful, partially because of their reduced filesize but mainly because I'm trying out a participant-level analysis that would require that resolution.

thanks

oesteban:
Adds the new `fov_mask` connection to the node that generates the registration target reference in T1w space, but with BOLD resolution.

This could also be added to the MNI resampling target, however, the MNI space is already restrictive enough in terms of FoV.

Closes #722

### DIFF
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -28,4 +28,4 @@ dependencies:
     - svgutils
     - nitime
     - nilearn
-    - git+https://github.com/poldracklab/niworkflows.git@cad7fb7cec09b4ac6ad39757697d71a8b1a8b144#egg=niworkflows-0.1.9-dev
+    - git+https://github.com/poldracklab/niworkflows.git@710b73301e391ef6f9e184384187cfd2d921fbe2#egg=niworkflows-0.1.9-dev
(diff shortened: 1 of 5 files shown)


oesteban:
You can tell that the image is being cropped by the CompCor plots.

I'll include copying the dtype to the reference in this PR.

effigies:
New pin: `git+https://github.com/poldracklab/niworkflows.git@710b73301e391ef6f9e184384187cfd2d921fbe2#egg=niworkflows-0.1.9-dev`

Should update in:

* `requirements.txt`
* `fmriprep/info.py`
* `docs/environment.yml`

Note this PR is pending testing on a partial FoV OpenFMRI dataset.

ds000108/sub-03 at least has a cropped FoV (a bit missing off the top.

oesteban:
Testing on ds000108/sub-03 right now.

PR #805, opened 2017-10-31
[ENH] Refactor BOLD workflows

oesteban:
Create a `bold` module that makes it more manageable contributing to FMRIPREP.

Generation of API for workflows has been revised (refs. #615).


fmriprep/workflows/bold/base.py
```diff
@@ -0,0 +1,785 @@
+# -*- coding: utf-8 -*-
+# emacs: -*- mode: python; py-indent-offset: 4; indent-tabs-mode: nil -*-
+# vi: set ft=python sts=4 ts=4 sw=4 et:
+"""
+Orchestrating the BOLD-preprocessing workflow
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. autofunction:: init_func_preproc_wf
+.. autofunction:: init_func_reports_wf
+.. autofunction:: init_func_derivatives_wf
```
effigies: Do you really want to include reports and derivatives in the API? I think I skipped them. But I suppose no harm, either.

fmriprep/workflows/bold/base.py
```diff
@@ -0,0 +1,785 @@
...
+            (bold_reference_wf, bold_hmc_wf, [('outputnode.bold_file', 'inputnode.bold_file')])])
+
+    # Cases:
+    # fmaps | use_syn | force_syn  |  ACTION
+    # ----------------------------------------------
+    #   T   |    *    |     T      | Fieldmaps + SyN
+    #   T   |    *    |     F      | Fieldmaps
+    #   F   |    *    |     T      | SyN
+    #   F   |    T    |     F      | SyN
+    #   F   |    F    |     F      | HMC only
```
effigies: If you're going to do this, should we make the choice of fieldmap unwarping and  SyN-based unwarping part of the fieldmaps workflows? Though, that may make more sense as a separate PR, as this one is already pretty big. Either way.

fmriprep/workflows/bold/base.py
oesteban: It generates two entries with one-liner description. May even help to newcomers.

fmriprep/workflows/bold/base.py
oesteban: Yes - the SyN-based workflow should be attached the same way we are currently doing with the fieldmap decision maker in `workflows.fieldmap.base` if we want to be consistent. But better in a different PR where that's the only target.

oesteban:
Seems like everything is in place. Gong ahead.

Issue #776, opened 2017-10-18
aCompCor - numpy.linalg.linalg.LinAlgError: SVD did not converge


oesteban:
DS105 / sub-6

Probably one that will pop up to nipype:
```
171018-06:23:12,950 workflow ERROR:
         Node acompcor failed to run on host nid00799.
171018-06:23:12,953 workflow ERROR:
         Saving crash info to /scratch/03763/oesteban/fmriprep-phase1/out/ds000105/fmriprep/sub-6/log/20171018-043321_cd7ae052-544c-497a-93d2-3b569f611033/crash-20171018-062312-oesteban-acompcor-ea1bbd45-3ed1-
4714-a7a1-52b68b13b70a.txt
Traceback (most recent call last):
  File "/usr/local/miniconda/lib/python3.6/site-packages/niworkflows/nipype/pipeline/plugins/multiproc.py", line 51, in run_node
    result['result'] = node.run(updatehash=updatehash)
  File "/usr/local/miniconda/lib/python3.6/site-packages/niworkflows/nipype/pipeline/engine/nodes.py", line 407, in run
    self._run_interface()
  File "/usr/local/miniconda/lib/python3.6/site-packages/niworkflows/nipype/pipeline/engine/nodes.py", line 517, in _run_interface
    self._result = self._run_command(execute)
  File "/usr/local/miniconda/lib/python3.6/site-packages/niworkflows/nipype/pipeline/engine/nodes.py", line 650, in _run_command
    result = self._interface.run()
  File "/usr/local/miniconda/lib/python3.6/site-packages/niworkflows/nipype/interfaces/base.py", line 1088, in run
    runtime = self._run_interface(runtime)
  File "/usr/local/miniconda/lib/python3.6/site-packages/niworkflows/common/report.py", line 55, in _run_interface
    ReportCapableInterface, self)._run_interface(runtime)
  File "/usr/local/miniconda/lib/python3.6/site-packages/niworkflows/nipype/algorithms/confounds.py", line 471, in _run_interface
    self.inputs.pre_filter, degree, self.inputs.high_pass_cutoff, TR)
  File "/usr/local/miniconda/lib/python3.6/site-packages/niworkflows/nipype/algorithms/confounds.py", line 1070, in compute_noise_components
    u, _, _ = linalg.svd(M, full_matrices=False)
  File "/usr/local/miniconda/lib/python3.6/site-packages/scipy/linalg/decomp_svd.py", line 119, in svd
[...]

PR #807, opened 2017-10-31
[ENH] Efforts towards keeping memory low

oesteban:
This PR reduces memory with two principal changes:

- [x] Improved and more granular memory annotations
- [x] Run confounds in native space:
  - [x] Added a new workflow that resamples the BOLD in native space (will be useful for #808)
  - [x] Adapted the confounds workflows to operate in native space: masks are converted into ROIs in the original T1w space (meaning, the erosions that Chris balanced are still happening at that resolution). A minor change is that the WM+CSF mask is now calculated from the TPMs (instead of calculating separate ROIs and then merging).
  - [x] Added the CSF ROI to the _confounds.tsv file.

For reviewing this PR, please have a detailed look into the new CompCor masks.


oesteban:
Edited the PR initial message.

fmriprep/interfaces/utils.py
```diff
@@ -87,49 +85,49 @@ def _run_interface(self, runtime):
...
+        data[data > 1] = 1.0
+        data[data < 0] = 0.0
 
-    ``in_file`` is resampled to match ``in_mask``, and the two concatenated
-    datasets are saved with the affine and header of ``ref_header``.
-    """
-    input_spec = ConcatROIsInputSpec
-    output_spec = ConcatROIsOutputSpec
+        out_file = fname_presuffix(first_fname, suffix='_tpmsum')
+        im.__class__(data, im.affine, im.header).to_filename(out_file)
```
effigies: I believe this is boolean? Might be worth making sure that it's saved as `np.uint8`.

fmriprep/workflows/bold/base.py
```diff
@@ -549,9 +562,60 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
                     ('outputnode.bold_mask', 'inputnode.bold_mask')]),
             ])
 
+    # Apply transforms in 1 shot
+    # Only use uncompressed output if AROMA is to be run
+    bold_bold_trans_wf = init_bold_preproc_trans_wf(
```
effigies: Why `bold_bold` on this and next workflow?

fmriprep/workflows/bold/confounds.py
```diff
@@ -71,20 +79,25 @@ def init_bold_confs_wf(bold_file_size_gb, use_aroma, ignore_aroma_err, metadata,
...
             SPM-formatted motion parameters file
         t1_mask
             Mask of the skull-stripped template image
         t1_tpms
             List of tissue probability maps in T1w space
-        bold_mask_t1
-            BOLD series mask in T1w space
+        t1_bold_xform
+            Affine matrix that maps the T1w space into alignment with
+            the native BOLD space
```
effigies: I think you're going to re-introduce issues with partial-FoV BOLD series if you're not masking with the BOLD mask after resampling the masks created in T1w space. (I think it's fine to use this transform. But I don't think it replaces masking.)

fmriprep/interfaces/utils.py
```diff
@@ -87,49 +85,49 @@ def _run_interface(self, runtime):
...
+            self._results['out_file'] = first_fname
+            return runtime
 
-class ConcatROIsOutputSpec(TraitedSpec):
-    out_file = File(exists=True, desc='output average file')
+        im = nb.load(first_fname)
+        data = im.get_data()
 
+        for idx in indexes[1:]:
+            data += nb.load(in_files[idx]).get_data()
```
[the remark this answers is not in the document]
oesteban: Not really, I was just avoiding the use of fsl.maths. This interface is just to add tissue probability maps which are small files by definition. Maybe I don't get where you want to get at...

fmriprep/workflows/bold/base.py
oesteban: well, its kind of bold-to-bold resampling. Since other are called bold_mni_ and bold_t1_ ...

fmriprep/workflows/bold/confounds.py
oesteban: Oh, I see. I'll make sure masks do not go off limits with partial-FoV in BOLD space.

oesteban:
I checked on the resulting _confounds.tsv files and compare to the former version. Everything is quite close to the originals, except for the vx-stdDVARS which are very different.

Since now we are working on native space, I'm inclined to think that the new values of the voxelwise standardized dvars are more trustworthy.

fmriprep/interfaces/utils.py
oesteban: Ok, I'm double checking. That should be a float. This interface works directly on the TPMs, and it should not change the original dtype. I think it is safe to leave it this way.

fmriprep/interfaces/utils.py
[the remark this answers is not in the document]
effigies: Just kind of wondering about the whole goal of this.

oesteban:
Confirmed this fixes #776.

fmriprep/workflows/bold/base.py
effigies: Okay, I'm going to need to reread, because I did not catch that this wasn't bold->T1w. So is this so that we have an output if it's `--func-only`?

oesteban: There are several targets with this PR:

1. Effectively, working on BOLD space will make it a lot easier to implement this functionality with `--func-only`. However, as it is now we still need segmentations from T1. Now it is just easier to replace them with other (e.g. coming from an EPI atlas).
2. Targets memory fingerprint, since the BOLD files we use to compute the confounds are as small as the original BOLD files. Resampled BOLD on T1 space were generally at least one order of magnitude bigger.
3. The idea of the new reportlet (std before and after resampling) we add some means to the reports to show if HMC worked well. You were right both on that the report I showed you mixes up all corrections (STC, HMC and SDC) but I'd expect that the largest contributor to differences in that plot is HMC.

fmriprep/workflows/bold/confounds.py
oesteban: Done

oesteban: Actually, I'm thinking this doesn't make any difference: we work on BOLD space (so it is still partial if it was partial in first place). Nothing will be considered if it is off FoV.

effigies: Yeah, that seems correct. I suppose I was thinking we were in `bold_space-T1w` still.

fmriprep/workflows/bold/base.py
oesteban: That's the thing: reduced FoV optimizes the tilt of the axial plane to minimize the number of slices you acquire, right.

If you bring that reduced FoV to the T1w, which is usually acquired to have the AC PC axis more or less aligned with the axial plane, then the projections of the brain mask to the planes are about the same in X and Y, but a lot larger in Z. This is because the axial plane of BOLD will be very tilted w.r.t. the axial plane of the T1w.

As a result: you have an image of the same resolution of the original BOLD, with about the same number of pixels in X and also in Y, but a lot more in Z. If the time-series is long, then the size of the image grows quickly.

Finally, an as I commented below w.r.t. the mask, the problems we were having with the limited FoV were derived from mapping those partial FoV to T1 and the poor coverage of the T1 brain mask. If you keep things in BOLD space, then you don't suffer from that problem.

PR #820, opened 2017-11-08
FIX: Normalize T1w image qform/sform matrices

### LINKED ISSUE #819: ANTs brain mask misaligned with FreeSurfer-conformed structurals
Resampling the ANTs-extracted brain mask into the FreeSurfer directory can fail pretty spectacularly. I believe this is happening because the input T1w has mismatched sform/qform affines, so ANTs skull-stripping produces outputs in the qform space, and FreeSurfer uses the sform space.

### `derivatives/fmriprep/sub-01/anat/sub-01_T1w_preproc.nii.gz`

preproc

### `derivatives/fmriprep/sub-01/anat/sub-01_T1w_brainmask.nii.gz`

brainmask

### `derivatives/freesurfer/sub-01/mri/T1.mgz`

t1

### `derivatives/freesurfer/sub-01/mri/brainmask.mgz`

brainmask2

effigies:
This PR normalizes header transforms to ensure that all software interprets the image space the same, particularly ANTs, which does not respect sform matrices.

The steps this takes are:

* Get best affine (nibabel selects from sform, qform and shape-derived)
* Push through quaternions to remove any shear (should have no effect, but just in case)
* Set qform and sform to normalized transform
* Set codes to 2 ("aligned to image" i.e. world space is not scanner or template coordinates)

Fixes #819.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,7 +2,9 @@ Next release
 ============
 
 * [FIX] Adopt new FreeSurfer license mechanism (#787)
+* [ENH] Output affine transforms from original T1w images to preprocessed anatomical (#726)
 * [FIX] Correct headers in AFNI-generated NIfTI files (#818)
+* [FIX] Normalize T1w image qform/sform matrices (#820)
 
 1.0.0-rc9 (2nd of November 2017)
 ================================
--- a/fmriprep/interfaces/__init__.py
+++ b/fmriprep/interfaces/__init__.py
@@ -6,7 +6,8 @@ from .bids import (
     ReadSidecarJSON, DerivativesDataSink, BIDSDataGrabber, BIDSFreeSurferDir, BIDSInfo
 )
 from .images import (
-    IntraModalMerge, InvertT1w, ValidateImage, TemplateDimensions, Conform, Reorient
+    IntraModalMerge, InvertT1w, ValidateImage, TemplateDimensions, Conform, Reorient,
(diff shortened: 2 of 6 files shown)


effigies:
@oesteban Just a note to look at the latest change. The earlier approach was wrong, since quaternions are pure rotations, and didn't include scale. I think the function I pulled out could be easily applied to the BOLD files, as well. If you think we should go ahead and do that within this PR, let me know.

oesteban:
Yeah, I think this PR is not too complicated yet so we can fix both at once.

effigies:
@oesteban Fixing the anatomical resolved all issues in that dataset. Synchronizing xforms in BOLD caused different issues, which I assume come from some non-obvious uses of unsynchronized BOLD files. Given that we're using the `c3d_affine_tool` to make sure the transforms work correctly for the time being, I'm inclined to leave this for another day.

Possibly the simplest thing would be to pass the original BOLD file to `Reorient()`. It would almost always make a new copy of the file, but it would also be more fool-proof than trying to pick clever times to sync the header.

oesteban:
Very reasonable, let's go ahead with this one.

Could you open issues for those problems derived from the BOLD synchronization of xforms?

effigies:
Are they issues, if we don't merge the changes that produce them? The short story is that the aCompCor masks are "centered", but the motion-corrected BOLD series is pushed out-of-center.

PR #825, opened 2017-11-10
[REF] Fine-grained logging (make FMRIPREP less shy)

oesteban:
Enables the `-v` flag for fine-grained tunning of FMRIPREP's quietness.
The default level is 25, only IMPORTANT messages or above (WARNING, ERROR)
will be written.
With one `-v` the level goes to 20 (INFO). Two (`-vv`) maps to the VERBOSE
new level. Finally, three (`-vvv`) is the DEBUG level.

### DIFF
--- a/wrapper/fmriprep_docker.py
+++ b/wrapper/fmriprep_docker.py
@@ -153,7 +153,7 @@ def merge_help(wrapper_help, target_help):
 
     # Make sure we're not clobbering options we don't mean to
     overlap = set(w_flags).intersection(t_flags)
-    expected_overlap = set(['h', 'v', 'w', 'output-grid-reference',
+    expected_overlap = set(['h', 'version', 'w', 'output-grid-reference',
                             'fs-license-file'])
     assert overlap == expected_overlap, "Clobbering options: {}".format(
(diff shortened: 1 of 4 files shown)


effigies:
Changing the `-v` flag is going to need to happen in `fmriprep-docker`, as well.

oesteban:
:+1: will do

wrapper/fmriprep_docker.py
```diff
@@ -214,7 +214,7 @@ def get_parser():
 
     parser.add_argument('-h', '--help', action='store_true',
                         help="show this help message and exit")
-    parser.add_argument('-v', '--version', action='store_true',
+    parser.add_argument('--version', action='store_true',
```
effigies: Also, line 156, in `expected_overlap`, `'v'` should become `'version'`.

oesteban:
@effigies could you double check that my last commit makes sense?

effigies:
It looks reasonable. Have you built an image and checked that `--help` and `--version` work? (Using the `-i` flag to specify the image should help with the weirdness of using it from the source directory.)

oesteban:
They work. Merging.

PR #828, opened 2017-11-13
[REF,ENH] Small refactor to the fieldmaps module and documentation

oesteban:
This PR adds two new functions that can be used to read fieldmap-correction related info from metadata.

Both are added with doctests that check they work in various use-cases.

Some changes have been proposed to the BIDS specification related to this issue. I'll submit an issue to INCF/BIDS-validator with some heuristics that can be used to check for sanity of echo spacing information in EPI scans.

Closes #149 

EDIT: after several trials on how to integrate this in the documentation, I found that the fieldmaps module is a bit disorganized. I tried to sort it out a bit, trying to make a more consistent documentation from it. Also, added the fieldmap module to the API of workflows (I think it disappeared at some point in time).

PR #830, opened 2017-11-14
DOC: Use non-hierarchical graph for bold_preproc_trans_wf

effigies:
Quick cleanup of the workflows page, noticed while reviewing #828.


effigies:
[Looks a little better](https://1006-78558509-gh.circle-artifacts.com/0/home/ubuntu/docs/workflows.html#pre-processed-bold-in-native-space)

Issue #833, opened 2017-11-15
Memory errors in sherlock too


oesteban:
The memory errors we were having in LS5 also happen in sherlock too, but in a far more complicated manner.

Since the overcommit policy of sherlock is set to `0`, the processes are allowed to allocate virtual memory beyond the physical memory limit, and more fmriprep instances will succeed without being killed. However, some of them still die more silently.

Checking on `dmesg` of one compute node of a hung fmriprep process, I find lots of:
```
[4696546.826293] Memory cgroup out of memory: Kill process 11189 (fmriprep) score 135 or sacrifice child
[4696839.164651] Memory cgroup out of memory: Kill process 11186 (fmriprep) score 142 or sacrifice child
[4764437.892109] Memory cgroup out of memory: Kill process 13718 (fmriprep) score 251 or sacrifice child
[4764803.104586] Memory cgroup out of memory: Kill process 13710 (fmriprep) score 191 or sacrifice child
[4764843.515150] Memory cgroup out of memory: Kill process 13724 (fmriprep) score 210 or sacrifice child
[4765217.241775] Memory cgroup out of memory: Kill process 13720 (fmriprep) score 199 or sacrifice child
[4765478.386302] Memory cgroup out of memory: Kill process 16036 (fmriprep) score 162 or sacrifice child
[4765626.640278] Memory cgroup out of memory: Kill process 13723 (fmriprep) score 148 or sacrifice child
[4765649.415484] Memory cgroup out of memory: Kill process 44419 (fmriprep) score 235 or sacrifice child
[4765881.317166] Memory cgroup out of memory: Kill process 7544 (fmriprep) score 133 or sacrifice child
[4766016.929903] Memory cgroup out of memory: Kill process 13721 (fmriprep) score 132 or sacrifice child
[4766182.712589] Memory cgroup out of memory: Kill process 13719 (fmriprep) score 154 or sacrifice child
[4766395.495706] Memory cgroup out of memory: Kill process 42172 (fmriprep) score 149 or sacrifice child
[4766530.188187] Memory cgroup out of memory: Kill process 13728 (fmriprep) score 146 or sacrifice child
[4766659.722944] Memory cgroup out of memory: Kill process 2832 (fmriprep) score 186 or sacrifice child
[4766805.647325] Memory cgroup out of memory: Kill process 13714 (fmriprep) score 165 or sacrifice child
[...]

PR #839, opened 2017-11-17
[FIX] Reduce image loads in local memory

effigies:
The goal of this PR is to systematically go through all nibabel image loads and image-data reads, and ensure that they do not happen in local memory. This is operating on the theory that there is a memory leak (or gc delay) in the "master" process, that is resulting in large `fork`ed processes, eventually causing crashes.

Minor cleanups while auditing.

Starting early so I can make comments as I find things that don't have obvious fixes.


effigies:
This code is run in the master process. There's no good way I can see around loading at least some data from this file.

https://github.com/poldracklab/fmriprep/blob/23f82eb0d08375105c907b8a97d22bf2aa612d98/fmriprep/workflows/bold/base.py#L839-L848

One possibility would be to wrap this in a subprocess, as nipype does, to shield the master process from its memory usage. That seems like a lot of boilerplate. Pushing a post-build/pre-run GC collection, to hopefully reduce the impact of calls like this.

effigies:
Audit complete. With the exception of the above function, I think all `nb.load` uses are now no longer being tagged with `run_without_submitting`, which I suspect is the biggest source of memory expansion during workflow runtime.

There could still be issues arising from nipype and niworkflows.

effigies:
Also, trying to pad my "lines deleted" stats.

deletions

oesteban:
Refs. #833

PR #842, opened 2017-11-20
[WIP] ENH: Conform offsets, relative to center-of-mass

### LINKED ISSUE #841: Merging T1ws failed
image

https://s3.amazonaws.com/openneuro.outputs/f85e0d83b3a5b45e3efd90854d6130a5/67629e33-bb62-41ff-8fbb-2068df24a514/fmriprep/sub-20.html

https://openneuro.org/datasets/ds000148/versions/00001?app=FMRIPREP&version=32&job=67629e33-bb62-41ff-8fbb-2068df24a514

effigies:
If someone wants to check my math, that'd be good. Going to run on ds114 subjects to exercise this code.

Closes #841.

### DIFF
--- a/fmriprep/interfaces/images.py
+++ b/fmriprep/interfaces/images.py
@@ -13,6 +13,7 @@ import os
 import numpy as np
 import nibabel as nb
 import nilearn.image as nli
+from scipy import ndimage
 
 from niworkflows.nipype import logging
 from niworkflows.nipype.utils.filemanip import fname_presuffix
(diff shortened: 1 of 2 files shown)


effigies:
This ended up being a little tricky, because the conformation has previously been pre-applied, relative to the original affine, while translation in RAS space is necessarily post-applied. That is, the new affine is

```
A' = T A R C
```

Where `T` is RAS translation, `A` is the original affine, `R` is the rotation induced by reorienting, and `C` is the rescaling induced by conformation. The fact that we were only doing pre-application allowed the earlier code to be a little unclear about exactly what was happening, so I've updated the comments to be very explicit about what each transform is doing.

effigies:
Changed back to WIP because attempting to `antsApplyTransform` the dtissue derivative back to the (second) original T1 space ended up misaligned. Have a suspicion that what we want is to construct transforms in RAS-RAS instead of XYZ-XYZ. I bet we're getting a double translation from the FSL format.

oesteban:
Can/should we transfer this to niworkflows?

oesteban:
paging @effigies (please see question above)

effigies:
It looks like the issue we were trying to fix (#841) was actually a difference of acquisition sequence, not an offset. I think we can just close this.

PR #847, opened 2017-11-22
[FIX] Copy sform over qform when qform is invalid

### LINKED ISSUE #845: [BUG] Negative quaternions break CopyXForm
DS000168/sub-08
DS000168/sub-27
DS000168/sub-30

/cc @effigies 
```
File "/usr/local/miniconda/lib/python3.6/site-packages/niworkflows/interfaces/utils.py", line 51, in _run_interface
message='CopyXForm (niworkflows v%s)' % __version__)
File "/usr/local/miniconda/lib/python3.6/site-packages/niworkflows/interfaces/utils.py", line 168, in _copyxform
qform, qform_code = orig.header.get_qform(coded=True)
File "/usr/local/miniconda/lib/python3.6/site-packages/nibabel/nifti1.py", line 917, in get_qform
quat = self.get_qform_quaternion()
File "/usr/local/miniconda/lib/python3.6/site-packages/nibabel/nifti1.py", line 892, in get_qform_quaternion
return fillpositive(bcd, self.quaternion_threshold)
File "/usr/local/miniconda/lib/python3.6/site-packages/nibabel/quaternions.py", line 100, in fillpositive
raise ValueError('w2 should be positive, but is %e' % w2)
ValueError: w2 should be positive, but is -3.845952e-07
```

oesteban:
Fixes #845

### DIFF
--- a/fmriprep/interfaces/images.py
+++ b/fmriprep/interfaces/images.py
@@ -360,13 +360,12 @@ class ValidateImage(SimpleInterface):
 
         # Check qform is valid
         valid_qform = True
+        old_qform_code = int(qform_code)
         try:
             img.get_qform()
         except ValueError:
             valid_qform = False
-            # if not valid, qform will be overwritten with sform
-            # if sform_code is 0, we want to fix the code
-            qform_code = sform_code
+            qform_code = nb.nifti1.xform_codes['unknown']
 
         # Check affine information is valid
         valid_codes = (qform_code, sform_code) != (0, 0)
@@ -397,16 +396,22 @@ class ValidateImage(SimpleInterface):
 """
         if not valid_qform:
             # Copy sform into qform
-            img.set_qform(img.get_sform(),
-                          int(img.header._structarr['sform_code']))
+            img.set_qform(img.get_sform(), nb.nifti1.xform_codes['aligned'])
+
+            extra_msg = ''
+            if old_qform_code == nb.nifti1.xform_codes['unknown']:
+                extra_msg = ('Additionally, please note that the original q-form code '
+                             'was 0 (unknown), which is potentially problematic.')
 
             snippet += """\
 <h3 class="elem-title">WARNING - Invalid q-form matrix</h3>
 <p class="elem-desc">Input file does not have a valid qform matrix.
-  To fix the q-form matrix, FMRIPREP copied the s-form matrix over.
+  To fix the q-form matrix, FMRIPREP copied the s-form matrix over, and
+  set 2 (aligned) as the q-form code. %s
   Analyses of this dataset MAY BE INVALID.
 </p>
-"""
+""" % extra_msg
+
 
         # Store new file and report
         img.to_filename(out_fname)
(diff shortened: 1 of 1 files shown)


effigies:
The logic here is a little weird. If the qform/sform codes are both 0, we want the old error. If the qform has negative quaternions, then we should have the new error. I don't think you can actually have both, and right now a valid qform with negative quaternions will display both errors.

fmriprep/interfaces/images.py
```diff
@@ -353,34 +354,64 @@ def _run_interface(self, runtime):
...
+  A left-right flip may have occurred.
+  Analyses of this dataset MAY BE INVALID.
+</p>
+"""
+        if not valid_qform:
+            # Copy sform into qform
+            img.set_qform(img.get_sform(),
+                          int(img.header._structarr['sform_code']))
+
+            snippet = """\
```
effigies: `+=`?

effigies:
I think this looks okay, but just to be explicit about the logic that I think we *should* have:

| valid quaternions | `qform_code > 0` | `sform_code > 0` | warnings |
|-|-|-|-|
| True | True | True | None |
| True | True | False | None |
| True | False | True | None |
| True | False | False | invalid codes |
| False | True | True | invalid qform |
| False | True | False | invalid codes + invalid qforms |
| False | False | True | None |
| False | False | False | invalid codes |

Do you agree?

oesteban:
Wouldn't it be...?

| valid quaternions | `qform_code > 0` | `sform_code > 0` | warnings |
|-|-|-|-|
| True | True | True | None |
| True | True | False | None |
| True | False | True | None |
| True | False | False | invalid codes |
| False | True | True | invalid qform |
| False | True | False | invalid codes + invalid qform |
| False | False | True | invalid qform |
| False | False | False | invalid codes + invalid qform |

We want to warn the user whenever we overwrite the qform. And we want to do that always if there is a negative quaternion. Right?

effigies:
My thought was we want to warn the user when the qform code says it's valid but the qform matrix can't be derived. Invalid quaternions don't seem to be a problem, if the qform code says there's no qform there.

Maybe here is where we should be doing our qform/sform syncrhonization.

1) If quaternions are bad, set qform code to zero, warning if non-zero
2) If qform and sform codes are zero, warn about invalid header
3) Set qform and sform to best affine (code 2 seems reasonable, but I could see an argument for 1)

The assumption then is that the output will be modified, so there isn't the danger of hitting invalid quaternions down the line, but the warning that quaternions are bad is only hit if the header indicates they should be valid.

WDYT?

oesteban:
I tried to capture your ideas in the last commit. One thing unclear to me is:

- As per 3), the modified s-form code would be 2 (aligned) in case both codes were 0?. If so, I'd need to update the sform update since we used to have code 1 (scanner).
- What if 1) quaternions are bad, should the qform code == sform code?. In that case (as the in previous bullet), 1 or 2?

oesteban:
It seemed weird to me having code 1 for the sform and 2 for the qform. I've temporarily merged this setting both to 1 (scanner) to run the new version on the failing datasets. I decided to keep code 1 since that is what we used to set for sform here.

Let me know if we need to revise this.

PR #851, opened 2017-11-27
[RFC] ENH: Always sync qforms, refactor error messaging

effigies:
Follow-up to #847, intended to clarify my opinions on this interface.

First, I think this is the easiest place to go ahead and sync qforms and sforms, unconditionally.

Second, I've refactored the warnings to follow this logic:

1) If qform/sform codes are bad, declare "WARNING - Invalid header"; if only qform is invalid, declare "WARNING - Invalid qform".
2) If the qform is invalid (and is not marked invalid), note it simply in the description.
3) If qform/sform are valid, but out of sync, make a note (without "WARNING") that they're mismatched.

Just as some after-the-fact commentary on #847: I think the warnings for invalid qforms are weird, in that a qform_code of 0 is treated as intensifying the problem with invalid qforms, rather than mitigating it. It's bad and marked as bad. That seems like some software did a reasonable thing, even if it's now something we have to clean up after.

Relatedly, I think the only time that we need to say analyses "MAY BE INVALID" is when there is no usable qform or sform. A good sform but bad qform is a totally reasonable situation to be in, if it weren't for *our choice* of using ANTs. It's thus our responsibility to handle that case, and make clear what we've done, but we know from the headers what the correct orientation of the image is.

I'm okay with keeping code 1 as the default. I think that makes sense when we use the default affine. However, if we sync the sform, we should use whatever code it has.

### DIFF
--- a/fmriprep/interfaces/images.py
+++ b/fmriprep/interfaces/images.py
@@ -370,47 +370,54 @@ class ValidateImage(SimpleInterface):
         # Check affine information is valid
         valid_codes = (qform_code, sform_code) != (0, 0)
 
+        # Matching affines
+        matching_affines = np.allclose(img.get_qform(), img.get_sform())
+
         # Both okay -> do nothing, empty report
-        if valid_qform and valid_codes:
+        if valid_qform and valid_codes and matching_affines:
             self._results['out_file'] = self.inputs.in_file
             open(out_report, 'w').close()
             self._results['out_report'] = out_report
             return runtime
 
-        snippet = ''
         out_fname = fname_presuffix(self.inputs.in_file, suffix='_valid', newpath=runtime.cwd)
         self._results['out_file'] = out_fname
 
+        warning = False
+        warning_txt = ''
+        description = ''
         # Fix sform if needed
         if not valid_codes:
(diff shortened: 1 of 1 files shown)


fmriprep/interfaces/images.py
```diff
@@ -370,47 +370,54 @@ def _run_interface(self, runtime):
...
-        if not valid_qform:
+        if not valid_qform and old_qform_code != 0:
+            if not warning:
+                warning = True
+                warning_txt = 'Invalid qform'
+            description += """\
+<p class="elem-desc">Input file does not have a valid qform matrix.</p>
+"""
+
+        if not (valid_qform and matching_affines):
```
oesteban: like before, if qform is not valid then it doesn't make much sense to see if it matches the sform (or, better said, you can't even check).

effigies: This is a short-circuit. We want to set the qform in the two cases: `not valid_qform` and `not matching_affines`.

oesteban: Yes, I just wanted to point out the fact that we need to set `matching_affines` in a different way for this switch to work as expected.

PR #856, opened 2017-11-29
[ENH] Use Nipype MultiProc even for sequential execution

### LINKED ISSUE #855: Use MultiProc even when --n_cpus=1
Remove Linear at all would fix memory issues in single-threaded execution.

oesteban:
After the experience of reducing the memory fingerprint of fmriprep, the Linear plugin is heavily affected by the memory explosion as it is implemented now. The easiest alternative is using MultiProc with `n_procs=1`.

Closes #855

### DIFF
--- a/fmriprep/cli/run.py
+++ b/fmriprep/cli/run.py
@@ -297,35 +297,35 @@ def build_workflow(opts, retval):
         logger.warning(msg)
 
     # Set up some instrumental utilities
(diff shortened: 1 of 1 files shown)


oesteban:
I'll take some risk here and merge without review, I'd like to release ASAP.

effigies:
Hi, sorry. Caught me around dinner/bedtime. This looks fine.

PR #863, opened 2017-11-30
MEM: Annotate t1_merge with approximate memory usage

effigies:
Ended up writing my own quick profiler, captured this memory profile, which seems to follow a 2N-1.5 rule, where N is the number of input images.

t1_merge_memory

This looks like an effect of registering each subsequent image to the first being embarrassingly parallel. If we hit more images than available threads, then I suspect the profile will end up looking a bit different, but I'm not sure it's worth trying to finesse this. If we get to a lot of input images, it will probably justifiably be the only thing running for a bit.

### DIFF
--- a/docs/workflows.rst
+++ b/docs/workflows.rst
@@ -66,7 +66,8 @@ T1w/T2w preprocessing
                               freesurfer=True,
                               longitudinal=False,
                               debug=False,
-                              hires=True)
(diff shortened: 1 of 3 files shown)


oesteban:
Ready?

effigies:
I'm happy with this if you are.

Issue #478, opened 2017-04-25
Reports are getting big again


chrisfilo:
A single report for Beast (with 5 runs and fieldmaps) can be 40Mb. Situation will get even worse in UH2 where there are 12 tasks across two sessions. Things to consider:

- Drop some visualisations (maybe we can make aCompCor and its visualisation optional?)
- Crop some plots (for example FAST)
- Split the report into multiple files (one per run + one for anat and fieldmaps)
- Try to improve compression (try [Guetzli](https://research.googleblog.com/2017/03/announcing-guetzli-new-open-source-jpeg.html) for example)
- Reduce number of tiles for some plots

oesteban:
From those 5 I would go for this (in this priority):
1. split reports in multiple files
2. remove aCompCor
3. reduce number of tiles for some plots (for example FAST, instead of cropping)

IanEisenberg:
I don't think we need one report per run, but maybe some would prefer that. Perhaps a flag indicating how you want the reports to be split? Per-run, per-session or just one report?

oesteban:
After considering this issue, I'm reaching to the following conclusions:

1. The size of the html is not a big deal per se. I find the most annoying that large html reports just crash the browser. So, maybe the right fight here is making the reports digestible by browsers, more than reducing the size. One idea would be this: https://github.com/jonathantneal/svg4everybody . We would have then the main html + one single, big svg, and this library loads the appropriate panel from the big file when you need to render it in the browser (meaning, only one svg loaded at a time).
2. That said, I think we can do some things without giving up on some of the reportlets:
  * Plot together: BOLD brainmask, aCompCor and tCompCor masks. Each contour with a different color. Having only one bitmap instead of 3 will definitely reduce sice.
  * Plot together: FSL FAST and T1w brain mask. Or better: FAST contours but rendered on the whole image (so we can see the skull).

WDYT?

chrisfilo:
The low hanging fruit of dealing with browser crashes is checking if disabling embedding of all the png and svg files (replacing it with relative links - in other words many files per report) solves the issue. My hunch is that it might help a lot.

oesteban:
A short investigation of this led me to think like that. Apparently, the browser does not render anything until it is shown in the viewport. The problem we might be hitting is the excessive number of elements hanging in the DOM. Any strategy to reduce them (e.g. writing isolated files) in theory will help.

[...]

PR #867, opened 2017-12-01
[ENH] Strip reportlets out of full report

oesteban:
A first stab at #478. Requires poldracklab/niworkflows#210 to be merged in first

The stripped svgs are stored in:

```
fmriprep/
 |- sub-01/
 |   |- anat/
 |   |- func/ 
 |   \- figures/
 |       \- *.svg  <------- HERE
 \- sub-01.html
```
we can move it if you think that is not right.


effigies:
Maybe `figures/` or `fig/`? Just to give a description besides the file type.

.circle/data/ds005_outputs.txt
```diff
@@ -45,6 +45,23 @@ ds005/out/fmriprep/sub-01/func/sub-01_task-mixedgamblestask_run-02_bold_space-MN
 ds005/out/fmriprep/sub-01/func/sub-01_task-mixedgamblestask_run-02_bold_space-MNI152NLin2009cAsym_preproc.nii.gz
 ds005/out/fmriprep/sub-01/func/sub-01_task-mixedgamblestask_run-02_bold_space-MNI152NLin2009cAsym_variant-smoothAROMAnonaggr_preproc.nii.gz
 ds005/out/fmriprep/sub-01.html
+ds005/out/fmriprep/sub-01/figures
+ds005/out/fmriprep/sub-01/figures/_scratch_reportlets_fmriprep_sub-01_anat_sub-01_T1w_reconall.svg
```
effigies: Can we drop the `_scratch_reportlets_fmriprep_`?

Also, if this is going to be a derivative, I think it makes sense to apply some naming conventions that can eventually be standardized. For instance, perhaps the type can be `_{static,anim}.svg` indicating whether it's a static or animated image. Here are some examples, but feel free to reject any that you don't like.

Structural:

* reconall.svg -> surfaceoutline_static.svg
* t1_2_mni.svg -> normalization_anim.svg
* t1_seg.svg -> segmentation_static.svg
* t1_skull_strip.svg -> skullstrip_static.svg

Functional:

* acompcor.svg -> acompcor_static.svg
* bbr.svg -> registration_anim.svg
* bold_mask.svg -> skullstrip_static.svg
* syn_sdc.svg -> synsdc_anim.svg
* tcompcor.svg -> tcompcor_static.svg

oesteban: I totally agree with all of this. Would it be fine with you if we leave this comment for the next iteration over #478 ? Just to make all these changes more iteratively, and risk less to breaking reports.

effigies: Missed this, sorry. Sure.

PR #872, opened 2017-12-04
[FIX] Combined ROI reportlets

oesteban:
This PR merges together:
  * [x] Anatomical - brain tissue segmentation + brain mask
  * [x] Functional - BOLD brain mask + aCompCor + rCompCorr.

One more step to reduce the report size (#478).

This PR also fixes the long prefixes added in [#867](https://github.com/poldracklab/fmriprep/pull/867#discussion_r154425984)

EDIT: To be merged after poldracklab/niworkflows#211 and correspondingly updating the niworkflows pin.


oesteban:
- https://450-53608443-gh.circle-artifacts.com/0/home/ubuntu/ds054/out/fmriprep/sub-100185.html
- https://450-53608443-gh.circle-artifacts.com/1/home/ubuntu/ds005/out/fmriprep/sub-01.html

fmriprep/workflows/anatomical.py
```diff
@@ -894,25 +900,21 @@ def init_anat_reports_wf(reportlets_dir, output_spaces,
...
-        name='ds_t1_seg_report', run_without_submitting=True)
-
     ds_t1_2_mni_report = pe.Node(
         DerivativesDataSink(base_directory=reportlets_dir, suffix='t1_2_mni'),
         name='ds_t1_2_mni_report', run_without_submitting=True)
 
-    ds_t1_skull_strip_report = pe.Node(
-        DerivativesDataSink(base_directory=reportlets_dir, suffix='t1_skull_strip'),
-        name='ds_t1_skull_strip_report', run_without_submitting=True)
+    ds_seg_brainmask = pe.Node(
```
effigies: For consistency and clarity, I'd prefer to keep `t1` and `report` in the name until/unless we decide to change all these nodes. Maybe `ds_t1_seg_mask_report` or similar?

fmriprep/workflows/anatomical.py
```diff
@@ -1073,3 +1073,22 @@ def init_anat_derivatives_wf(output_dir, output_spaces, template, freesurfer,
...
+    labels = nii.get_data()
+
+    out_files = []
+    for i in range(1, 4):
+        ldata = np.zeros_like(labels)
+        ldata[labels == i] = 1
+        out_files.append(op.abspath('label%d.nii.gz' % i))
+        nii.__class__(ldata, nii.affine, nii.header).to_filename(out_files[-1])
+
+    return out_files
```
effigies: If this splitting is necessary, and we're not using it elsewhere, why not have it internal to the reporting interface? You're already loading all of the ROIs in memory; is there an advantage to writing them to file and then reloading them in a separate interface?

oesteban: We could do without splitting if we leverage the `levels` input of `add_contours` from nilearn. However, for simplicity, I did not provide an input for that to the interface. We can do that in the next iteration - https://github.com/poldracklab/niworkflows/issues/212
I would not invest time on doing the splitting within the interface instead.

oesteban: Yep, with levels we would pass along the segmentation. No extraction of regions needed at all

PR #880, opened 2017-12-06
[FIX] Collision between Multi-T1w and --no-freesurfer

### LINKED ISSUE #879: --fs-license-file and --no-freesurfer are conflicting for multi-T1w datasets
Multiple-T1 datasets require `mri_robust_template` from FreeSurfer. However, when run with `--no-freesurfer` the `FS_LICENSE` env var is not set, and fmriprep crashes.

This issue requires:
  - `--fs-license-file` to become mandatory
  - `--no-freesurfer` will be deprecated and replaced with `--fs-no-reconall`

I guess these changes will affect the wrapper in some ways.

oesteban:
Fixes #879

### DIFF
--- a/fmriprep/cli/run.py
+++ b/fmriprep/cli/run.py
@@ -162,7 +162,8 @@ def get_parser():
 
     # FreeSurfer options
     g_fs = parser.add_argument_group('Specific options for FreeSurfer preprocessing')
-    g_fs.add_argument('--no-freesurfer', action='store_false', dest='freesurfer',
+    g_fs.add_argument('--fs-no-reconall', '--no-freesurfer',
+                      action='store_false', dest='run_reconall',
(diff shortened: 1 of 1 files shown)


effigies:
I agree we should replace `--no-freesurfer`, as it's no longer accurate, but given that the program can run without a FreeSurfer license (unless you have multiple T1s), it seems a little strong to make it a strict requirement.

It's pretty low effort to get a license, but I expect we're going to get a flurry of bug reports when people start using it because their old CLI invocations will start throwing errors.

oesteban:
Marriage with FreeSurfer comes with some downsides.

effigies:
Yeah. If it seems worth it, go for it. The alternative is just checking for multiple T1s during setup.

oesteban:
Again, this is looking good. I'm going to push for release.

Issue #878, opened 2017-12-06
Compression of reports in resampling workflows & T1w derivatives.


oesteban:
From https://github.com/poldracklab/fmriprep/pull/876#discussion_r155250612:

@effigies said:
> I suspect this is a holdover from `bold_mni_trans_wf`. The reasoning there would be that ICA-AROMA will be able to benefit from mmapping, but if we're not going to run ICA-AROMA then the output isn't used except as a derivative, so we might as well go ahead and compress.
>
> If I understand the confounds situation now, the T1w-resampled BOLD series are only used for derivatives, so we want:
>
>   * `bold_bold_trans_wf`: `use_compression=not low_mem`
>   * `bold_reg_wf`: `use_compression=False`
>   * `bold_mni_trans_wf`: `use_compression=not (low_mem and use_aroma)`
>
> Also, I've noticed that resampling to T1w is not contingent on T1w derivatives. If there's no T1w output then we should probably drop this, to save the time and space of creating those files.

oesteban:
Is this still an issue @effigies ?

effigies:
I don't even know... I'm honestly not sure how this pertains to reports. Let me have a look at the `Merge` nodes.

PR #882, opened 2017-12-06
[FIX] Setting use_compression on resampling wfs

oesteban:
Addresses #878 partially


oesteban:
I'll go ahead given the low risk of merging and the high interest of releasing today.

PR #883, opened 2017-12-06
[ENH] Add resource monitor flag

oesteban:


effigies:
Looks good. Want to bump the niworkflows version to 0.2.2? I think that's needed to get VMS stats from the profiler.

oesteban:
Yes, I was already working on that

PR #885, opened 2017-12-08
DOC: fix example in installation ants-nthreads -> omp-nthreads

mvdoc:
And allow me to use this PR to thank you guys for making this amazing piece of software :)


effigies:
Good catch. Thanks!

PR #876, opened 2017-12-05
 [ENH] Estimate motion parameters before STC

oesteban:
This PR focuses on estimating head-motion parameters before temporal filtering of the signal, as suggested in (Power, 2017) - see #849.

If slice-timing correction (STC) is done, then all processing is tailored to STC. Otherwise, the original data are used.

PR #891, opened 2017-12-14
[ENH] Handle first echo separately for meepi

emdupre:
The current integration of multi-echo processes all echos separately (unless T2* coregistration is selected). Ideally, the motion parameters should only be computed for the first echo and then applied across echos, to match what's done in ME-ICA.

This PR builds on #876, where head motion parameters were estimated before (and separately from) STC. Here, motion params should be estimated with only the first echo, but all echos should be fed into STC and subsequent processing.


oesteban:
Let me know when this is ready for review ;)

emdupre:
Thanks for the catch (and for being so on top of things)!

emdupre:
Ok, I think this is ready for review. I'd appreciate any feedback, particularly on the nipype workflow logic!

oesteban:
I'll try to allocate time for this review tomorrow. Sorry for the wait

emdupre:
Don't worry about the delay. I'm reverting this to WIP, as the outputs look off on one data set. Will update.

emdupre:
Ok, this is (finally) ready for review ! Sorry about the delay. This has underscored for me the need to get multi-echo into integrated testing, since my local test environment was substantially slower than expected.

This ended up being a bigger PR than I would have liked because I had to add in a new function from ME-ICA (`get_opt_comb`). Rather than skull-strip the T2* volume, the optimal combination of all supplied echos is now being calculated and used for skull-stripping.

This has been checked on `///openfmri/ds000210`, but will not run in its current form since I have not replaced the grouping code in `utils/bids.py`. Since this is such a large PR already, I thought I'd open another one to address that issue, but please let me know if you have another preferred workflow !

emdupre:
Checking in: is this waiting on adding mutli-echo tests, or is there anything else you need from me here ? Thanks ! 🙂

oesteban:
Sorry Elizabeth, it's been two hasty weeks an now I have some days off
ahead. But I have this as top 2 in my priority list (1 is another issue
related to fMRIprep, you'll hear about it soon).

Sorry for the wait

emdupre:
No rush !! I just wanted to make sure you weren't waiting on me. Thanks again !

oesteban:
Ok, I'm slowly coming back. Can I ask you to merge with current master?

fmriprep/interfaces/multiecho.py
```diff
@@ -12,25 +12,82 @@
...
+from niworkflows.nipype.utils.filemanip import (split_filename, fname_presuffix)
 from niworkflows.nipype.interfaces.base import (
     traits, TraitedSpec, File, InputMultiPath, SimpleInterface,
     BaseInterfaceInputSpec)
 
 LOGGER = logging.getLogger('interface')
 
 
+class FirstEchoInputSpec(BaseInterfaceInputSpec):
+    in_files = InputMultiPath(File(exists=True), mandatory=True,
```
oesteban: I'm not 100% this suggestion would work, but I'd love if you gave it a try:

```Python
in_files = InputMultiPath(File(exists=True), mandatory=True, minlen=2,
                          desc='multi-echo BOLD EPIs')
```

[In theory](https://github.com/enthought/traits/blob/080b722f72299acb63d539037107bd630f07ae35/traits/trait_types.py#L2292) List traits (as the InputMultiPath) allow for a `minlen` argument.

fmriprep/interfaces/multiecho.py
```diff
@@ -12,25 +12,82 @@
...
+                              desc='multi-echo BOLD EPIs')
+    ref_imgs = InputMultiPath(File(exists=True), mandatory=True,
```
oesteban: same here (`minlen`)

fmriprep/interfaces/multiecho.py
```diff
@@ -175,12 +241,81 @@ def define_t2s_map(echo_list, tes, last_emask, two_emask):
...
+        Masked array of shape (nx*ny*nz[, Ne[, nt]])
+    mask
+        Boolean array of shape (nx, ny, nz)
+
+    **Outputs**
+
+    ndarray
+        Array of shape (nx, ny, nz[, Ne[, nt]])
+    """
+    s = data.shape
```
oesteban:
```Python
new_s = tuple([np.prod(data.shape[:3])] + list(data.shape[3:]))
```

fmriprep/interfaces/reports.py
```diff
@@ -115,8 +115,13 @@ def _generate_segment(self):
 
         # Add list of tasks with number of runs
         bold_series = self.inputs.bold if isdefined(self.inputs.bold) else []
+
+        if any(isinstance(series, list) for series in bold_series):
```
oesteban: Isn't this equivalent to:
```Python
bold_series = [s[0] if isinstance(s, list) else s for s in bold_series]
```

fmriprep/workflows/base.py
```diff
@@ -353,10 +353,10 @@ def init_single_subject_wf(subject_id, task_id, name,
         raise Exception("No T1w images found for participant {}. "
                         "All workflows require T1w images.".format(subject_id))
 
-    if t2s_coreg and not layout.get_echos():
-        raise Exception("No multiecho BOLD images found for participant {}. "
-                        "T2* coregistration requires multiecho BOLD.").format(
-                            subject_id)
+    # if t2s_coreg and not layout.get_echos():
```
oesteban: Can we remove these commented out lines?

fmriprep/workflows/bold/base.py
```diff
@@ -220,8 +221,10 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
         ref_file = bold_file
     else:
         if isinstance(bold_file, list):  # if multi-echo data
```
oesteban:
```Python
multiecho = isinstance(bold_file, list)
ref_file = bold_file[0] if multiecho else bold_file
```

fmriprep/workflows/bold/base.py
```diff
@@ -243,15 +246,19 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
             'magnitude2': 'sub-03/ses-2/fmap/sub-03_ses-2_run-1_magnitude2.nii.gz'
         }]
         run_stc = True
+        multiecho = False
         bold_pe = 'j'
     else:
-        if isinstance(bold_file, list):  # For multiecho data, grab TEs
+        if multiecho is True:  # For multiecho data, grab TEs
```
oesteban: could multiecho not be a boolean? If it cannot:

```
if multiecho:
   ...
```

fmriprep/workflows/bold/base.py
```diff
@@ -243,15 +246,19 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
...
             tes = [layout.get_metadata(echo)['EchoTime'] for echo in bold_file]
+            fmap_file = bold_file[0]
+        else:
+            fmap_file = bold_file
         # Since all other metadata is constant
         metadata = layout.get_metadata(ref_file)
 
         # Find fieldmaps. Options: (phase1|phase2|phasediff|epi|fieldmap)
-        fmaps = layout.get_fieldmap(bold_file, return_list=True) \
+        fmaps = layout.get_fieldmap(fmap_file, return_list=True) \
```
oesteban: `ref_file`? and remove lines 254-256?

fmriprep/workflows/bold/base.py
```diff
@@ -507,21 +542,23 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
                 ('outputnode.itk_t1_to_bold', 'inputnode.in_xfm')]),
         ])
     elif not use_syn:
+        LOGGER.warn('No fieldmaps found or they were ignored, building base workflow '
+                    'for dataset %s.', ref_file)
+        summary.inputs.distortion_correction = 'None'
+
         if t2s_coreg is True:
```
oesteban: type check necessary here? (why not just `if t2s_coreg:`)

fmriprep/workflows/bold/base.py
emdupre: I could just keep using `isinstance(bold_file, list)` as a proxy, but since it's used four times I thought a boolean would be easier to interpret. If you'd prefer a non-boolean, though, I can stick to the `isinstance` solution !

fmriprep/workflows/base.py
emdupre: I had an exception there, but I thought better of it. I still think something should be provided to the user, though, if they request `--t2s_coreg` with non-ME data. What do you think of a warning?

fmriprep/interfaces/multiecho.py
```diff
@@ -12,25 +12,82 @@
...
+
+
+class FirstEchoOutputSpec(TraitedSpec):
+    first_image = File(exists=True,
+                       desc='BOLD EPI series for the first echo')
+    first_ref_image = File(exists=True, desc='generated reference image for '
+                                             'the first echo')
+
+
+class FirstEcho(SimpleInterface):
```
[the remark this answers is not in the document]
emdupre: I added something here, but I'm not sure this is exactly what you had in mind !

emdupre:
I went ahead and updated this to start address your review comments ! I added in #921 to be able to run locally, but can remove if you're not happy with the code I've sketched out there.

fmriprep/interfaces/multiecho.py
```diff
@@ -8,44 +8,157 @@
...
+    =======
+    >>> from fmriprep.interfaces import multiecho
+    >>> first_echo = multiecho.FirstEcho()
+    >>> first_echo.inputs.in_files = ['sub-01_run-01_echo-1_bold.nii.gz', \
+                                      'sub-01_run-01_echo-2_bold.nii.gz', \
+                                      'sub-01_run-01_echo-3_bold.nii.gz']
+    >>> first_echo.inputs.ref_imgs = ['sub-01_run-01_echo-1_bold.nii.gz', \
+                                      'sub-01_run-01_echo-2_bold.nii.gz', \
+                                      'sub-01_run-01_echo-3_bold.nii.gz']
+    >>> res = first_echo.run() # doctest: +SKIP
```
oesteban: Since this is only checking names, you can do:

```Python
>>> res = first_echo.run()  # removing the +SKIP flag
>>> res.outputs.first_image
'sub-01_run-01_echo-1_bold.nii.gz'
>>> res.outputs.first_ref_image
'sub-01_run-01_echo-1_bold.nii.gz'
```

oesteban:
Okay, this is almost there! Only two tests failed:

ds005 - you need to merge upstream/master again to update the outputs.
ds000210 - you need to add an "outputs file" [here](https://github.com/poldracklab/fmriprep/tree/260872273a1f4ef02de2cae20dd7d6948b531c4b/.circleci). For you, the contents of this output file would [were generated in your latest build](https://1777-53175327-gh.circle-artifacts.com/0/tmp/ds000210/test/outputs.out)

oesteban:
Just to clarify a bit: fmriprep checks that the expected outputs exist in the output directory. These files just help us developers to keep track of the generated files, and make sure none is missing or extra files have appeared without explicit intention.

emdupre:
Thanks for linking those artifacts ! I had the file, but the listed files were from my local build and so didn't match those generated with some of the updates incorporated here. I'll merge master now and patch the build conflicts. 

To confirm, do you have thoughts on issuing a warning for users who request `--t2s-coreg` with non-multi-echo data?

fmriprep/workflows/base.py
oesteban: A warning would be great!

fmriprep/workflows/bold/base.py
```diff
@@ -352,17 +352,23 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
...
-                                       joinsource='inputnode',
-                                       joinfield='echo_files',
-                                       name='join_split_echos')
-
-        # create a T2* map
-        bold_t2s_wf = init_bold_t2s_wf(echo_times=tes,
-                                       name='bold_t2s_wf',
-                                       mem_gb=mem_gb['filesize'],
-                                       omp_nthreads=omp_nthreads)
+        if not multiecho:
```
oesteban: Indentation can be reduced by one level:
```Python
if t2s_coreg and not multiecho:
    LOGGER.warning(...)
    t2s_coreg = False

if t2s_coreg:
    join_split_echos = ...
```
Please also note the `LOGGER.warning` instead of `LOGGER.warn` that is deprecated.

fmriprep/interfaces/reports.py
```diff
@@ -71,7 +71,7 @@ class SubjectSummaryInputSpec(BaseInterfaceInputSpec):
     t2w = InputMultiPath(File(exists=True), desc='T2w structural images')
     subjects_dir = Directory(desc='FreeSurfer subjects directory')
     subject_id = Str(desc='Subject ID')
-    bold = InputMultiPath(File(exists=True), desc='BOLD functional series')
+    bold = traits.List(desc='BOLD functional series')
```
oesteban: Why is this change necessary?

fmriprep/utils/bids.py
```diff
@@ -171,21 +171,21 @@ def collect_data(dataset, participant_label, task=None):
...
+            return x
+        echo = re.search("_echo-\\d*", x).group(0)
+        return x.replace(echo, "_echo-?")
+
+    if subj_data["bold"] is not []:
+        bold_sess = subj_data["bold"]
+
+        if any(['_echo-' in bold for bold in bold_sess]):
+            ses_uids = [list(bold) for _, bold in groupby(bold_sess, key=_grp_echos)]
+            ses_uids = list(map(lambda x: x[0] if len(x) == 1 else x, ses_uids))
```
oesteban: Can we replace the lambda with a list comprehension here?

fmriprep/workflows/bold/base.py
```diff
@@ -518,21 +551,23 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
...
+                    ('echo_files', 'inputnode.echo_split')]),
                 (bold_hmc_wf, bold_t2s_wf, [
-                    ('outputnode.xforms', 'inputnode.xforms')]),
+                    ('outputnode.xforms', 'inputnode.hmc_xforms')]),
                 (bold_t2s_wf, bold_reg_wf, [
                     ('outputnode.t2s_map', 'inputnode.ref_bold_brain'),
-                    ('outputnode.t2s_mask', 'inputnode.ref_bold_mask')])
-            ])
+                    ('outputnode.oc_mask', 'inputnode.ref_bold_mask')])
+                ])
```
oesteban: remove 4 spaces

oesteban:
After looking at the report, I have a couple of thoughts:

- We should probably generate only one EPI-to-T1 plot, right? If I understand the PR correctly, this is generating a T2* map to perform the co-registration. Is that correct? That way, we would also get to visualize the T2* map in the reports so the users can assess it looks alright.

- ROIs plot:
  1. the aCompCor mask (pink) is a lot smaller it usually is. That could indicate a problem on the transformation of the segmentation from T1 (and therefore because the downsampled data) or that the EPI mask is very small.
  2. Can you think of a way of calculating both masks only once? These masks, like HMC and co-registration, should be fundamentally the same for all echoes.

fmriprep/interfaces/reports.py
emdupre: Because the `bold_list` it's being passed is a list that can contain sub-lists, requiring `File(exists=True)` was not working for the sub-lists. Unless you can think of another way to check this!

emdupre:
Thanks for reviewing this ! To answer your questions:

- Yes, this is generating a T2* map and using it in the coregistration-- the generated T2* map is the only "EPI" shown for all echos in the EPI-T1 coregistration plots. So yes, having only one (non-redundant) plot would be ideal !

- A non-downsampled report is available [here](https://www.dropbox.com/sh/ip6xpd6oc5awnbm/AADp92KIKev9h6Yb8fIZsYu7a?dl=0). The aCompCor mask in a non-downsampled dataset looks normal to my viewing, but obviously you have much more experience with these ! Would love to know your thoughts. 

- Re: calculating both masks once, I hadn't considered that, but you're right that since they're coming from the anatomical scan they'll still only be calculated once. I'll look through now to see how we could reduce the number of times they're created.

emdupre:
To follow-up on the last point, though, tCompCor _could_ differ across echos. Would we then be splitting those reportlets ?

oesteban:
Sorry, I needed some time to maturate my answer:

- Do you plan to reduce the multiplicity of EPI-T1 registration plots? I'd say that'd be part of this PR. Let me know if you need any pointers 👍 .
- Your high-resolution report looks good indeed! Thanks!
- Regarding the t/aCompCor: is CompCor necessary or even applicable if you plan to use ME-ICA? Maybe we are splitting hairs here. That said, I think we can easily merge tCompCor masks calculating the union of all echoes, or maybe calculate it only using the echo closest to 30ms (the typical TE of a single-echo EPI).

In general, I think we should fine tune the contrast range of plots using the T2*, since only the CSF is really visible.

emdupre:
Re: reducing EPI-T1 plots; I'll look into it tomorrow and get back to you, thanks !

For CompCor, it isn't really applicable with ME-ICA. However, this integration with fmriprep would still allow users to consider the preprocessed echos individually, so I'm happy to support CompCor for now if you'd prefer. If so, I'd advocate for calculating tCompCor only from the middle echo, since that's closest to where the measure was validated.

emdupre:
Ok, the number of plots is down to one when using T2* maps for coregistration ! Any pointers on where the contrast in the plot is generated ?

fmriprep/workflows/bold/base.py
```diff
@@ -221,7 +221,7 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
         ref_file = bold_file
     else:
         multiecho = isinstance(bold_file, list)
-        ref_file = bold_file[0] if multiecho else bold_file
+        ref_file = sorted(bold_file)[0] if multiecho else bold_file
```
oesteban: Good catch!

fmriprep/workflows/bold/base.py
```diff
@@ -352,23 +358,31 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
...
-                                           joinfield='echo_files',
-                                           name='join_split_echos')
-
-            # create a T2* map
-            bold_t2s_wf = init_bold_t2s_wf(echo_times=tes,
-                                           name='bold_t2s_wf',
-                                           mem_gb=mem_gb['filesize'],
-                                           omp_nthreads=omp_nthreads)
+        # use a joinNode to gather all preprocessed echos
+        join_split_echos = pe.JoinNode(niu.IdentityInterface(fields=['echo_files']),
```
oesteban: Can we move the instantiation of these interfaces to after line 572 (right after the second `if t2s_coreg:`) so we remove this first branching step?

emdupre:
I think I'd prefer to open another issue, especially since I'm also trying to understand exactly why we're no longer able to use BBR for the T2* image&mdash; since the T2* map is really emphasizing the boundaries, and should be best with BBR ! WDYT ?

oesteban:
Sounds good to me. Let me have a final review to make sure everything looks good?

fmriprep/workflows/bold/base.py
```diff
@@ -935,6 +1009,8 @@ def init_func_derivatives_wf(output_dir, output_spaces, template, freesurfer,
 
 def _get_series_len(bold_fname):
     from niworkflows.interfaces.registration import _get_vols_to_discard
+    if isinstance(bold_fname, list):  # Multi-echo data
+        bold_fname = bold_fname[0]
```
oesteban: `sorted(bold_fname)[0]` perhaps? is it relevant the order at all here?

fmriprep/workflows/bold/base.py
```diff
@@ -791,11 +857,19 @@ def _bold_reg_suffix(fallback, freesurfer):
         (inputnode, ds_bold_rois_report, [('source_file', 'source_file'),
                                           ('bold_rois_report', 'in_file')]),
         (inputnode, ds_bold_reg_report, [
-            ('source_file', 'source_file'),
             ('bold_reg_report', 'in_file'),
             (('bold_reg_fallback', _bold_reg_suffix, freesurfer), 'suffix')]),
     ])
 
+    if t2s_coreg:
+        workflow.connect([
```
oesteban: maybe worth collapsing this if..else into a one liner with the main connect block we had previously:

```Python
workflow.connect([
    # ... connections
    (inputnode, ds_bold_reg_report, [
        ('first_echo' if t2s_coreg else 'source_file', 'source_file')]),
    # ... connections
])
```

fmriprep/workflows/bold/base.py
```diff
@@ -279,11 +278,23 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
                 't1_2_fsnative_forward_transform', 't1_2_fsnative_reverse_transform']),
         name='inputnode')
 
-    if isinstance(bold_file, list):
-        inputnode.inputs.iterables = ("bold_file", bold_file)
+    if multiecho:  # if multi-echo data
+        # iterate most preprocessing steps over each echo
+        inputnode.iterables = ('bold_file', bold_file)
+
+        me_first_echo = pe.JoinNode(interface=FirstEcho(),
```
oesteban: I think this would be more readable if the interface instantiation happens right before it is connected (~line 460)

fmriprep/workflows/bold/base.py
```diff
@@ -279,11 +278,23 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
                 't1_2_fsnative_forward_transform', 't1_2_fsnative_reverse_transform']),
         name='inputnode')
 
-    if isinstance(bold_file, list):
-        inputnode.inputs.iterables = ("bold_file", bold_file)
+    if multiecho:  # if multi-echo data
+        # iterate most preprocessing steps over each echo
+        inputnode.iterables = ('bold_file', bold_file)
```
oesteban: Could you check if we can completely remove this if .. else?:

1. Check that one can set `inputnode.inputs.bold_file = bold_file` for all workflows, and then
2. Check that setting `inputnode.iterables = ('bold_file', bold_file)` shadows the setting in 1).

I don't think we have used that approach ever, so I don't know how nipype will understand that. But, if it works, then the iterables could be set within the multiecho block in line 460.

fmriprep/workflows/bold/base.py
emdupre: The order wouldn't be relevant here, as all echo time series should have the same number of volumes !

fmriprep/workflows/bold/base.py
emdupre: I didn't know you could do that ! 🌟

fmriprep/workflows/bold/base.py
emdupre: I would never have thought of that, but it seems to have worked ! 🎉

fmriprep/workflows/bold/base.py
```diff
@@ -453,6 +445,30 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
         (summary, func_reports_wf, [('out_report', 'inputnode.summary_report')]),
     ])
 
+    # if multiecho data, select middle echo for hmc correction
```
oesteban: This is something I have overlooked so far. Talking today with Chris I learned that the order of echoes (`echo-[1-3]`) may not align with the acquisition order. Therefore, selecting the first echo does not ensure that you are using a particular echo time.

Additionally, the comment says middle echo but the interface selects the first. Am I correct?

(I think this is the very last comment from me)

emdupre:
That code comment should read "first echo" ! Thanks for the catch.

It's news to me, though, that the echo ordering is allowed to diverge from
acquisition time in BIDS standard. I find that rather surprising. Any
insight as to why @chrisfilo? I'm struggling to think of a situation where
that would aid interpretation, if "echo-2" has the first echo time. Or do
you mean that they could be labeled as e.g. "echo-first"?

Elizabeth

chrisfilo:
It was a rather nitpicky comment on my side. There is nothing in the spec saying that `echo-1` corresponds to shortest echo rather than the longest. Even though the opposite is unlikely might be better to pick the appropriate echo based on the JSON metadata (easy with `get_metadata` function from pybids).

But this is just a tiny detail that should be left to another PR. I am way to excited to see MEICA integrated into FMRIPREP to delay the current PR any longer :)

oesteban:
@emdupre could you update the comment to say the "first echo" making sure you use the `[skip ci]` tag so I merge this now?

oesteban:
I said it before, but I feel the need to repeat it: this is astonishing work. For some reason the Circle tests were triggered but I'll merge now anyways.

chrisfilo:
+1000!

Best,
Chris

PR #892, opened 2017-12-18
[FIX] Check for FS license only if run_reconall

emdupre:
Patches run.py so FS license is checked only if running reconall

### DIFF
--- a/fmriprep/cli/run.py
+++ b/fmriprep/cli/run.py
@@ -212,18 +212,19 @@ def main():
     opts = get_parser().parse_args()
 
     # FreeSurfer license
-    default_license = op.join(os.getenv('FREESURFER_HOME', ''), 'license.txt')
-    # Precedence: --fs-license-file, $FS_LICENSE, default_license
-    license_file = opts.fs_license_file or os.getenv('FS_LICENSE', default_license)
-    if not os.path.exists(license_file):
-        raise RuntimeError(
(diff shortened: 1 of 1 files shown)


effigies:
This is basically a reversion of #880. The problem with this proposal is that, in the case of multiple T1w input images or pepolar fieldmaps, we use FreeSurfer's `mri_robust_template`, which also requires the license. While it's possible to run fmriprep without FreeSurfer tools, there are now three conditions in which it is required. Rather than maintain a series of checks in `run.py`, which may grow if we find other situations where FreeSurfer tools are the best choice, we decided that it was a reasonably low cost to ask people to get a FreeSurfer license. 

I'm going to close this for now, but if you feel strongly and want to code up the checks (perhaps a `_check_freesurfer` that takes a `reconall` boolean and a BIDSLayout to check on the inputs), we can reconsider this decision.

emdupre:
Thanks for pointing that out! I hadn't seen #880. I understand the rational, now, but I think the documentation should then be updated to indicate that Freesurfer is required to run `fmriprep`.

As it stands, it suggests that a license file is only required if running `recon-all`, particularly [here](http://fmriprep.readthedocs.io/en/latest/usage.html). I'll open a new issue to discuss either updating the documentation / options or adding checks.

effigies:
Great, thanks!

Issue #893, opened 2017-12-18
Update docs for freesurfer as a strict requirement


emdupre:
I opened #892 due to confusion about the requirement of a Freesurfer license to run `fmriprep`, as discussed in #880. 

Is the current preference to make Freesurfer a strict requirement, and update associated documentation? Particularly the [usage notes](http://fmriprep.readthedocs.io/en/latest/usage.html), [installation guide](http://fmriprep.readthedocs.io/en/latest/installation.html) and `--no-freesurfer` option. Alternatively, should Freesurfer remain a requirement for only some options, in which case we would implement a series of checks in `run.py`?

Personally I'd lean towards the checks, but I understand concerns that they could grow over time.

oesteban:
Hi @emdupre, in the latest version of FMRIPREP we made FreeSurfer a mandatory dependency. The issue is that, even using the `--no-freesurfer` option, when more than one T1 are found then FreeSurfer is used regardless the option.

In that situation, we had the option of writing an alternative workflow for merging the T1s (without FS) or making FS mandatory. We took the easy way and turned the `--no-freesurfer` option into `--no-reconall` to better reflect what is exactly being disabled.

We probably should move that into the `--disable` option, but for now, it is mandatory.

However, nothing of this is final: considering your opinion we will try to identify the situation where we don't want to use FS but we find several T1s (and raise error in that case). In the mean time, I agree the documentation is currently very unclear regarding this.

emdupre:
I'm happy to update the documentation, if that's the preferred path forward ! I just think the present situation creates unnecessary confusion (at least it did for me :)

oesteban:
I totally agree, right now it is very confusing. If you update the documentation it will be very very much appreciated!

effigies:
I agree that moving to `--ignore surface` or something like that would be good. I'll review #894.

PR #894, opened 2017-12-19
[DOC] Update docs re: freesurfer

emdupre:
In response to #893, updates documentation to reflect that Freesurfer is a required dependency. Also removes the command line option ``--no-freesurfer`` to minimize ambiguity. Internally replaces most instances of the `freesurfer` variable with `reconall` for code clarity.


docs/installation.rst
```diff
@@ -130,7 +127,8 @@ Manually Prepared Environment
 
 Make sure all of fmriprep's `External Dependencies`_ are installed.
 These tools must be installed and their binaries available in the
-system's ``$PATH``.
+system's ``$PATH``. In particular, FreeSurfer requires a license
```
effigies: Please start new sentences on new lines. (Particularly in larger paragraphs, this makes it easier to see changes in the diffs restricted to the relevant sentence.)

docs/installation.rst
```diff
@@ -146,11 +144,10 @@ If you have your data on hand, you are ready to run fmriprep: ::
 The FreeSurfer license
 ======================
 
-FMRIPREP will run FreeSurfer unless the ``--no-freesurfer`` command-line argument is
-provided. Therefore, make sure there is a valid FreeSurfer available to FMRIPREP
-or opt-out otherwise.
+FMRIPREP will requires Freesurfer to run. Therefore, make sure there is a
+valid FreeSurfer available to FMRIPREP.
```
effigies: I'd simplify this:

```
FMRIPREP uses FreeSurfer tools, which require a license to run.
```

docs/installation.rst
```diff
@@ -146,11 +144,10 @@ If you have your data on hand, you are ready to run fmriprep: ::
...
 
-Getting a FreeSurfer license is free, register a new key at
+Getting a FreeSurfer license is free, simply register a new key at
```
effigies: This whole sentence reads a little awkwardly to me. Maybe:

```
To obtain a FreeSurfer license, simply register for free at
```

Or similar.

docs/workflows.rst
```diff
@@ -25,7 +25,7 @@ is presented below:
                                 longitudinal=False,
                                 t2s_coreg=False,
                                 omp_nthreads=1,
-                                freesurfer=True,
+                                reconall=True,
```
effigies: I would leave these be for now. (I wouldn't use `reconall` anyway, since turning off `freesurfer` also turns off bbregister and surface sampling). Once we decide the final form of this option will be time enough to change parameter names. (Users don't see this, anyway.)

fmriprep/cli/run.py
```diff
@@ -168,9 +168,9 @@ def get_parser():
 
     # FreeSurfer options
     g_fs = parser.add_argument_group('Specific options for FreeSurfer preprocessing')
-    g_fs.add_argument('--fs-no-reconall', '--no-freesurfer',
```
effigies: I'd keep this around for the time being, so that people can keep using their command line options. We can take it out in a version or two, once the docs have been changed for a bit.

@oesteban feel free to override me on this.

fmriprep/cli/run.py
```diff
@@ -168,9 +168,9 @@ def get_parser():
...
+    g_fs.add_argument('--fs-no-reconall',
                       action='store_false', dest='run_reconall',
-                      help='disable FreeSurfer preprocessing')
+                      help='disable FreeSurfer recon-all preprocessing')
```
effigies: Can you make this "surface" instead of "recon-all"? As noted above, it also disables bbregister and mri_vol2surf.

fmriprep/workflows/base.py
```diff
@@ -33,15 +33,15 @@
...
+                     reconall, output_spaces, template, medial_surface_nan, hires,
                      use_bbr, bold2t1w_dof, fmap_bspline, fmap_demean, use_syn, force_syn,
                      use_aroma, ignore_aroma_err, output_grid_ref):
     """
     This workflow organizes the execution of FMRIPREP, with a sub-workflow for
     each subject.
 
-    If FreeSurfer is to be used, a FreeSurfer derivatives folder is created and
-    populated with any needed template subjects.
+    If FreeSurfer's recon-all is to be used, a FreeSurfer derivatives folder is
```
effigies: "used" -> "run"

fmriprep/cli/run.py
emdupre: What do you think about adding something like this in the description?
```
Note : `--no-freesurfer` is deprecated and will be removed in 1.2.
Use `--fs-no-reconall` instead.
```

effigies: That seems reasonable to me.

effigies:
Great, thanks!

Issue #907, opened 2017-12-19
Too conservative BOLD skullstripping inds000116


chrisfilo:
![image](https://user-images.githubusercontent.com/238759/34183079-d5a32386-e4cd-11e7-8ab6-e943ddcbd7d6.png)

ds000116/reportlets/fmriprep/sub-12/func/sub-12_task-visualoddballwithbuttonresponsetotargetstimuli_run-02

Could be related to #905 #904 #902

Issue #905, opened 2017-12-19
Too conservative BOLD skullstripping in ds000133


chrisfilo:
![image](https://user-images.githubusercontent.com/238759/34182978-52465e7c-e4cd-11e7-8f79-d4999cd44f50.png)

ds000133 sub-06_ses-post_task-rest_run-01

Could be related to #904 and #902

Issue #895, opened 2017-12-19
Poor BOLD skulls tripping in ds000323


chrisfilo:
![image](https://user-images.githubusercontent.com/238759/34181452-09e4c0ac-e4c7-11e7-8b7f-ce54a3ed773c.png)

sub-02 all runs

PR #911, opened 2017-12-22
[FIX] Robuster EPI mask

### LINKED ISSUE #902: Bad BOLD skullstripping in ds000157
image

ds000157 sub-23_task-passiveimageviewing

oesteban:
This PR:

- [x] Replaces the default Otsu mask of `N4BiasFieldCorrection` by setting a mask from nilearn
- [x] Runs `N4BiasFieldCorrection` to only one thread, as parallelization is not really necessary here and results are not run-to-run reproducible when multithreaded.

Closes #902, #905, #907. Ref. #895.

### DIFF
--- a/docs/workflows.rst
+++ b/docs/workflows.rst
@@ -232,12 +232,14 @@ BOLD reference image estimation
 
 This workflow estimates a reference image for a
 :abbr:`BOLD (blood-oxygen level-dependent)` series.
-If T1-saturation effects ("dummy scans" or non-steady state volumes) are
-detected they are used as reference due to their superior tissue contrast.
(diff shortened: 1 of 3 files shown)


oesteban:
I just confirmed that the mask issues of #895 will not be fixed by this PR. It improved SDC-SyN though. Close #905, close #907, and #902.

Check out the reports -> https://drive.google.com/file/d/1V9P0Fc-SKJQoTZZ3GP_3yX9_x4PS2Cbc/view?usp=sharing

For very limited FoV,  possible heuristic would be: 1) detecting limited FoV (i.e. size_z < some number) and 2) not using BET nor 3dAutomask, just nilearn + dilation.



WDYT?

PR #912, opened 2017-12-27
[FIX] Refine antsBrainExtraction if recon-all is run

### LINKED ISSUE #431: Refine ANTs brain mask with FreeSurfer

oesteban:
Use the `aseg.mgz` output of FreeSurfer to derive a refined mask to replace the (generally less conservative) mask calculated with `antsBrainExtraction.sh`.

Closes #431

### DIFF
--- a/fmriprep/interfaces/freesurfer.py
+++ b/fmriprep/interfaces/freesurfer.py
@@ -337,7 +337,7 @@ def grow_mask(anat, aseg, ants_segs, ww=7, zval=2.0, bw=4):
     indices = np.argwhere(newrefmask > 0)
     for pixel in indices:
         # When ATROPOS identified the pixel as GM, set and carry on
-        if ants_segs[tuple(pixel)] == 3:
+        if ants_segs[tuple(pixel)] == 2:
             refined[tuple(pixel)] = 1
             continue
 
--- a/fmriprep/interfaces/freesurfer.py
+++ b/fmriprep/interfaces/freesurfer.py
@@ -228,6 +228,12 @@ class RefineBrainMaskOutputSpec(TraitedSpec):
 
 
 class RefineBrainMask(SimpleInterface):
+    """
+    Refine the brain mask implicit in the ``aseg.mgz``
+    file to include possibly missing gray-matter voxels
+    and deep, wide sulci.
+    """
+
     input_spec = RefineBrainMaskInputSpec
     output_spec = RefineBrainMaskOutputSpec
 
--- a/fmriprep/interfaces/freesurfer.py
+++ b/fmriprep/interfaces/freesurfer.py
@@ -219,6 +219,8 @@ class RefineBrainMaskInputSpec(BaseInterfaceInputSpec):
                    desc='input anatomical reference (INU corrected)')
     in_aseg = File(exists=True, mandatory=True,
                    desc='input ``aseg`` file, in NifTi format.')
+    in_ants = File(exists=True, mandatory=True,
+                   desc='brain tissue segmentation generated with antsBrainExtraction.sh')
 
 
 class RefineBrainMaskOutputSpec(TraitedSpec):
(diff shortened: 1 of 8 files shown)


oesteban:
Hold on with the review, I'm adding a little improvement.

fmriprep/interfaces/freesurfer.py
```diff
@@ -246,3 +285,70 @@ def detect_inputs(t1w_list, t2w_list=None, hires_enabled=True):
...
+    aseg[aseg == 42] = 3  # Collapse both hemispheres
+    gm = anat.copy()
+    gm[aseg != 3] = 0
+
+    refined = refine_aseg(aseg)
+    newrefmask = sim.binary_dilation(refined, selem) - refined
+    indices = np.argwhere(newrefmask > 0)
+    for pixel in indices:
+        # When ATROPOS identified the pixel as GM, set and carry on
+        if ants_segs[tuple(pixel)] == 2:
```
oesteban: These lines incorporate Arno's concepts for relabeling pixels excluded by FreeSurfer. However, my impression is that this adds a lot of pixels wrongly identified as GM by ATROPOS during `antsBrainExtraction.sh` (the segmentation comes from there).

Let me know if we want to keep it or discard it.

chrisfilo: It would be useful to see an example of this in action (in other words which voxels are wrongly relabeled).

chrisfilo:
Let me know when this is ready for review.

fmriprep/interfaces/freesurfer.py
oesteban: Okay, I'll get some examples.

I have to say that Arno's segmentations are probably a lot better, since he gets them from `antsCorticalThickness.sh`. The idea is probably great, but currently we can't apply it properly.

oesteban:
Here, two examples.

**Legend**: yellow is the original ANTs mask; red is with Arno's relabeling; green is the refined mask without Arno's relabeling.

### ds000114/sub-01 (the first example of #431).

ds114


### ds000017/sub-2 (second example)
ds017-1
ds017-2

It seems like, despite my first impression, the version with Arno's trick may work a bit better for these two cases. In general, both solutions seem better than the original mask (yellow contour).

fmriprep/interfaces/freesurfer.py
oesteban: https://github.com/oesteban/notebooks/blob/f5cf852d212867b32e7bf38bda2424e7cc132a35/FMRIPREP-issue431.ipynb - discussion moved to the main thread.

chrisfilo:
Both solutions look comparable (each makes mistakes at different locations). The improvement over ants is great.

One more thing I would check is how this new - tighter - mask works with normalization. If the template mask is more liberal it could be problematic.

oesteban:
Sweet. Should we just run 10 subjects to check the impact on spatial normalization? WDYT?

chrisfilo:
Sounds like a good plan.

Best,
Chris

oesteban:
Done, please wait for Circle to generate documentation. The diff is not particularly helpful since I introduced some line breaks for readability. Some changes are unrelated to this PR, but I added them because I judged it preferable (even if it means holding off merging for a little while):

  - [x] Added reference to Nilearn.
  - [x] Better described what are tools and what are packages (using the `<code>tool</code>` tags for tools)
  - [x] Brain tissue segmentation moved to the anatomical processing
  - [x] Rewrote the CompCor passage.
  - [x] Minor nits here and there (e.g. removing the `[Internet]` marker added after the journal names on references).

Also, updated the CHANGES file with all PR done so far since 1.0.0, with the idea of making it easier to brew a new release quickly and soon.

oesteban:
I'm not seeing any major problem with the extra 10 subjects and their masks. I haven't identified any difference in spatial normalization either. Merging...

PR #925, opened 2018-01-09
[FIX] INU correction before merging several T1w

### LINKED ISSUE #924: [BUG] INU correction should be run before t1_merge

oesteban:
Otherwise, we get funky composite bias fields that N4 can't recover.
This will also make coregistration between T1s robuster.

Fixes #924


fmriprep/workflows/anatomical.py
```diff
@@ -420,52 +420,74 @@ def init_anat_template_wf(longitudinal, omp_nthreads, num_t1w, name='anat_templa
...
+            (t1_reorient, concat_affines, [('transform', 'mat_CtoD')]),
+        ])
+    else:
+
+        def _get_first(in_list):
+            return in_list[0]
+
+        workflow.connect([
+            (t1_conform, t1_reorient, [(('out_file', _get_first), 'in_file')]),
+            (t1_reorient, concat_affines, [('transform', 'mat_BtoC')]),
```
effigies: `t1_reorient` is not necessary if `t1_merge` is not used. Conformation orients to RAS, but for reasons I don't understand, merging sometimes changes the orientation. Then you can get rid of `concat_affines` and just use the conformation affine.

oesteban:
I've been working at times in a pure-nipype implementation of `antsBrainExtraction.sh`. However, I haven't got any decent prototype yet (lack of time).

I'm not particularly worried about running N4 twice though, as long as it gets us sensible brain masks. But I agree, it'd be better not doubling up.

effigies:
Okay, so if we're fairly confident that the quality of the brain mask would be unaffected, but don't want to risk any artifacts of doubling up N4, we could use the following sequence in all cases:

* Conform
* N4-correct
* Merge
* Skull-strip
* Apply brain mask to merged image

This would mean we stop using the `BrainExtractionBrain` and `N4Corrected0` outputs of brain extraction directly. We could slide the extraction into the `skullstrip_ants_wf` to keep the interfaces consistent.

WDYT?

oesteban:
Yes, though I would give it a try to what we have right now.

My real, deep interest in deconstructing `antsBrainExtraction.sh` is to be able to connect stuff in the middle, without modifying the nipype interface, and to allow for tweaks. Great things we could have:
  - Find a mapping from OASIS (or NKI) to MNI and initialize spatial normalization with the precomputed transform of brain masking (remove 20 min from runtime).
  - Run the brain extraction on all T1s separately:
    1. Run an antsTemplate -like workflow to replace the fs.RobustTemplate.
    2. Initialize second and subsequent transforms with the first OASIS-T1w transform followed by the affine found in i).
    3. Final brainmask is the consensus of all masks.

But I can't see this happening anytime soon.

effigies:
Alright. So we can keep this if branch for now, and maybe create an issue to move to avoid having double N4 correction in skull-stripping outputs. It shouldn't be a huge task, but it will still be some work.

oesteban:
I'd go for that. I'm simplifying the workflow following your suggestions.

CHANGES.rst
```diff
@@ -1,6 +1,7 @@
 1.0.3 (3rd of January 2018)
 ===========================
 
+* [FIX] INU correction before merging several T1w (#925).
```
effigies: Should go in a "Next release" section.

fmriprep/data/itkIdentityTransform.tfm
```diff
@@ -0,0 +1,5 @@
+#Insight Transform File V1.0
+#Transform 0
+Transform: MatrixOffsetTransformBase_double_3_3
+Parameters: 1 0 0 0 1 0 0 0 1 0 0 0
+FixedParameters: 0 0 0
```
effigies: I think this file will need to be added to `setup.py`?

fmriprep/workflows/anatomical.py
```diff
@@ -420,60 +422,86 @@ def init_anat_template_wf(longitudinal, omp_nthreads, num_t1w, name='anat_templa
...
-    # Reorient template to RAS, if needed (mri_robust_template may set to LIA)
+    # 2. Reorient template to RAS, if needed (mri_robust_template may set to LIA)
     t1_reorient = pe.Node(Reorient(), name='t1_reorient')
 
-    lta_to_fsl = pe.MapNode(fs.utils.LTAConvert(out_fsl=True), iterfield=['in_lta'],
-                            name='lta_to_fsl')
-
     concat_affines = pe.MapNode(
-        ConcatAffines(3, invert=True), iterfield=['mat_AtoB', 'mat_BtoC'],
+        ConcatAffines(1 + min(num_t1w, 2), invert=True),
```
effigies: I think you can revert `lta_to_fsl` and `concat_affines` entirely, cleaning up the diff a little.

In any event, this no longer needs the `1 + min(...)` construct.

fmriprep/data/itkIdentityTransform.tfm
oesteban: yes

PR #933, opened 2018-01-12
[DOC] fix doi

jdkent:
Small fix:
 - DOI for 17 was hyperlinked to 16's paper, this fixes the hyperlink to go to the correct landing area.


jdkent:
For future reference:
changes just to the docs shouldn't invoke ci/circleci, correct?
Should I have named my commit `[docs only]: fix doi`?

effigies:
I tend to do `[docs only]` at the end, but yeah, that would do the right thing. Circle does build the docs every time, but `[docs only]` tells it not to do anything else.

Thanks.

PR #942, opened 2018-01-27
[FIX] Rerun tCompCor interface in case of MemoryError

### LINKED ISSUE #940: tCompCor - numpy.linalg.linalg.LinAlgError: SVD did not converge
#776 (aka the revenant) is back. Now on tCompCor. I'll apply the same solution.

oesteban:
Fixes #940


oesteban:
Tests were passing here: https://circleci.com/gh/poldracklab/fmriprep/1377

PR #943, opened 2018-01-28
[ENH] Migrate to CircleCI 2.0 and workflows

oesteban:
Hopefully, this PR will:

  - [x] Make use of docker caching (and stop inflating the docker pull count)
  - [x] Shorten the overall run time. From 1h 45m down to 1h. Additionally, jobs are individual so scheduling can be more efficient. Runtime upper bound: 2h 40min.
  - [x] Test caching (the anatomical workflows of ds005 and ds054 are pre-cached and updated if the workflow changes).
  - [x] The `[docs only]` marker is **not supported** anymore. However, only documentation branches can be built using branch that start with `doc/` or `docs/`.

Please merge with `--squash`, since the try-and-error with Circle took +70 commits.


effigies:
Have you tried building without cache? It would be good to get an upper bound of build time.

oesteban:
Hopefully this one will work. The docker image was cached, but there is no real difference between building the image from scratch and using cache (30m vs. 28m respectively).

oesteban:
Merge #944 first.

effigies:
So did you actually manage to save time by pre-building the Docker image? I didn't follow everything you did here, but in nipype I was finding that the persisting to workspace + reloading was actually taking longer than building the images in the individual test blocks. If you want to critique nipy/nipype#2386 with that in mind, I'd appreciate that. (When you get some minutes. No rush.)

oesteban:
Nope, I'm seeing the same thing: https://github.com/poldracklab/fmriprep/pull/943#issuecomment-361084628

But I thought it is a good testing exercise to keep using the cache and then persisting to the workspace. Maybe in a future, they optimize these processes at Circle (more bandwidth, mostly) and we start having some improvements. For now, it just protects us from downtimes in FreeSurfer and Neurodebian servers: build time is more predictable with the cache.

Other than that, I agree this is not a breakthrough in our testing. At the very least we'll be ready for retirement of CircleCI 1.0.

I'll check nipy/nipype#2386, actually that was the origin of my interest to start this PR.

PR #951, opened 2018-01-29
Import nibabel in get_ees

sitek:
When running with a field map without defining `EffectiveEchoSpacing` in the json, getting
```
File "<string>", line 59, in get_ees
NameError: name 'nb' is not defined
```

### DIFF
--- a/fmriprep/interfaces/fmap.py
+++ b/fmriprep/interfaces/fmap.py
@@ -314,6 +314,8 @@ def get_ees(in_meta, in_file=None):
 
     """
 
+    import nibabel as nb
+
     # Use case 1: EES is defined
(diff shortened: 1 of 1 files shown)


oesteban:
Thanks for this fix. Please consider adding your name in at this point of the [zenodo](https://github.com/poldracklab/fmriprep/blob/eb7ff9e920829d979c63c4e3bbec692a00c45c34/.zenodo.json#L56) contributors list. If you do so, please recall to include the mark `[skip ci]` in the first line of your commit so that tests are not triggered again and I can merge this quickly. Otherwise, I'll merge right after you confirm the zenodo file does not need to be updated.

PR #953, opened 2018-01-30
CI: Do not deploy on branches, only tags

effigies:
This should keep it from trying to push a new image with every commit to master.


oesteban:
Sorry I was on to the same thing, working directly on master.

effigies:
Ah. We doubled up the "ignore" statement.

oesteban:
Sorry again. I'll fix that.

Issue #934, opened 2018-01-14
adding multi-echo data to circleci tests


emdupre:
I'd like to add mutli-echo data to [circleci integration testing](https://github.com/poldracklab/fmriprep/blob/master/circle.yml). However, it seems that all of the used data is hosted on an OSF project that I don't have access to. 

Is there a preferred method for adding new tests to FMRIPREP?

oesteban:
OSF is the preferred way of adding data for testing. Please send me an email with your OSF login and I will give you access :)

oesteban:
Closed with #891

PR #956, opened 2018-02-01
[TST] Multi-echo testing

emdupre:
As discussed in #934, introduces testing for multi-echo data using openfMRI ds000210.
To be merged after #891!


.circleci/config.yml
```diff
@@ -122,11 +132,12 @@ jobs:
       - run:
           name: Create Nipype config files
           command: |
-            mkdir -p /tmp/ds005 /tmp/ds054
+            mkdir -p /tmp/ds005 /tmp/ds054 tmp/ds000210
```
oesteban: `/tmp/ds000210` (missing first slash)

emdupre:
I went ahead and merged master, here ! I tried to implement the only `meepi` branches feature, but it wasn't executing tests even for this PR (which is on a `meepi-testing` branch). It's now failing where I'd expect it to without merging #891.

emdupre:
That works for me. It's merged in now !

Issue #450, opened 2017-04-18
Use subject specific BOLD template for T1w coregistration


chrisfilo:
Such template would:
- improve within subject overlap
- improve coregistration with T1w

How to create it:
- If SBRefs are present use them exclusively
- If SBRefs are not present use reference_images from all runs

Affine unbiased coregistration seems a good method for creating such template.

oesteban:
This idea would require susceptibility distortion correction, otherwise the references are likely to have different realizations of distortion and the fused reference will be smoother than the individual cases.

What did you mean by affine _unbiased_ coregistration?

oesteban:
Rescued from the duplicate issue (Chris M.):

> Thinking about what's involved in this:
> 
> - Single workflow that takes all BOLD series
> - Each series needs to be SDC unwarped
> - Resulting reference files then merged with `StructuralReference`, and (EPI -> mean EPI) transform saved
> - Mean EPI aligned to T1w image
> 
> This will involve a fairly major refactor from the current, isolated processing of each BOLD series. One problem with this approach is that you can get different results based on whether your analyze all tasks or just a subset in the same process.
> 
> However, if we took an EPInorm approach, then there's a common space to which all EPIs are registered that doesn't depend on which subset of tasks have been specified.
> 
> WDYT?

PR #957, opened 2018-02-01
[ENH] Output aseg and aparc in T1w and BOLD spaces

### LINKED ISSUE #937: aparc+aseg outputs
@satra requested FreeSurfer aparc+aseg parcellations sampled into the output spaces, BOLD resolution.

It probably makes the most sense to treat these comparably to brain masks, i.e. one for each preprocessed BOLD series, even though these should be identical, assuming that all BOLD series have the same voxel size.

This may be very easy thanks to #431.

oesteban:
This PR generates additional outputs: the FreeSurfer's Aseg template and the FreeSurfer parcellation in both T1 and BOLD spaces.

Closes #937


.circleci/ds005_outputs.txt
```diff
@@ -14,6 +14,8 @@ fmriprep/sub-01/anat/sub-01_T1w_midthickness.R.surf.gii
 fmriprep/sub-01/anat/sub-01_T1w_pial.L.surf.gii
 fmriprep/sub-01/anat/sub-01_T1w_pial.R.surf.gii
 fmriprep/sub-01/anat/sub-01_T1w_preproc.nii.gz
+fmriprep/sub-01/anat/sub-01_T1w_rois-fsaparcaseg.nii.gz
+fmriprep/sub-01/anat/sub-01_T1w_rois-fsaseg.nii.gz
```
effigies: What's the advantage of including `aseg` in addition to `aparc+aseg`? `aseg` is just `aparc+aseg` with all cortical labels collapsed into "gray matter".

It's not going to be a huge cost in space, so that's fine.

Also, the [derivatives spec](https://docs.google.com/document/d/1Wwc4A6Mow4ZPPszDIWfCUCRNstn7d_zzaWPcfcHmgI4/edit#heading=h.9146wuepclkt) says that the type of these files should be `roi`, so they should end with `_roi.nii.gz`. And instead of `rois-`, I would use `label-`, to be consistent with that spec.

Finally, I think `aseg` and `aparcaseg` are just as unambiguous as `fsaseg` and `fsaparcaseg` and a little easier to read.

.circleci/ds005_outputs.txt
```diff
@@ -42,6 +44,8 @@ fmriprep/sub-01/func
 fmriprep/sub-01/func/sub-01_task-mixedgamblestask_run-01_bold_AROMAnoiseICs.csv
 fmriprep/sub-01/func/sub-01_task-mixedgamblestask_run-01_bold_confounds.tsv
 fmriprep/sub-01/func/sub-01_task-mixedgamblestask_run-01_bold_MELODICmix.tsv
+fmriprep/sub-01/func/sub-01_task-mixedgamblestask_run-01_bold_rois-fsaparc.nii.gz
+fmriprep/sub-01/func/sub-01_task-mixedgamblestask_run-01_bold_rois-fsaseg.nii.gz
```
effigies: `space-T1w_label-aparcaseg_roi.nii.gz`

fmriprep/workflows/anatomical.py
```diff
@@ -704,7 +748,10 @@ def init_surface_recon_wf(omp_nthreads, hires, name='surface_recon_wf'):
 
     autorecon_resume_wf = init_autorecon_resume_wf(omp_nthreads=omp_nthreads)
     gifti_surface_wf = init_gifti_surface_wf()
-    refine_brainmask_wf = init_refine_brainmask_wf()
+
+    asegnat = init_segs_to_native_wf()
+    aparcnat = init_segs_to_native_wf(segmentation='aparc_aseg')
```
effigies: To fit conventions a little better, I'd prefer:

```Python
aseg_to_native_wf = init_segs_to_native_wf()
aparc_to_native_wf = init_segs_to_native_wf(segmentation='aparc_aseg')
```

fmriprep/workflows/anatomical.py
```diff
@@ -1010,33 +1041,37 @@ def init_refine_brainmask_wf(name='refine_brainmask'):
...
 
     workflow.connect([
-        (inputnode, refine, [('in_file', 'in_anat'),
-                             ('ants_segs', 'in_ants')]),
-        (inputnode, get_aseg, [('subjects_dir', 'subjects_dir'),
-                               ('subject_id', 'subject_id')]),
+        (inputnode, fssource, [
+            ('subjects_dir', 'subjects_dir'),
+            ('subject_id', 'subject_id')]),
         (inputnode, tonii, [('in_file', 'reslice_like')]),
```
effigies: I suspect this `mri_convert` approach won't work so simply if using a pre-computed FreeSurfer reconstruction with more than one input T1. This is the case that we were checking against when we ran `mri_robust_register` against the `T1.mgz` and `T1w_preproc.nii.gz` in order to get the `fsnative` transforms.

The easiest way to check this case is to get a multi-session dataset (e.g. ds114) and run a FS recon in the opposite order (`recon-all -s sub-07 -i $DS114/sub-07/ses-test/sub-07_ses-test_T1w.nii.gz -i $DS114/sub-07/ses-retest/sub-07_ses-retest_T1w.nii.gz -all`) and use that as your FreeSurfer directory. There should be a coregistration matrix that encodes more than just a reslice.

fmriprep/workflows/bold/confounds.py
```diff
@@ -286,8 +300,24 @@ def _pick_wm(files):
...
+        aparc_tfm = pe.Node(ApplyTransforms(interpolation='NearestNeighbor', float=True),
+                            name='aparc_tfm', mem_gb=0.1)
+        workflow.connect([
+            # FreeSurfer ROIs
+            (inputnode, aseg_tfm, [('bold_mask', 'reference_image'),
+                                   ('t1_aseg', 'input_image'),
+                                   ('t1_bold_xform', 'transforms')]),
+            (inputnode, aparc_tfm, [('bold_mask', 'reference_image'),
+                                    ('t1_aparc', 'input_image'),
+                                    ('t1_bold_xform', 'transforms')]),
```
effigies: This is probably fine, but I wonder if we can do a single-shot transform from `fsnative` to `bold_space-T1w`.

fmriprep/workflows/anatomical.py
oesteban: I'll check before merging. Thanks

fmriprep/workflows/bold/confounds.py
oesteban: Not sure if it is worth spending time on a single transform. Let's go ahead with this and re-iterate if necessary.

.circleci/ds005_outputs.txt
```diff
@@ -49,6 +53,8 @@ fmriprep/sub-01/func/sub-01_task-mixedgamblestask_run-01_bold_space-MNI152NLin20
 fmriprep/sub-01/func/sub-01_task-mixedgamblestask_run-01_bold_space-MNI152NLin2009cAsym_variant-smoothAROMAnonaggr_preproc.nii.gz
 fmriprep/sub-01/func/sub-01_task-mixedgamblestask_run-02_bold_AROMAnoiseICs.csv
 fmriprep/sub-01/func/sub-01_task-mixedgamblestask_run-02_bold_confounds.tsv
+fmriprep/sub-01/func/sub-01_task-mixedgamblestask_run-02_bold_label-aparcaseg_roi.nii.gz
+fmriprep/sub-01/func/sub-01_task-mixedgamblestask_run-02_bold_label-aseg_roi.nii.gz
```
effigies: @satra Do you want to have a look at these new proposed outputs when you get a second? Does this look like what you had in mind?

satra: @effigies - looks good to me.  and just to clarify all runs are realigned to each other during fmriprep processing or left in some self-realigned state?

oesteban:
They are left in a self aligned state. We could align them if you think
it's necessary

satra:
@oesteban - it could be useful, but not a high-priority at the point. i can always map to the individual surface and extract the volumetric structures with these new parcellations to a create a common individual space across all runs.

effigies:
Related discussion: #253, #450

.circleci/ds005_outputs.txt
effigies: Satra's comment here made me realize we still need to specify the target space for the BOLD-sampled ROIs here with a `space-T1w` entity.

oesteban:
I don't think so: those ROIs are in native BOLD space

.circleci/ds005_outputs.txt
satra: @effigies - aren't the bold sampled rois in the functional space of whatever run was processed?

effigies:
But we don't output BOLD series in native BOLD space, do we? The BOLD series are motion/distortion corrected within run, and then each aligned to T1w. So what we want, it seems to me, would be a T1w-aligned ROI image, but sampled to BOLD space so it can be directly usable with the BOLD series.

satra:
> So what we want, it seems to me, would be a T1w-aligned ROI image, but sampled to BOLD space so it can be directly usable with the BOLD series.

indeed.

effigies:
I should say "BOLD resolution". Since the alignment is already the same.

oesteban:
I don't think it is a good idea to resample the bold signal in T1w space
AND to keep the resolution of BOLD just for the purpose of sampling on the
ROIs. Without having checked quantitatively, it seems to me that
interpolation errors will be larger in T1 space.

On Feb 2, 2018 18:01, "Chris Markiewicz" <notifications@github.com> wrote:

I should say "BOLD resolution". Since the alignment is already the same.

—
<https://github.com/poldracklab/fmriprep/pull/957#issuecomment-362765097>,
or mute the thread

satra:
>  don't think it is a good idea to resample the bold signal in T1w space
AND to keep the resolution of BOLD just for the purpose of sampling on the
ROIs.

i think what @effigies is saying is that BOLD and T1w are aligned but different resolution. the ROI would be also aligned but in BOLD resolution. so BOLD is not resampled to T1.

.circleci/ds005_outputs.txt
chrisfilo: The current name suggest that they are in the space of the corresponding functional scan, but @effigies will know for sure.

effigies:
Sorry if I'm retreading ground, but just to try to make my understanding of this situation (and my corresponding conclusions) as clear as I can:

The individual-run BOLD space is purely for calculating confounds without resampling beyond that necessary to do motion/distortion correction, while the actual functional derivatives are all either in T1w or MNI space. There is, as far as I can see, no end-user use for a BOLD-space ROI image.

For users like @satra, there are two options:

1. They can resample the `T1w_label-aparcaseg_roi.nii.gz` to BOLD resolution (since it's already in alignment)
2. We can do that for them

So, in any event, the current BOLD-space outputs should be removed. The question is entirely whether to replace them with images that are directly usable.

Am I summarizing the situation properly?

----

If so, and we want to provide ROI images that may be directly used with preprocessed functional series, then there are some additional issues that I don't think we've addressed.

At present we're producing one such image per run, but given that these are aligned to T1w, and it is highly likely that (at least within task) the series have the same resolution/FoV, we find ourselves producing many identical ROI images.

I see a few possibilities here:

1. Go ahead and produce one ROI image per run, as these volumes are pretty small on disk.
2. Produce one ROI image per task, collapsing across runs and sessions. (`sub-<subj>_task-<taskid>_bold_space-T1w_label-aparcaseg_roi.nii.gz`)
3. Produce one ROI image per resolution, ignoring task altogether. (`sub-<subj>_bold_space-T1w_res-<XxYxZ>_label-aparcaseg_roi.nii.gz`) It would then be the job of the end user to read the preprocessed functionals, get the zooms, and find the correct ROI image.

I think 2 is likely to be the easiest to use, and give us a pretty solid reduction in redundancy, but 3 would be guaranteed to have no more outputs than necessary.

----

If resampling the ROI image to BOLD resolution is the user's responsibility, ignore that last block.

fmriprep/workflows/anatomical.py
oesteban: Why not we just leave that edge case out? Is there a way to check if FreeSurfer's T1.mgz and T1w_preproc.nii.gz are not consistent (and just fail and recommend a wipe out)? cc @chrisfilo

chrisfilo: Is the edge case any dataset with more than one T1w per subject?

effigies: The edge case is any pre-computed FreeSurfer derivatives that don't happen to align perfectly with our preprocessed T1.

satra:
> There is, as far as I can see, no end-user use for a BOLD-space ROI image.

@effigies - given the lab you came from this sounds strange :)

effigies:
I can't tell if i'm wrong or unclear. By "BOLD-space" I mean "distortion-corrected BOLD series that is aligned only to itself". And by "no end-user use", I mean "such an image would not be aligned to any of our existing outputs, so cannot be directly applied without digging into the nipype working directory."

@satra, perhaps it would be useful for you to describe exactly what output you're looking for, at what resolutions, and aligned to what images.

satra:
@effigies - i think what you described earlier made sense to me, having the ROI label mask in T1w space and/or resampled to individual BOLD space. i only meant that individual ROIs in bold space are exactly what gets used for the kind of analyses we (jay/jason/myself/others) did in frank's lab. we never went to mni space if we didn't have to.

further, i thought this was about creating these roi files. as long as the individual BOLD functional series are aligned with the T1w, i am ok with the roi files only being there in T1w space. if i have to transform them for each individual, preprocessed functional run, then i would like to have them transformed into functional space. hope that makes sense.

satra:
ps. this would also be useful to build the individual specific cifti file as well.

effigies:
Sorry, I feel like there's still some confusion, though it might be entirely on my end. As far as I know, we weren't discussing MNI at all. Below, I'll list the spaces in BIDS style, where `space-native` means the space of the original input file with the same name:

* `*_T1w_space-native_*` - Merged T1w images, aligned to lexically first T1w image
  * N.B. This is equivalent to `*_T1w_space-T1w_*`
* `*_T1w_space-MNI*` - Merged T1w images, SyN-registered to MNI
* `*_bold_space-native_*` - Distortion corrected, aligned to reference image derived from original series
* `*_bold_space-T1w_*` - Distortion corrected, aligned to `T1w_preproc`
  * N.B. This is the same space as `*_T1w_space-native_*`, but with the resolution as `*_bold_space-native_*`
* `*_bold_space-MNI*` - Distortion corrected, aligned with

What @oesteban is *currently* producing are `*_bold_space-native_*_roi.nii.gz` files. I've been arguing that we should produce `*_bold_space-T1w_*_roi.nii.gz` files, because we do not produce any other `*_bold_space-native_*` derivatives, and `*_bold_space-T1w_*_roi.nii.gz` should be directly usable as in:

```Python
import nibabel as nb

func = nb.load('derivatives/fmriprep/sub-01/func/sub-01_task-rest_bold_space-T1w_preproc.nii.gz')
labels = nb.load('derivatives/fmriprep/sub-01/func/sub-01_task-rest_bold_space-T1w_label-aparcaseg_roi.nii.gz')

roi_series = {
    roi_id: func.get_data()[labels == roi_id].mean(axis=0).mean(axis=0).mean(axis=0)
    for roi_id in np.unique(labels.get_data())
    if roi_id != 0}
```

If you want `*_bold_space-native_label-*_roi.nii.gz`, then it should be understood that they will not be directly usable with any of our other outputs, either T1w or MNI space. If you need BOLD outputs that are not aligned to T1w, that is another feature that needs to be added.

If you simply use `*_T1w_space-native_label-*_roi.nii.gz`, then you'll need something to reslice. I believe @chrisfilo said that `nilearn` masking handles this automatically, so if that's your approach, we can drop the `*_bold_*_roi.nii.gz` altogether.

Again, the MNI space outputs were listed above purely for the sake of distinguishing the spaces. There is no proposal at this time to output ROIs in MNI.

Hopefully this is clearer. Apologies to everybody if I'm restating the obvious, and even more profuse apologies if I'm restating the incorrect.

effigies:
Just to be absurdly explicit about what we need to know:

* What resolution do you want your BOLD series?
* What alignment do you want your BOLD series?
* What resolution do you want your ROIs?
* What alignment do you want your ROIs?

The answer to each of these questions can be "native", "T1w", or "MNI".

satra:
@effigies - thanks for the clarification - i think that's very useful.

i agree with you that the 80% individual use case would indeed be: 

> *_bold_space-T1w_* - Distortion corrected, aligned to T1w_preproc
N.B. This is the same space as *_T1w_space-native_*, but with the resolution as *_bold_space-native_*

rather than bold_space-native. 

if this change was made, then having the roi as: `*_T1w_space-native_label-*_roi.nii.gz` makes most sense.

emdupre:
Just to jump in here (thanks to @effigies for clarifying the logic, and making this possible!), I'd like to +1 having these aligned to the T1 rather than bold_space-native. Thinking about the multi-echo integration, one of the benefits of ME is increased subcortical SNR, so making it easy for users to plug-and-play with a T1-aligned CIFTI volume would be great.

oesteban:
Okay, this is ready to merge.

This PR generates the following aseg/aparc files:

* `*_bold_space-T1w_label-[aseg|aparcaseg]_roi.nii[.gz]`
* `*_T1w_space-native_*` - even though the actual names are `T1w_label-[aseg|aparcaseg]_roi` to be consistent with other derivatives such as the brain mask. I beg you not to enter this discussion because then the brain mask should also carry the `space-native` and possibly a `label-oasis` as it is strictly an atlas-based segmentation.

I'd appreciate a quick review, I'm rushing to release and this PR would be great to have.

chrisfilo:
I would not worry about the names too much since they are likely to change in the future (`_roi` was intended for individual ROI files not parcellations, which are not properly covered by BIDS Derivatives yet). LGTM

oesteban:
Sweet!

Issue #963, opened 2018-02-08
 BOLD brain mask depicted in the ROIs reportlet does not correspond to the mask produced in the outputs


chrisfilo:
Background: https://neurostars.org/t/fmriprep-1-0-6-2-issues-semi-random-brain-mask-failure-and-super-bright-areas-in-epi/1303/8

EDIT OE: (moved to new issue #982)

reportlet:
![image](https://user-images.githubusercontent.com/238759/35985579-d029b5ea-0cab-11e8-9f5e-4776a03e1d70.png)

output mask:
![image](https://user-images.githubusercontent.com/238759/35985603-dcb4c9bc-0cab-11e8-8eff-63a746708e48.png)

chrisfilo:
![image](https://user-images.githubusercontent.com/238759/35994571-a674a6aa-0cc5-11e8-8bec-f0c506c79c33.png)
Seems like a 3dUnifize issue.

chrisfilo:
Similar problem is also visible in: https://openneuro.org/datasets/ds001216/versions/00002?app=FMRIPREP&version=48&job=5a7b6fd4831753000151c620 

https://s3.amazonaws.com/openneuro.outputs/89832c01e43fee4b165e81a351af7b26/37c908e6-f88d-46e3-a0e3-d9c7e2dab4ac/fmriprep/sub-30009.html

oesteban:
I can confirm that this is also happening on our Phase 2 test base. I'm seeing this on ds000237/sub-0[3,6] and some edge case on sub-01.

oesteban:
Identified the source - this mask calculation is failing horribly:

https://github.com/poldracklab/fmriprep/blob/3e7d830e0322435532e875ec8c51213f9f6fec3f/fmriprep/workflows/bold/util.py#L176

chrisfilo:
This might explain #964, but why this mask calculation fail is not visible in the ROIs reportlet?

chrisfilo:
I don't understand why you renamed this issue. I believe the original problem is still unresolved - the mask plotted in the report comes from a different estimation process than the one (two?) leading to _brainmask file in the outputs.

oesteban:
Sorry for the name change

oesteban:
Sorry for the name change. I didn't understand the issue.

I've opened and closed #982, which is the issue that has been really addressed, restored the name and reopened this issue.

Also, I hope this message triggers a notification that makes all aware of these changes.

[...]

Issue #964, opened 2018-02-08
Overzelous BOLD skullstripping


chrisfilo:
Background: https://neurostars.org/t/fmriprep-1-0-6-2-issues-semi-random-brain-mask-failure-and-super-bright-areas-in-epi/1303/8

![image](https://user-images.githubusercontent.com/238759/35985654-fbe80cf4-0cab-11e8-8290-0a4cefb03784.png)

related to #963

oesteban:
Fixed through #966+#978

PR #966, opened 2018-02-09
[FIX] Robuster BOLD mask

oesteban:
Several improvements to make BOLD masks robuster.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,5 +1,40 @@
-./work*
-./out*
-data
-.git
+# python cache
+__pycache__/**/*
+__pycache__
+*.pyc
+
+# python distribution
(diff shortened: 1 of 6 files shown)


chrisfilo:
Wow - this was quick! Did you test it on problematic cases (see https://github.com/poldracklab/fmriprep/issues?utf8=%E2%9C%93&q=is%3Aissue+label%3Askullstripping+). Does it also solve the reporting issue (#963)?

oesteban:
After checking ~2000 masks (of 10800 masks I could compute using all bold files from OpenfMRI by Sept 2017), I think this okay to merge.

This indeed fixes #963. It doesn't with #964.

PR #962, opened 2018-02-08
updating niworkflows to 0.3.1

chrisfilo:

PR #967, opened 2018-02-09
[FIX] Write latest BOLD mask out (space-T1w)

oesteban:
As reported in #964, the BOLD brain mask written to the derivatives folder did not match the one presented in the BOLD ROIs reportlet.

This PR resamples that mask into T1w space and connects the output to the appropriate data sink, what should fix the issue.

### DIFF
--- a/fmriprep/workflows/bold/base.py
+++ b/fmriprep/workflows/bold/base.py
@@ -19,6 +19,9 @@ from niworkflows.nipype import logging
 from niworkflows.nipype.interfaces.fsl import Split as FSLSplit
 from niworkflows.nipype.pipeline import engine as pe
 from niworkflows.nipype.interfaces import utility as niu
+from niworkflows.interfaces.fixes import (
+    FixHeaderApplyTransforms as ApplyTransforms
+)
 
 from ...interfaces import (
     DerivativesDataSink,
@@ -374,6 +377,10 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
         name='bold_bold_trans_wf'
     )
 
+    # Map new BOLD mask (from ``bold_bold_transf_wf``) into T1w space
+    boldmsk_tfm = pe.Node(ApplyTransforms(interpolation='NearestNeighbor', float=True),
+                          name='boldmsk_tfm', mem_gb=0.1)
+
     workflow.connect([
         (inputnode, bold_reference_wf, [('bold_file', 'inputnode.bold_file')]),
         (bold_reference_wf, bold_hmc_wf, [
@@ -406,8 +413,7 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
             ('outputnode.melodic_mix', 'melodic_mix'),
             ('outputnode.nonaggr_denoised_file', 'nonaggr_denoised_file'),
         ]),
-        (bold_reg_wf, outputnode, [('outputnode.bold_t1', 'bold_t1'),
-                                   ('outputnode.bold_mask_t1', 'bold_mask_t1')]),
+        (bold_reg_wf, outputnode, [('outputnode.bold_t1', 'bold_t1')]),
         (bold_confounds_wf, func_reports_wf, [
             ('outputnode.rois_report', 'inputnode.bold_rois_report'),
             ('outputnode.ica_aroma_report', 'inputnode.ica_aroma_report')]),
@@ -416,6 +422,12 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
         (summary, func_reports_wf, [('out_report', 'inputnode.summary_report')]),
         (bold_split, bold_bold_trans_wf, [
             ('out_files', 'inputnode.bold_split')]),
+        (bold_bold_trans_wf, boldmsk_tfm, [
+            ('outputnode.bold_mask', 'input_image')]),
+        (bold_reg_wf, boldmsk_tfm, [
+            ('outputnode.bold_mask_t1', 'reference_image'),
+            ('outputnode.itk_bold_to_t1', 'transforms')]),
+        (boldmsk_tfm, outputnode, [('output_image', 'bold_mask_t1')])
     ])
 
     # bool('TooShort') == True, so check True explicitly
@@ -517,7 +529,7 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
                 (bold_t2s_wf, bold_reg_wf, [
                     ('outputnode.t2s_map', 'inputnode.ref_bold_brain'),
                     ('outputnode.t2s_mask', 'inputnode.ref_bold_mask')])
-                ])
+            ])
         else:
             LOGGER.warn('No fieldmaps found or they were ignored, building base workflow '
                         'for dataset %s.', ref_file)


oesteban:
Hold on with this one, I have to check the MNI mask is correct.

oesteban:
I had to remove `--write-graph` from the ds054 test build (see #970).

I'm about to test on ds224 (has fieldmaps) to double check this problem. If it runs correctly on 224, are you okay with merging this and releasing (incl. #962)?

effigies:
Is the issue just that the existing `bold_mask_t1` never got the fieldmap update? If so, this seems like it can be fixed in `bold_reg_wf` by allowing `mask_t1w_tfm` to take a list of transforms (similar to how we do for the whole BOLD series).

Or is there something besides the fieldmap that makes the `bold_bold` workflow mask different? (I'm also not sure why the BOLD-MNI mask needed changing, but I haven't really been following this issue.)

oesteban:
> Is the issue just that the existing bold_mask_t1 never got the fieldmap update? If so, this seems like it can be fixed in bold_reg_wf by allowing mask_t1w_tfm to take a list of transforms (similar to how we do for the whole BOLD series).
> Or is there something besides the fieldmap that makes the bold_bold workflow mask different? (I'm also not sure why the BOLD-MNI mask needed changing, but I haven't really been following this issue.)

We are doing this in the bold_to_bold workflow. And this PR implements this approach in the bold_to_mni one.

effigies:
It looks like you have two different approaches.

* T1w
  * bold to bold (field map correction)
  * bold to T1w applied to mask

* MNI
  * Single shot bold to MNI
  * Generate reference image from MNI BOLD series

oesteban:
You are right. Now I recall how I got here. First I fixed the bold mask in T1 space, but then realized that the mask in #964 would probably be the MNI one.

Then, I thought that recalculating the mask was probably the best option, and failed to update the T1 workflow with that thinking.

If we can agree that recalculating the reference and mask in every new space is probably the best option, then I'll fix the T1 path. Ideas? (cc @chrisfilo)

oesteban:
BTW, `--write-graph` worked on ds224.

chrisfilo:
Let me try to understand this. Assuming user requested T1w and template spaces:

Option 1:
- Calculate the bold mask after SDC in reference bold space.
- Use this mask to estimate T1w transformations.
- Plot this mask in ROIs plot.
- Apply T1w transformations to move the mask to T1w space.
- Apply template transformation to move the mask to template space.

Option 2:
- Calculate the bold mask (v1) after SDC in reference bold space.
- Use this mask to estimate T1w transformations.
- Plot this mask in ROIs plot.
- Apply T1w transformations to the reference.
- Estimate new mask (v2a) using reference in T1w space.
- Apply template transformation to move the the reference.
- Estimate new mask (v2b) using reference in T1w space.

I prefer option 1 because it makes things easier to debug (you plot the mask that is being used) and does not require recalculating the mask multiple times. The most important thing is that mask should be estimated after SDC (which happens in both options).

oesteban:
This is correct in general, except some details. The current implementation goes:

1. Run [`init_bold_reference_wf`](http://fmriprep.readthedocs.io/en/latest/api/index.html#fmriprep.workflows.bold.util.init_bold_reference_wf) to get a reference for HMC and SDC.
2. Corrections and registrations are run
2b. If SDC was run, calculate a new reference with `init_bold_reference_wf` again. This will only be used for BBR.
3. Apply 1-shot resampling of the BOLD series into target spaces (including native BOLD)
4a (BOLD in BOLD space). Run `init_bold_reference_wf` again, generate undistorted reference with corresponding mask. This mask is used for the ROIs plot.
4b (BOLD in T1w space). BOLD reference and mask are transformed (from 1!).
4c (BOLD in MNI space). BOLD reference and mask are transformed (from 1!).

In the current implementation, the mask and the BOLD reference used here are from 1). For the reference this does not seem very relevant. But, for the masks, the concurrence of binary moving and target, large voxels, and nearest-neighbor resamplings produces a lot less accurate results than calculating reference and mask from the `_preproc` time-series.

I would recommend Option 2.

Ref. the debugging comment - even though this would not address the core of your comment (it is easier to debug what we can see on the reports), I was thinking that we should add one reportlet for the MNI space. It would show MNI-template's contours (inspired by the "glass-brain") like those I generated for the variability figure in the paper, AND the new BOLD mask in red color. WDYT?

chrisfilo:
+1 for the "glass-brain" template space reportlet (although this is indepent of the current discussion - might be worth opening an issue for it).

I remain weakly unconvinced and a bit confused:
1. Why do we calculate new reference from data after SDC instead of applying the estimated warp to the original reference?
2. What is the difference between the reference in 2b and 4a?
3. What do you mean by 'concurrence of binary moving and target'?
4. Could you show some examples of accuracy loss when transforming a mask vs recalculating it?
5. Finally when you say 'I would recommend option 2' you mean calculating the mask three times as described inhttps://github.com/poldracklab/fmriprep/pull/967#issuecomment-365116144 not the 'current implementation' described in https://github.com/poldracklab/fmriprep/pull/967#issuecomment-365121351.

I am a bit skeptical about the 'loss of accuracy' argument but if it is true Option 2 makes sense. The problem is we would have to add reportlets for all three brain extraction procedures.

oesteban:
I've opened #975. As for your questions:

> 1. Why do we calculate new reference from data after SDC instead of applying the estimated warp to the original reference?

After revising the workflow, I can see that I was wrong. 2b is false, since we don't use `init_bold_reference_wf`. But we use [`init_enhance_and_skullstrip_bold_wf `](https://github.com/poldracklab/fmriprep/blob/1eae2c47ca2f83915578b783271a30bf0d74c5b6/fmriprep/workflows/fieldmap/unwarp.py#L158) on the reference image instead, so we calculate a new mask (instead of resampling).

> 2. What is the difference between the reference in 2b and 4a?

I don't think there's a spatial difference.

> 3. What do you mean by 'concurrence of binary moving and target'?

The moving image is binary. The target image (another mask) is binary. "The concurrence" refers to the three factors (binary masks, big voxels, nearest interpolations).

> 4. Could you show some examples of accuracy loss when transforming a mask vs recalculating it?

screen shot 2018-02-12 at 6 53 33 pm
Left is the mask proposed in this PR (ds054, run 01) from CircleCI artifacts. Right would be the former implementation (transforming a mask) in master artifacts. Right seems more hairy (sorry I couldn't hold the joke).

Looking at both masks, now I realize of an additional plus: mathematical morphology of edges is better for the newly calculated mask. But also, some pixels that were isolated in mask from raw data may become more integrated after distortion correction. Those are likely to change if the mask is calculated again. Conversely, transforming the mask would keep that pixel out of it. When voxels are big, then this error is big.

> 5. Finally when you say 'I would recommend option 2' you mean calculating the mask three times as described in https://github.com/poldracklab/fmriprep/pull/967#issuecomment-365116144 not the 'current implementation' described in https://github.com/poldracklab/fmriprep/pull/967#issuecomment-365121351.

Option 2 of your comment.

> I am a bit skeptical about the 'loss of accuracy' argument but if it is true Option 2 makes sense. The problem is we would have to add reportlets for all three brain extraction procedures.

I would agree on this, if we have the same definition of "reportlet" in mind. If reportlet here also means that it should be rendered in the report, I disagree. The reports are meant to show the most relevant points of the preprocessing, so supporting masks like these would not be crucial and they would definitely clutter the reports.

If here you meant that we generate reportlets and leave them in the work directory (or move them to the reportlets folder, both work) for us to quickly assess things without need of opening niftis, then +1.

chrisfilo:
Wow - this does indeed look bad. I'm surprised - have not seen such a bad case when looking at FMRIPREP outputs. I look at some historical CircleCI results and things have been similarly bad for a while. I think this is especially exacerbated by 4mm voxels. I'm convinced!

As for the reports - I think we should at least include the two final masks (one for T1w and one for the MNI space - as we saw before masks could fail independently).

PS Is there any difference between NN interpolation and linear interpolation followed by 0.5 thresholding? I wonder if there are some lessons here that need to be applied when interpolating segmentations - see #957

chrisfilo:
Some relevant references I came across re interpolations
- http://ieeexplore.ieee.org/xpls/icp.jsp?arnumber=7080516
- https://sourceforge.net/p/advants/discussion/840261/thread/8f4f91f8/

oesteban:
> As for the reports - I think we should at least include the two final masks (one for T1w and one for the MNI space - as we saw before masks could fail independently).

Agreed.

> PS Is there any difference between NN interpolation and linear interpolation followed by 0.5 thresholding? I wonder if there are some lessons here that need to be applied when interpolating segmentations

I don't know. But my intuition would be yes, as where the 0.5 threshold falls will depend a lot on the interpolation kernel.

For the aseg interpolation things are more complicated, as more knowledge would be needed to collapse several labels that fall within the same voxel. Freesurfer first transfers the labels from the atlas but then applies a bayesian segmentation approach with that initialization, per subject.

This is one reason for me to be a bit discouraged to implement the aseg in T1-space with BOLD resolution. But after thinking it thorough, I'd say the difference in that case is very very minimal and probably not worth the pain.

PR #976, opened 2018-02-13
[experiment] replace NearestNeighbors with MultiLabel

chrisfilo:
Checking if this will deal with 'jaggies'. Relevant to #967 discussion.

### DIFF
--- a/fmriprep/workflows/anatomical.py
+++ b/fmriprep/workflows/anatomical.py
@@ -271,13 +271,13 @@ def init_anat_preproc_wf(skull_strip_template, output_spaces, template, debug,
     # Resample the brain mask and the tissue probability maps into mni space
     mni_mask = pe.Node(
         ApplyTransforms(dimension=3, default_value=0, float=True,
-                        interpolation='NearestNeighbor'),
+                        interpolation='MultiLabel'),
         name='mni_mask'
     )
 
     mni_seg = pe.Node(
         ApplyTransforms(dimension=3, default_value=0, float=True,
-                        interpolation='NearestNeighbor'),
+                        interpolation='MultiLabel'),
         name='mni_seg'
     )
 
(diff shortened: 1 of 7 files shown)


chrisfilo:
This resolves the jagged edges problem
image

However runtime increased substantially. Things to consider:
- Do we need MultiLabel everywhere?
- Maybe linear + thresholding at 0.5 will be faster and as good?

oesteban:
I'll use MultiLabel in the Aseg resampling.

For all those resamplings of masks in the confounds workflows I'd say that linear + thresholding would work.

For brain masks, I would keep generating new reference and mask as interpolation is not the only factor that influences the result.

chrisfilo:
But what's wrong with this mask? Regenerating mask twice will require two plots and increase cognitive load required to review the reports.

chrisfilo:
It's a tradeoff between pipeline complexity combined  with cognitive load due to need to verify additional information in the reports (it's not just the number of plots, but number of things you need to check) and quality of the output mask. I was concerned about jagged edges but with smarter interpolation you can achieve good results without the need to recompute the mask. I don't think recomputing the mask is justified in this case. Have a look at the mask
image

At the end of the day I will stop arguing, because we are sort of splitting hairs here. It's your call.

chrisfilo:
It does indeed look marginally better (although I expect the differences to diminish with 2 or 3mm resolution). Which masks do you plan to add to the reports? Is ROIs plot already in T1w space?

> The only check we may want to add here is the overlap of the T1w mask in BOLD resolution and this mask from BOLD.  
Could you clarify? I don't understand.

oesteban:
I'm going to merge #978 since we agree that it is a correct way of doing things and it is a smaller iteration over what it is implemented. I'll leave #967 open, ready to merge if I manage to elaborate a more convincing argument.

About the comparison: the idea would be calculating the overlap of both masks (resampled with MultiLabel or recalculated with the reference workflow) to the corresponding (anatomical) brainmask calculated on the T1w data (or the MNI template if checking on that space). The procedure that gets better overlap with the anatomical mask.

PR #987, opened 2018-02-15
Prepopulate tag when posting neurostars questions.

chrisfilo:


oesteban:
I'm going to cancel builds to free resources. I'll merge without tests. One hint: using `docs/*` as branch name will only trigger the build_docs test.

chrisfilo:
Cool - thanks!

Best,
Chris

Issue #794, opened 2017-10-25
Generate CIFTI derivatives


oesteban:
@satra said (#658)
> * saving to CIFTI

@effigies:
> This seems reasonable, but out of my experience so far. If you have example code that converts the type of data you'd be interested in, that would be useful. (I assume CIFTI can handle either structural or functional data, as standard NIfTI can...)

This issue propose storing derivatives to CIFTI format.

effigies:
Just a bump on this, at @satra's request. This is pretty high priority for their group.

We have a couple options:

- Include either finding somebody using nibabel's CIFTI support to generate images
- Wrap `connectome_wb`

Satra says his group can contribute some effort, so just as a guess, ccing @mgxd.

mgxd:
Quick question - is there a labels atlas for the MNI template fmriprep uses - `MNI152NLin2009cAsym`, similar to mindboggle's [OASIS-TRT-20 joint fusion atlas](https://osf.io/d2cmy/?action=download&version=1)?

oesteban:
We could map that one into MNI if you'd need it

chrisfilo:
What's the role of such labels in context of CIFTI files?

mgxd:
@oesteban @chrisfilo we could use the labels to generate the [corresponding brain structure label](https://github.com/nipy/nibabel/blob/master/nibabel/cifti2/tests/test_cifti2io.py#L218-L238) and ijk coordinates

chrisfilo:
I see - thanks for clarifying. It might be worth splitting this task to cover different CIFTI outputs in separate PRs.

mgxd:
we should always have the output CIFTI's subcortical structures in the same space as the template - as more are added (#487) we'll have to generate new labels for that template, so I agree this should be broken up in separate PRs. I'm hoping to add support for common space (fsaverage5 surf, MNI152NLin2009cAsym vols) and hopefully subject space (piggybacking once #957 is merged)

@oesteban how would we go about doing this mapping?

oesteban:
The first I'd do is running a registration using: [`niworkflows.interfaces.mni.RobustMNINormalization`](https://github.com/poldracklab/niworkflows/blob/f3ebdf5479ed8e03143677017d5d084f5a81ae90/niworkflows/interfaces/mni.py#L66). Default settings should be pretty good. I assume the template comes with a good brain mask, so everything should be pretty straightforward.

[...]

PR #1001, opened 2018-02-21
ENH: generate CIFTI derivatives

mgxd:
ref. #794 

* converts bold/gii data to CIFTI dtseries file
  * `fsaverage5`/`fsaverage6` surfaces + `MNI2009NLin2009cAsym` subcortical volumes
* adds `compress` input to `DerivativesDataSink`
* increases required niworkflows to 0.3.7

###### TODOs
* [x] complete interface
* [x] add to bold workflow
* [x] add testing

###### To be added in another PR
* create HCP 91282 greyordinate space CIFTIs (#1025)
* support `fsnative` surface + subject space `aparc+aseg`
  * number of indices in fsaverage5 (~10k) vs fsnative (+100k) will make these files fairly large


fmriprep/interfaces/cifti.py
```diff
@@ -0,0 +1,219 @@
...
+class GenerateCiftiInputSpec(BaseInterfaceInputSpec):
+    bold_file = File(mandatory=True, exists=True, desc="input BOLD file")
+    volume_target = traits.Enum("MNI152NLin2009cAsym", mandatory=True, usedefault=True,
+                                desc="CIFTI volumetric output space")
+    surface_target = traits.Enum("fsaverage5", "fsaverage6", "fsaverage", mandatory=True,
+                                 usedefault=True, desc="CIFTI surface target space")
+    subjects_dir = Directory(mandatory=True, desc="FreeSurfer SUBJECTS_DIR")
+    TR = traits.Float(mandatory=True, desc="repetition time")
+    gifti_files = traits.List(File(exists=True), mandatory=True,
+                              desc="surface geometry files")
```
effigies: Probably worth specifying that this should be length 2 with order `(left, right)`.

fmriprep/interfaces/cifti.py
```diff
@@ -0,0 +1,219 @@
...
+        """Converts inputspec to files"""
+        if self.surface_target == "fsnative":
+            raise NotImplementedError
+
+        annotation_files = sorted(glob(os.path.join(self.subjects_dir,
+                                                    self.surface_target,
+                                                    'label',
+                                                    '*h.aparc.annot')))
+        # TODO: fetch label_file
+        return annotation_files
```
effigies: Does it make sense to use a `FreeSurferSource` to find the `aparc.annot` files?

Also, do we want to consider the possibility people will want to use DKTatlas or a2009s (at least to the extent of being aware of whether we're making a move in that direction easy or hard)?

fmriprep/interfaces/cifti.py
```diff
@@ -0,0 +1,219 @@
...
+    def _run_interface(self, runtime):
+        self.annotation_files, self.label_file = self._fetch_data()
+        self._results["out_file"] = create_cifti_image(
+            self.bold_file,
+            self.label_file,
+            self.annotation_files,
+            self.gifti_files,
+            self.volume_target,
+            self.surface_target,
+            self.TR)
```
effigies: `self.inputs.*`

fmriprep/interfaces/cifti.py
```diff
@@ -0,0 +1,219 @@
...
+
+        return runtime
+
+    def _fetch_data(self):
+        """Converts inputspec to files"""
+        if self.surface_target == "fsnative":
```
effigies: `self.inputs.surface_target`

fmriprep/interfaces/cifti.py
```diff
@@ -0,0 +1,219 @@
...
+            raise NotImplementedError
+
+        annotation_files = sorted(glob(os.path.join(self.subjects_dir,
+                                                    self.surface_target,
```
effigies: `self.inputs.*`

fmriprep/workflows/bold/base.py
```diff
@@ -676,6 +681,18 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
...
+        # CIFTI output
+        if cifti_output and 'template' in output_spaces:
+            cifti = pe.MapNode(GenerateCifti(), iterfields=["surface_target"], name="gen_cifti")
+            cifti.inputs.TR = metadata.get("RepetitionTime")
+            workflow.connect([
+                (bold_surf_wf, cifti, [
+                    ('targets.out', 'surface_target'),
+                    ('outputnode.surfaces', 'gifti_files')]),
+                (inputnode, cifti, [('subjects_dir', 'subjects_dir')]),
+                (bold_mni_trans_wf, outputnode, [('outputnode.bold_mni', 'bold_mni')]),
```
effigies: I suspect this is meant to be:

```Python
(cifti, outputnode, [('out_file', 'bold_cifti')])
```

fmriprep/interfaces/cifti.py
```diff
@@ -0,0 +1,219 @@
...
+    idx_offset = 0
+    brainmodels = []
+    bm_ts = np.empty((timepoints, 0))
+
+    for structure, labels in CIFTI_STRUCT_WITH_LABELS.items():
+        if not labels:  # surface model
+            model_type = "CIFTI_MODEL_TYPE_SURFACE"
+            # use the corresponding annotation
+            hemi = structure.split('_')[-1][0]
+            annots = [fl for fl in annotation_files if ".aparc.annot" in fl]
```
effigies: While probably fine, this could be made a little more precise:

```Python
annots = [fl for fl in annotation_files if fl.endswith('.aparc.annot')]
```

fmriprep/interfaces/cifti.py
```diff
@@ -0,0 +1,219 @@
...
+            annot = (nb.freesurfer.read_annot(annots[0]) if hemi == "LEFT"
+                     else nb.freesurfer.read_annot(annots[1]))
+            # currently only supports L/R cortex
+            gii = (nb.load(gii_files[0]) if hemi == "LEFT"
+                   else nb.load(gii_files[1]))
```
effigies:
```Python
gii = nb.load(gii_files[hemi == 'RIGHT'])
```

fmriprep/interfaces/cifti.py
```diff
@@ -0,0 +1,219 @@
...
+
+class GenerateCifti(SimpleInterface):
+    """
+    Generate CIFTI image from BOLD file in target space
+    """
+    input_spec = GenerateCiftiInputSpec
+    output_spec = GenerateCiftiOutputSpec
+
+    def _run_interface(self, runtime):
+        self.annotation_files, self.label_file = self._fetch_data()
```
effigies: No clear reason to assign these to instance variables.

fmriprep/interfaces/cifti.py
```diff
@@ -0,0 +1,219 @@
...
+    bm_ts = np.empty((timepoints, 0))
+
+    for structure, labels in CIFTI_STRUCT_WITH_LABELS.items():
+        if not labels:  # surface model
+            model_type = "CIFTI_MODEL_TYPE_SURFACE"
+            # use the corresponding annotation
+            hemi = structure.split('_')[-1][0]
+            annots = [fl for fl in annotation_files if ".aparc.annot" in fl]
+            annot = (nb.freesurfer.read_annot(annots[0]) if hemi == "LEFT"
+                     else nb.freesurfer.read_annot(annots[1]))
```
effigies:
```Python
annot = nb.freesurfer.read_annot(annots[hemi == 'RIGHT'])
```

oesteban:
> we should probably store the label_file template to make this process faster - perhaps to the OSF project used in niworkflows?

👍  I'll give you access to the OSF repository. What is your username/orcid/handle?

mgxd:
@oesteban orcid # 0000-0002-7252-7771

fmriprep/interfaces/cifti.py
mgxd: We could but we'd still have to parse the results - I can change if you feel strongly.

Could you elaborate on your second point? Use how?

effigies: FreeSurfer now automatically produces

```Shell
% ls /data/out/ds000114/derivatives/freesurfer/sub-07/label/lh.aparc*.annot 
/data/out/ds000114/derivatives/freesurfer/sub-07/label/lh.aparc.a2009s.annot
/data/out/ds000114/derivatives/freesurfer/sub-07/label/lh.aparc.annot
/data/out/ds000114/derivatives/freesurfer/sub-07/label/lh.aparc.DKTatlas.annot
```

It seems plausible that someone might want to generate their CIFTI files with labels from another parcellation. Again, not necessarily that we need to code it up right now, but just to consider how easy it would be to augment.

(It may also be that I'm horribly wrong and CIFTI pretty much demands vanilla aparc.)

fmriprep/interfaces/cifti.py
```diff
@@ -0,0 +1,221 @@
...
+    'CIFTI_STRUCTURE_ACCUMBENS_RIGHT': [58],
+    # 'CIFTI_STRUCTURE_ALL_WHITE_MATTER': [],
+    # 'CIFTI_STRUCTURE_ALL_GREY_MATTER': [],
+    'CIFTI_STRUCTURE_AMYGDALA_LEFT': [18],
+    'CIFTI_STRUCTURE_AMYGDALA_RIGHT': [54],
+    'CIFTI_STRUCTURE_BRAIN_STEM': [16],
+    'CIFTI_STRUCTURE_CAUDATE_LEFT': [11],
+    'CIFTI_STRUCTURE_CAUDATE_RIGHT': [50],
+    'CIFTI_STRUCTURE_CEREBELLAR_WHITE_MATTER_LEFT': [7],
+    'CIFTI_STRUCTURE_CEREBELLAR_WHITE_MATTER_RIGHT': [46],
```
satra: i don't think we need the white matter voxels included.

fmriprep/interfaces/cifti.py
```diff
@@ -0,0 +1,221 @@
...
+    # 'CIFTI_STRUCTURE_CEREBELLUM': [],
+    'CIFTI_STRUCTURE_CEREBELLUM_LEFT': [6],
+    'CIFTI_STRUCTURE_CEREBELLUM_RIGHT': [45],
+    'CIFTI_STRUCTURE_DIENCEPHALON_VENTRAL_LEFT': [28],
+    'CIFTI_STRUCTURE_DIENCEPHALON_VENTRAL_RIGHT': [60],
```
satra: or the diencephalon at present. i'm not sure these are labeled by the atlas.

fmriprep/interfaces/cifti.py
```diff
@@ -0,0 +1,221 @@
...
+class GenerateCiftiInputSpec(BaseInterfaceInputSpec):
+    bold_file = File(mandatory=True, exists=True, desc="input BOLD file")
+    volume_target = traits.Enum("MNI152NLin2009cAsym", mandatory=True, usedefault=True,
+                                desc="CIFTI volumetric output space")
+    surface_target = traits.Enum("fsaverage5", "fsaverage6", "fsaverage", mandatory=True,
+                                 usedefault=True, desc="CIFTI surface target space")
+    subjects_dir = Directory(mandatory=True, desc="FreeSurfer SUBJECTS_DIR")
+    TR = traits.Float(mandatory=True, desc="repetition time")
+    gifti_files = traits.List(File(exists=True), mandatory=True,
+                              desc="list of surface geometry files (length 2 with order [L,R])")
```
satra: @effigies  - does the gifti file contain any info about the surface target, in which case surface_target should be removed and just derived from the gifti file.

fmriprep/interfaces/cifti.py
```diff
@@ -0,0 +1,221 @@
...
+    'CIFTI_STRUCTURE_THALAMUS_LEFT': [10],
+    'CIFTI_STRUCTURE_THALAMUS_RIGHT': [49],
+}
+
+
+class GenerateCiftiInputSpec(BaseInterfaceInputSpec):
+    bold_file = File(mandatory=True, exists=True, desc="input BOLD file")
+    volume_target = traits.Enum("MNI152NLin2009cAsym", mandatory=True, usedefault=True,
+                                desc="CIFTI volumetric output space")
+    surface_target = traits.Enum("fsaverage5", "fsaverage6", "fsaverage", mandatory=True,
```
satra: probably remove fsaverage for now as well since it's going to be really large.

fmriprep/interfaces/cifti.py
effigies: So these are functional time series, and it does not appear so.

```Shell
% mris_info /data/out/ds000114/derivatives/fmriprep/sub-07/ses-test/func/sub-07_ses-test_task-fingerfootlips_bold_space-fsaverage5.L.func.gii
gifti_image struct
    version    = 1.0
    numDA      = 184
gim->meta nvpairs struct, len = 4 :
    nvpair: 'AnatomicalStructurePrimary' = 'CortexLeft'
    nvpair: 'UserName' = 'root'
    nvpair: 'Date' = 'Mon Oct 16 18:34:03 2017'
    nvpair: 'gifticlib-version' = 'gifti library version 1.09, 28 June, 2010'

gim->labeltable giiLabelTable struct, len = 0 :
--------------------------------------------------
gim->darray[0] giiDataArray struct
    intent   2001 = NIFTI_INTENT_TIME_SERIES
    datatype   16 = NIFTI_TYPE_FLOAT32
    ind_ord     1 = RowMajorOrder
    num_dim       = 1
    dims          = 10242, 0, 0, 0, 0, 0
    encoding    3 = GZipBase64Binary
    endian      2 = LittleEndian
    ext_fname     = 
    ext_offset    = 0
darray->meta nvpairs struct, len = 1 :
    nvpair: 'TimeStep' = '2500.000000'

darray->coordsys giiCoordSystem struct
    dataspace  = NIFTI_XFORM_UNKNOWN
    xformspace = NIFTI_XFORM_UNKNOWN
    xform[0] :  1.000000  0.000000  0.000000  0.000000
    xform[1] :  0.000000  1.000000  0.000000  0.000000
    xform[2] :  0.000000  0.000000  1.000000  0.000000
    xform[3] :  0.000000  0.000000  0.000000  1.000000
    data       = <set>
    nvals      = 10242
    nbyper     = 4
    numCS      = 1
darray->ex_atrs nvpairs struct, len = 0 :
--------------------------------------------------
--------------------------------------------------
gim->darray[1] giiDataArray struct
    intent   2001 = NIFTI_INTENT_TIME_SERIES
    datatype   16 = NIFTI_TYPE_FLOAT32
    ind_ord     1 = RowMajorOrder
    num_dim       = 1
    dims          = 10242, 0, 0, 0, 0, 0
    encoding    3 = GZipBase64Binary
    endian      2 = LittleEndian
    ext_fname     = 
    ext_offset    = 0
darray->meta nvpairs struct, len = 1 :
    nvpair: 'TimeStep' = '2500.000000'

darray->coordsys giiCoordSystem struct
    dataspace  = NIFTI_XFORM_UNKNOWN
    xformspace = NIFTI_XFORM_UNKNOWN
    xform[0] :  1.000000  0.000000  0.000000  0.000000
    xform[1] :  0.000000  1.000000  0.000000  0.000000
    xform[2] :  0.000000  0.000000  1.000000  0.000000
    xform[3] :  0.000000  0.000000  0.000000  1.000000
    data       = <set>
    nvals      = 10242
    nbyper     = 4
    numCS      = 1
darray->ex_atrs nvpairs struct, len = 0 :
--------------------------------------------------
[...]
--------------------------------------------------
gim->darray[183] giiDataArray struct
    intent   2001 = NIFTI_INTENT_TIME_SERIES
    datatype   16 = NIFTI_TYPE_FLOAT32
    ind_ord     1 = RowMajorOrder
    num_dim       = 1
    dims          = 10242, 0, 0, 0, 0, 0
    encoding    3 = GZipBase64Binary
    endian      2 = LittleEndian
    ext_fname     = 
    ext_offset    = 0
darray->meta nvpairs struct, len = 1 :
    nvpair: 'TimeStep' = '2500.000000'

darray->coordsys giiCoordSystem struct
    dataspace  = NIFTI_XFORM_UNKNOWN
    xformspace = NIFTI_XFORM_UNKNOWN
    xform[0] :  1.000000  0.000000  0.000000  0.000000
    xform[1] :  0.000000  1.000000  0.000000  0.000000
    xform[2] :  0.000000  0.000000  1.000000  0.000000
    xform[3] :  0.000000  0.000000  0.000000  1.000000
    data       = <set>
    nvals      = 10242
    nbyper     = 4
    numCS      = 1
darray->ex_atrs nvpairs struct, len = 0 :
--------------------------------------------------
gifti_image struct
    swapped    = 0
    compressed = 1
 -- darray totals: 7 MB
gim->ex_atrs nvpairs struct, len = 0 :
==================================================
```

fmriprep/interfaces/cifti.py
mgxd: I used the labels provided on [mindboggle's page](http://mindboggle.readthedocs.io/en/latest/labels.html) - I'll remove these though

satra: ok you can leave these in place. do you have an example file on openmind that i can take a look at?

fmriprep/interfaces/cifti.py
effigies: This seems fine to me to leave, since `fsaverage` is a valid (though off-by-default) target for FMRIPREP.

satra: i'm just worried about the file size - also the reason i suggested leaving out cerebellar wm voxels.

mgxd:
sits on top of poldracklab/niworkflows#228

effigies:
Oh, sorry. You'll need to pin the niworkflows commit. I can do that on your branch, as it's probably the easiest way to show you how.

Also, there are some Flake8 complaints:

https://travis-ci.org/poldracklab/fmriprep/builds/346981847?utm_source=github_status&utm_medium=notification

satra:
@oesteban - while this is fine for now, i did ask @mgxd to put interfaces into nipype itself, so this may move over into misc.

oesteban:
I'll cut a new niworkflows release. When migration to nipype is done we'll release niworkflows again.

oesteban:
@mgxd niworkflows-0.3.5 has been released just now.

oesteban:
Seems like some labeling is off limits:

```
	 Saving crash info to /out/fmriprep/sub-01/log/20180301-154659_7893bebd-00a5-4233-8b0f-c5cb62462a29/crash-20180301-161154-root-gen_cifti-e62b8388-e788-4555-b0fc-c73227a821d8.txt
  File "/usr/local/miniconda/lib/python3.6/site-packages/niworkflows/nipype/pipeline/plugins/multiproc.py", line 68, in run_node
  File "/usr/local/miniconda/lib/python3.6/site-packages/niworkflows/nipype/pipeline/engine/nodes.py", line 487, in run
  File "/usr/local/miniconda/lib/python3.6/site-packages/niworkflows/nipype/pipeline/engine/nodes.py", line 1265, in _run_interface
    self.config['execution']['stop_on_first_crash'])))
  File "/usr/local/miniconda/lib/python3.6/site-packages/niworkflows/nipype/pipeline/engine/nodes.py", line 1187, in _collate_results
    (self.name, '\n'.join(msg)))
Exception: Subnodes of node: gen_cifti failed:
Subnode 0 failed
Error: Traceback (most recent call last):

  File "/usr/local/miniconda/lib/python3.6/site-packages/niworkflows/nipype/pipeline/engine/utils.py", line 106, in nodelist_runner
    result = node.run(updatehash=updatehash)

  File "/usr/local/miniconda/lib/python3.6/site-packages/niworkflows/nipype/pipeline/engine/nodes.py", line 487, in run

  File "/usr/local/miniconda/lib/python3.6/site-packages/niworkflows/nipype/pipeline/engine/nodes.py", line 571, in _run_interface

  File "/usr/local/miniconda/lib/python3.6/site-packages/niworkflows/nipype/pipeline/engine/nodes.py", line 650, in _run_command

  File "/usr/local/miniconda/lib/python3.6/site-packages/niworkflows/nipype/interfaces/base/core.py", line 516, in run

  File "/usr/local/miniconda/lib/python3.6/site-packages/fmriprep/interfaces/cifti.py", line 90, in _run_interface
    download_link)

  File "/usr/local/miniconda/lib/python3.6/site-packages/fmriprep/interfaces/cifti.py", line 190, in create_cifti_image
    ts.append(bold_data[ijk])

IndexError: index 70 is out of bounds for axis 1 with size 58
```

The automated tests are failing because you need to add a `cifti_output=False` argument here: 

https://github.com/poldracklab/fmriprep/blob/4314431f98e1ec68d4666ae03be684b414785d22/fmriprep/workflows/tests/test_base.py#L14-L41

mgxd:
@oesteban yeah, I believe the problem is that the atlas needs to match the same resolution as the bold image (both in `MNI152NLin`). WDYT of using nilearn's `resample_to_img` to match both the atlas/bold's resolution?

oesteban:
One question (cc @satra): is there any particular reason not to use FreeSurfer's DKT atlas for this PR? I say this because FMRIPREP already has the gears in place to provide it with the appropriate resolution (#957).

Right now we are using the atlas mapped into OASIS and then mapped into MNI. It does not look like a very robust idea.

satra:
@oesteban -  it depends on what one is trying to do. 

if the cortical stuff is in fsaverage space, then using the atlas makes most sense. however, fsaverge5's subcortical things are not as good as mapping the mni template to an atlas.

if the cortical target is individual, then using the freesurfer atlas for subcortical things make sense.

an alternative would be to use the OASIS mni template as the target.

oesteban:
> if the cortical stuff is in fsaverage space, then using the atlas makes most sense.

What you mean by "cortical stuff"? the BOLD sampled on the surface?

> however, fsaverge5's subcortical things are not as good as mapping the mni template to an atlas.

You mean, the subcortical segmentation of the DKT atlas is superior to the fsaverage?

> an alternative would be to use the OASIS mni template as the target.

We are actually using OASIS as target for the brain extraction (default `antsBrainExtraction.sh`), so we could use directly the atlas.

satra:
> What you mean by "cortical stuff"? the BOLD sampled on the surface?

correct, cifti doesn't make sense for volumetric files, except for subcortical structures.

> You mean, the subcortical segmentation of the DKT atlas is superior to that of fsaverage?

yes. it's not perfect by any means, but better than using fsaverage to extract these regions. both are DKT, but i think (soon to be tested), the subcortical registrations for ANTs is pretty decent.

however, the ideal atlas would be created by registering 1mm anatomical targets and then bringing over the atlas and resampling it to 2mm.

> We are actually using OASIS as target for the brain extraction (default antsBrainExtraction.sh), so we could use directly the atlas.

why use two different templates? why not use the OASIS template through and through. we could always regenerate the template and atlas with newer versions of ANTs.

oesteban:
> why use two different templates? why not use the OASIS template through and through. we could always regenerate the template and atlas with newer versions of ANTs.

For long, we've been thinking of refining spatial normalization from/to MNI through the OASIS. However, that would require us to retrieve the spatial normalization to OASIS from `antsBrainExtraction.sh` to concatenate then an OASIS-to-MNI warping (that could be precomputed). However, we never found the time.

I actually started a nipype implementation of `antsBrainExtraction.sh` (instead of modifying the interfaces) to check if we got some gains in parallelization and accrue from the extra flexibility of nipype to expose all intermediate results.

I think we can get this merged with the DKT template that @mgxd added to niworkflows and refine the spatial normalization later on.

WDYT?

mgxd:
Actually, before I make any resampling changes I want to clarify the resampling already done within fmriprep.

If I have BOLD data at a greater resolution than the `MNI152NLin` template used, it looks like my data will be downsampled to the template's resolution. And if the data is of lower resolution, the image is not upsampled, but the derived BOLD in template has an altered resolution (I'm comparing ds005 raw/derived data) - what accounts for this change?

fmriprep/interfaces/bids.py
```diff
@@ -162,6 +162,8 @@ class DerivativesDataSinkInputSpec(BaseInterfaceInputSpec):
     source_file = File(exists=False, mandatory=True, desc='the input func file')
     suffix = traits.Str('', mandatory=True, desc='suffix appended to source_file')
     extra_values = traits.List(traits.Str)
+    force_uncompress = traits.Bool(False, usedefault=True,
+                                   desc="save as uncompressed nifti")
```
effigies: I might just do this as

```Python
compress = traits.Bool(desc="output file as compressed (True) or uncompressed (False) (default: unchanged from input)")
```

Or something similar. That seems more natural to me.

mgxd:
this is ready for further review

fmriprep/interfaces/bids.py
```diff
@@ -201,9 +203,10 @@ def __init__(self, out_path_base=None, **inputs):
     def _run_interface(self, runtime):
         src_fname, _ = _splitext(self.inputs.source_file)
         _, ext = _splitext(self.inputs.in_file[0])
-        compress = ext == '.nii'
-        if compress:
+        if self.inputs.compress is True:
             ext = '.nii.gz'
+        elif self.inputs.compress is False:
+            ext = '.nii'
```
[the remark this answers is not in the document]
satra: just a note: gifti files can be compressed, but the compression is inside the file in the data blob, rather than external. (i do think this sort of logic should make it's way to nibabel, with perhaps a `compress=???` option in to_filename).

.circleci/ds005_outputs.txt
```diff
@@ -47,6 +47,7 @@ fmriprep/sub-01/func/sub-01_task-mixedgamblestask_run-01_bold_MELODICmix.tsv
 fmriprep/sub-01/func/sub-01_task-mixedgamblestask_run-01_bold_space-fsaverage5.L.func.gii
 fmriprep/sub-01/func/sub-01_task-mixedgamblestask_run-01_bold_space-fsaverage5.R.func.gii
 fmriprep/sub-01/func/sub-01_task-mixedgamblestask_run-01_bold_space-MNI152NLin2009cAsym_brainmask.nii.gz
+fmriprep/sub-01/func/sub-01_task-mixedgamblestask_run-01_bold_space-MNI152NLin2009cAsym_preproc.dtseries.nii
```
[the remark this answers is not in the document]
satra: i'm not sure the `+` character is allowed, but the general idea sounds ok for fmriprep. for cifti itself, i could craft up data that could be in any number of spaces.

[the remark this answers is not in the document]
mgxd: how about `space-MNI152NLin2009cAsym_surf-fsaverage5`?

also, keep in mind both spaces are embedded in the metadata
https://github.com/mgxd/fmriprep/blob/ed8b6424b7e971f0b5eb33daed099b6d18142cb4/fmriprep/interfaces/cifti.py#L222-L226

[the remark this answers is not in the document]
satra: or `space-multiple` and the interpretation would be left to the filetype/file+json.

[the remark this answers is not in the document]
effigies: @mgxd surf isn't yet a valid entity key, and I'd rather not proliferate those more than we have to. Space currently can indicate either volumetric or surface spaces, and there's not a principled reason that I can see to choose one over the other as the type of space that gets an additional key.

@satra You're right that `+` is not currently supported... I think labels are defined to be alphanumeric. I would be okay with trying to push to enable `+`, or we can just come up with another name to refer to this space. I'm reluctant to use `space-multiple`, as that isn't any more informative than `.dtseries.nii`, which indicates that it's a CIFTI, and thus the spaces can be retrieved from metadata. We would not be able to support multiple CIFTI output spaces, since the filenames would collide.

If we're okay with this, and want to rely on metadata, I would suggest dropping `space-` entirely. Alternately, let's come up with a name that indicates the spaces are `MNI152NLin2009cAsym` and `fsaverage5`, and add it to the derivatives spec.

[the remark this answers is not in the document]
effigies: As I think more about this, I find the `+` approach least objectionable. I'm going to propose to the BIDS discussion list permitting `+` as a delimiter in cases where multiple values reasonably apply to a key.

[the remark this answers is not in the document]
satra: i generally prefer that we don't encode metadata in filenames (unless it's a 90% usecase that would make it more useful than not). there was a discussion on the mailing list on separators. may be worthwhile to look that one up.

in this particular case, since one could generate cifti files in different combination of spaces, it would be useful to encode the difference in the filename. how that happens, i'm a bit more ambivalent about.

effigies:
@mgxd Have you verified that you can correctly view the generated CIFTI files in the HCP viewer? I believe that's the only CIFTI-enabled viewer at the moment, so making sure that they can correctly read and display images that are not in their custom spaces would be a worthwhile effort for both FMRIPREP and connectome_wb.

satra:
@effigies - I don't think the workbench will open this file properly, but worth trying. The workbench requires hcp style cifti files to map properly into the surfaces and volumes it displays.

For that a different making would have to be done for both surface and volume.

effigies:
Can you not load your own surfaces and volumes in `connectome_wb`? I believe I was able to do that last summer. I did not have CIFTI data to load on top, however.

satra:
@effigies - i guess if they were gifti that could work. @mgxd - could you try it out?

mgxd:
after a quick fix to this code, `wb_view` can now load up the output

.circleci/ds005_outputs.txt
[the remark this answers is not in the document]
mgxd: linking further discussion: https://groups.google.com/forum/#!topic/bids-discussion/ASHycf6JB_0

fmriprep/interfaces/cifti.py
```diff
@@ -117,122 +135,146 @@ def _fetch_data(self):
         download_link = getters.OSF_PROJECT_URL + getters.OSF_RESOURCES[label_space][0]
         return annotation_files, label_file, download_link
 
-
-def create_cifti_image(bold_file, label_file, annotation_files, gii_files,
-                       volume_target, surface_target, tr, download_link=None):
+    def _create_cifti_image(bold_file, label_file, annotation_files, gii_files,
```
effigies: This either needs to be a `@staticmethod` or have `self` as an initial parameter.

mgxd:
re: the nibabel crash - perhaps this line should be set to False?
https://github.com/poldracklab/fmriprep/blob/5b5e5eb823fed0beae0bdbe9253603e9f50a7679/fmriprep/cli/run.py#L22

effigies:
Does that resolve the issue? That looks like a nasty bug ([permalink](https://circleci.com/gh/poldracklab/fmriprep/2409) for reference), and my guess is it replicates rarely. Re-running that test to check.

effigies:
Well, it's reliable. This doesn't seem related to this PR, though. Bumping builds to see whether this is specific to your branch or not.

mgxd:
I'm trying to build a docker container off this branch, but am running into a peculiar error.

#### First attempt - build a new docker image from my repo.
```
docker build -t cifti-fmriprep .
```
Once successfully built, I tried running a subject, but was given this error:
```
180405-15:54:42,857 workflow IMPORTANT:
	 
    Running fMRIPREP version :
      * BIDS dataset path: /data.
      * Participant list: ['BANDA001'].
      * Run identifier: 20180405-155442_f08b40e4-9a37-425a-9c3e-d1ec8ef8c21d.
    
/usr/local/miniconda/lib/python3.6/site-packages/grabbit/core.py:410: UserWarning: No valid root directory found for domain 'derivatives'. Falling back on the Layout's root directory. If this isn't the intended behavior, make sure the config file for this domain includes a 'root' key.
  "'root' key." % config['name'])
180405-15:54:43,853 workflow IMPORTANT:
	 Creating bold processing workflow for "/data/sub-BANDA001/func/sub-BANDA001_task-conflict_run-01_bold.nii.gz" (0.32 GB / 280 TRs). Memory resampled/largemem=1.28/2.17 GB.
Process Process-2:
  File "/usr/local/miniconda/lib/python3.6/multiprocessing/process.py", line 249, in _bootstrap
    self.run()
  File "/usr/local/miniconda/lib/python3.6/multiprocessing/process.py", line 93, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/local/miniconda/lib/python3.6/site-packages/fmriprep/cli/run.py", line 451, in build_workflow
    ignore_aroma_err=opts.ignore_aroma_denoising_errors,
  File "/usr/local/miniconda/lib/python3.6/site-packages/fmriprep/workflows/base.py", line 199, in init_fmriprep_wf
    ignore_aroma_err=ignore_aroma_err)
  File "/usr/local/miniconda/lib/python3.6/site-packages/fmriprep/workflows/base.py", line 450, in init_single_subject_wf
    ignore_aroma_err=ignore_aroma_err)
  File "/usr/local/miniconda/lib/python3.6/site-packages/fmriprep/workflows/bold/base.py", line 397, in init_func_preproc_wf
    bold_sdc_wf.inputs.inputnode.template = template
AttributeError: 'TraitedSpec' object has no attribute 'inputnode'
```

However, using the `poldracklab/fmriprep:latest` image did not result in the error.

#### Second attempt - bind cifti fmriprep fork within `poldracklab/fmriprep:latest`

```
docker run --rm -it -v /code/datasets/banda:/data \
-v /code/fmriprep/fmriprep:/usr/local/miniconda/lib/python3.6/site-packages/fmriprep:ro \
-v /home/mathias/code/workdir/niworkflows/niworkflows/data:/usr/local/miniconda/lib/python3.6/site-packages/niworkflows/data:ro poldracklab/fmriprep:latest \
/data /data/derivatives participant --participant_label BANDA001 \
--fs-license-file /data/code/.freesurfer_license.txt --ignore slicetiming
```

This results in the same error (but now version is shown as `0+unknown`)

```
180405-16:05:14,264 workflow IMPORTANT:
	 
    Running fMRIPREP version 0+unknown:
      * BIDS dataset path: /data.
      * Participant list: ['BANDA001'].
      * Run identifier: 20180405-160514_2ebd676a-c387-445a-b20c-9100a66eb9f0.
    
/usr/local/miniconda/lib/python3.6/site-packages/grabbit/core.py:410: UserWarning: No valid root directory found for domain 'derivatives'. Falling back on the Layout's root directory. If this isn't the intended behavior, make sure the config file for this domain includes a 'root' key.
  "'root' key." % config['name'])
180405-16:05:15,307 workflow IMPORTANT:
	 Creating bold processing workflow for "/data/sub-BANDA001/func/sub-BANDA001_task-conflict_run-01_bold.nii.gz" (0.32 GB / 280 TRs). Memory resampled/largemem=1.28/2.17 GB.
Process Process-2:
  File "/usr/local/miniconda/lib/python3.6/multiprocessing/process.py", line 249, in _bootstrap
    self.run()
  File "/usr/local/miniconda/lib/python3.6/multiprocessing/process.py", line 93, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/local/miniconda/lib/python3.6/site-packages/fmriprep/cli/run.py", line 451, in build_workflow
    ignore_aroma_err=opts.ignore_aroma_denoising_errors,
  File "/usr/local/miniconda/lib/python3.6/site-packages/fmriprep/workflows/base.py", line 199, in init_fmriprep_wf
    ignore_aroma_err=ignore_aroma_err)
  File "/usr/local/miniconda/lib/python3.6/site-packages/fmriprep/workflows/base.py", line 450, in init_single_subject_wf
    ignore_aroma_err=ignore_aroma_err)
  File "/usr/local/miniconda/lib/python3.6/site-packages/fmriprep/workflows/bold/base.py", line 397, in init_func_preproc_wf
    bold_sdc_wf.inputs.inputnode.template = template
AttributeError: 'TraitedSpec' object has no attribute 'inputnode'
```

#### Third attempt - binding my fork of master within `poldracklab/fmriprep:latest`

Again, the same error - I'm not sure why this is happening...

However, all tests pass on the CI, so this leads me to believe it's something with my machine.

@oesteban any chance you could give this another (final?) review

effigies:
`bold_sdc_wf.inputs.inputnode.template = template` is a problem. I think that would need to be `bold_sdc_wf.get_node('inputnode').inputs.template = template`, or similar.

mgxd:
@effigies I forgot to mention that I had also tried that:
```
180405-15:28:50,759 workflow IMPORTANT:
	 
    Running fMRIPREP version 0+unknown:
      * BIDS dataset path: /data.
      * Participant list: ['BANDA001'].
      * Run identifier: 20180405-152850_10fccda0-6e5c-4e21-8f3e-dc8a289c4f68.
    
180405-15:28:51,751 workflow IMPORTANT:
	 Creating bold processing workflow for "/data/sub-BANDA001/func/sub-BANDA001_task-conflict_run-01_bold.nii.gz" (0.32 GB / 280 TRs). Memory resampled/largemem=1.28/2.17 GB.
Process Process-2:
  File "/usr/local/miniconda/lib/python3.6/multiprocessing/process.py", line 249, in _bootstrap
    self.run()
  File "/usr/local/miniconda/lib/python3.6/multiprocessing/process.py", line 93, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/local/miniconda/lib/python3.6/site-packages/fmriprep/cli/run.py", line 451, in build_workflow
    ignore_aroma_err=opts.ignore_aroma_denoising_errors,
  File "/usr/local/miniconda/lib/python3.6/site-packages/fmriprep/workflows/base.py", line 199, in init_fmriprep_wf
    ignore_aroma_err=ignore_aroma_err)
  File "/usr/local/miniconda/lib/python3.6/site-packages/fmriprep/workflows/base.py", line 450, in init_single_subject_wf
    ignore_aroma_err=ignore_aroma_err)
  File "/usr/local/miniconda/lib/python3.6/site-packages/fmriprep/workflows/bold/base.py", line 397, in init_func_preproc_wf
    bold_sdc_wf.get_node('inputnode').inputs.template = template
AttributeError: 'NoneType' object has no attribute 'inputs'
```

oesteban:
For the `version 0+unknown` problem:

* Update your tags: `git fetch upstream --tags`
* And build with a version: 
  ```
 FMRIPREP_VER=1.0.9-dev0; docker build --rm -t poldracklab/fmriprep:${FMRIPREP_VER} --build-arg BUILD_DATE=`date -u +"%Y-%m-%dT%H:%M:%SZ"` --build-arg VCS_REF=`git rev-parse --short HEAD` --build-arg VERSION=${FMRIPREP_VER} .
  ```

For the other issue, please merge upstream first.

oesteban:
Let's check those things and then I'll try to merge this in today :+1:

oesteban:
DS210 is green :+1:

fmriprep/cli/run.py
```diff
@@ -140,6 +140,9 @@ def get_parser():
         '--medial-surface-nan', required=False, action='store_true', default=False,
         help='Replace medial wall values with NaNs on functional GIFTI files. Only '
         'performed for GIFTI files mapped to a freesurfer subject (fsaverage or fsnative).')
+    g_conf.add_argument(
+        '--cifti-output', required=False, action='store_true', default=False,
```
oesteban: What happens if ``--cifti-output`` and ``--fs-no-reconall`` are both set? Is there any error handling here?

fmriprep/interfaces/cifti.py
```diff
@@ -0,0 +1,282 @@
...
+
+        Returns
+            out_file : BOLD data as CIFTI dtseries
+        """
+
+        bold_img = nb.load(bold_file)
+        label_img = nb.load(label_file)
+        if not bold_img.shape[:3] == label_img.shape:
+            # we need these images to have the same resolution
+            bold_img = resample_to_img(bold_img, label_img, interpolation="nearest")
```
oesteban: Is the image resampled this way then used when sampling on the surface? (in other words, are we not using the promised Lanczos interpolation on an important resampling?)

requirements.txt
```diff
@@ -1,2 +1,2 @@
-niworkflows>=0.3.4
+niworkflows>=0.3.5
```
oesteban: niworkflows>=0.3.7

fmriprep/interfaces/cifti.py
mgxd: @oesteban I don't think this shouldn't affect the surfaces, since those are acquired from the GIFTI files. This resampling is only done to ensure the subcortical structures can be indexed by the same coordinates found in the atlas.

oesteban: Sure, so basically you need to pack ((i, j, k, t), bold_value) within the format right? Thus, that bold_value will be the result of this interpolation. Correct?

mgxd: ^ yes, but only for subcortical - I believe surface voxels (i,j,k) are stored via GIFTI.

oesteban: Okay, I'll have to have a deeper look into this. If I can't check on it anytime soon, I'll open an issue and merge this as is :+1:

fmriprep/cli/run.py
```diff
@@ -330,6 +330,11 @@ def build_workflow(opts, retval):
             raise RuntimeError(msg)
         logger.warning(msg)
 
+    # Cannot output CIFTI if no freesurfer surfaces
+    if opts.cifti_output and not opts.run_reconall:
```
satra: instead of logging, could this return an error in the command line itself. `argparse` has xor type options.

.circleci/ds005_outputs.txt
```diff
@@ -56,6 +57,7 @@ fmriprep/sub-01/func/sub-01_task-mixedgamblestask_run-01_bold_space-T1w_preproc.
 fmriprep/sub-01/func/sub-01_task-mixedgamblestask_run-02_bold_AROMAnoiseICs.csv
 fmriprep/sub-01/func/sub-01_task-mixedgamblestask_run-02_bold_confounds.tsv
 fmriprep/sub-01/func/sub-01_task-mixedgamblestask_run-02_bold_MELODICmix.tsv
+fmriprep/sub-01/func/sub-01_task-mixedgamblestask_run-02_bold_space-cifti_variant-space1_preproc.dtseries.nii
```
effigies: Isn't there supposed to be a `json` file that describes `space1`?

fmriprep/cli/run.py
```diff
@@ -173,13 +173,18 @@ def get_parser():
                        help='EXPERIMENTAL/TEMPORARY: Use SyN correction in addition to '
                        'fieldmap correction, if available')
 
+    # Surface generation
+    g_surfs = parser.add_mutually_exclusive_group()
```
effigies: I will say this is a little dissatisfying. Is it possible to make subgroups, as in `g_surfs = g_fs.add_mutually_exclusive_group`?

screen shot 2018-04-06 at 10 25 38

effigies: Just checked:

```Python
In [5]: parser = argparse.ArgumentParser()

In [6]: g_surf = parser.add_argument_group('Surface-preprocessing options')

In [7]: g_surf_xor = g_surf.add_mutually_exclusive_group()

In [8]: g_surf.add_argument('--no-submm-recon', action='store_false', dest='hires',
   ...:                        help='disable sub-millimeter (hires) reconstruction')
   ...:                        
Out[8]: _StoreFalseAction(option_strings=['--no-submm-recon'], dest='hires', nargs=0, const=False, default=True, type=None, choices=None, help='disable sub-millimeter (hires) reconstruction', metavar=None)

In [9]: g_surf_xor.add_argument('--fs-no-reconall', '--no-freesurfer',
   ...:                          action='store_false', dest='run_reconall',
   ...:                          help='disable FreeSurfer surface preprocessing.
   ...: '
   ...:                          ' Note : `--no-freesurfer` is deprecated and will be removed in 1.2.'
   ...:                          ' Use `--fs-no-reconall` instead.')
   ...:                          
Out[9]: _StoreFalseAction(option_strings=['--fs-no-reconall', '--no-freesurfer'], dest='run_reconall', nargs=0, const=False, default=True, type=None, choices=None, help='disable FreeSurfer surface preprocessing. Note : `--no-freesurfer` is deprecated and will be removed in 1.2. Use `--fs-no-reconall` instead.', metavar=None)

In [10]: g_surf_xor.add_argument('--cifti-output', action='store_true', default=
    ...: False,
    ...:                          help='output BOLD files as CIFTI dtseries')
    ...:                          
Out[10]: _StoreTrueAction(option_strings=['--cifti-output'], dest='cifti_output', nargs=0, const=True, default=False, type=None, choices=None, help='output BOLD files as CIFTI dtseries', metavar=None)

In [12]: print(parser.format_help())
usage: ipython [-h] [--no-submm-recon] [--fs-no-reconall | --cifti-output]

optional arguments:
  -h, --help            show this help message and exit

Surface-preprocessing options:
  --no-submm-recon      disable sub-millimeter (hires) reconstruction
  --fs-no-reconall, --no-freesurfer
                        disable FreeSurfer surface preprocessing. Note :
                        `--no-freesurfer` is deprecated and will be removed in
                        1.2. Use `--fs-no-reconall` instead.
  --cifti-output        output BOLD files as CIFTI dtseries
```

fmriprep/workflows/bold/base.py
```diff
@@ -647,6 +658,22 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
             (bold_surf_wf, outputnode, [('outputnode.surfaces', 'surfaces')]),
         ])
 
+        # CIFTI output
+        if cifti_output and 'template' in output_spaces:
+            cifti = pe.MapNode(GenerateCifti(), iterfield=["surface_target", "gifti_files"],
+                               name="gen_cifti")
```
effigies: Node "name" fields should match the node variable names. Makes it easier to associate code with locations in the working directory. `gen_cifti` seems more descriptive, so I'd go with that for the variable name.

effigies:
Sorry about not formatting as a single review. I'm done now.

.circleci/ds005_outputs.txt
mgxd: I was under the impression the space would be embedded within the file's metadata - I'm open to generating a `json` file but it seems a bit overkill for how few permutations we currently have

effigies: @satra Can you clarify your intent here? I was under the impression that space variants were to require a defining metadata file.

From my perspective, if we're only depending on internal metadata, then `variant-X` provides no information beyond `space-cifti`, except in that one hopes that the naming convention is consistent across datasets.

> it seems a bit overkill for how few permutations we currently have

Sure, but we have to consider the possibility that someone else is producing files and doesn't use the same `space1` and `space2` conventions.

mgxd:
@effigies re: https://github.com/poldracklab/fmriprep/pull/1001#issuecomment-379000810, it looks like `inputnode` is never added to the `workflow` within `init_sdc_wf` if *only* `epi` fieldmap types are available, so the line https://github.com/poldracklab/fmriprep/blob/1755986606806207fdc665a2a469aff2cbfd2dba/fmriprep/workflows/bold/base.py#L387 should fail. I'll add a patch within this PR

effigies:
Oh, wow. Good catch.

oesteban:
the inputnode should be added in here:

https://github.com/poldracklab/fmriprep/blob/1755986606806207fdc665a2a469aff2cbfd2dba/fmriprep/workflows/fieldmap/base.py#L158

for the case no fieldmap is found. Maybe it is only a case of real fieldmaps?

oesteban:
Sorry, I misunderstood. You are right!

fmriprep/interfaces/cifti.py
mgxd: @oesteban, did you get a chance to think about this? This is *hopefully* the last thing before merging

oesteban: I don't have time dive into this currently, so I just opened an issue for the future :+1: . I think this is good to go!

mgxd:
i think we're good to go :+1: !

fmriprep/interfaces/cifti.py
effigies: Okay, we had a discussion over this line, and we think it should be fine if you just drop the `interpolation='nearest'`. The issue is that in many cases this is going to be upsampling BOLD data to 2mm^3 voxels, which should really be interpolated.

fmriprep/interfaces/cifti.py
```diff
@@ -0,0 +1,287 @@
...
+            tr : repetition timeseries
+            download_link : URL to download label_file
+
+        Returns
+            out_file : BOLD data as CIFTI dtseries
+        """
+
+        bold_img = nb.load(bold_file)
+        label_img = nb.load(label_file)
+        if not bold_img.shape[:3] == label_img.shape:
```
effigies: You can drop this check. `nilearn.image.resample_to_img` will just return a copy if the shape and affine match.

So I'd actually write the whole loading section in two lines:

```Python
label_img = nb.load(label_file)
bold_img = resample_to_img(bold_file, label_img)
```

fmriprep/interfaces/cifti.py
mgxd: forgive my ignorance, but how is continuous interpolation applied? I understand linear and nearest

effigies: Cubic spline interpolation. It sets `order=3` in [`ndimage.affine_transform()`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.affine_transform.html).

effigies:
Awesome. In it goes.

chrisfilo:
This is great! Thank you so much @mgxd and everyone else for the feedback.

PR #1007, opened 2018-02-26
 [ENH] Adopt versioneer to compose version names

oesteban:
This not only makes more detailed the version that is used in development (particularly when using containers), it also simplifies several steps of our release workflow (no need to manually handle versions anymore).

The system has been tested on mriqc for a while now, and it is simple:

- In dev environments (`pip install -e`): versioneer parses the version and gives you the version string:
  ```
  fmriprep --version
  fmriprep v1.0.8+3.gea418f9
  ```
  A new `get_version.py` file under the root directory reads the version without having fmriprep installed:
  ```
  python get_version.py 
  1.0.8+3.gea418f9
  ```
  When my system is 3 commits after tag 1.0.8, HEAD pointing to `ea418f9`. For some reason, there is always a `g` before the commit. Haven't looked into versioneer to know why is that.

- When deployed, a `VERSION` file is written under the `fmriprep/` folder. If the file is found, that is the version reported. This file is generated before building the docker image or uploading to Pypi with the appropriate tag if releasing.


oesteban:
Okay, this is working for CircleCI. Now it is possible to know the exact version of FMRIPREP people are using - https://1310-53608443-gh.circle-artifacts.com/0/tmp/ds005/derivatives/fmriprep/sub-01.html says

```
FMRIPREP version: 1.0.8+10.gdfd8dae4
```

Which is the latest commit I submitted.

effigies:
Seems fine to me. I assume that the two new big files are basically boilerplate?

If it builds, I'm okay with it.

oesteban:
Yes, just in case versioneer is not installed

effigies:
Looks like this is breaking RTD. Can look into it if you don't have time.

oesteban:
I can check

oesteban:
Back on track

PR #1017, opened 2018-03-02
[MAINT] Refactor fieldmap heuristics

oesteban:
Handle all fieldmaps (including syn) together, so they are easier to maintain (thinking of integrating the ME-EPI with fieldmap correction and also of the `--func-only`).

When no fieldmap correction is posible, then a naive workflow connecting inputs and outputs  is set up.

Some other "better practices" are introduced in this PR:

- [x] Try to avoid passing along BIDS layout objects
- [x] Reduce the number of arguments a workflow takes
- [x] Place DataSinks for reportlets closer to their generation, but avoiding passing `reportlets_dir` on (this is done post-hoc, collecting all nodes which name starts with `'ds_report_'` and setting the proper `base_directory` and `source_file` inputs for all at once). Eventually, we will be able to remove the whole workflow of datasinks for reports.


oesteban:
Closes #1013

fmriprep/workflows/fieldmap/base.py
```diff
@@ -36,48 +36,223 @@
...
+        workflow.connect([
+            (inputnode, outputnode, [('bold_ref', 'bold_ref'),
+                                     ('bold_mask', 'bold_mask'),
+                                     ('bold_ref_brain', 'bold_ref_brain')]),
+        ])
+        return workflow
+
+    # In case there are multiple fieldmaps prefer EPI
+    fmaps.sort(key=lambda fmap: FMAP_PRIORITY[fmap['type']])
+    fmap = fmaps[0]
```
effigies: I feel like the selection logic is a bit distributed. I'd rather make decisions in one place, and then just work through the consequences. So here, I would look for two things: (1) Which primary SDC should we run? (2) Are we running SyN?

```Python
fmap = min(fmaps, key=lambda...)
run_syn = any(fm['type'] == 'syn' for fm in fmaps)
```

Yeah, it's iterating twice, but at `n < 4`, it doesn't matter. If you want to sort and use `fmaps[0]` and `fmaps[-1]`, that's fine, but I feel like this is slightly clearer in meaning.

fmriprep/workflows/fieldmap/base.py
```diff
@@ -36,48 +36,223 @@
...
+
+    # PEPOLAR path
+    if fmap['type'] == 'epi':
+        setattr(workflow, 'sdc_method', 'PEB/PEPOLAR (phase-encoding based / PE-POLARity)')
```
effigies:
```Python
workflow.sdc_method = '...'
```

fmriprep/workflows/fieldmap/base.py
```diff
@@ -36,48 +36,223 @@
...
+                     for fmap_ in fmaps if fmap_['type'] == 'epi']
+        sdc_unwarp_wf = init_pepolar_unwarp_wf(
+            bold_meta=bold_meta,
+            epi_fmaps=epi_fmaps,
+            omp_nthreads=omp_nthreads,
+            name='pepolar_unwarp_wf')
+
+    # FIELDMAP path
+    if fmap['type'] in ['fieldmap', 'phasediff']:
+        setattr(workflow, 'sdc_method', 'FMB (%s-based)' % fmap['type'])
```
effigies:
```Python
workflow.sdc_method = '...'
```

fmriprep/workflows/fieldmap/base.py
```diff
@@ -36,48 +36,223 @@
...
+                ('bold_ref_brain', 'inputnode.in_reference_brain'),
+                ('bold_mask', 'inputnode.in_mask')]),
+            (fmap_estimator_wf, sdc_unwarp_wf, [
+                ('outputnode.fmap', 'inputnode.fmap'),
+                ('outputnode.fmap_ref', 'inputnode.fmap_ref'),
+                ('outputnode.fmap_mask', 'inputnode.fmap_mask')]),
+        ])
+
+    # FIELDMAP-less path
+    if fmaps[-1]['type'] == 'syn':
```
effigies:
```Python
if run_syn:
```

fmriprep/workflows/fieldmap/base.py
```diff
@@ -36,48 +36,223 @@
...
+            (inputnode, syn_sdc_wf, [
+                ('t1_brain', 'inputnode.t1_brain'),
+                ('t1_2_mni_reverse_transform', 'inputnode.t1_2_mni_reverse_transform'),
+                ('bold_ref', 'inputnode.bold_ref'),
+                ('bold_ref_brain', 'inputnode.bold_ref_brain'),
+                ('template', 'inputnode.template')]),
+        ])
+
+        # XXX Eliminate branch when forcing isn't an option
+        if len(fmaps) == 1:  # --force-syn was called
```
effigies: I think this is clearer:

```Python
if fmap['type'] != 'syn':
```

fmriprep/workflows/fieldmap/base.py
```diff
@@ -36,48 +36,223 @@
...
+            setattr(workflow, 'sdc_method', 'FLB ("fieldmap-less" based) - SyN')
```
effigies:
```Python
workflow.sdc_method = '...'
```

Also, I don't know that "fieldmap-less based" makes sense. Maybe "fieldmap-less, SyN-based", or just "fieldmap-less".

fmriprep/workflows/fieldmap/pepolar.py
```diff
@@ -81,28 +77,22 @@ def init_pepolar_unwarp_wf(fmaps, bold_file, omp_nthreads, layout=None,
...
-        if fmap_pe[0] == bold_file_pe[0]:
-            if len(fmap_pe) != len(bold_file_pe):
-                add_list = usable_fieldmaps_opposite_pe
-            else:
-                add_list = usable_fieldmaps_matching_pe
-            add_list.append(fmap)
+    usable_fieldmaps_matching_pe = []
+    usable_fieldmaps_opposite_pe = []
+    for fmap, fmap_pe in epi_fmaps:
+        fmap_pe = fmap_pe
```
effigies: This line should be dropped.

fmriprep/workflows/fieldmap/syn.py
```diff
@@ -150,46 +145,46 @@ def init_nonlinear_sdc_wf(bold_file, freesurfer, bold2t1w_dof,
...
+            ('output_image', 'inputnode.in_file')]),
+        (unwarp_ref, outputnode, [('output_image', 'out_reference')]),
         (skullstrip_bold_wf, outputnode, [
             ('outputnode.skull_stripped_file', 'out_reference_brain'),
             ('outputnode.mask_file', 'out_mask')]),
-        (syn_rpt, outputnode, [('out_report', 'out_warp_report')])])
+    ])
 
     return workflow
+
```
effigies: Also, PEP8: Two blank lines before/after functions.

fmriprep/workflows/fieldmap/base.py
oesteban: done at the only branching where the `run_syn` is tested, thanks!

oesteban:
Testing right now on a dataset without fieldmaps at all.

Also, removed the opaque way of passing the SDC type along as an object attribute. Nipype has means for that :).

fmriprep/workflows/fieldmap/base.py
oesteban: Now it uses the workflow's outputnode

fmriprep/workflows/fieldmap/base.py
```diff
@@ -36,48 +36,223 @@
...
-        fmap
-          The estimated fieldmap itself IN UNITS OF Hz.
-        fmap_ref
-          the anatomical reference for the fieldmap (magnitude image, corrected SEm, etc.)
-        fmap_mask
-          a brain mask for the fieldmap
+    **Inputs**
+        name_source
+            Original BOLD run filename to fetch metadata (TODO: should
+            be replaced with the ``bold_meta`` and ``fmaps`` input)
```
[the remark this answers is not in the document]
oesteban: While I totally agree with this comment, I ran the extra mile and actually fixed this TODO.

oesteban:
Okay, seems to work fine with and without `--use-syn`/`--force-syn`. Thumbs up @chrisfilo, @effigies ?

effigies:
:+1:

Issue #1009, opened 2018-02-27
Multi-echo: look at the TE to select first echo


oesteban:
Since the BIDS spec does not impose an ordering with the naming (last-to-first or first-to-last), TEs should be looked into to select the "first" (shortest) echo.

Originally derived from #891.

emdupre:
Is there any appetite for bringing this up as a discussion point for the BIDS spec ? I think it's reasonable to enforce a standardize ordering&mdash; it's an issue that's come up already in https://github.com/ME-ICA/me-ica/pull/5.

Regardless, I agree that having a built-in check can only be an additional benefit. I'll address this ASAP, but wanted to see if it was worth expanding the discussion to the BIDS list. cc @chrisfilo

chrisfilo:
I think this would be a relatively small and not controversial change. If
you could send a quick email to the mailing list we could get the ball
rolling. Thanks for your continuing work on ME!

Best,
Chris

emdupre:
Emailed the list, and opened #1018 to hopefully address this for FMRIPREP !

oesteban:
Closed with #1018

PR #1018, opened 2018-03-02
[FIX] Use metadata to select shortest echo as ref_file

emdupre:
As discussed in #1009, the BIDS spec does not currently enforce that `*_echo-1*` is the shortest TE of all collected echos. This should patch the current sorting procedure to grab the shortest echo and treat as the `ref_file`.

### DIFF
--- a/fmriprep/workflows/bold/base.py
+++ b/fmriprep/workflows/bold/base.py
@@ -450,9 +450,8 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
     if multiecho:
         inputnode.iterables = ('bold_file', bold_file)
 
-        me_first_echo = pe.JoinNode(interface=FirstEcho(),
-                                    joinfield=['in_files',
-                                               'ref_imgs'],
+        me_first_echo = pe.JoinNode(interface=FirstEcho(te_list=tes),
+                                    joinfield=['in_files', 'ref_imgs'],
                                     joinsource='inputnode',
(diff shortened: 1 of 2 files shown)


emdupre:
Sorry, I forgot to put WIP in the tag title before, but I think this is ready now !

fmriprep/workflows/bold/base.py
```diff
@@ -222,7 +222,11 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
         ref_file = bold_file
     else:
         multiecho = isinstance(bold_file, list)
-        ref_file = sorted(bold_file)[0] if multiecho else bold_file
+        if multiecho:
```
oesteban: Lines 219-230 can be simplified:

```Python
    ref_file = bold_file
    mem_gb = {'filesize': 1, 'resampled': 1, 'largemem': 1}
    bold_tlen = 10
    multiecho = isinstance(bold_file, list)

    if multiecho:
        tes = [layout.get_metadata(echo)['EchoTime'] for echo in bold_file]
        ref_file = dict(zip(tes, bold_file))[min(tes)]

    if os.path.isfile(ref_file):
        bold_tlen, mem_gb = _create_mem_gb(ref_file)
```

emdupre: But then there's no "/completely/made/up/path/" ! 😆 Updating now.

Issue #998, opened 2018-02-21
Very short TE EPIs and other adjustments of SDC-SyN


oesteban:
[Here](https://github.com/poldracklab/fmriprep/issues/997#issuecomment-367129621) I suggested that, in situations like DS000148 where EPIs are acquired with very short TEs, it might be interesting not to invert the T1w image.

@effigies also suggested using T2w images [whenever possible](https://github.com/poldracklab/fmriprep/issues/997#issuecomment-367165775). I argued that in general I totally agree, but didn't see it clear for the particular case of very short TE EPIs.

Additionally, I suggested that, when we are running BBR after a successful SDC method (regardless which one) then the BBR should use only 6 dof. The rationale is that any affine/nonlinear deformation w.r.t. the T1 contours should have been captured by the SDC. Enabling 3 additional DoFs is *asking for trouble* plagiarizing the expression from Chris (even though he used it in a close but different context).

PR #1021, opened 2018-03-03
[MAINT,WIP] DataSinks within their workflows

oesteban:
Considerably eases reading workflows and particularly the base workflow. (This is based off of #1017, so it depends on that one).


.circleci/ds000210_outputs.txt
```diff
@@ -22,8 +22,6 @@ fmriprep/sub-02/figures/sub-02_T1w_seg_brainmask.svg
 fmriprep/sub-02/figures/sub-02_T1w_t1_2_mni.svg
 fmriprep/sub-02/figures/sub-02_task-cuedSGT_run-01_echo-1_bold_flirt.svg
 fmriprep/sub-02/figures/sub-02_task-cuedSGT_run-01_echo-1_bold_rois.svg
-fmriprep/sub-02/figures/sub-02_task-cuedSGT_run-01_echo-2_bold_rois.svg
```
oesteban: I'm aware of this. CompCor does not make much sense for ME datasets as we plan to integrate MEICA

fmriprep/workflows/bold/base.py
```diff
@@ -542,10 +518,10 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
                                            mem_gb=mem_gb['filesize'],
                                            omp_nthreads=omp_nthreads)
 
-            subset_reg_reports = pe.JoinNode(niu.Select(index=0),
-                                             name='subset_reg_reports',
-                                             joinsource=inputnode,
-                                             joinfield=['inlist'])
+            # subset_reg_reports = pe.JoinNode(niu.Select(index=0),
```
oesteban: Temporary patches while I review the multiecho path

effigies:
Reportlet sinking within subworkflows seems reasonable. Splitting out into their own workflow was mostly to make the graphs easier to read.

For derivatives, I do still prefer to keep them somewhat separate for the reason that if an output should be in the derivatives, it should be accessible to nodes downstream of the workflow. The derivatives-sinking workflows help to enforce that.

It looks like you're modifying reportlets only, so I'm fine with this in general. Do you have particular things you want comment on, or just a general code review?

oesteban:
> It looks like you're modifying reportlets only, so I'm fine with this in general. Do you have particular things you want comment on, or just a general code review?

Yes, this idea only applies to the reportlets (and I totally agree with you w.r.t. derivatives). Just a general code review will suffice.

oesteban:
All green here :)

fmriprep/workflows/bold/base.py
```diff
@@ -706,89 +677,28 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
...
-    if use_aroma:
-        workflow.connect([
-            (inputnode, ds_ica_aroma_report, [('source_file', 'source_file'),
-                                              ('ica_aroma_report', 'in_file')]),
-        ])
+    # Fill-in datasinks of reportlets seen so far
+    for node in workflow.list_node_names():
+        if node.split('.')[-1].startswith('ds_report'):
+            workflow.get_node(node).inputs.base_directory = reportlets_dir
+            workflow.get_node(node).inputs.source_file = ref_file
```
effigies: This seems a little fragile, in the sense that, if somebody forgets to start their datasink node with `ds_report` some time down the way, it may take some time to track down. But if it makes life a lot easier, I guess this is fine.

fmriprep/workflows/bold/base.py
effigies: What's the story with multiecho? Was it merged prematurely, or is it just related to this PR? (I haven't been paying much attention.)

If the latter, does it make sense to hold this PR until that's resolved?

oesteban: This particular joinnode can be removed: it merged the 3 echos in only one EPI-to-T1w reportlet. Since now that reportlet is generated within the workflow (and one of the features added in that PR is joining the echoes), then only one reportlet is added. When the `--t2-coreg` option is not set, it naturally generates one registration plot per echo (which we may want to change and register only the first echo, but that was outside the scope of that PR).

The rest of commented out comment refers to the ROIs plot, which will be revisited with the upcoming implementation of MEICA.

fmriprep/workflows/bold/base.py
oesteban: The alternative is passing in the base_directory and the source_file to all depths of workflows (which basically led us to place these datasinks at the top level in first place).

.circleci/ds000210_outputs.txt
emdupre: For ME datasets, we're only maintaining the first echo's CompCor for now? Why the first over the second echo, or whichever echo is closest to 30ms?

fmriprep/workflows/bold/base.py
oesteban: Okay, I added this to the code-conventions wiki https://github.com/poldracklab/fmriprep/wiki/Code-conventions

oesteban: Would you feel more confortable now that this constraint is documented?

On the other hand, missing the `ds_report` naming will cause the reportlet to be dismissed and that should be easy to spot on the final report.

.circleci/ds000210_outputs.txt
oesteban: Yes, you are right. I started this before we got your PR merged. I'll also fix this before merging.

The other open thread about ME is fieldmap correction. Making that easy was part of the reasoning under #1017.

oesteban: Okay @emdupre, I've revised this code and it should:
- Enable fieldmap correction on the T2* map
- Generate CompCor masks for the shortest echo (with your PR)
- Generate one _preproc derivative per echo.

Please give it a quick check :)

emdupre: https://github.com/oesteban/fmriprep/blob/39113d9614e6ee62425c074811fd06351cee46a8/fmriprep/workflows/bold/base.py#L526 
It seems like this node is no longer connected ? Unless I'm overlooking it. Is that intended to connect to the confounds wf?

emdupre:
This looks great, @oesteban ! I'm curious how fieldmap corrections on the T2* map will look&mdash; I've never actually seen them combined.

emdupre:
Ok, looking at the [derived report](https://1606-53608443-gh.circle-artifacts.com/0/tmp/ds000210/derivatives/fmriprep/sub-02.html), I'm actually a little nervous about these changes. I'm not sure if it's just the down-sampled data again, but it seems as though the SDC correction is removing a lot of the benefits of the T2* map in terms of clearly defining tissue boundaries. Have you run this on non-down-sampled data, by chance ?

oesteban:
@emdupre what about this:

- We use the first echo for SDC (whatever method is run).
- We then feed corrected images into the T2* calculation

is this theoretically consistent with the multiecho foundations?

emdupre:
I'm not sure how to answer without knowing: How do you think SDC-SyN will interact with very short TEs ? I'm unclear from #998 what the conclusion was, there.

In terms of theoretically consistent, if the correction is applied across echos (s.t. echos are in the same space), it should be ok. In practice, I've never seen it used. The idea of the T2* map is that you calculate which echo shows the strongest signal at a voxelwise level, so that the resulting image shows high tissue contrast across cortex. This also allows us to capture regions where we usually see strong signal dropout, like OFC. 

That being said, I might just not be familiar with it as I've never collected multi-band, multi-echo data-- I'm not sure if @prantikk has and has seen field maps (or a SDC-SyN like correction) used, though.

oesteban:
Regarding #998 we could use some echo with larger T2* contrast for SyN, but I'd would try that only if we see that the results with the first echo are always bad. That issue is more concerned about the EPI-to-T1 registration, as BBR methods require T2 contrast. It is relevant for ME when `--t2-coreg` is not set. In that case we would have two options:
 - Make `--t2-coreg` the default (I would prefer this)
 - Use the echo closest to ~30ms.

But all of that can be addressed in different PRs.

For this one, let's check that SDC can be estimated on the first echo (as we are doing for HMC), then apply it for all echos and finally check that the T2* still looks good.

Please @prantikk do not hesitate to jump in if you see we are doing something wrong.

.circleci/ds000210_outputs.txt
oesteban: Okay, just could check on this. So you extracted the echo for the reports. But now they happen within their generating workflow. So it wasn't necessary anymore. I'll remove this node.

oesteban:
Hi, I think I found the problem @emdupre was pointing out before:

> Ok, looking at the derived report, I'm actually a little nervous about these changes. I'm not sure if it's just the down-sampled data again, but it seems as though the SDC correction is removing a lot of the benefits of the T2* map in terms of clearly defining tissue boundaries. Have you run this on non-down-sampled data, by chance ?

Looking at your code, I've found you were using nearest interpolation:

https://github.com/poldracklab/fmriprep/blob/e842cd1ce477063e7d8299c978ce4a0f9ad9bd33/fmriprep/workflows/bold/t2s.py#L76

When I've switched this to Lanczos, the contrast is looking bad again. My intuition is that Lanczos introduces some small negative values and those are making exponentials blow up. Is that possible?

On the other hand, applying nearest interpolation to correct for head motion may defeat the purpose of head motion correction. I would suggest using lanczos, but with the following updates to the interface that calculates the T2*:

- First remove any value below zero from all echoes.
- Enable passing in a mask to the interface, so the T2* is calculated only within mask.

WDYT?

oesteban:
One more thing to add to the T2* calculation:
- Crop out outliers: some very bright pixels force a compression of the histogram. Identify those and set them to a saturation value (a high percentile). Make sure that the output T2* fits in a int16 image.

prantikk:
Could someone fill me in on the SDC correction?  Be very careful about
modifications to the T2* map, or the data that is input to the map
calculation.  Doing so will influence the Kappa and Rho calculation, which
may in turn affect component selection.

oesteban:
> Could someone fill me in on the SDC correction?

Sure, the question is whether the different echoes should be corrected for susceptibility distortions before constructing the T2* map. Since the dropout is proportional to the TE, with the current implementation there is a very clear gradient on the T2* map corresponding to this difference in signal dropout.

FMRIPREP would calculate a field of displacements that accounts for such deformations, and then this nonlinear warping is composed with the affine calculated by head-motion correction. This way, FMRIPREP is able to account for both corrections with only one resampling step.

> Be very careful about modifications to the T2* map, or the data that is input to the map calculation.

What would your general opinion about, e.g. slice-timing correction? If I interpret this correctly, you would not apply any data transformation to the input echoes, is that correct?

Particularly, as implemented right now, the distribution of the resulting T2* map generally shows an extremely high kurtosis, because of these very bright outliers found around the brain edge. I'd say it is safe to remove those outliers, they would probably go away if the T2* map were calculated only within the brain mask.

Please, let me know if these comments make sense.

oesteban:
Okay, as regards the T2* map, this PR will not modify the current implementation (it wasn't the idea in first place):

- this PR: https://1691-53608443-gh.circle-artifacts.com/0/tmp/ds000210/derivatives/fmriprep/sub-02.html
- current master: https://2324-53175327-gh.circle-artifacts.com/0/tmp/ds000210/derivatives/fmriprep/sub-02.html

The only difference is that now, it shows just one ROIs plot for the reference echo. However, preprocessed versions of all echoes are still stored and the T2* map looks the same.

I think we can merge (after fixing the extra newline in one file) this and keep the conversation about SDC in #1031 and the modifications to the echo calculations in #1034.

PR #1024, opened 2018-03-05
[ENH] Dynamic citation boilerplate

oesteban:
Compose a citation boilerplate in a per-workflow basis, with up-to-date software versions and generated based on the actual workflows being run.

This PR is more of a request for comments from @effigies, @chrisfilo.

Maybe this can be done already with duecredit (cc @yarikoptic), but the basic idea here is to place a mechanism to build literate descriptions of the workflow. My plan would be then collect all the references with duecredit.

How it works. Workflows in FMRIPREP will have a `__predesc__` and a `__postdesc__` attributes (in this version the pre- dunder is just `__desc__` as I was just prototyping). When writing a new workflow, these fields can be used to add the corresponding literate description of what that workflow does (allowing to dynamically replace version strings, optional parameters, etc).

Let me know if you think this is worth exploring. The most challenging culprit happens with inhomogeneous datasets (e.g. fieldmap has `IntendedFor` only on 3 out of 5 runs). In that case the citation boilerplate will have two descriptions, as there's is not just one.

Any comments will be very welcome.

EDIT

# Results

This is how it looks
boilerplate

# Links to more results
In the context of the reports
- https://2290-53608443-gh.circle-artifacts.com/0/tmp/ds054/derivatives/fmriprep/sub-100185.html#boilerplate
- https://2292-53608443-gh.circle-artifacts.com/0/tmp/ds005/derivatives/fmriprep/sub-01.html#boilerplate
- https://2294-53608443-gh.circle-artifacts.com/0/tmp/ds210/derivatives/fmriprep/sub-02.html


chrisfilo:
I have a feeling this would be very challenging to do if the goal is to provide a snippet of text ready to paste into the manuscript. A solution based on conditional statements mirroring the one we did on the website might be easier to implement (even though harder to maintain).

oesteban:
IMHO implementation is fairly natural since each `Workflow` factory function can (this is not mandatory) set the `__predesc__` and `__postdesc__` attributes and they are automatically picked up the same way nipype generates the graph. Having them set right after the workflow instantiation feels pretty natural, just like adding another docstring.

For this PR I didn't get to the bottom and generate the exact citation boilerplate we have currently online just for laziness, but this prototype is able to do that already.

I think we could leverage duecredit to generate the lookup table of references (@yarikoptic can correct me if I'm wrong).

On the other hand, I see some positive side effects:
  - Trivial to append the boilerplate to the report (minimizing work for the user)
  - Trivial to maintain the citing.rst. Also, the citing.rst can be generated automatically, becoming an actual test for this feature.
  - Automatically include these new dunder fields when generating the documentation will make it very easy to highlight these excerpts with their corresponding citation in documentation proper.

I'll let this sit here while I focus on more urgent issues.

oesteban:
Hey @satra, what do you think about this extension to nipype?. Particularly the magic:

https://github.com/oesteban/fmriprep/blob/b349694455eac1fbbaa7c6ee1b6a133b71c980e4/fmriprep/engine/workflows.py#L14-L46

satra:
@oesteban - i think this a reasonable starting point that could then be edited by a human. i agree with @chrisfilo that this can get complicated. but this is at least a simple mechanism by which certain workflows could be described. how about adding a flag to allow ignoring nested workflows.

oesteban:
> how about adding a flag to allow ignoring nested workflows

Seems like a good idea.

satra:
@oesteban - for the references it may be nice if this supported duecredit as well (perhaps @yarikoptic can chime in on how). then for other use cases we can use the references as well. also this will help move the references to the underlying interfaces since nipype does have support for duecredit.

oesteban:
That's why I initially pinged @yarikoptic here.

At this point, I think that for fmriprep it is easier to maintain a single bibtex file. But, if we decide to upstream to nipype then I'm sure that coordinating with duecredit is a good idea.

satra:
@oesteban - sounds good.

oesteban:
This is getting there.

Now, we can provide some text within the reports (minor issues still open, such as why the zenodo references of fmriprep and nipype do not work) and also a new CITATION.md file with markdown. The markdown version is used when pandoc is not installed.

I also want to generate a Latex version, for copy&paste.

WDYT?

chrisfilo:
I think there is an issue with one of the citations: 
image

fmriprep/data/boilerplate.bib
```diff
@@ -0,0 +1,260 @@
...
+	author = {Esteban, Oscar and Markiewicz, Christopher and Blair, Ross W and Moodie, Craig and Isik, Ayse Ilkay and Erramuzpe Aliaga, Asier and Kent, James and Goncalves, Mathias and DuPre, Elizabeth and Snyder, Madeleine and Oya, Hiroyuki and Ghosh, Satrajit and Wright, Jessey and Durnez, Joke and Poldrack, Russell and Gorgolewski, Krzysztof Jacek},
+	title = {FMRIPrep: a robust preprocessing pipeline for functional MRI},
+	year = {2018},
+	doi = {10.1101/306951},
+	journal = {bioRxiv}
+}
+
+@article{fmriprep2,
+ 	author = {Esteban, Oscar and Blair, Ross and Markiewicz, Christopher J. and Berleant, Shoshana L. and Moodie, Craig and Ma, Feilong and Isik, Ayse Ilkay and Erramuzpe, Asier and Kent, James D. andGoncalves, Mathias and DuPre, Elizabeth and Sitek, Kevin R. and Gomez, Daniel E. P. and Lurie, Daniel J. and Ye, Zhifang and Poldrack, Russell A. and Gorgolewski, Krzysztof J.},
+ 	title = {FMRIPrep 1.1.2},
```
chrisfilo: We need to figure out how to cite the right version here. Especially considering that the zenodo handle is created after the release is made.

chrisfilo:
overall looks very cool!

fmriprep/data/boilerplate.bib
oesteban: I think I may have the solution. I just found this on the Zenodo record of fmriprep:

> **Cite all versions?** You can cite all versions by using the DOI 10.5281/zenodo.852659. This DOI represents all versions, and will always resolve to the latest one. Read more.

We could reset the version in the bib file during deployment. Would that reply this concern?

oesteban:
Ready for review!

oesteban:
> The display looks off for Markdown and LaTeX

Yep, I'm working this out right now.

oesteban:
Okay, with my last commit this should be ready to go.

- [x] Fixed visualization
- [x] Fixed @effigies' minor comments
- [x] Fixed @effigies' major comment about the grayordinates file - I have actually added the HCP pipelines reference and cited the paper in two contexts: the grayordinates files and the epi unwarp with phase difference fieldmap (since that workflow is inspired in HCP pipelines).

This is how it looks now (locally, awaiting for tests in circle):
new-boilerplate-0

and
new-boilerplate-1

.circleci/ds005_outputs.txt
```diff
@@ -1,5 +1,7 @@
 fmriprep
 fmriprep/logs
+fmriprep/logs/CITATION.html
+fmriprep/logs/CITATION.md
```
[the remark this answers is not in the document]
oesteban: Done!

fmriprep/workflows/bold/resampling.py
```diff
@@ -92,6 +92,8 @@ def init_bold_surf_wf(mem_gb, output_spaces, medial_surface_nan, name='bold_surf
         workflow.__desc__ = """\
 The BOLD time-series, were resampled to surfaces on the following
 spaces: {out_spaces}.
+*Grayordinates* files [@hcppipelines], which combine surface-sampled
+data and volume-sampled data, were also generated.
```
effigies: Do we want to make this conditional on `cifti_output`? In which case, it might go better back in `init_func_preproc_wf`, which has access to that variable.

oesteban: Thanks, I've just added it to `init_func_preproc_wf` :+1:

oesteban:
I'm merging this.

> @chrisfilo 
> We need to figure out how to cite the right version here. Especially considering that the zenodo handle is created after the release is made.

I'm assuming this comment is not meant to block merging this. I've opened #1229 to follow up on this.

PR #1026, opened 2018-03-06
[WIP] PIN: pybids/grabbit-fixes

effigies:
Testing grabbit/pybids refactor to verify incf/pybids#135 doesn't break FMRIPREP.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -247,12 +247,6 @@ jobs:
     steps:
       - attach_workspace:
           at: /tmp
-      - restore_cache:
-          keys:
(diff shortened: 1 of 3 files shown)


oesteban:
To discard a cache in a commit, just update the version of the cache.

Instead of deleting entries for the key `ds005-anat-v4`, replace them with `ds005-anat-v5` and that will discard the cache.

effigies:
Yeah, I understand how to use it. But there's no reason to increment the cache here. I don't want to pollute the name space, I just want to drop the cache to verify there are no issues.

oesteban:
:+1:

PR #1033, opened 2018-03-21
Revert "[ENH] Add a new submodule to maintain notebooks"

oesteban:
Reverts poldracklab/fmriprep#1030


oesteban:
Read the docs does not support filtering submodules out (https://github.com/rtfd/readthedocs-build/issues/30), so I'll revert this addition until we can make the target repo public.

Issue #1035, opened 2018-03-26
Document BBR fallback.


danlurie:
It would be useful to have this feature mentioned in the documentation, including how `fmriprep` decides when to apply it.

effigies:
Related: #681, #694, nipy/nipype#2198

PR #1036, opened 2018-03-26
Add BBR fallback to user docs.

danlurie:
Addressed #1035


effigies:
Oh, also include `[docs only]` in your commit messages to skip time-consuming tests.

oesteban:
Alternatively, you would need to use all the skip * patterns: `[skip ds005][skip ds054][skip ds210][skip tests] ...`

effigies:
Ugh. This failure looks like a nibabel bug.

PR #1037, opened 2018-03-30
CI: Minor change to trigger builds

effigies:
Seeing if error in gh-1001 is hitting master

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -101,3 +101,5 @@ For more details, please see :ref:`citation`.
 .. image:: https://badges.gitter.im/poldracklab/fmriprep.svg
    :alt: Join the chat at https://gitter.im/poldracklab/fmriprep
    :target: https://gitter.im/poldracklab/fmriprep?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge
+
+


mgxd:
@effigies same error is present in #1036 as well

effigies:
@oesteban @emdupre Any notion as to why the ds210 tests are failing recently? We may need to upstream this bug to nibabel, but a minimal reproduction would be good to have first.

effigies:
@mgxd I thought I'd seen this issue before. Apparently I commented on it there, too.

emdupre:
I'm really not sure what triggered it to start failing&mdash; it does seem like an introduced dependency error to me, but only niworkflows cut a version recently.

I'll keep playing around with it in #1038 and try pinning the dependencies, but given that [we don't even have STC implemented for multi-echo](https://github.com/poldracklab/fmriprep/blame/83c4f1e0682b788cc1bac7831b218d5bf322821f/fmriprep/workflows/bold/base.py#L270), I'm erring towards amending that check for now. WDYT?

oesteban:
https://github.com/poldracklab/fmriprep/pull/1001#issuecomment-379034447

PR #1038, opened 2018-03-31
[WIP, TST] Pass ref_file to STC check

emdupre:
Equating STC check for multi-echo and single-echo bold data-- trying to debug the error in ds210 tests seen in #1037 and #1036.


oesteban:
https://github.com/poldracklab/fmriprep/pull/1001#issuecomment-379034447

PR #1040, opened 2018-04-04
[ENH] Revise resampling grid for template outputs

### LINKED ISSUE #337: Add option to us 2mm voxels size for the MNI transformation

oesteban:
This PR resurrects the dicussion about native or other output resolutions. ~~In particular, this PR CHANGES the current default (the "native" space of the input BOLD file) and sets the template-2mm version~~.

Implications:

  * As off this PR, any new template added to FMRIPREP will require having a 1mm and a 2mm resolution versions. This was probably necessary already, but from this PR on it will be mandatory for sure.
  * The default behavior until v1.0.8 can be obtained using ``--output-grid-reference native``.
  * The option has been renamed: from ``--output-grid-reference`` to ``--template-resampling-grid``, as it is more descriptive.

Justification:

  * Comparison studies need not only a standard (physical) space, but also a standard voxel size (digital space) for the good of analysis.
  * Cluster analyses on images with very large voxels may be very hard to interpret.

Since it's been a long overdue request, I thought it was time to change this default. Closes #337.

Modified ds005 test on circleci with ``--template-resampling-grid native``, so that we test both options.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -417,6 +417,7 @@ jobs:
                 /tmp/data/ds054 /tmp/ds054/derivatives participant \
                 --fs-no-reconall --debug --force-syn \
                 --output-space T1w template \
+                --template-resampling-grid native \
                 --mem_mb 4096 --nthreads 2 -vv
       - run:
           name: Checking outputs of fMRIPrep
@@ -520,6 +521,7 @@ jobs:
                 --config $PWD/nipype.cfg -w /tmp/ds000210/work \
                 /tmp/data/ds000210 /tmp/ds000210/derivatives participant \
                 --fs-no-reconall --t2s-coreg --use-syn-sdc \
+                --template-resampling-grid native \
                 --debug --write-graph --mem_mb 4096 --nthreads 2 -vv 
       - run:
           name: Checking outputs of fMRIPrep
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -330,7 +330,7 @@ jobs:
                 --debug --write-graph --use-syn-sdc --use-aroma \
                 --ignore-aroma-denoising-errors --mem_mb 4096 \
                 --output-space T1w template fsaverage5 \
-                --template-resampling-grid 2mm \
+                --template-resampling-grid native \
                 --nthreads 2 -vv
             sudo find /tmp/ds005/work -not -name "*.svg" -not -name "*.html" -not -name "*.rst" \
                 -not -name "*.mat" -not -name "*.lta" -type f -delete
@@ -417,7 +417,7 @@ jobs:
                 /tmp/data/ds054 /tmp/ds054/derivatives participant \
                 --fs-no-reconall --debug --force-syn \
                 --output-space T1w template \
-                --template-resampling-grid native \
+                --template-resampling-grid 2mm \
                 --mem_mb 4096 --nthreads 2 -vv
       - run:
           name: Checking outputs of fMRIPrep
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -330,7 +330,7 @@ jobs:
                 --debug --write-graph --use-syn-sdc --use-aroma \
                 --ignore-aroma-denoising-errors --mem_mb 4096 \
                 --output-space T1w template fsaverage5 \
-                --template-resampling-grid native \
+                --template-resampling-grid 2mm \
                 --nthreads 2 -vv
             sudo find /tmp/ds005/work -not -name "*.svg" -not -name "*.html" -not -name "*.rst" \
                 -not -name "*.mat" -not -name "*.lta" -type f -delete
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -330,6 +330,7 @@ jobs:
                 --debug --write-graph --use-syn-sdc --use-aroma \
                 --ignore-aroma-denoising-errors --mem_mb 4096 \
                 --output-space T1w template fsaverage5 \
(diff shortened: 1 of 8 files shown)


docs/workflows.rst
```diff
@@ -45,7 +45,7 @@ is presented below:
                                 fmap_demean=True,
                                 use_syn=True,
                                 force_syn=True,
-                                output_grid_ref=None,
+                                output_grid_ref='native',
```
chrisfilo: we should make this variable name consistent with the command line option name

fmriprep/cli/run.py
```diff
@@ -132,7 +132,7 @@ def get_parser():
         choices=['MNI152NLin2009cAsym'], default='MNI152NLin2009cAsym',
         help='volume template space (default: MNI152NLin2009cAsym)')
     g_conf.add_argument(
-        '--output-grid-reference', required=False, action='store', default=None,
+        '--template-resampling-grid', required=False, action='store', default='2mm',
         help='Grid reference image for resampling BOLD files to volume template space. '
              'It determines the field of view and resolution of the output images, '
              'but is not used in normalization.')
```
chrisfilo: filename or keyword - we should list allowed keywords

oesteban:
I can set `'native'` to be the default, no problem (although I'm not quite convinced that it is better). Particularly, I'm not sure that the default of MNI 2mm will cause "_increasing the size of the output images (quite a bit for multi-band sequences)_". This is very true for (now legacy) datasets with very big voxel sizes. But multiband datasets are usually around the 2mm voxel size... so the size increase may not be as large.

For me, the biggest reason for the "more standard" 2mm space is random field theory analysis.

Meanwhile, I'll work on those specific comments on the code above 👍

chrisfilo:
True, but not all multiband are 2mm (many are 2.2) and they usually have smaller field of view than the MNI template - multiply the difference in voxel count by 900 volumes (short TR) and you get a large spike in size. We run into this with the Beast quite a bit.

As for voxel size - it's important we think about this, but random field theory should not be dependent on the input resolution. Do you have any references to show that RFT is only valid if inputs are 2mm? I find it quite unlikely... Would that make RFT not applicable to ultra high field (1.5mm) fMRI?

oesteban:
I can remember the discussion about size for the T1w space output (and I think that was a good call). However, I don't think we ever thought about it for the MNI space. Basically we extended the conclusions to the MNI space (probably thinking of the 1mm version, only). Then @poldrack suggested #337.

RFT does not depend on the output resolution, however field modeling may break if the field is not sampled densely enough. So, I didn't mean that you can only use RFT iff the voxel size is 2mm, I meant that it may not be useful, for instance, along Z when you have only 40 slices of 4mm thickness.

I don't know enough to say what would be a reasonable threshold. But I'd say that for RFT it is better to have 110 voxels @ 2.0mm than 100 @ 2.2mm (and even better 220 voxels @ 1mm).

Of course, you can always run your analysis at the lower (native) resolution and then upsample the final stats maps (which are not as big) to 2mm isotropic using an apt interpolation (b-splines?). However, this would mean changing the way people run their analyses (and may confront with the aspirations of FMRIPREP not to modify their ways and be limited just to preprocessing).

chrisfilo:
Interesting - do you know of any papers that support the thesis that RFT somehow improves when you upsample the input data prior to estimating smoothness and fitting the random field model?

chrisfilo:
This seems relevant (but not conclusive): https://www.frontiersin.org/articles/10.3389/fnhum.2017.00345/full

chrisfilo:
Also the thread with AFNI simulations https://twitter.com/Neuro_Skeptic/status/886620819803303936?s=19

oesteban:
@effigies 

> just go for the nearest isotropic voxel size, so `3x3x4` -> `3x3x3`.

This seems like a great idea to me.

@chrisfilo @effigies, what about the default? should I roll back to `'native'`?

effigies:
I would say that if we're doing a 1.0.9, I'd keep `'native'` as default. For 1.1, I think it's a question of whether we think 2x2x2 is actually a better approach.

.circleci/config.yml
```diff
@@ -330,6 +330,7 @@ jobs:
                 --debug --write-graph --use-syn-sdc --use-aroma \
                 --ignore-aroma-denoising-errors --mem_mb 4096 \
                 --output-space T1w template fsaverage5 \
+                --template-resampling-grid 2mm \
```
effigies: How about we set this option on to `native` on ds054, just so that if/when we change this, we don't have to remember to set one back to native for testing.

oesteban:
@chrisfilo a more recent related discussion: https://twitter.com/CyrilRPernet/status/980865846217203712

.circleci/config.yml
oesteban: done

effigies:
LGTM. This gets access to 2x2x2 resolution for tests. Maybe let's punt on the `'nearest'` option and finalizing the 1.1.0+ default to another issue/PR?

oesteban:
wfm!

chrisfilo:
@oesteban this paper talks about dimensionality (1D, 2D, 3D) not resolution (mm) right?

PR #1046, opened 2018-04-10
[FIX] Connect inputnode to SDC for pepolar images

oesteban:
As a result of refactoring, this node got disconnected for that particular use-case.

This PR fixes that (https://github.com/poldracklab/fmriprep/pull/1001#issuecomment-379000810)

Thanks @mgxd for spotting this problem.


oesteban:
Well, this is the only SDC path that we don't test in Circle.

PR #1048, opened 2018-04-11
CI: Re-run ds005 with only one BOLD run

effigies:
Based on discussion in #1047, I thought perhaps we could run a 1-run test on CircleCI without adding much to the runtime. This makes a copy of the working directory, deletes one run from the source data, and re-runs.

Ideally the ds005 test should fail pre-#1047 and pass afterwards.


oesteban:
I guess this is the pre-#1047. Could you merge that one in here?

effigies:
This isn't quite failing for the right reasons. There are permissions issues to address, first.

oesteban:
Gotcha

effigies:
Okay, failing at the right spot. Merging #1047.

PR #1070, opened 2018-04-26
[DOC] Add preprint fig. 1 to About

oesteban:


effigies:
This looks good in the docs, but not so much on the GitHub front page.

https://github.com/oesteban/fmriprep/tree/docs/add-resources-from-preprint#fmriprep-a-robust-preprocessing-pipeline-for-fmri-data

Is there some way to make this work for both?

oesteban:
This may work out :+1:

effigies:
Yup, works in both places.

PR #1050, opened 2018-04-12
[ENH/WIP] Robust EPI Mask for heavily nonuniform image

ZhifangYe:
Ref #1000 

When processing heavily nonuniform bold data (eg, from 64 channels head coils), the MaskEPI step in init_enhance_and_skullstrip_bold_wf often give bad results. 

This PR tweaks some parameters in MaskEPI and adds an addition N4 correction step before MaskEPI. I have tested the code on a multiband dataset and several legacy images. Generally it gives better or equal results with little additional computation cost.

Issue #997, opened 2018-02-20
Very short TE EPIs are challenging for BBR


oesteban:
DS000148 has a TE of 18ms, what yields a contrast like:
<img width="383" alt="screen shot 2018-02-20 at 12 28 00 pm" src="https://user-images.githubusercontent.com/598470/36448308-0e73e0a2-163c-11e8-8bb7-8e3c1caa4c5b.png">

This makes it very challenging for BBR to find the WM/GM boundary.

effigies:
Looks like a pretty intense bias field, too. How does the enhance and skull strip do?

chrisfilo:
Would be useful to see the contrast enhanced and bias corrected reference volume that BBR "sees".

oesteban:
This is one of the few runs that worked in sub-09 (run-02):
<img width="545" alt="ds148-bbr" src="https://user-images.githubusercontent.com/598470/36450524-0e08a5e2-1643-11e8-9152-2f53847916e8.png">

Most of the runs fell back to the `mri_coreg` initialization. As you can see, the 9 parameters caused that bbregister expanded the image.

Two things we should consider:
- if SDC was run, then limit the BBR to 6 parameters (don't allow those 3 free parameters to overfit an inexistent deformation after SDC).
- for SDC-SyN: when the TE is this short and the intensity of CSF is below the WM and GM then bypass the T1 inversion.

Other than that, taking into consideration the original contrast, I don't think the reference generation is performing badly.

oesteban:
Please note that in the above picture I made sure the reportlet was 100% at the "moving" frame.

oesteban:
@effigies I see what you meant before about the inhomogeneity, would you agree that it is not present anymore in the reference?

effigies:
Yes, but is that from the same series as the first image in this thread?

oesteban:
This would be DS148/sub-09 (run-02), the [same as in my previous comment](https://github.com/poldracklab/fmriprep/issues/997#issuecomment-367129621).
<img width="385" alt="screen shot 2018-02-20 at 3 36 11 pm" src="https://user-images.githubusercontent.com/598470/36455172-d494a958-1653-11e8-802c-a0ec1b1e6bb9.png">

oesteban:
Less INU, but still present

oesteban:
@effigies, what do you think about not inverting the T1 in SDC-SyN for these kind of images?

effigies:
[...]

PR #1074, opened 2018-04-26
TEST: Add enhance-and-skullstrip regression tests

effigies:
Just to track progress on this, I think this is the rough framework we need for testing.

I picked this symmetric overlap metric (taken from [doi:10.1371/journal.pone.0007200](http://journals.plos.org/plosone/article?id=10.1371/journal.pone.0007200#article1.body1.sec4.sec3.p3)) because it's sensitive to increases and decreases in mask size, as well as shifts.

Next I'll need a list of images to make reference files from.

### DIFF
--- a/fmriprep/workflows/bold/tests/test_util.py
+++ b/fmriprep/workflows/bold/tests/test_util.py
@@ -25,7 +25,9 @@ def symmetric_overlap(img1, img2):
     (os.path.join(os.getenv('FMRIPREP_REGRESSION_SOURCE', ''),
                   base_fname),
      fname_presuffix(base_fname, suffix='_mask', use_ext=True,
-                     newpath=os.getenv('FMRIPREP_REGRESSION_TARGETS', '')))
+                     newpath=os.path.join(
+                         os.getenv('FMRIPREP_REGRESSION_TARGETS', ''),
+                         os.path.dirname(base_fname))))
     for base_fname in (
         'ds000116/sub-12_task-visualoddballwithbuttonresponsetotargetstimuli_run-02_bold.nii.gz',
         # 'ds000133/sub-06_ses-post_task-rest_run-01_bold.nii.gz',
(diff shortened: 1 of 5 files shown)


effigies:
@oesteban Do you have a list of images? Otherwise I can randomly sample OpenFMRI on datalad.

oesteban:
Oh, sorry. I think we can start with:

ds000237/sub-0[3,6] (#963)
ds000116/sub-12_task-visualoddballwithbuttonresponsetotargetstimuli_run-02 (#907)
ds000133/sub-06_ses-post_task-rest_run-01 (#905)
ds000140/sub-32_task-heatpainwithregulationandratings_run-02 (#904)
ds000157/sub-23_task-passiveimageviewing (#902)

I would also add the three echos from any of the participants in ME datasets (ds000210, ds000216).

And there was ds000148 with very short TEs which could be hard (#997).

You would probably need some reference that works. I'd take them from ds000030 which is a dataset we heavily massaged and generally works very well.

chrisfilo:
Could you also add the case from #1181?

fmriprep/workflows/bold/tests/test_util.py
```diff
@@ -0,0 +1,56 @@
...
+def test_masking(input_fname, expected_fname):
+    bold_reference_wf = init_bold_reference_wf(omp_nthreads=1, enhance_t2=True)
+    bold_reference_wf.inputs.inputnode.bold_file = input_fname
+    res = bold_reference_wf.run(plugin='MultiProc')
+
+    combine_masks = [node for node in res.nodes if node.name.endswith('combine_masks')][0]
+    overlap = symmetric_overlap(expected_fname,
+                                combine_masks.result.outputs.out_file)
+
+    assert overlap > 0.95, input_fname
```
effigies: My main remaining question is whether 0.95 is a good threshold. A 95% symmetric overlap means 95% of voxels overlap if the size stays the same; for an inflation where the old mask is a subset of the new, the inflation could be up to 110.8% (1/.95^2) the original size; or a deflation to as low as 90.25% (.95^2) of the original size.

oesteban: Seems like a good heuristic to start with :+1:

effigies:
Okay, I'll go ahead and merge. We can refine the threshold along with getting #1050 to a mergeable state.

Issue #1075, opened 2018-04-27
[URGENT] Degenerate derivatives


oesteban:
Since #1001 was merged in, the bold derivatives (at least the `space-MNI...` ones) are broken.

My intuition is that these changes on the Datasink introduced some problem https://github.com/poldracklab/fmriprep/compare/744de95e8c21...677c55647547#diff-7bf7d3ca48aa501c95ee753280fb3ee1

If I track back the files to the node that feeds the datasink, they look okay.

effigies:
It's this bit:

https://github.com/poldracklab/fmriprep/blob/c76acd4efb89918fce517b3845688068705b3aa5/fmriprep/interfaces/bids.py#L252-L255

It assumes that the input is uncompressed.

oesteban:
@mgxd can you/would you like to submit a fix?

effigies:
I think if we do `ext.endswith('.gz') and not fname.endswith('.gz')`, that would work, but I guess we should actually make a smarter copy, like:

```Python
def copy_any(src, dst):
    src_gz = src.endswith('.gz')
    dst_gz = dst.endswith('.gz')
    if src_gz == dst_gz:
        copy(src, dst)
    src_open = gzip.open if src_gz else open
    dst_open = gzip.open if dst_gz else open
    with src_open(src, 'rb') as f_in:
        with dst_open(dst, 'wb') as f_out:
            copyfileobj(f_in, f_out)
```

And if we want to permit hardlinks, nipype's `utils.filemanip.copyfile` would be a good alternative to `shutil.copy`.

oesteban:
Closed via #1077

oesteban:
Some of the reportlets are broken since we merged #1077:

https://2844-53175327-gh.circle-artifacts.com/0/tmp/ds000210/derivatives/fmriprep/sub-02.html

PR #1077, opened 2018-04-27
[FIX] Derivatives datasink handling of compression

mgxd:
not fully tested - adapted from @effigies  https://github.com/poldracklab/fmriprep/issues/1075#issuecomment-385057946

### DIFF
--- a/fmriprep/interfaces/bids.py
+++ b/fmriprep/interfaces/bids.py
@@ -24,7 +24,7 @@ import os.path as op
 import re
 import simplejson as json
 import gzip
-from shutil import copy, copytree, rmtree, copyfileobj
+from shutil import copytree, rmtree, copyfileobj
 
 from niworkflows.nipype import logging
 from niworkflows.nipype.interfaces.base import (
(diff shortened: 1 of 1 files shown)


fmriprep/interfaces/bids.py
```diff
@@ -420,3 +416,15 @@ def _splitext(fname):
         fname, ext2 = op.splitext(fname)
         ext = ext2 + ext
     return fname, ext
+
+
+def _copy_any(src, dst):
+    src_isgz = src.endswith('.gz')
+    dst_isgz = dst.endswith('.gz')
+    if src_isgz == dst_isgz:
+        copyfile(src, dst, use_hardlink=True)
```
effigies: Need `copy=True` in there, otherwise we might end up with symlinks.

effigies:
Looks good, thanks. @oesteban do we want to add any sanity checks, such as making sure that all image files in the output can at least be `nibabel.load`ed?

Actually, maybe we should add `--low-mem` to one of the tests, so that the compression pathway actually gets hit.

oesteban:
Adding `--low-mem` to one of the tests seems like a good idea :+1: .

I think the `nibabel.load` test can be done in a different PR. We probably want to check the hash of the outputs and force ourselves to update manually the hashes when they change (as it is done now for MRIQC).

oesteban:
I've opened #1078. I'll merge this in since it is quite urgent.

PR #1086, opened 2018-04-30
[DOC] Prefer pre-print over Zenodo doi in boilerplate

oesteban:


chrisfilo:
Both should be cited. New contributors will come along and should get credit.

PR #1088, opened 2018-04-30
[ENH] Add usage tracking and centralized error reporting

### LINKED ISSUE #1041: Usage tracking and centralized error reporting
[Sentry](http://sentry.io) is a platform used to track usage and lock exceptions for web applications. It can be used from desktop clients. We could use it to:

- track FMRIPREP popularity (useful for grant applications)
- see what command line options are most popular (useful for triaging features/issues)
- see which data types/properties are used (useful for triaging features/issues)
- automatically log exceptions (with tracebacks and variable context)

Sentry is free for open source projects - https://sentry.io/for/open-source/

oesteban:
Closes #1041


fmriprep/cli/run.py
```diff
@@ -282,6 +288,26 @@ def main():
...
+                '21f3c516491847af8e4ed249b122c4af@sentry.io/1137693',
+                release=__version__)
+            client.captureMessage(message=msg,
+                                  level='debug' if dev_user else 'info',
+                                  tags={
+                                      'run_id': run_uuid,
+                                      'npart': len(subject_list),
+                                      'type': 'ping',
+                                      'dev': dev_user})
+        except Exception:
```
chrisfilo: Not a huge fan of this catch all statement. Maybe we can improve it in the future.

fmriprep/cli/run.py
```diff
@@ -282,6 +288,26 @@ def main():
...
+                'https://d5a16b0c38d84d1584dfc93b9fb1ade6:'
+                '21f3c516491847af8e4ed249b122c4af@sentry.io/1137693',
+                release=__version__)
+            client.captureMessage(message=msg,
+                                  level='debug' if dev_user else 'info',
+                                  tags={
+                                      'run_id': run_uuid,
+                                      'npart': len(subject_list),
+                                      'type': 'ping',
+                                      'dev': dev_user})
```
chrisfilo: Are we capturing all of the command line arguments?

fmriprep/cli/run.py
```diff
@@ -282,6 +288,26 @@ def main():
     if opts.reports_only:
         sys.exit(int(retcode > 0))
 
+    # Sentry tracking
+    if not opts.notrack:
+        try:
+            from raven import Client
```
chrisfilo: So even though this is the only place raven is imported tracebacks are automatically captured no matter where they happen?

fmriprep/cli/run.py
```diff
@@ -217,6 +217,11 @@ def get_parser():
     g_other.add_argument('--stop-on-first-crash', action='store_true', default=False,
                          help='Force stopping on first crash, even if a work directory'
                               ' was specified.')
+    g_other.add_argument('--notrack', action='store_true', default=False,
+                         help='Opt-out of sending tracking information of this run to '
+                              'the FMRIPREP developers. This information helps to '
+                              'improve FMRIPREP and provides an indicator of real '
+                              'world usage crucial for obtaining funding.')
```
effigies: Not to delay this PR, but maybe by the next release it would be good to add a URL to an entry in the docs where we explicitly state everything that's included in the tracking message, perhaps with the full text of an example message.

I think maximal transparency about things like this will help mollify people who might otherwise opt out.

fmriprep/cli/run.py
effigies: If we are going to capture args, for data, output and work directories, it would be good to replace all but the last path element with "..." (i.e. `/path/to/bids/ds000XXX` becomes `.../ds000XXX`). It's unimportant to us, and it's information about someone's system that just doesn't need to be shared.

oesteban: Sentry captures sys.argv by default. I have to read the Data filtering module they have to do what @effigies suggests

fmriprep/cli/run.py
oesteban: If the exception is raised yes. I'm seeing that some are submitted but others are not. It will probably require some effort on integration with nipype

oesteban:
> Also please add a changelog entry.

That is added automatically with the changelog script 👍

effigies:
FWIW all my suggestions should be considered non-blocking. But if you're not going to put them in, I think opening an issue to remind ourselves would be good.

fmriprep/cli/run.py
chrisfilo: Is there an issue open for this? If not could you open one?

oesteban: Here it is - #1101

PR #1095, opened 2018-05-02
[FIX] Compression in DataSinks

### LINKED ISSUE #1075: [URGENT] Degenerate derivatives

oesteban:
This PR tries to fix #1075 for good.

### DIFF
--- a/fmriprep/interfaces/bids.py
+++ b/fmriprep/interfaces/bids.py
@@ -423,8 +423,11 @@ def _copy_any(src, dst):
     dst_isgz = dst.endswith('.gz')
     if src_isgz == dst_isgz:
         copyfile(src, dst, copy=True, use_hardlink=True)
+        return True
+
     src_open = gzip.open if src_isgz else open
     dst_open = gzip.open if dst_isgz else open
     with src_open(src, 'rb') as f_in:
         with dst_open(dst, 'wb') as f_out:
             copyfileobj(f_in, f_out)
+    return True
--- a/fmriprep/interfaces/bids.py
+++ b/fmriprep/interfaces/bids.py
@@ -255,10 +255,6 @@ class DerivativesDataSink(SimpleInterface):
                 out_file = out_file.format(extra_value=self.inputs.extra_values[i])
             self._results['out_file'].append(out_file)
             self._results['compression'].append(_copy_any(fname, out_file))
-
-        if len(self._results['compression']) == 1:
-            self._results['compression'] = self._results['compression'][0]
-
         return runtime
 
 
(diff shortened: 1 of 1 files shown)


fmriprep/interfaces/bids.py
```diff
@@ -423,8 +423,11 @@ def _copy_any(src, dst):
...
     if src_isgz == dst_isgz:
         copyfile(src, dst, copy=True, use_hardlink=True)
+        return True
+
     src_open = gzip.open if src_isgz else open
     dst_open = gzip.open if dst_isgz else open
     with src_open(src, 'rb') as f_in:
         with dst_open(dst, 'wb') as f_out:
             copyfileobj(f_in, f_out)
+    return True
```
effigies: Unless there's a use for `True` (and some case in which we might return `False`), I would just use `return` with no argument.

oesteban: We could add one output for debugging: return `False` if no compress/uncompress was performed, `True` otherwise. WDYT?

effigies: Sure. Maybe we should go ahead and add a unit test for this.

effigies: (I recognize that I'm using "we" somewhat disingenuously. I can't stop.)

oesteban: Let's make it a real "we". @mgxd would you like to add that unit test?

(actually, we probably may want to see this interface in nipype proper, WDYT? cc/ @satra)

effigies:
I think `BIDSDataSink` is probably going to want a refactor with some of the new `pybids` functionality before it gets upstreamed...

fmriprep/interfaces/bids.py
```diff
@@ -170,6 +170,9 @@ class DerivativesDataSinkInputSpec(BaseInterfaceInputSpec):
 
 class DerivativesDataSinkOutputSpec(TraitedSpec):
     out_file = OutputMultiPath(File(exists=True, desc='written file path'))
+    compression = traits.List(
+        traits.Bool, desc='whether ``in_file`` was compressed/uncompressed '
+                          'or `it was copied directly.')
```
effigies: I would just use `OutputMultiPath`, here. There's no good reason to make it a list when `out_file` isn't.

fmriprep/interfaces/bids.py
```diff
@@ -256,6 +256,9 @@ def _run_interface(self, runtime):
             self._results['out_file'].append(out_file)
             self._results['compression'].append(_copy_any(fname, out_file))
 
+        if len(self._results['compression']) == 1:
+            self._results['compression'] = self._results['compression'][0]
```
effigies: `OutputMultiPath` automatically does this.

PR #1114, opened 2018-05-05
[ENH] Carpetplot + confounds plot

### LINKED ISSUE #174: Include confound plots in the reports
To save space lets limit the confounds to the following line plots:
1. X, Y, Z, roll, pitch, yaw on one plot (mm on left axis, degrees on the right)
1. FD
1. Standardized DVARS
1. CompCorr (all components on one plot)
1. Grey, white, CSF, and global signal (all on one plot)

plot horizontal lines with: max, median and min

### LINKED ISSUE #877: Carpetplot in reports
It would be great to generate two carpetplots: one before any preprocessing and another after all preprocessing. Then, we would use something like https://zurb.com/playground/twentytwenty to allow users scroll and compare while saving space in the (already lengthy) reports.

oesteban:
A first version of the carpetplot. Close #174, close #877.


oesteban:
Refinements to this PR have been filed: #1119 and #1120.

oesteban:
Links to reports with these carpetplots:

- https://2025-53608443-gh.circle-artifacts.com/0/tmp/ds005/derivatives/fmriprep/sub-01.html
- https://2028-53608443-gh.circle-artifacts.com/0/tmp/ds054/derivatives/fmriprep/sub-100185.html
- https://2030-53608443-gh.circle-artifacts.com/0/tmp/ds210/derivatives/fmriprep/sub-02.html

I think the awkward fontsizes and width should be fixed in niworkflows proper (although I'm not sure, I should have a look into that).

effigies:
Visually, I like these. My main issues are:

* I cannot read the numbers without zooming in heavily, and the number in light gray on the left (intercept?) is hard even then.
* It is unclear how visible spikes will be when they fall behind the titles. As it is, different amounts of each confound graph are visible. I wonder if we could align the bottom of the text (both title and stats) to the max point on the graph, without collateral aesthetic damage.
* Maybe we could reduce the coverage by changing `FramewiseDisplacement [mm]` to `FD` or `FD [mm]`? And `GlobalSignal` to `Global`.

Textually:

* "global signals ('GlobalSignal', 'WM', 'GM')" doesn't match the reported signals, which includes CSF, not GM.
* Where possible, it would be nice to link to something that goes into the interpretation of the stats being shown. For example [standardized DVARS](https://doi.org/10.1016/j.neuroimage.2017.12.098).
* As long as we describe the figure from top-to-bottom, I don't think we need to say "top" and "bottom" in the text.

Just as a suggestion, here's the same information in a little bit more narrative style (I've included some shortened names, to try out how they'd look):

> ## BOLD Summary
> Summary statistics are plotted which may reveal trends or artifacts in the BOLD data. Global, WM and CSF show the mean BOLD signal in the entire brain mask, and the white matter and CSF segmentation maps, respectively. stdDVARS and FD show the [standardized DVARS](https://doi.org/10.1016/j.neuroimage.2017.12.098) and [framewise displacement](https://doi.org/10.1016/j.neuroimage.2011.10.018) measures for each time point.
> 
> A "carpet plot" shows the time series for all voxels within the brain mask. Voxels are grouped into cortical (blue) and subcortical (orange) gray matter, cerebellum (green), and white matter and CSF (red), indicated by the color map on the left-hand side.

What do you think?

fmriprep/workflows/bold/confounds.py
```diff
@@ -270,6 +270,55 @@ def _pick_wm(files):
...
+        return list(reversed(inlist))
+
+    def _flags(inlist):
+        return [True] * len(inlist)
+
+    workflow = pe.Workflow(name=name)
+    workflow.connect([
+        (inputnode, resample_parc, [
+            ('bold_mask', 'reference_image'),
+            (('bold_mni_transforms', _reversed), 'transforms'),
```
effigies: Is this going to be correct, given that the ANTs warps aren't invertible? We actually need `t1_2_mni_reversed_transform`.

oesteban:
> - I cannot read the numbers without zooming in heavily, and the number in light gray on the left (intercept?) is hard even then.

Yeah, this is a big issue, although I think it is a problem of the generation itself. I'm going to try using css to zoom in to 100% width and I expect the fonts to stay fixed in size. This is some heritage from the original MRIQC implementation.

> - It is unclear how visible spikes will be when they fall behind the titles. As it is, different amounts of each confound graph are visible. I wonder if we could align the bottom of the text (both title and stats) to the max point on the graph, without collateral aesthetic damage.

I played around this for a while, and to me it was acceptable when spikes fell below the label. However we could give it some more love. I'll check if there's some low hanging fruit here.

> Maybe we could reduce the coverage by changing FramewiseDisplacement [mm] to FD or FD [mm]? And GlobalSignal to Global.

Yep, that would require extra inputs to map the names of columns in the tsv file to labels. But I think this one is pretty straightforward to implement.

Textually:

> "global signals ('GlobalSignal', 'WM', 'GM')" doesn't match the reported signals, which includes CSF, not GM.

Oh, that is a mistake, will fix.

> Where possible, it would be nice to link to something that goes into the interpretation of the stats being shown. For example standardized DVARS.

👍 

> As long as we describe the figure from top-to-bottom, I don't think we need to say "top" and "bottom" in the text.

👍 

> Just as a suggestion, here's the same information in a little bit more narrative style (I've included some shortened names, to try out how they'd look):
>
> **BOLD Summary**
> Summary statistics are plotted which may reveal trends or artifacts in the BOLD data. Global, WM and CSF show the mean BOLD signal in the entire brain mask, and the white matter and CSF segmentation maps, respectively. stdDVARS and FD show the standardized DVARS and framewise displacement measures for each time point.
>
> A "carpet plot" shows the time series for all voxels within the brain mask. Voxels are grouped into cortical (blue) and subcortical (orange) gray matter, cerebellum (green), and white matter and CSF (red), indicated by the color map on the left-hand side.

👍 

> What do you think?

Cool! thanks for all these ideas. Will work on them and let you know when done.

effigies:
Also, I'll go ahead and say we shouldn't have quotes around any part of carpet plot. It's a common enough term at this point, and we're pretty clearly indicating what we mean by it, so let's just let it be a term. (We should decide whether it's one word or two, though. I feel like I've seen both. I incline to two, but do not feel strongly.)

oesteban:
Okay, the latest commit should pass. If so, this is ready to merge.

The problem of label sizes and look'n'feel must be addressed in niworkflows (https://github.com/poldracklab/niworkflows/issues/237).

oesteban:
Forgot this: https://github.com/poldracklab/fmriprep/pull/1114#pullrequestreview-118878846 let me fix that.

fmriprep/workflows/bold/confounds.py
oesteban: Okay, now I recall why I did it this way.

In principle, you are right: we should use the `t1_2_mni_reversed_transform` and the reversed transform of the SDC.

However, I preferred to set the `invert_transform_flags` and have ApplyTransform to calculate the inverses. You are right that some transforms may not be invertible (although SyN is diffeomorphic so the T1-to-MNI should be invertible), but ApplyTransform will approximate the transforms and this way was very convenient. Otherwise, we would need to keep track of which transforms are inverted. Also, I'm not sure whether we generate the warped-to-corrected transform for SDC. If not, then we will need to approximate anyways (and it will be likely not invertible for EPI and fieldmap SDCs).

So far, with the examples I've seen, the estimated transforms seem accurate enough to resample the atlas in bold space for the purpose of the carpet plot.

Sounds reasonable?

effigies: Sure. Approximate seems fine, here.

Issue #881, opened 2017-12-06
Move anatomical workflow out into niworkflows


oesteban:
The anatomical workflow is mature enough to be detached from fmriprep's base code.

This would make the anatomical workflow more visible and will allow:
  1) Move the anatomical workflow tests to niworkflows.
  2) Cache all the anatomical workflow between circle builds (speed up build time of fmriprep).

WDYT?

(option 2 could happen before 1, that would also be okay)

effigies:
I guess my question is whether it makes sense outside of fmriprep. As it is, we pipe a lot of options in through workflow arguments, which, on a first glance would seem to make it difficult to abstract to usability in another program.

oesteban:
Let's start with 2 for now. I personally would have interest in seeing the anatomical workflows in niworkflows, but it is true that I know how to use them too well. Other users might be clueless.

oesteban:
As we are considering to branch a dMRIPrep out, I guess this makes more sense now.

oesteban:
See https://github.com/poldracklab/smriprep

PR #1115, opened 2018-05-06
RF: Create T1-MNI normalization workflow

effigies:
Uses poldracklab/niworkflows#236.


oesteban:
It'd be nice to move all anatomical workflows to niworkflows (#881). Would you like to move this workflow there?

oesteban:
Do you mind submitting this PR to smriprep (when it starts to work out... which is on its way)?

oesteban:
Let's keep tracking on this under poldracklab/smriprep#1. After #1482 this will need to happen there anyways.

Issue #1125, opened 2018-05-13
Bad connection for  --medial-surface-nan option


ZhifangYe:
When I try to output bold data on surface with --medial-surface-nan option, I encountered this error. 

> 180513-14:41:58,207 workflow IMPORTANT:
	 Creating BOLD surface-sampling workflow.
Process Process-2:
Traceback (most recent call last):
  File "/usr/local/miniconda/lib/python3.6/multiprocessing/process.py", line 258, in _bootstrap
    self.run()
  File "/usr/local/miniconda/lib/python3.6/multiprocessing/process.py", line 93, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/local/miniconda/lib/python3.6/site-packages/fmriprep/cli/run.py", line 510, in build_workflow
    ignore_aroma_err=opts.ignore_aroma_denoising_errors,
  File "/usr/local/miniconda/lib/python3.6/site-packages/fmriprep/workflows/base.py", line 202, in init_fmriprep_wf
    ignore_aroma_err=ignore_aroma_err)
  File "/usr/local/miniconda/lib/python3.6/site-packages/fmriprep/workflows/base.py", line 458, in init_single_subject_wf
    ignore_aroma_err=ignore_aroma_err)
  File "/usr/local/miniconda/lib/python3.6/site-packages/fmriprep/workflows/bold/base.py", line 674, in init_func_preproc_wf
    name='bold_surf_wf')
  File "/usr/local/miniconda/lib/python3.6/site-packages/fmriprep/workflows/bold/resampling.py", line 154, in init_bold_surf_wf
    (medial_nans, merger, [('out', 'in1')]),
  File "/usr/local/miniconda/lib/python3.6/site-packages/niworkflows/nipype/pipeline/engine/workflows.py", line 214, in connect
    '\n'.join(['Some connections were not found'] + infostr))
Exception: Some connections were not found
Module medial_nans has no output called out

I think the line 154 in resampling.py should be `(medial_nans, merger, [('out_file', 'in1')]),` instead of `(medial_nans, merger, [('out', 'in1')]),`.

ZhifangYe:
Sure. Just submitted #1128.

PR #1128, opened 2018-05-14
Fix bad connection for --medial-surface-nan option

ZhifangYe:
This PR solved #1125.


effigies:
Want to add yourself to the `.zenodo.json` file, just above Russ?

https://github.com/poldracklab/fmriprep/blob/0d9aa92db967f7dbe6d46905cacbfae644497b0f/.zenodo.json#L66-L75

PR #1133, opened 2018-05-16
RF: Resume external nipype dependency

effigies:
Purges `niworkflows.nipype` in favor of `nipype`.

Pins poldracklab/niworkflows#241.

----

#### Original text:

This is just to validate poldracklab/niworkflows#238. Should be a functionally inert refactor, so can wait for the next niworkflows release to actually update the pin.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -296,10 +296,10 @@ jobs:
           at: /tmp
       - restore_cache:
           keys:
-            - ds005-anat-v6-{{ .Branch }}-{{ epoch }}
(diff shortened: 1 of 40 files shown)


effigies:
@oesteban There's another ICA-AROMA bug popping up. I think you've been paying closer attention to these than I have... Could you take a quick look to see if it's an old problem reemerging or something I'm introducing?

oesteban:
Oh yeah, look at the command line, the output arg is just `-o out`.

effigies:
Gotcha. Merging nipype/master in. Didn't realize I hadn't.

Issue #1131, opened 2018-05-15
Possible bug or user error? Echoes look suspitiously similar after fmriprep.


dangom:
Hi guys,

we have an issue here but it's possible that we are the ones doing something wrong.
In one of our experiments we acquire a 3-echo dataset and preprocess it with fmriprep 1.0.13 running the following command:

> FMRIPrep command: /home/mrphys/zahfaz/anaconda3/bin/fmriprep --participant-label sub-07 --low-mem --use-syn-sdc --ignore fieldmaps slicetiming --use-aroma --aroma-melodic-dimensionality 100 --no-submm-recon --output-space T1w template -w /data/fmripreptest-task-sub-07 --nthreads 12 --fs-license /home/mrphys/zahfaz/usr/local/freesurfer/license.txt --fs-no-reconall /project/3015046.02/task-experiment-datasets /project/3015046.02/task-derivatives/temp participant

For completion, here's how the func directory looks like:

```
$ tree
.
|-- sub-07_ses-01_task-stroop_acq-mb_bold.json
|-- sub-07_ses-01_task-stroop_acq-mb_bold.nii.gz
|-- sub-07_ses-01_task-stroop_acq-mbme_echo-01_bold.json
|-- sub-07_ses-01_task-stroop_acq-mbme_echo-01_bold.nii.gz
|-- sub-07_ses-01_task-stroop_acq-mbme_echo-02_bold.json
|-- sub-07_ses-01_task-stroop_acq-mbme_echo-02_bold.nii.gz
|-- sub-07_ses-01_task-stroop_acq-mbme_echo-03_bold.json
`-- sub-07_ses-01_task-stroop_acq-mbme_echo-03_bold.nii.gz
```
BIDS-Validator says things are OK except for a missing 'TaskName', but I wouldn't expect that to be relevant to the issue at hand.

Our problem is that after processing, all of the three echoes look extremely alike, as you can see in the shots below:

Echo 1 **raw**
![echo-01](https://user-images.githubusercontent.com/7881842/40056632-478db956-584c-11e8-9704-c7ef494af72d.png)
Echo 1 **fmriprep**
![echo-01_preproc](https://user-images.githubusercontent.com/7881842/40056637-4a3d3b86-584c-11e8-8072-5c1fa48a286d.png)
Echo 2 **raw**
![echo-02](https://user-images.githubusercontent.com/7881842/40056639-4bb9e608-584c-11e8-8589-e852a72c8b0f.png)
[...]

PR #1134, opened 2018-05-16
[WIP, FIX] Patch boldbuffer for ME

emdupre:
As reported in #1131, recent refactors (post v. 1.0.9) of the bold workflow were yielding duplicate results across echos. This PR should increase the legibility of early portions of the BOLD workflow and patch the multi-echo workflow (specifically the `boldbuffer` connections) so that correct results are achieved for each echo.

### DIFF
--- a/fmriprep/workflows/bold/base.py
+++ b/fmriprep/workflows/bold/base.py
@@ -385,9 +385,9 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
     if run_stc is True:  # bool('TooShort') == True, so check True explicitly
         bold_stc_wf = init_bold_stc_wf(name='bold_stc_wf', metadata=metadata)
         workflow.connect([
-            (bold_stc_wf, boldbuffer, [('outputnode.stc_file', 'bold_file')]),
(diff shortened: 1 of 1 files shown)


emdupre:
I just compared the circle outputs for echo1 and echo2 in ds210 using `np.allclose` and they're now giving different results, so I think this is patched ! 

I'd like to add this to the overall testing suite (seeing if the results are consistent across refactors, maybe using the cache), but this is a bigger idea that I think we can discuss in a separate issue, at least if we'd like to implement it across data sets.

effigies:
Should there be multiple functional entries in the [report](https://3136-53175327-gh.circle-artifacts.com/0/tmp/ds210/derivatives/fmriprep/sub-02.html#Functional)?

emdupre:
Sorry, to confirm, do you want me to add anything to this PR ? I accidentally opened this on my master, so if I should edit the reports I'll go ahead and close this and open a new PR with this code and whatever additions are requested !

oesteban:
I think this PR is pretty well focused. I've opened #1135 with @effigies suggestion.

Issue #1145, opened 2018-05-24
Precaching cifti atlas in docker


ZhifangYe:
When output CIFTI files, I found it needs to download atlas files from OSF server. Many HPC computational nodes like ours doesn't have active Internet connection. I think it's reasonable to prepare those files in the container instead of downloading during running.

effigies:
Agreed. Would you like to submit a PR?

ZhifangYe:
Yeah, I could make some time for a PR this weekend.

chrisfilo:
Great - we much appreciate your contribution!

ZhifangYe:
@effigies I realize that the oasis_dtk31 atlas folder on OSF storage is named `oasis_trt20_DKT31-MNI152NLin` instead of the key used in niworkflows (`oasis_dkt31_mni152`). This leads to `niworkflows.data.get_oasis_dkt31_mni152` shows undesired behaviors and hard to parse the atlas filename.

I think the simplest solution here is rename the folder inside `oasis_trt20_DKT31-MNI152NLin.tar.gz `. I'm not sure whether I could modify file on OSF storage. Could you help me re-upload it?

effigies:
I'm a little surprised, as this seems like something that should have been caught before, as we do have some people using the pipeline. @mgxd perhaps you can comment, here?

@ZhifangYe It might help if you describe the approach you're taking. Just to be clear, what I was suggesting was to add a few lines to `docker/scripts/get_templates.sh`.

ZhifangYe:
Indeed I tried to add the atlas file address in `docker/scripts/get_templates.sh` at first.  But it can't prevent the attempt to re-download files from OSF.

When I pre-cached the atlas, the extracted folder name is `oasis_trt20_DKT31-MNI152NLin`. The code use this atlas is in line 134. Because the folder name is hard coded in line 135, it will not cause any problem during runtime. But it still tries to connect OSF server.
https://github.com/poldracklab/fmriprep/blob/5dd78197a828de566d754b327f0f7d663bfe42f0/fmriprep/interfaces/cifti.py#L133-L138

If I'm right, the expected atlas data should be in path `$CRN_SHARED_DATA/oasis_dkt31_mni152` instead of current `oasis_trt20_DKT31-MNI152NLin`. During running, line 134 will always try to download the tar file even the files are already in `oasis_trt20_DKT31-MNI152NLin`. 

[...]

PR #1159, opened 2018-06-06
ENH: Pre-cache DKT31 template in Docker image

ZhifangYe:
This PR solved #1145 by cache DKT31 atlas in container. It also modified `cifti.py` line133-136 to respect the naming convention.

I think the check will fail until poldracklab/niworkflows#245 merged.


docker/scripts/get_templates.sh
```diff
@@ -8,6 +8,8 @@ OASIS_TEMPLATE="https://files.osf.io/v1/resources/fvuh8/providers/osfstorage/584
 OASIS_SHA256="d87300e91346c16f55baf6f54f5f990bc020b61e8d5df9bcc3abb0cc4b943113"
 NKI_TEMPLATE="https://files.osf.io/v1/resources/fvuh8/providers/osfstorage/59cd90f46c613b02b3d79782"
 NKI_SHA256="4bba067f6675d15be96b205cb227e18a540673fd7e4577e13feedcef3a6f0ec5"
+OASIS_DTK31_TEMPLATE="https://files.osf.io/v1/resources/fvuh8/providers/osfstorage/5b16f17aeca4a80012bd7542"
```
effigies: You use `OASIS_DTK31_TEMPLATE` here, but `$OASIS_DTK31_ATLAS` below. Either one is fine with me.

ZhifangYe: Thanks.
Fixed now.

effigies:
I added pins so that your modifications to niworkflows get included in the build. If this passes, then we will merge poldracklab/niworkflows#245, and we'll be able to release and update this PR to depend on it.

effigies:
Travis apparently maintains its own pip cache, so it's failing to build. As long as Circle builds, I'm happy to merge, and we can get a final test on master. @chrisfilo Sound okay, or should we wait on the whole build?

PR #1160, opened 2018-06-06
[ENH] Use upstream TShift improvements

effigies:
* First commit: Verify that nipy/nipype#2608 does not break existing usage
* Upcoming commit: Use `TShift.inputs.slice_timing` instead of helper `Function` interface


effigies:
I don't think we actually use slice timing in any of our tests, so we can't go by Circle on this...

effigies:
Status update: nipy/nipype#2608 will be in 1.1.1, so I'll update this and we can merge after 1.1.1.

I tested on ds000030/sub-10159. Output files of the TShift step differed by timestamp.

PR #1161, opened 2018-06-06
RF: Replace InvertT1w with upstream Rescale(invert=True)

effigies:
Testing nipy/nipype#2599.


effigies:
This is ready for review.

PR #1163, opened 2018-06-07
[MAINT] Updating core dependencies

chrisfilo:


effigies:
I wonder if at some point we should upgrade the base image to 18.04 LTS. We'll need to check on NeuroDebian status for 18.04, first.

chrisfilo:
Thanks for cleaning this up for me!

PR #1178, opened 2018-07-02
exclude sourcedata

### LINKED ISSUE #1177: Is looking for T1s in places where it shouldn't
The 'life' movie dataset (http://datasets.datalad.org/?dir=/labs/haxby/life) carries non-defaced T1w images in a `sourcedata/` subdirectory. However, their content is not available (due to privacy issues), so those files appear to be dead symlinks. When running fmriprep it picks these files up and then crashed when opening them. I think:

- it should not be looking in `sourcedata/` at all
- it could treat dead symlinks as non-existent files

(of course that dataset has defaced T1w images in the respective BIDS-compliant locations too)

chrisfilo:
Closes #1177

### DIFF
--- a/fmriprep/utils/bids.py
+++ b/fmriprep/utils/bids.py
@@ -151,7 +151,7 @@ def collect_data(dataset, participant_label, task=None):
 
 
     """
-    layout = BIDSLayout(dataset, exclude='derivatives')
+    layout = BIDSLayout(dataset, exclude=['derivatives', 'sourcedata'])
     queries = {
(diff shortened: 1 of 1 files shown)


effigies:
I believe this should be getting resolved in pybids soon, with derivatives, sourcedata and code not getting indexed by default. I'll need to acquaint myself with what's happened since the hackathon started...

oesteban:
Anyways, this will not hurt at all right?

effigies:
If it's not failing at least one of the tests, I don't think so...

Issue #1174, opened 2018-06-19
pybids performance


chrisfilo:
Context https://neurostars.org/t/possible-to-skip-bids-validation/1851

Ideas:

- turn off pybids validation (`validation=False`)
- update pybids and grabbits (there were recent performance related improvements)

effigies:
I don't think we do validation, but both of these sound reasonable. There were some API changes, but we might not use those aspects, so it might be as simple as pinning a new version.

PR #1179, opened 2018-07-02
PIN: pybids 0.6.3

effigies:
Attempts to address #1174. At the very least we'll have access to the latest changes and see if we need to change our invocation at all.


effigies:
Cancelled to give #1180 priority. Feel free to restart if it finishes and I don't.

oesteban:
This can be merged right?

PR #1180, opened 2018-07-02
PIN: nipype 1.1.0

effigies:
Testing nipy/nipype#2598 on the fmriprep battery.

### DIFF
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -21,6 +21,7 @@ dependencies:
 - python-dateutil
 - pydot>=1.2.3
 - cython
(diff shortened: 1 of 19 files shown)


effigies:
I suspect we're going to run into some compatibility issues with 1.1.0, as there has been some removal of deprecated components and the logging has been made hierarchical. We may need another niworkflows release, as well, post-1.1.0.

effigies:
Well, that took too many tries. Incidentally, we need to keep niworkflows dev versions up-to-date to make sure pinning works correctly with the Docker build.

Issue #1182, opened 2018-07-02
Make reusing freesurfer outputs more explicit


chrisfilo:
Reusing precomputed freesurfer outputs is currently supported, but requires a very specific approach that is often confusing for users. To make things more transparent I propose adding `--reuse-freesurfer-dir` flag pointing to a SUBJECTS_DIR which will be reused. When that flag is set recon-all should not be attempted.

effigies:
I think this is reasonable. :+1:

emdupre:
Could we also update the documentation to reference that FS outputs can be reused ? I wasn't able to find reference of it in the documentation already, though I imagine it might go best about [here](http://fmriprep.readthedocs.io/en/latest/workflows.html#surface-preprocessing).

effigies:
Near the end of that section:

> In order to bypass reconstruction in `fmriprep`, place existing reconstructed subjects in `<output dir>/freesurfer` prior to the run. `fmriprep` will perform any missing `recon-all` steps, but will not perform any steps whose outputs already exist.

emdupre:
Thanks ! I worry there are others who share my short-sightedness, though :smile:

WDYT of including at least a note in the first paragraph, right after where it says 

> All surface preprocessing may be disabled with the --fs-no-reconall flag.

The implication there (again just to me) is that it will be run if that flag is not passed-- but including the derivatives short-circuits that !

effigies:
I agree, this has been hard for many users to find, so I'm very open to improving this. If you have suggestions for the wording and some time to spare, could you submit a PR?

PR #1189, opened 2018-07-03
[DOC] Better document re-using FS derivatives

emdupre:
Addresses #1182.

Iterates on documentation for re-using freesurfer derivatives directory. I've switched to a Sphinx note (since it seems that this issue has come up repeatedly, and is worth highlighting), and I also moved the text up to a more prominent spot in the paragraph.

Happy to make any suggested changes, here !


effigies:
screen shot 2018-07-03 at 20 47 59

Looks great! Thanks!

PR #1204, opened 2018-07-12
[ENH] Change WORKDIR of Docker image

oesteban:
## Changes proposed in this pull request

This PR makes some improvements to minor details of the Dockerfile. The most relevant is changing the working directory.

## Documentation that should be reviewed

No documentation changes are needed.

## Pull-request guidelines:

1. We invite you to list yourself as a *fMRIPrep* contributor, so if your name 
   is not already mentioned, please modify the 
   [``.zenodo.json``](https://github.com/poldracklab/fmriprep/blob/master/.zenodo.json)
   file with your data right above Russ' entry. Example:
   ```
   {
      "name": "Contributor, New FMRIPrep",
      "affiliation": "Department of fMRI prep'ing, Open Science Made-Up University",
      "orcid": "<your id>"
   },
   {
      "name": "Poldrack, Russell A.",
      "affiliation": "Department of Psychology, Stanford University",
      "orcid": "0000-0001-6755-0259"
   },
   ```
   
2. By submitting this request you acknowledge that your contributions are available under the BSD 3-Clause license.

3. Use a descriptive prefix, between brackets for your PR: ``ENH`` (enhancement), ``FIX``, ``TST``, ``DOC``, ``STY``,
   ``REF`` (refactor), ``CI`` (continous integration), ``MAINT`` (maintenance). Example:
   ```
   [ENH] Support for SB-reference in multi-band datasets
   ```
   For works-in-progress, add the ``WIP`` tag in addition to the descriptive prefix. 
   Pull-requests tagged with ``[WIP]`` will not merged in until the tag is removed.

4. Your PR will be reviewed according to the following
   [template](https://github.com/poldracklab/fmriprep/wiki/Reviewing-a-Pull-Request).

5. Documentation is a fundamental aspect to the *glass-box* philosophy that *fMRIPrep* abides by.
   Please understand that the *fMRIPrep* team may (are likely to) request you to improve the documentation
   provided with this PR, within the PR or in future PRs.
   

**Please review and check the following**:

  - [x] I have read the [guidelines for contributions](https://github.com/poldracklab/fmriprep/blob/master/CONTRIBUTING.md).
  - [x] I understand that my contributions will not be merged unless the work is
        finished (i.e. no ``[WIP]`` tag remains in the title of my PR) and tests pass.
  - [x] The proposed code follows the
        [coding style](https://github.com/poldracklab/fmriprep/blob/master/CONTRIBUTING.md#fmriprep-coding-style-guide),
[...]

PR #1218, opened 2018-07-24
[FIX] Corrects Dockerfile WORKDIR.

pvelasco:

It sets the last WORKDIR in the docker image to /tmp, instead of /root/src/fmriprep, so now you can run the docker container as a user different from root.  

As far as I can tell, if no working directory is specified by the `-w` parameter, fmriprep will try to create the folder `work` under `/root/src/fmriprep`, and only root has access to that folder.

This should fix problems people have found when running docker specifying a user (e.g., ref #1060).

**Please review and check the following**:

  - [x] \(Optional\) I opt-out from being listed in the `.zenodo.json` file.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -161,7 +161,7 @@ RUN echo "${VERSION}" > /root/src/fmriprep/fmriprep/VERSION && \
 
 RUN ldconfig
 
-WORKDIR /root/src/fmriprep
+WORKDIR /tmp/
 
 ENTRYPOINT ["/usr/local/miniconda/bin/fmriprep"]
 


oesteban:
Hi @pvelasco, would this #1204 fix your issue? Any particular reason to use `/tmp` instead of `/scratch` ?

effigies:
@oesteban I understand the problem here to be that permissions issues when running Docker with the `-u <UID>` flag. You should make sure that `/scratch` is world-writable, so working directories can be written within the container. Perhaps we should go ahead and run one of our tests with the `-u` flag.

pvelasco:
@oesteban, I tried using `/scratch`, but it doesn't work because it is created by root and any other users don't have writing privileges; you would have to `chmod` in the Dockerfile.
Why not using `/tmp`?  That's precisely its purpose: temporary scratch space readable/writable by everybody.

oesteban:
@pvelasco, you are right. I was overthinking this.

@effigies I am testing that on a different branch. Will send PR when working.

PR #1236, opened 2018-08-06
FIX: Verify first word of _cmd in dependency check

### LINKED ISSUE #1235: Singularity build missing dependencies (1.1.3)
```
Cannot run fMRIPrep. Missing dependencies:
    3dQwarp -prefix Qwarp.nii.gz -plusminus (Interface: QwarpPlusMinus)
    3dQwarp -prefix Qwarp.nii.gz -plusminus (Interface: QwarpPlusMinus)
    3dQwarp -prefix Qwarp.nii.gz -plusminus (Interface: QwarpPlusMinus)
    3dQwarp -prefix Qwarp.nii.gz -plusminus (Interface: QwarpPlusMinus)
    3dQwarp -prefix Qwarp.nii.gz -plusminus (Interface: QwarpPlusMinus)
    3dQwarp -prefix Qwarp.nii.gz -plusminus (Interface: QwarpPlusMinus)
    3dQwarp -prefix Qwarp.nii.gz -plusminus (Interface: QwarpPlusMinus)
    3dQwarp -prefix Qwarp.nii.gz -plusminus (Interface: QwarpPlusMinus)
    3dQwarp -prefix Qwarp.nii.gz -plusminus (Interface: QwarpPlusMinus)
    3dQwarp -prefix Qwarp.nii.gz -plusminus (Interface: QwarpPlusMinus)
    3dQwarp -prefix Qwarp.nii.gz -plusminus (Interface: QwarpPlusMinus)
    3dQwarp -prefix Qwarp.nii.gz -plusminus (Interface: QwarpPlusMinus)
```

[Full log here](https://pastebin.com/ahTvmjJ6)
Container was built from dockerhub using Singularity `2.4.5` and `2.5.1`
`singularity build fmriprep.simg docker://poldracklab/fmriprep:1.1.3`

effigies:

In the event that an `interface._cmd` has spaces in it - *i.e.*, pre-set flags in addition to the command name, the `which` check will fail. This PR uses `_cmd.split()[0]` to use only the command up to the first space when checking for the command.

A command of the form `python /some/path/some_script` will merely check the existence of `python`, and not `/some/path/some_script`, but that would not be correctly handled by the existing check. To my knowledge, we do not have any such nodes at this time, so this is theoretical, but we may want to consider a more robust long-term solution, but I don't believe that should hold up this fix.

Fixes #1235.

None.

## Pull-request guidelines:

  - [ ] \(Optional\) I opt-out from being listed in the `.zenodo.json` file.

### DIFF
--- a/fmriprep/cli/run.py
+++ b/fmriprep/cli/run.py
@@ -37,7 +37,7 @@ def check_deps(workflow):
         (node.interface.__class__.__name__, node.interface._cmd)
         for node in workflow._get_all_nodes()
         if (hasattr(node.interface, '_cmd') and
-            which(node.interface._cmd) is None))
+            which(node.interface._cmd.split()[0]) is None))
 
 
 def get_parser():


effigies:
@mgxd Are you able to check build a Singularity image off of this branch, prior to release? It would be nice to have a check.

The issue should arise in Docker, as well, given the same inputs. The error criterion, I believe, should be whether a `QwarpPlusMinus` interface is instantiated in our workflow, and has nothing to do with Singularity.

mgxd:
@effigies @oesteban this is a good short-term solution, but should probably be reverted once the relevant interfaces are fixed and released upstream. 

One case that comes to mind is `wb_command <command>`, in which the <command> could vary between workbench versions. This way of checking will always assume the command works (as long as wb_command is found in the PATH)

effigies:
Reverting would break `wb_command <command>` even further. We don't pretend to check whether the arguments will work, just whether we can hit the executable.

PR #1247, opened 2018-08-13
ENH: Write derivatives dataset_description.json

effigies:

* Add a `dataset_description.json` to the output directory. This is required by the [BIDS derivatives spec](https://docs.google.com/document/d/1Wwc4A6Mow4ZPPszDIWfCUCRNstn7d_zzaWPcfcHmgI4/edit#heading=h.y0r14oqbfj8z), and pybids complains when loading a layout from fMRIPrep derivatives.

**Please review and check the following**:

  - [ ] \(Optional\) I opt-out from being listed in the `.zenodo.json` file.


fmriprep/utils/bids.py
```diff
@@ -193,3 +193,52 @@ def _grp_echos(x):
...
+    from fmriprep import __version__
+
+    desc = {
+        'Name': 'fMRIPrep output',
+        'BIDSVersion': '1.1.1',
+        'PipelineDescription': {
+            'Name': 'fMRIPrep',
+            'Version': __version__,
+            # We could parse the version to reference a commit or tag
+            'CodeURL': 'https://github.com/poldracklab/fmriprep',
```
oesteban: We can use the [`DOWNLOAD_URL`](https://github.com/poldracklab/fmriprep/blob/master/fmriprep/info.py#L49)

effigies: Does that work for non-tagged versions?

fmriprep/utils/bids.py
```diff
@@ -193,3 +194,52 @@ def _grp_echos(x):
...
+            },
+        'CodeURL': 'https://github.com/poldracklab/fmriprep',
+        'HowToAcknowledge': 'https://github.com/poldracklab/fmriprep',
```
oesteban: Perhaps something more like current BIDS (sec 8.1.1)?:

```JSON
"HowToAcknowledge": "Please cite our paper (https://doi.org/10.1101/306951), and include the generated citation boilerplate within the Methods section of the text",
"ReferencesAndLinks": [ list generated from the boilerplate ],
"PipelineDOI": "https://doi.org/10.5281/zenodo.852659"
```

fmriprep/utils/bids.py
oesteban: Yeah, of course not. But we could fall back to the repo in that case.

fmriprep/utils/bids.py
```diff
@@ -193,3 +194,52 @@ def _grp_echos(x):
...
+        }
+
+    # Keys that can only be set by environment
+    if 'FMRIPREP_DOCKER_TAG' in os.environ:
+        desc['DockerHubContainerTag'] = os.environ['FMRIPREP_DOCKER_TAG']
```
oesteban: Do you suggest to set these env variables when building the docker image?

fmriprep/utils/bids.py
effigies: But with versioneer we should be able to get a specific URL for all actual commits.

oesteban: Maybe I'm wrong, but the general case here is that the version is tagged.

I did not mean that the specific URL for non-tagged versions would not be best. Sure it is, and if versioneer provides it cheap, much better :). I proposed the `DOWNLOAD_URL` because it is already there for you to use in this PR without any complication.

fmriprep/utils/bids.py
effigies: If possible. Or perhaps `fmriprep-docker` would get the actual SHA256 hash and set the environment variable...

These are mostly just stubs indicating there's room to be clever and try to figure out how to get precise container references.

fmriprep/utils/bids.py
effigies: PipelineDOI is not defined in the BIDS spec or the derivatives draft spec. ReferencesAndLinks is described as "List of references to publication that contain information on the dataset, or links.", which is different from how you seem to want to use it here. I updated `HowToAcknowledge`, as that seems fine. I'm a little inclined to try to move these things into `PipelineDescription` in the draft spec, to more clearly distinguish descriptions of the data and the tools that have been applied to it.

oesteban: You are right, I meant both `ReferencesAndLinks` and `PipelineDOI` to live under `PipelineDescription`. If moved there, then `PipelineDOI` might better be represented by `SoftwareDOI` (or similar, the idea being people setting the zenodo DOI here).

.circleci/ds005_outputs.txt
```diff
@@ -1,4 +1,5 @@
 fmriprep
+fmriprep/dataset_description.json
```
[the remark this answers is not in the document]
effigies: Sorry for missing that. Fixed.

oesteban:
WDYT about https://github.com/poldracklab/fmriprep/pull/1247#discussion_r210380907 ? I would agree that `ReferencesAndLinks` might be a bit cumbersome to fill in within this PR, but the pipeline DOI would be pretty straigthforward (hardcode the DOI to the latest version).

Also, it could be interesting to add setting the env variables you proposed in this PR.

effigies:
@oesteban How about we merge this (barring any objections) and consider *additional* metadata separately?

oesteban:
I'm fine, after all we are talking right now about this. Let's wait for the meeting to consider additional metadata.

BTW, have you checked this https://app.codacy.com/app/code_3/fmriprep/pullRequest?prid=2070359?

PR #959, opened 2018-02-03
[WIP,FIX]: fix deriving confounds when using ICA-AROMA

jdkent:
## Summary
Including the confounds derived from the original bold image file and using those confounds in the same model as the denoised bold image file may re-introduce noise back into the image. To hopefully remedy this situation, this pull request replaces inputs where the original bold file would go with a denoised bold file if `--use-aroma` was specified.

## Addresses
 #817, ~~and I would like it to fix #870 while I'm doing this, but I didn't immediately see the problem connection.~~

## TODO
- [X] connect native space denoised output to outputnode
^^ decide what what file types should be passed on, since using the original bold file is no longer appropriate if the `confounds.tsv` represents signals extracted from the denoised image.

PR #1255, opened 2018-08-20
[ENH]: add nonaggressively denoised images in different output spaces

jdkent:

ref #959 
I have worked on applying nonaggressive denoising to other outputs  in a [previous pull request](https://github.com/poldracklab/fmriprep/pull/959#issuecomment-364814090), and there was a [recent comment](https://neurostars.org/t/fmriprep-ica-aroma-output-always-in-mni152-2mm/2288) about the outputs only being in MNI space.

The following changes to the files add the functionality so that if a user specifies `--use-aroma` the output in the selected output spaces will have the `variant-AROMAnonaggr` key-value pair.

The important exception is the CIFTI outputs since I didn't see a clear path towards marking those files as different (the variant key appears to already be taken).
If `--use-aroma` is specified, the CIFTI filenames will be the same as if `use-aroma` hadn't been specified, but underlying data will still be denoised. 
As a consequence, `--keep-non-denoised` currently does not work on CIFTI outputs.
(This may change with help on how to best change the cifti outputs)
EDIT: ^^Keeping this as is unless someone else has input

Also, I feel like there has to be a cleaner/more clever way of adding this functionality in than the bunch of `if` statements I added, so feedback is much appreciated if anyone has ideas about that.
EDIT: ^^Keeping as is unless someone else has input

Finally, I have a question on whether I should pass a mask to the `FilterRegressor` interface.
I need to verify whether passing a mask just denoised that part of the image or if the output removes (makes  zero) all the voxels that are outside the mask.
EDIT:^^I should not pass a mask since that masks the output.

- `.circleci/config.yml`: add --keep-non-denoised to the ds005 fmriprep runs
- `.circleci/ds005_outputs.txt`: add the new denoised outputs in the specified output spaces
- `.circleci/ds005_partial_outputs.txt`: same as above
- `docs/workflows.rst`: add the keep_non_denoised parameter to the functions
- `fmriprep/cli/run.py`: add the --keep-non-denoised commandline option
- `fmriprep/workflows/base.py`: pass the keep_non_denoised parameter along and update docstrings
- `fmriprep/workflows/bold/base.py`: add if conditionals everywhere to incorporate the desired functionality of this commit
- `fmriprep/workflows/tests/test_base.py`: add keep_non_denoised as a parameter to the test workflow

Currently, the only change in the documentation is adding the `keep_non_denoised` parameter to the workflow functions.


fmriprep/workflows/bold/base.py
```diff
@@ -146,6 +147,11 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
         template_out_grid : str
             Keyword ('native', '1mm' or '2mm') or path of custom reference
             image for normalization
+        keep_non_denoised : bool
```
emdupre: This language might get a little confusing as we're also integrating tedana denoising. Should we clarify if AROMA vs tedana denoising was used, rather than just "non-denoised", generally ?

jdkent: I agree, that will get confusing. So this could be `keep_aroma_non_denoised`?

And then other variables with `denoised` with have `aroma` inserted into their names so that will not lead to conflict.

Is there an opinion on any shortening of the word denoised that would still be clear? (just so the variables don't become overly long, but still descriptive)
1. `dnsd`
2. `densd`
3. `dnsed`

effigies: Are there situations in which someone will perform both tedana and AROMA denoising? And if not, would you want to keep a variant that is denoised by one and no the other?

If not, this could be a generic flag.

emdupre: I'd argue that we should make denoising by tedana and AROMA mutually exclusive, since at least tedana will not function properly if performed after AROMA denoising

In the case of tedana, then, this would be keeping the optimal combination in addition to the denoised time series ? We do this by default, but I'm fine to make it optional :+1:

effigies: That seems fine to me. Am I correct in understanding "optimal combination" to be a synthetic BOLD series that estimates the true signal as closely as possible from the multiple echoes?

I think there are two (potentially) competing impulses here. The first is to give people the outputs they're going to use and not (especially if large) those they won't, while the second is to provide consistent behavior when analogous algorithms are applied.

The assumption with AROMA is that people want the non-aggressively denoised variant, and aren't going to bother with the standard preprocessed BOLD series except in exploratory situations. I'm not sure what the equivalent assumptions are with tedana. If they're too different, we can break the second rule, but my guess is that you'll generally want one or the other, so it'll mostly be a question of what's the default and what the best way of indicating that is. (e.g. perhaps we want `--tedana=combined` and `--tedana=denoised`)

Getting fairly off-topic for this PR, though. Sorry.

effigies:
@jdkent At the recent BIDS convergence, `variant` has been renamed to `desc`. In the next couple weeks, we're going to be reworking fMRIPrep derivatives to match the upcoming derivatives spec release candidate. So just a warning that you may have a moving target.

And there is the question of whether we should be denoising before detrending at all, as discussed on [Twitter](https://twitter.com/SudoNeuroSci/status/1037037692729679878)/[NeuroStars](https://neurostars.org/t/testing-different-noise-models-output-from-fmriprep-unclear-results-with-ica-aroma/2352/12). Pending this decision, I'm inclined to hold off on this PR for the time being. Please let me know if you disagree. (@chrisfilo: Same to you.)

fmriprep/workflows/bold/base.py
```diff
@@ -689,50 +710,104 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
...
+                workflow.connect([
+                    (bold_surf_wf, gen_cifti, [
+                        ('targets.out', 'surface_target'),
+                        ('outputnode.surfaces', 'gifti_files')]),
+                    (inputnode, gen_cifti, [('subjects_dir', 'subjects_dir')]),
+                    (bold_mni_trans_wf, gen_cifti, [('outputnode.bold_mni', 'bold_file')]),
+                    (gen_cifti, outputnode, [('out_file', 'bold_cifti'),
+                                             ('variant', 'cifti_variant'),
+                                             ('variant_key', 'cifti_variant_key')]),
+                ])
```
effigies: Almost all of this is the same in both the `if` and `else` block. You could keep the original and change one line:

```Diff
  workflow.connect([
-     (bold_surf_wf, gen_cifti [
+     (bold_surf_denoised_wf if use_aroma else bold_surf_wf, gen_cifti [
          ('targets.out', 'surface_target'),
          ('outputnode.surfaces', 'gifti_files')]),
```

jdkent:
@effigies I see this as a moving target as well, so I am good with waiting until the other items are resolved.

I'm currently leaning towards providing a 4D voxelwise regressor in place of the non-aggressive outputs so people can still use multiple denoising strategies on their outputs (which would still mean providing the ability to generate the 4D regressor in multiple output spaces, but would simplify the logic since there wouldn't be a `both` option to contend with).

oesteban:
Hi @jdkent, I can allocate some time by the end of next week to take a look on this. Could you resync with master?

oesteban:
For the last month or so we are seeing many AROMA-related issues creeping up (#1816, #1833, #1766, #1389, etc.). In #1784, we consider whether we should have a simpler and custom implementation of AROMA that works on native space.

As I see it now, I don't think it makes much sense to generate AROMA conversions on spaces other than MNI152NLin6Asym, and even less so if we finally do the filtering on native space.

Would you agree with closing this PR which has become pretty stale and maybe join all hands to try to get #1784 done?

jdkent:
Yes, I am okay with closing this pull request, the main (eventual) goal for me was to have ICA-AROMA run on native space so we could recalculate the relevant confounds so they would be orthogonal to the denoised data.

PR #1263, opened 2018-08-28
[ENH] Update ME-EPI workflow to create optimal combination

emdupre:

- Removes in-package implementation of T2* map, wraps tedana in a nipype style interface

- IMPORTANT ! Integrates optimal combination into fmriprep workflow as a default. Previously, users would supply multi-echo files and receive all processed, individual echos back. Now, users will receive a single, optimally combined time series when using multi-echo files. This is in response to queries like [this one on NeuroStars](https://neurostars.org/t/fmriprep-combining-multi-echo-data-error-with-unwarping/2310/3), and several offline requests. Crucially, however, this effectively removes the ability for users to run fmriprep and _then_ tedana, as tedana denoising will now need to be _integrated_ into fmriprep entirely, probably by subbing out the AROMA connection. I would, though, like to have this working before opening another PR for adding in denoising.

Update docstring for `T2SMap`, also updated workflow description for `bold_t2s_wf`. In both cases, removed graphs (as I considered these unhelpful, since so many of the steps are wrapped within `tedana`).

### DIFF
--- a/docs/workflows.rst
+++ b/docs/workflows.rst
@@ -574,7 +574,7 @@ this workflow uses the `tedana`_ `T2* workflow`_ to generate an adaptive T2* map
 and optimally weighted combination of all supplied single echo time series.
 This optimaly combined time series is then carried forward for all subsequent
 preprocessing steps.
-Optionally, if the ``--t2s_coreg`` flag is supplied, the T2* map is then used
+Optionally, if the ``--t2s-coreg`` flag is supplied, the T2* map is then used
 in place of the :ref:`BOLD reference image <bold_ref>` to
 :ref:`register the BOLD series to the T1w image <bold_reg>` of the same subject.
 
--- a/docs/workflows.rst
+++ b/docs/workflows.rst
@@ -372,21 +372,15 @@ T2* Driven Coregistration
 ~~~~~~~~~~~~~~~~~~~~~~~~~
 :mod:`fmriprep.workflows.bold.t2s.init_bold_t2s_wf`
 
-.. workflow::
-    :graph2use: colored
-    :simple_form: yes
-
-    from fmriprep.workflows.bold import init_bold_t2s_wf
-    wf = init_bold_t2s_wf(
-        bold_echos=['echo1', 'echo2', 'echo3'],
-        echo_times=[13.6, 29.79, 46.59],
-        mem_gb=3,
-        omp_nthreads=1)
-
 If the ``--t2s-coreg`` command line argument is supplied with multi-echo
-:abbr:`BOLD (blood-oxygen level-dependent)` data, a T2* map is generated.
+:abbr:`BOLD (blood-oxygen level-dependent)` data, an adaptive T2* map is generated
+using the `tedana`_ `T2* workflow`_.
 This T2* map is then used in place of the :ref:`BOLD reference image <bold_ref>`
 to ref:`register the BOLD series to the T1w image of the same subject <bold_reg>`.
+
+.. _tedana: https://github.com/me-ica/tedana
+.. _`T2* workflow`: https://tedana.readthedocs.io/en/latest/generated/tedana.workflows.t2smap_workflow.html#tedana.workflows.t2smap_workflow  # noqa
+
 
 Susceptibility Distortion Correction (SDC)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -456,7 +450,9 @@ matter boundary (FreeSurfer's ``?h.white`` surfaces).
 
 If FreeSurfer processing is disabled, FSL ``flirt`` is run with the
 :abbr:`BBR (boundary-based registration)` cost function, using the
-``fast`` segmentation to establish the gray/white matter boundary. After :abbr:`BBR (boundary-based registration)` is run, the resulting affine transform will be compared to the initial transform found by FLIRT. Excessive deviation will result in rejecting the BBR refinement and accepting the original, affine registration.
+``fast`` segmentation to establish the gray/white matter boundary.
+After :abbr:`BBR (boundary-based registration)` is run, the resulting affine transform will be compared to the initial transform found by FLIRT.
+Excessive deviation will result in rejecting the BBR refinement and accepting the original, affine registration.
 
 EPI to MNI transformation
 ~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/docs/workflows.rst
+++ b/docs/workflows.rst
@@ -128,7 +128,7 @@ to be run through ``fmriprep``.
 
 Longitudinal processing
 ~~~~~~~~~~~~~~~~~~~~~~~
-In the case of multiple T1w images (across sessions and/or runs), T1w images are 
+In the case of multiple T1w images (across sessions and/or runs), T1w images are
 merged into a single template image using FreeSurfer's `mri_robust_template`_.
 This template may be *unbiased*, or equidistant from all source images, or
 aligned to the first image (determined lexicographically by session label).
@@ -573,7 +573,7 @@ This optimaly combined time series is then carried forward for all subsequent
 preprocessing steps.
 Optionally, if the ``--t2s_coreg`` flag is supplied, the T2* map is then used
 in place of the :ref:`BOLD reference image <bold_ref>` to
-ref:`register the BOLD series to the T1w image <bold_reg>` of the same subject.
+:ref:`register the BOLD series to the T1w image <bold_reg>` of the same subject.
 
 .. _tedana: https://github.com/me-ica/tedana
 .. _`T2* workflow`: https://tedana.readthedocs.io/en/latest/generated/tedana.workflows.t2smap_workflow.html#tedana.workflows.t2smap_workflow  # noqa
--- a/docs/workflows.rst
+++ b/docs/workflows.rst
@@ -270,15 +270,6 @@ BOLD preprocessing
 Preprocessing of :abbr:`BOLD (blood-oxygen level-dependent)` files is
 split into multiple sub-workflows described below.
 
-In the case of multi-echo :abbr:`BOLD (blood-oxygen level-dependent)` data,
-each echo is processed independently. The two exceptions to this occur for
-:ref:`head-motion estimation <bold_hmc>` and :ref:`T2* map creation <bold_t2s>`.
-
-For the :ref:`head-motion estimation workflow <bold_hmc>`, only the first echo
-is submitted as this echo is expected to have the highest contrast between gray
-and white matter. For :ref:`T2* map creation <bold_t2s>`, all echos are
-considered jointly to look at voxel-wise T2* decay.
-
 .. _bold_ref:
 
 BOLD reference image estimation
@@ -365,22 +356,6 @@ command line argument.
 If a :abbr:`BOLD (blood-oxygen level-dependent)` series has fewer than
 5 usable (steady-state) volumes, slice time correction will be disabled
 for that run.
-
-.. _bold_t2s:
-
-T2* Driven Coregistration
-~~~~~~~~~~~~~~~~~~~~~~~~~
-:mod:`fmriprep.workflows.bold.t2s.init_bold_t2s_wf`
-
-If the ``--t2s-coreg`` command line argument is supplied with multi-echo
-:abbr:`BOLD (blood-oxygen level-dependent)` data, an adaptive T2* map is generated
-using the `tedana`_ `T2* workflow`_.
-This T2* map is then used in place of the :ref:`BOLD reference image <bold_ref>`
-to ref:`register the BOLD series to the T1w image of the same subject <bold_reg>`.
-
-.. _tedana: https://github.com/me-ica/tedana
-.. _`T2* workflow`: https://tedana.readthedocs.io/en/latest/generated/tedana.workflows.t2smap_workflow.html#tedana.workflows.t2smap_workflow  # noqa
-
 
 Susceptibility Distortion Correction (SDC)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -583,3 +558,22 @@ A visualization of the AROMA component classification is also included in the HT
     Right hand side of each map: time series (top in seconds),
     frequency spectrum (bottom in Hertz).
     Components classified as signal in green; noise in red.
+
+
+.. _bold_t2s:
+
+T2* Driven Coregistration
+~~~~~~~~~~~~~~~~~~~~~~~~~
+:mod:`fmriprep.workflows.bold.t2s.init_bold_t2s_wf`
+
+If multi-echo :abbr:`BOLD (blood-oxygen level-dependent)` data is supplied,
+this workflow uses the `tedana`_ `T2* workflow`_ to generate an adaptive T2* map
+and optimally weighted combination of all supplied single echo time series.
+This optimaly combined time series is then carried forward for all subsequent
+preprocessing steps.
+Optionally, if the ``--t2s_coreg`` flag is supplied, the T2* map is then used
+in place of the :ref:`BOLD reference image <bold_ref>` to
+ref:`register the BOLD series to the T1w image <bold_reg>` of the same subject.
+
+.. _tedana: https://github.com/me-ica/tedana
+.. _`T2* workflow`: https://tedana.readthedocs.io/en/latest/generated/tedana.workflows.t2smap_workflow.html#tedana.workflows.t2smap_workflow  # noqa
--- a/fmriprep/__about__.py
+++ b/fmriprep/__about__.py
@@ -104,6 +104,7 @@ REQUIRES = [
     'indexed_gzip>=0.8.2',
     'scikit-image',
     'versioneer',
+    'tedana',
 ]
 
 LINKS_REQUIRES = [
--- a/fmriprep/__about__.py
+++ b/fmriprep/__about__.py
@@ -104,10 +104,11 @@ REQUIRES = [
     'indexed_gzip>=0.8.2',
     'scikit-image',
     'versioneer',
-    'tedana',
 ]
 
 LINKS_REQUIRES = [
+    'git+https://github.com/me-ica/tedana.git@'
+    'cb01d725d05f0102e2da89ad50c32905c68a79fd#egg=tedana',
 ]
 
 TESTS_REQUIRES = [
--- a/fmriprep/interfaces/multiecho.py
+++ b/fmriprep/interfaces/multiecho.py
@@ -15,23 +15,23 @@ Change directory to provide relative paths for doctests
 >>> os.chdir(datadir)
 
 """
-import tedana
+import os
 import numpy as np
 
 from nipype import logging
 from nipype.interfaces.base import (
-    traits, TraitedSpec, File, InputMultiPath, SimpleInterface,
-    BaseInterfaceInputSpec)
+    traits, TraitedSpec, File, CommandLine,
+    SimpleInterface, BaseInterfaceInputSpec, CommandLineInputSpec)
 
 LOGGER = logging.getLogger('nipype.interface')
 
 
 class FirstEchoInputSpec(BaseInterfaceInputSpec):
-    in_files = InputMultiPath(File(exists=True), mandatory=True, minlen=2,
-                              desc='multi-echo BOLD EPIs')
-    ref_imgs = InputMultiPath(File(exists=True), mandatory=True, minlen=2,
-                              desc='generated reference image for each '
-                              'multi-echo BOLD EPI')
+    in_files = traits.List(File(exists=True), mandatory=True, minlen=3,
+                           desc='multi-echo BOLD EPIs')
+    ref_imgs = traits.List(File(exists=True), mandatory=True, minlen=3,
+                           desc='generated reference image for each '
+                           'multi-echo BOLD EPI')
     te_list = traits.List(traits.Float, mandatory=True, desc='echo times')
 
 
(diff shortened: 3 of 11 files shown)


fmriprep/workflows/bold/base.py
```diff
@@ -561,28 +560,31 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
...
+                (bold_bold_trans_wf, bold_reg_wf, [
                     ('outputnode.t2s_map', 'inputnode.ref_bold_brain'),
-                    ('outputnode.oc_mask', 'inputnode.ref_bold_mask')]),
+                    ('outputnode.t2s_mask', 'inputnode.ref_bold_mask')]),
             ])
 
+    workflow.connect([
+        # Connect bold_bold_trans_wf -- either original or T2S shadow
+        (inputnode, bold_bold_trans_wf, [
+            ((if multiecho 'ref_file' else 'bold_file'), 'inputnode.name_source')]),
```
emdupre: I'm a little stuck here. I seem to remember seeing something like this at one point (maybe @effigies or @oesteban had done something similar ?), but I can't figure out a nipype appropriate way to do this. Any pointers would be appreciated !

effigies: This is a ternary operator with a minor syntax error. Try `'ref_file' if multiecho else 'bold_file'`.

emdupre: That was it ! Sorry, I had it set up like an if-statement. Thanks !

emdupre:
I've tried to be very explicit in this code (so far) about disconnecting and re-connecting workflows if multi-echo data is provided. I'd like to think there's a cleaner way to do this, but it seems that it would involve shadowing the original `bold_trans_trans_wf` and I'm not sure that's best practice. Any thoughts one way or another would be appreciated !

Currently I'm running up against [an unhelpful error message](https://github.com/nipy/nipype/issues/2245#issue-267309769) for the multi-echo test case, but I'll pick back up again with this tomorrow.

fmriprep/interfaces/multiecho.py
```diff
@@ -78,320 +74,44 @@ def _run_interface(self, runtime):
...
-    for echo in range(2, necho + 1):
+        # tedana expects echo times in milliseconds, rather than seconds
+        echo_times = [te * 1000 for te in self.inputs.te_list]
 
-        # multiply by 1000, so in ms rather than s
-        # ΔS/S = ΔS0/S0 − ΔR2 * TE, so take neg TEs
-        neg_tes = [-1000 * te for te in tes[:echo]]
+        (self._results['t2sv'], self._results['s0v'],
+         self._results['t2svG'], self._results['s0vG'],
+         self._results['ts_OC']) = tedana.t2smap_workflow(self.inputs.in_files, echo_times)
```
effigies: I'm wondering if the issue is here. [`t2smap_workflow`](https://github.com/ME-ICA/tedana/blob/e7f04ff299412befbf309ceece17cd6e4889038e/tedana/workflows/t2smap.py#L78-L194) does not seem to return anything.

I'm not sure why this would result in a failure like the issue you linked to, but sometimes things cascade weirdly.

fmriprep/interfaces/multiecho.py
```diff
@@ -78,320 +74,44 @@ def _run_interface(self, runtime):
...
-        self._results['masked_t2s'] = fname_presuffix(
-            self.inputs.image, suffix='_masked' + ext, newpath=runtime.cwd, use_ext=False)
-        masked_image.to_filename(self._results['masked_t2s'])
-        return runtime
-
-
 class T2SMapInputSpec(BaseInterfaceInputSpec):
-    in_files = InputMultiPath(File(exists=True), mandatory=True,
+    in_files = InputMultiPath(File(exists=True), mandatory=True, minlen=2,
                               desc='multi-echo BOLD EPIs')
```
effigies: If there's a minimum length of 2, then this should just be a `traits.List`, not an `InputMultiPath`.

fmriprep/interfaces/multiecho.py
```diff
@@ -78,320 +74,44 @@ def _run_interface(self, runtime):
...
 
 
 class T2SMapOutputSpec(TraitedSpec):
-    t2s_vol = File(exists=True, desc='T2* map')
-    opt_comb = File(exists=True, desc='optimal combination of echos')
+    t2sv = File(exists=True, desc='limited T2* map')
+    s0v = File(exists=True, desc='limited s0 map')
+    t2svG = File(exists=True, desc='adaptive T2* map')
+    s0vG = File(exists=True, desc='adaptive s0 map')
+    ts_OC = File(exists=True, desc='optimally combined ME-EPI time series')
```
effigies: I would prefer more descriptive output names, such as `t2star_map`, `s0_map`, etc.

fmriprep/interfaces/multiecho.py
emdupre: It looks like `InputMultiPath` this use in the [`FirstEcho` interface](https://github.com/poldracklab/fmriprep/blob/f5b502c329daef0a1d0766c1553d08cb65115af9/fmriprep/interfaces/multiecho.py#L34) as well -- should I update both ?

fmriprep/interfaces/multiecho.py
emdupre: Yes, I can definitely do that ! Sorry, these were taken from tedana itself.

fmriprep/interfaces/multiecho.py
emdupre: You're totally right -- it writes out the files rather than returning a value. I'll go back to some of the other nipype interfaces and check to see how they're capturing outputs.

fmriprep/interfaces/multiecho.py
effigies: Yes.

emdupre:
Ok, I've updated the interface to address @effigies review comments (thank you !), but I'm still receiving the same error. I'm still mostly sure it's related to the `JoinNode`, or at least an `IdentityInterface`. Will update as I work through this !

emdupre:
Just as an update, .@effigies was able to solve my error ! From the slack channel:
> My guess is what’s happening is that when you disconnect `bold_bold_trans_wf`, the nodes still exist in the workflow but aren’t receiving inputs. If that’s the case, probably what’s needed is to conditionally connect, rather than connect and then conditionally disconnect.
Also if that’s true, it’s a really weird way for the error to express, and probably worth seeing if we can detect situations like that better in nipype.

This was exactly the issue -- I've rearranged the `bold_t2s_wf` so that now `bold_bold_trans_wf` is called outside of the T2* workflow and unconnected nodes are no longer present.

I'm now working on (what I think is an unrelated issue):
```
Standard error:
  File "/usr/local/miniconda/bin/t2smap", line 11, in <module>
    sys.exit(_main())
  File "/usr/local/miniconda/lib/python3.6/site-packages/tedana/workflows/t2smap.py", line 200, in _main
    if options.debug and not options.quiet:
AttributeError: 'Namespace' object has no attribute 'debug'
```

Will report back.

effigies:
Via https://mail.nmr.mgh.harvard.edu/pipermail/freesurfer/2018-September/058436.html

> Unfortunately, the disk that hosts the ftp data has died, so resolving this might take a bit longer than expected. The new disk should be repopulated within the next few days.

emdupre:
OK, I know this isn't passing CI because of the Freesurfer FTP issue, but I think this is actually ready for review !

It's passing locally (see included screenshot) and now we're down to one output file for all provided echos; i.e., we're getting the optimal combination back, so users could call AROMA or whatever they'd like on it until we integrate tedana. Slice timing is also finally incorporated for MEEPI data, so I'm sure users will be happy to be past that error.

The T2* coreg flag is still performing worse than I would have expected, but better than before. I'd prefer to address that in a different PR, since this one is so big already ! But LMK what you think is easiest / best.

screenshot_2018-09-05 screenshot

fmriprep/workflows/bold/base.py
```diff
@@ -539,48 +528,69 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
...
     if multiecho:
-        inputnode.iterables = ('bold_file', bold_file)
+        from .util import init_skullstrip_bold_wf
+        inputnode.inputs.bold_file = ref_file  # Replace reference w first echo
+
+        if run_stc:
+            workflow.disconnect([
+                (bold_reference_wf, bold_stc_wf,
+                    [('outputnode.bold_file', 'inputnode.bold_file')])])
+            meepi_echos = boldbuffer.clone(name='meepi_echos')
```
emdupre: Is there a way to pass iterables to `LiterateWorkflow` directly ?

emdupre: Pinging @effigies as the nipype guru :)

effigies: Treat `LiterateWorkflow` the same as any other `Workflow`. Or is the question how to pass iterables to any workflow?

emdupre: Yes, please.

emdupre: Resolved in person -- the used approach is best for now.

romainVala:
Hi

I am very interested by this pull request, and so thanks for sharing your work
I came to fmriprep to preprocess the multiecho data in order to run the tedana denoising. So I am surprise by your choice

>     * however, this effectively removes the ability for users to run fmriprep and _then_ tedana, as tedana denoising will now need to be _integrated_ into fmriprep entirely, probably by subbing out the AROMA connection.

I understand that it will be easier to have all step include in fmriprep, but it is important to have some critical preprocessing step in order to be able to run different pipeline combination

I do not agree to have only the optimal echo recombination (by the way what is the optimal combination : the SNR weighted sum of the exp(-T2/TE) weighted sum ?). 
Why not just produce both (each preprocessed echo and the  optimally combined echo ) and later the denoise version. 
May be just a command line argument that will give you the choice to have all.

May be I am wrong, and I can understand the main purpose of fmriprep which is to produce the 'best preprocessed' finale time series, but I thing it is not incompatible with the possibilité to provide specific intermediary results.

Cheers

emdupre:
Hi @romainVala, thanks for sharing your thoughts ! This PR is not meant to be standalone; the plan is that we will *also* incorporate tedana denoising directly into fMRIPrep in a separate PR.

To make this a little clearer (and to prevent releasing a version with only partial multiecho support), @effigies has created a `multiecho` branch where I'll continue this work. The thought, then, would be that if you would like the individual echos returned, we'll add in an `--echo_id` flag (similar to the `--task_id` flag) where you can select individual echos for processing.

Let me know if this addresses your concerns !

effigies:
Alas, conflicts to merge...

emdupre:
Assuming I handled this merge conflict correctly, I'm going to add one more commit re: using #1278 and then I think this is ready for review !

fmriprep/workflows/bold/base.py
```diff
@@ -533,19 +556,55 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
...
+        workflow.disconnect([
+            (bold_sdc_wf, bold_reg_wf, [
+                ('outputnode.bold_ref_brain', 'inputnode.ref_bold_brain'),
+                ('outputnode.bold_mask', 'inputnode.ref_bold_mask')]),
+        ])
+        workflow.connect([
+            (bold_t2s_wf, bold_reg_wf, [
+                ('outputnode.bold_ref', 'inputnode.ref_bold_brain'),
+                ('outputnode.bold_mask', 'inputnode.ref_bold_mask')]),
+        ])
```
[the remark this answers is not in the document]
effigies: Just a quick note, it's the now-missing `bold_reg_wf.inputnode.ref_bold_brain` that's causing the test failures. So you may want to grep through for other cases like this.

emdupre:
I used to call `bold_bold_trans_wf` _in_ `bold_t2s_wf`, since we need the individual echos to be motion corrected before creating the optimal combination. I moved it to just before `bold_t2s_wf` to get around the `TypeError("'_Undefined' object does not support indexing ...")` issue we were seeing earlier when I wasn't fully disconnecting the node.

As to why to not feed the inputs into `bold_bold_trans_wf`, it would effectively be trying to derive motion parameters on motion-corrected files, so I thought it would be better to avoid that extra interpolation. Otherwise we'll have to run it twice (once before and once after `bold_t2s_wf`). 

Does that make sense ? WDYT ?

effigies:
Ah, right. Sorry, I forgot about the whole motion correction thing. Let me read through again with that in mind.

fmriprep/utils/misc.py
```diff
@@ -24,6 +24,26 @@ def fix_multi_T1w_source_name(in_files):
...
+    ...     'sub-01_run-01_echo-1_bold.nii.gz',
+    ...     'sub-01_run-01_echo-2_bold.nii.gz',
+    ...     'sub-01_run-01_echo-3_bold.nii.gz',])
+    'sub-01_run-01_bold.nii.gz'
+
+    """
+    import os
+    from nipype.utils.filemanip import filename_to_list
+    base, in_file = os.path.split(filename_to_list(in_files)[0])
+    keys_pre_echo = in_file.split("_")[:-2]
```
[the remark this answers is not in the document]
emdupre: in the current spec, echo is guaranteed to be last ! happy to update if you'd like to future-proof, though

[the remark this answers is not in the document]
effigies: Either way. If this works for all valid datasets, then I'm okay with it.

fmriprep/workflows/bold/base.py
```diff
@@ -438,20 +432,32 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
...
-        workflow.connect([
-            (bold_reference_wf, boldbuffer, [
-                ('outputnode.bold_file', 'bold_file')]),
-        ])
+        if not multiecho:
+            workflow.connect([
+                (bold_reference_wf, boldbuffer, [
+                    ('outputnode.bold_file', 'bold_file')])])
+        else:  # for meepi, iterate over all meepi echos to boldbuffer
+            boldbuffer.iterables = ('bold_file', bold_file)
```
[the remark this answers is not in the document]
emdupre: Sorry, I don't understand this suggestion-- I don't think I can have an `elif / else` block without an `if` statement ?

fmriprep/workflows/bold/base.py
```diff
@@ -471,6 +477,24 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
...
+                                 joinsource=('meepi_echos' if run_stc is True else 'boldbuffer'),
+                                 joinfield=['bold_files'],
+                                 name='join_echos')
+
+        # create optimal combination, adaptive T2* map
+        bold_t2s_wf = init_bold_t2s_wf(echo_times=tes,
+                                       mem_gb=mem_gb['resampled'],
+                                       omp_nthreads=omp_nthreads,
+                                       name='bold_t2smap_wf')
+
```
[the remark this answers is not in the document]
emdupre: Moved !

emdupre:
The T2* map creation and the optimal combination are done in one step, so each echo needs to be SDC'ed separately. Right now I'm trying to think through the best way to combine the transforms, though, since I need the optimal combination to be in the same space as the T1w..

fmriprep/workflows/bold/base.py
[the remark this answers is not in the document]
effigies: Sorry for being unclear. I mean the current structure is:

```Python
if run_stc is True:
    # block A
else:
    if not multiecho:
        # block B
    else:
        # block C
```

This could become either:

```Python
if run_stc is True:
    # block A
elif not multiecho:
    # block B
else:
   # block C
```

or:

```Python
if run_stc is True:
    # block A
elif multiecho:
    # block C
else:
    # block B
```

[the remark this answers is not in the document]
emdupre: I see, but there are really four options, no ? It's all combinations of `run_stc is True` and `multiecho is True`. So I could do something like:

```
if run_stc is True and multiecho is True:
    # block A
elif run_stc is False and multiecho is True:
    # block C
...
```

fmriprep/workflows/bold/base.py
```diff
@@ -451,13 +451,13 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
...
             workflow.connect([
                 (meepi_echos, bold_stc_wf, [('bold_file', 'inputnode.bold_file')])])
-    else:  # bypass STC from original BOLD to the splitter through boldbuffer
-        if not multiecho:
-            workflow.connect([
-                (bold_reference_wf, boldbuffer, [
-                    ('outputnode.bold_file', 'bold_file')])])
-        else:  # for meepi, iterate over all meepi echos to boldbuffer
-            boldbuffer.iterables = ('bold_file', bold_file)
+    elif (run_stc is False) and not multiecho:
```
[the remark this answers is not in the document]
emdupre: You're right ! Sorry, fixed.

effigies:
> The T2* map creation and the optimal combination are done in one step, so each echo needs to be SDC'ed separately.

Okay, then let's keep things as they are, for now.

> Right now I'm trying to think through the best way to combine the transforms, though, since I need the optimal combination to be in the same space as the T1w.

Not knowing details of tedana, I apologize for any absurdities that follow:

Perhaps it could be written as a two-step process, where the T2* map is created as well as the transform (mixing matrix?) from the BOLD series, but the optimal combination is not generated. Then we could run SDC and coregistration on the T2*, and pass in the motion parameters, warp field, and affine transform (and possibly T1-template warp) to perform a single shot optimal-combination construction.

effigies:
If I'm reading that last correctly, you're just disabling SDC for multiecho? If so, then it probably makes sense to just do that at the bold workflow level, and let the existing logic take care of handling the transforms.

emdupre:
I don't think it's being totally disabled (if I'm understanding [the processing tree](https://fmriprep.readthedocs.io/en/stable//workflows-4.pdf) correctly) since SDC is still feeding into `bold_bold_trans_wf`, which is what I'm using to derive processed echos from which the optimal combination is generated. Does that match your understanding ?

effigies:
Oh, gotcha. It is hard to keep all these things straight...

emdupre:
It really, really is :sweat_smile: I'm happy to talk through this to make sure we're all on the same page before merging ! I'm hitting a memory error locally so I want to see if it replicates on circle, but I think this is close....

emdupre:
Ok, this is finally ready ! It's brought up a few subsequent issues I'd like to address (the occasional T2* coregistration seeming to have bad q/sform codes, updating the boilerplate, when to skullstrip the optimal combination), but overall this is a big step forward I think.

Thanks so much to  everyone, and in particular to @effigies for all of the time you've already taken to review / help this PR :hugs:

docs/workflows.rst
```diff
@@ -592,3 +561,22 @@ A visualization of the AROMA component classification is also included in the HT
...
+T2* Driven Coregistration
+~~~~~~~~~~~~~~~~~~~~~~~~~
+:mod:`fmriprep.workflows.bold.t2s.init_bold_t2s_wf`
+
+If multi-echo :abbr:`BOLD (blood-oxygen level-dependent)` data is supplied,
+this workflow uses the `tedana`_ `T2* workflow`_ to generate an adaptive T2* map
+and optimally weighted combination of all supplied single echo time series.
+This optimaly combined time series is then carried forward for all subsequent
+preprocessing steps.
+Optionally, if the ``--t2s_coreg`` flag is supplied, the T2* map is then used
```
effigies: `t2s-coreg` (hyphen, not underscore)

https://github.com/poldracklab/fmriprep/blob/8a4de9fccd28d06ce765df1a5aea4e4abb9ed0f2/fmriprep/cli/run.py#L117

fmriprep/__about__.py
```diff
@@ -108,6 +108,8 @@
 ]
 
 LINKS_REQUIRES = [
+    'git+https://github.com/me-ica/tedana.git@'
+    'cb01d725d05f0102e2da89ad50c32905c68a79fd#egg=tedana',
```
effigies: Since we're merging to `multiecho` and not `master`, this isn't urgent, but it would be good to depend on a release version of tedana before we have fMRIPrep releases that depend on this commit.

fmriprep/workflows/bold/base.py
```diff
@@ -301,17 +298,15 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
...
         t2s_coreg = False
 
-    # Switch stc off
-    if multiecho and run_stc is True:
-        LOGGER.warning('Slice-timing correction is not available for '
-                       'multiecho BOLD data (not implemented).')
-        run_stc = False
+    # By default, force-bbr for t2s_coreg unless user specifies otherwise
+    if t2s_coreg and (use_bbr is not False):
+            use_bbr = True
```
effigies: Overindented.

fmriprep/workflows/bold/base.py
```diff
@@ -301,17 +298,15 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
...
                        "Using standard EPI-T1 coregistration.")
         t2s_coreg = False
 
-    # Switch stc off
-    if multiecho and run_stc is True:
-        LOGGER.warning('Slice-timing correction is not available for '
-                       'multiecho BOLD data (not implemented).')
-        run_stc = False
+    # By default, force-bbr for t2s_coreg unless user specifies otherwise
+    if t2s_coreg and (use_bbr is not False):
```
effigies: How about `if t2s_coreg and use_bbr is None:`? If it's `True`, this branch has no effect anyway, and we can be more explicit that this is the "default" behavior. Also, can you add a note to the docstring (L97-99)?

fmriprep/workflows/bold/registration.py
```diff
@@ -297,31 +299,47 @@ def init_bold_t1_trans_wf(freesurfer, mem_gb, omp_nthreads, use_fieldwarp=False,
...
+        ])
 
     workflow.connect([
-        (inputnode, merge_xforms, [('itk_bold_to_t1', 'in1')]),
-        (merge_xforms, bold_to_t1w_transform, [('out', 'transforms')]),
-        (inputnode, merge, [('name_source', 'header_source')]),
-        (inputnode, bold_to_t1w_transform, [('bold_split', 'input_image')]),
         (gen_ref, bold_to_t1w_transform, [('out_file', 'reference_image')]),
+        # merge 3D volumes into 4D time series
+        (inputnode, merge, [('name_source', 'header_source')]),
```
effigies: Sorry, could you move this back above, just to reduce the diff to actually changed behavior?

fmriprep/workflows/bold/registration.py
```diff
@@ -297,31 +299,47 @@ def init_bold_t1_trans_wf(freesurfer, mem_gb, omp_nthreads, use_fieldwarp=False,
...
-    workflow.connect([
-        (inputnode, merge_xforms, [('hmc_xforms', 'in%d' % nforms)])
-    ])
-
-    if use_fieldwarp:
-        workflow.connect([
-            (inputnode, merge_xforms, [('fieldwarp', 'in2')])
-        ])
+    # merge 3D volumes into 4D timeseries
+    merge = pe.Node(Merge(compress=use_compression), name='merge', mem_gb=mem_gb)
```
effigies: Could you move this back below `bold_to_t1w_transform`?

fmriprep/workflows/bold/registration.py
emdupre: It is changed behavior here, I think -- now it's called regardless of whether MEEPI data is provided, whereas before it was only called for non-MEEPI data

fmriprep/__about__.py
emdupre: :+1: yes, there's a growing list of things `tedana` needs, but cutting a release soon is definitely one of them

fmriprep/workflows/bold/registration.py
effigies: Sorry, I just meant "above  the `(gen_ref, bold_to_t1w_transform, ` line". It's identical with the removed line 322.

emdupre: ok, I can get rid of the identical comment ! sorry, I didn't realize I had duplicated that. or will removing that cause more headache for the diff ?

effigies: Sorry, I don't understand the question. The latest update looks good, though.

effigies:
Sorry, thought I'd already merged this...

Issue #1150, opened 2018-05-28
Coregistration failure - Rotation ?


emdupre:
As promised in #999, I'm opening a new issue for rotated coregistration with multi-echo. This is a problem I'm still seeing with the latest release (v. 1.0.15) and the same data set as shared previously on OpenNeuro:

<img width="1125" alt="screen shot 2018-05-28 at 16 38 13" src="https://user-images.githubusercontent.com/15017191/40628870-9ad52664-6295-11e8-8b9e-9163692c685d.png">

The only thing I could think that's differing from my local execution and on OpenNeuro is that I'm using a previously computed FreeSurfer directory, though I'm not sure how it could be causing this issue. 

Any (other) ideas where to start looking would be deeply appreciated !

oesteban:
Can you point me to these data on OpenNeuro?

emdupre:
This subject should be available [here](https://openneuro.org/datasets/ds001239) !

PR #1270, opened 2018-09-04
ENH: Use single-band reference images when available

### LINKED ISSUE #449: Add support for SBRef images
All it takes is that the optimal reference estimation procedure to return the SBRef image.

effigies:

For each BOLD file, this approach finds the associated `sbref` file, if available, and passes it to `EstimateReferenceImage`.

One hangup here is that we need to validate both the bold file and sbref if available. To avoid weird conditionals, I put a second, post-reference-estimation validation step. If the image is estimated from the BOLD series, it should have no effect. If the sbref is returned, it will get the same treatment as the BOLD series did prior to estimation.

An alternative approach would be to generate the reference image, and then validate both the BOLD series and the reference image afterwards. I don't think it would make any computational effects (unless the reference process implicitly depends on good affine codes), but it might be a little easier to understand for future maintainers.

Given the possibility of implicit dependence on affine codes, I chose not to take that alternative, but content to be overruled.

Closes #449.

## To-do

* [x] Add `--ignore sbref` option to CLI to retain current behavior

Added a couple lines to [BOLD reference image estimation](https://fmriprep.readthedocs.io/en/stable/workflows.html#bold-reference-image-estimation).

### DIFF
--- a/fmriprep/workflows/bold/base.py
+++ b/fmriprep/workflows/bold/base.py
@@ -242,6 +242,7 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
                     'Memory resampled/largemem=%.2f/%.2f GB.'),
                ref_file, mem_gb['filesize'], bold_tlen, mem_gb['resampled'], mem_gb['largemem'])
 
+    sbref_file = None
     # For doc building purposes
     if layout is None or bold_file == 'bold_preprocesing':
         LOGGER.log(25, 'No valid layout: building empty workflow.')
@@ -257,7 +258,6 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
             'magnitude2': 'sub-03/ses-2/fmap/sub-03_ses-2_run-1_magnitude2.nii.gz',
         }]
         run_stc = True
-        sbref_file = None
         multiecho = False
     else:
         # Find associated sbref, if possible
--- a/fmriprep/workflows/bold/base.py
+++ b/fmriprep/workflows/bold/base.py
@@ -265,7 +265,10 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
         entities['type'] = 'sbref'
         files = layout.get(**entities)
         refbase = os.path.basename(ref_file)
-        if files:
+        if files and multiecho:
+            LOGGER.warning("Single-band reference found, but not supported in "
+                           "multi-echo workflows at this time. Ignoring.")
+        elif files:
             sbref_file = files[0].filename
             sbbase = os.path.basename(sbref_file)
             if len(files) > 1:
--- a/fmriprep/workflows/bold/base.py
+++ b/fmriprep/workflows/bold/base.py
@@ -257,8 +257,26 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
             'magnitude2': 'sub-03/ses-2/fmap/sub-03_ses-2_run-1_magnitude2.nii.gz',
         }]
         run_stc = True
+        sbref_file = None
         multiecho = False
     else:
+        # Find associated sbref, if possible
+        entities = layout.parse_file_entities(ref_file)
+        entities['type'] = 'sbref'
+        files = layout.get(**entities)
+        refbase = os.path.basename(ref_file)
+        if files:
+            sbref_file = files[0].filename
+            sbbase = os.path.basename(sbref_file)
+            if len(files) > 1:
+                LOGGER.warning(
+                    "Multiple single-band reference files found for {}; using "
+                    "{}".format(refbase, sbbase))
+            else:
+                LOGGER.log(25, "Using single-band reference file {}".format(sbbase)
+        else:
+            LOGGER.log(25, "No single-band-reference found for {}".format(refbase))
+
         metadata = layout.get_metadata(ref_file)
 
         # Find fieldmaps. Options: (phase1|phase2|phasediff|epi|fieldmap|syn)
@@ -312,13 +330,15 @@ Non-gridded (surface) resamplings were performed using `mri_vol2surf`
 """
 
     inputnode = pe.Node(niu.IdentityInterface(
-        fields=['bold_file', 'subjects_dir', 'subject_id',
+        fields=['bold_file', 'sbref_file', 'subjects_dir', 'subject_id',
                 't1_preproc', 't1_brain', 't1_mask', 't1_seg', 't1_tpms',
                 't1_aseg', 't1_aparc',
                 't1_2_mni_forward_transform', 't1_2_mni_reverse_transform',
                 't1_2_fsnative_forward_transform', 't1_2_fsnative_reverse_transform']),
         name='inputnode')
     inputnode.inputs.bold_file = bold_file
+    if sbref_file is not None:
+        inputnode.inputs.sbref_file = sbref_file
 
     outputnode = pe.Node(niu.IdentityInterface(
         fields=['bold_t1', 'bold_mask_t1', 'bold_aseg_t1', 'bold_aparc_t1', 'cifti_variant',
@@ -441,7 +461,8 @@ Non-gridded (surface) resamplings were performed using `mri_vol2surf`
     # MAIN WORKFLOW STRUCTURE #######################################################
     workflow.connect([
         # Generate early reference
-        (inputnode, bold_reference_wf, [('bold_file', 'inputnode.bold_file')]),
+        (inputnode, bold_reference_wf, [('bold_file', 'inputnode.bold_file'),
+                                        ('sbref_file', 'inputnode.sbref_file')]),
         # BOLD buffer has slice-time corrected if it was run, original otherwise
         (boldbuffer, bold_split, [('bold_file', 'in_file')]),
         # HMC
--- a/fmriprep/workflows/bold/base.py
+++ b/fmriprep/workflows/bold/base.py
@@ -573,7 +573,8 @@ Non-gridded (surface) resamplings were performed using `mri_vol2surf`
         # Replace reference with the echo selected with FirstEcho
         workflow.disconnect([
             (inputnode, bold_reference_wf, [
-                ('bold_file', 'inputnode.bold_file')]),
+                ('bold_file', 'inputnode.bold_file'),
+                ('sbref_file', 'inputnode.sbref_file')]),
             (bold_reference_wf, boldbuffer, [
                 ('outputnode.bold_file', 'bold_file')]),
         ])
--- a/fmriprep/workflows/bold/base.py
+++ b/fmriprep/workflows/bold/base.py
@@ -273,7 +273,7 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
                     "Multiple single-band reference files found for {}; using "
                     "{}".format(refbase, sbbase))
             else:
-                LOGGER.log(25, "Using single-band reference file {}".format(sbbase)
+                LOGGER.log(25, "Using single-band reference file {}".format(sbbase))
         else:
             LOGGER.log(25, "No single-band-reference found for {}".format(refbase))
 
--- a/fmriprep/workflows/bold/base.py
+++ b/fmriprep/workflows/bold/base.py
@@ -265,7 +265,9 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
         entities['type'] = 'sbref'
         files = layout.get(**entities)
         refbase = os.path.basename(ref_file)
-        if files and multiecho:
+        if 'sbref' in ignore:
+            LOGGER.info("Single-band reference files ignored.")
+        elif files and multiecho:
             LOGGER.warning("Single-band reference found, but not supported in "
                            "multi-echo workflows at this time. Ignoring.")
         elif files:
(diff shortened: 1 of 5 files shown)


effigies:
@emdupre Would this screw things up in a multi-echo context?

emdupre:
I personally haven't worked with `sbref` files, so I'm not sure how they'll interact with fast TE acquisitions like the first echo (which we're using as the `ref_file` in the MEEPI workflow). Is there an example dataset available with both that we could test on ?

effigies:
https://drive.google.com/drive/folders/1sZ2hNeGbvfyqjrM3Jqrw9JHrWis3VQNl?usp=sharing (via https://neurostars.org/t/fmriprep-coregistration-and-unwarping-error/2377)

So if you have a `ref_file = sub-01/func/sub-01_task-rest_echo-1_bold.nii.gz`, then we will use a single-band reference `sub-01/func/sub-01_task-rest_echo-1_sbref.nii.gz`, if available. At least as the spec is currently written, the only valid sbref for a BOLD series matches all other entities. So it should be the same TE and all other parameters (unless you're using a multiband sequence for your BOLD series).

effigies:
So we're in a weird place where @oesteban changed how some templates work, and not others. I'm working around it by copying the old `TEMPLATE_MAP` to preserve how registration works.

This means that niworkflows 0.4.3 will not work with fmriprep 1.1.4 and earlier.

chrisfilo:
Could you test this PR on data from https://neurostars.org/t/fmriprep-coregistration-and-unwarping-error/2377?

effigies:
So that one doesn't work because there's no sbref for `echo-1`. @emdupre's multi-echo contributions mean we use the shortest echo time as reference. Not really sure what the best way to deal with this is.

effigies:
They have sbrefs for echoes 2 (TE=0.0343s) and 3 (TE=0.0562s).

effigies:
Would the best approach be to use the shortest TE sbref, regardless of whether it matches the shortest TE BOLD? We would effectively be trusting the dataset curator to include the best contrast sbrefs. If we want to make this more explicit, we could push for an `IntendedFor` field in the spec that would allow curators to specifically override the default assumption that sbrefs must match bold files exactly.

A much more complicated approach would be to attempt to directly compare gray/white contrasts in the automatically estimated reference file and the sbref. I'd rather stick to available data as heuristics, if we can do so reasonably.

chrisfilo:
On Thu, Sep 6, 2018, 5:01 AM Chris Markiewicz <notifications@github.com>
wrote:

> Would the best approach be to use the shortest TE sbref, regardless of
> whether it matches the shortest TE BOLD?
>
I did not expect such discrepancy. Is it the case with the ME dataset you
are using?

emdupre:
It looks like my email didn't go through ! Should have been in response to @effigies 's last comment:

Ah I see ! Sorry, I had tried to look at the google drive link on mobile, which was apparently a poor choice.

So essentially what they would like is to iterate through every echo (except the first) and apply the sbref image. This makes sense since the first echo should not need an sbref (and, indeed, is why we use it as the reference). The idea then would be that any distortion seen in the later echos could be corrected and allow for a better optimal combination / T2* map creation.

The trouble is going to be that in my standing PR I just went through and _stopped_ us from iterating over all the echos so that we only get one file (the optimal combination, and / or eventually the tedana denoised image) as an output.

Though, right now I don't have T2* map hooking into the SDC workflow, so maybe that could be a good next step.

effigies:
> I did not expect such discrepancy. Is it the case with the ME dataset you are using?

If you're referring to the one you asked me to test against, yes.

ihuijsmans:
Hello, the data that you were looking at in the google drive (https://drive.google.com/drive/folders/1sZ2hNeGbvfyqjrM3Jqrw9JHrWis3VQNl?usp=sharing) is collected for a project with me and a master student. 

On neurostarts.org https://neurostars.org/t/fmriprep-coregistration-and-unwarping-error/2377? @chrisfilo suggested that the problems with coregistration might be due to low contrast in the raw data. However, when I looked at the raw data, I was surprised with this suggestion. The image you see below is the first echo of the first volume for run 1 and run 3. Visually it seems that the contrast in run 3 (no problems) and run 1 (coregistration problems) are very similar. Could there be a different problem?

rawdatasub-024

emdupre:
Thanks for reposting, @ihuijsmans !

I think that the coregistration issue is more likely arising from the `--t2s_coreg` flag than the image contrast. As you report, we've seen sporadic failures with that method (see #1150). I've opened #1263 to start addressing this issue by calling `tedana` directly, but my immediate suggestion would be to try those subjects without the `--t2s_coreg` flag.

chrisfilo:
1. Thanks for catching the issue with `--t2s_coreg` @emdupre. @ihuijsmans could you run the problematic participants data without that option and report back? If indeed it makes a difference it would be good for @emdupre to test #1263 using this data.

2. I would still continue the work on SBRef unless there are known regressions. This might be a valuable addition even if initially implemented only for non ME data. I wonder if we should test it on the HCP data which is not ME, but does suffer from low tissue contrast. BIDSified version can be found at https://osf.io/6429f/

3. I also looked at the dataset provided by @ihuijsmans. I initially misread the message thinking that corresponding bold and sbref data for the same run/echo do not have the same EchoTime. This is not the case. Sorry for the confusion.

effigies:
Apparently this is having a bad interaction with ME as is.

```
180906-18:15:14,463 nipype.workflow ERROR:
	 Saving crash info to /out/fmriprep/sub-02/log/20180906-181323_37649c8b-dfcd-4b94-b970-1d68f4d53d27/crash-20180906-181514-root-t2s_map-9f3daadd-1aed-4b58-8da3-5bcdb204f4d2.txt
  File "/usr/local/miniconda/lib/python3.6/site-packages/nipype/pipeline/engine/nodes.py", line 933, in _slot_value
    return getattr(self._inputs, slot_field)
AttributeError: 'DynamicTraitedSpec' object has no attribute 'in_filesJ4'

During handling of the above exception, another exception occurred:

  File "/usr/local/miniconda/lib/python3.6/site-packages/nipype/pipeline/plugins/multiproc.py", line 69, in run_node
  File "/usr/local/miniconda/lib/python3.6/site-packages/nipype/pipeline/engine/nodes.py", line 471, in run
  File "/usr/local/miniconda/lib/python3.6/site-packages/nipype/pipeline/engine/nodes.py", line 555, in _run_interface
  File "/usr/local/miniconda/lib/python3.6/site-packages/nipype/pipeline/engine/nodes.py", line 879, in _run_command
    self._collate_join_field_inputs()
  File "/usr/local/miniconda/lib/python3.6/site-packages/nipype/pipeline/engine/nodes.py", line 889, in _collate_join_field_inputs
    val = self._collate_input_value(field)
  File "/usr/local/miniconda/lib/python3.6/site-packages/nipype/pipeline/engine/nodes.py", line 919, in _collate_input_value
    for idx in range(self._next_slot_index)
  File "/usr/local/miniconda/lib/python3.6/site-packages/nipype/pipeline/engine/nodes.py", line 919, in <listcomp>
    for idx in range(self._next_slot_index)
  File "/usr/local/miniconda/lib/python3.6/site-packages/nipype/pipeline/engine/nodes.py", line 938, in _slot_value
    field, index, e))
AttributeError: The join node fmriprep_wf.single_subject_02_wf.func_preproc_task_cuedSGT_run_01_echo_1_wf.bold_t2s_wf.t2s_map does not have a slot field in_filesJ4 to hold the in_files value at index 3: 'DynamicTraitedSpec' object has no attribute 'in_filesJ4'
```

@emdupre Does this look familiar and easy to fix?

emdupre:
It's not a familiar error, and this would be a very odd way for it to express, but I wonder if it's because not we're not fully disconnecting the `inputnode` and `bold_reference_wf` ? In particular, you could update [this line](https://github.com/poldracklab/fmriprep/pull/1270/files#diff-e705526667f071ca0c3b04c2bd96272bR571) to now additionally disconnect the `sbref_file`.

emdupre:
Glad to hear that helped, @ihuijsmans !

I don't think I'll be able to run the provided data through #1263 (as @chrisfilo suggested) as only two echos for each run were included in the google drive, and we have a check in place to make sure at least three echos are available (otherwise [`tedana`](https://github.com/me-ica/tedana) won't work).

I also am not sure that it will (yet) solve the `--t2s_coreg` problem, as it's really designed to revamp the workflow to provide the optimal combination -- and in the process it seems to be helping to stabilize `--t2s_coreg`. 

Happy to continue the discussion there, though ! I know this is diverging pretty significantly from the purpose of this PR, sorry @effigies !

effigies:
@emdupre @chrisfilo How do you feel about simply not attempting to use sbrefs in a multi-echo context for this PR, to be resolved in #1263? It seems like anything else is going to just cause a massive merge conflict for no real gain.

emdupre:
That's alright by me. If we do tackle ME sbref's here, I'd prefer to merge #1263 first -- so just skipping them in this PR (maybe you could throw in a more helpful error, @effigies ?), sounds fine.

chrisfilo:
Sounds good to me

effigies:
> maybe you could throw in a more helpful error

I'll just post a warning. Hopefully we can get both of these in before the next release, so that nobody but us has to see it.

effigies:
Hmm. I guess it makes sense that this didn't fix the issue, as we never tried to use an sbref in ds210. I still have no idea why this is a problem now but wasn't before.

emdupre:
I assume [this](https://github.com/poldracklab/fmriprep/pull/1270#issuecomment-419271183) did not help ?

effigies:
Somehow I completely failed to read that in full and take appropriate action. Let me try that...

fmriprep/workflows/bold/util.py
```diff
@@ -87,7 +87,8 @@ def init_bold_reference_wf(omp_nthreads, bold_file=None, name='bold_reference_wf
 First, a reference volume and its skull-stripped version were generated
 using a custom methodology of *fMRIPrep*.
```
oesteban: Can we update the boilerplate description with a conditional? That will likely require you to add a new argument to the function, and with it maybe avoid the re-run of validation.

I think it is relevant to write that fMRIPrep used the SBRef to the boilerplate.

fmriprep/workflows/bold/base.py
```diff
@@ -259,6 +260,26 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
         run_stc = True
         multiecho = False
     else:
+        # Find associated sbref, if possible
+        entities = layout.parse_file_entities(ref_file)
```
oesteban: This is great, I'll check whether we can use it in other places and in MRIQC 👍

fmriprep/workflows/bold/base.py
```diff
@@ -259,6 +260,26 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
...
+        entities['type'] = 'sbref'
+        files = layout.get(**entities)
+        refbase = os.path.basename(ref_file)
+        if files and multiecho:
```
oesteban: Just to double check: is the conflict for all ME or only if ``--t2-coreg`` is set?

fmriprep/workflows/bold/base.py
```diff
@@ -259,6 +260,26 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
...
+            LOGGER.warning("Single-band reference found, but not supported in "
+                           "multi-echo workflows at this time. Ignoring.")
+        elif files:
+            sbref_file = files[0].filename
+            sbbase = os.path.basename(sbref_file)
+            if len(files) > 1:
+                LOGGER.warning(
+                    "Multiple single-band reference files found for {}; using "
```
oesteban: What about merging all SBRefs in one 4D file? `gen_ref` would use that correctly, I guess. (BTW, without thinking too much I'd say I'd place the merge within `init_bold_reference_wf`)

fmriprep/workflows/bold/base.py
effigies: The conflict is basically that things are going to change in #1263, and the multi-echo w/ sbref case doesn't have a full set of sbrefs, so let's not mess with it now.

fmriprep/workflows/bold/base.py
effigies: I think the more likely case is that a bold file without an `acq-` tag finds an sbref with and another without. I don't know that in general we want to merge those.

effigies:
> [ ] The new functionalities in this PR are covered by the existing tests

I would check this one. ds054 has sbrefs.

fmriprep/workflows/bold/base.py
oesteban: Yep, we don't want to merge those. However, we should make sure we pick the right SBRef in that case or otherwise the distortion will be present in the opposite PE direction.

In general, I agree on that it is probably not worth splitting hairs by merging SBRefs 👍

fmriprep/workflows/bold/util.py
effigies: For this workflow, the `sbref_file` is a workflow input, not a function input. That makes making it conditional somewhat difficult.

fmriprep/workflows/bold/base.py
effigies: Sure. The point is to warn so that users check that it's right.

The simpler route would just be to `s/bold/sbref/` the `bold_file`, and check for existence.

fmriprep/workflows/bold/util.py
effigies: What's the expected boilerplate if some series have sbref and others don't?

effigies:
@oesteban The main thing holding this up is the boilerplate, which as noted are going to be somewhat complicated. To avoid having a massive merge-conflict-in-waiting hanging over #1263, I'm going to merge and add an issue to update the boilerplate.

Issue #1268, opened 2018-08-31
Sanitize long paths in LTA files


romainVala:
Hello

I hope it is the wright place to report the bug : 
I get systematic Segmentation fault for mri_vol2vol.bin when performing the corregistration of the bold data onto the freesurfer T1.

```
File  "/network/lustre/iss01/apps/lang/anaconda/fmriprep/lib/python3.6/site-packages/niworkflows/nipype/interfaces/freesurfer/base.py", line 144, in run
    return super(FSCommand, self).run(**inputs)
  File "/network/lustre/iss01/apps/lang/anaconda/fmriprep/lib/python3.6/site-packages/niworkflows/nipype/interfaces/base/core.py", line 520, in run
    runtime = self._run_interface(runtime)
  File "/network/lustre/iss01/apps/lang/anaconda/fmriprep/lib/python3.6/site-packages/niworkflows/interfaces/report_base.py", line 60, in _run_interface
    self._post_run_hook(runtime)
  File "/network/lustre/iss01/apps/lang/anaconda/fmriprep/lib/python3.6/site-packages/niworkflows/interfaces/registration.py", line 263, in _post_run_hook
    res = mri_vol2vol.run()
  File "/network/lustre/iss01/apps/lang/anaconda/fmriprep/lib/python3.6/site-packages/niworkflows/nipype/interfaces/freesurfer/base.py", line 144, in run
    return super(FSCommand, self).run(**inputs)
  File "/network/lustre/iss01/apps/lang/anaconda/fmriprep/lib/python3.6/site-packages/niworkflows/nipype/interfaces/base/core.py", line 520, in run
    runtime = self._run_interface(runtime)
  File "/network/lustre/iss01/apps/lang/anaconda/fmriprep/lib/python3.6/site-packages/niworkflows/nipype/interfaces/base/core.py", line 1020, in _run_interface
    self.raise_exception(runtime)
  File "/network/lustre/iss01/apps/lang/anaconda/fmriprep/lib/python3.6/site-packages/niworkflows/nipype/interfaces/base/core.py", line 957, in raise_exception
    ).format(**runtime.dictcopy()))
RuntimeError: Command:
[...]

PR #1274, opened 2018-09-07
FIX: Remove long paths from all LTA output files

romainVala:
Sanitize long paths in LTA files

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -88,6 +88,10 @@
       "affiliation": "Department of Psychology, Stanford University",
       "orcid": "0000-0003-3321-7583"
     }
+    {
+      "name": "Romain, Valabregue",
+      "affiliation": "CENIR, ICM, Paris"
+    }
   ],
   "keywords": [
     "neuroimaging",
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -84,13 +84,14 @@
       "orcid": "0000-0001-6755-0259"
     },
     {
+      "name": "Valabregue, Romain",
+      "affiliation": "Cenir, ICM, Paris",
+      "orcid": "0000-0002-1814-9570"
+    },
+    {
       "name": "Gorgolewski, Krzysztof J.",
       "affiliation": "Department of Psychology, Stanford University",
       "orcid": "0000-0003-3321-7583"
-    }
-    {
-      "name": "Romain, Valabregue",
-      "affiliation": "CENIR, ICM, Paris"
     }
   ],
   "keywords": [
--- a/fmriprep/interfaces/freesurfer.py
+++ b/fmriprep/interfaces/freesurfer.py
@@ -34,7 +34,7 @@ from nipype.interfaces.base import (
 )
 from nipype.interfaces import freesurfer as fs
 from nipype.interfaces.base import SimpleInterface
-from nipype.interfaces.freesurfer.preprocess import ConcatenateLTA
+from nipype.interfaces.freesurfer.preprocess import ConcatenateLTA, RobustRegister
 from nipype.interfaces.freesurfer.utils import LTAConvert
 from niworkflows.interfaces.registration import BBRegisterRPT, MRICoregRPT
 
@@ -193,15 +193,23 @@ class FSDetectInputs(SimpleInterface):
 
 
 class TruncateLTA(object):
+    """Mixin to ensure that LTA files do not store overly long paths,
+    which lead to segmentation faults when read by FreeSurfer tools.
+
+    See the following issues for discussion:
+
+    * https://github.com/freesurfer/freesurfer/pull/180
+    * https://github.com/poldracklab/fmriprep/issues/768
+    * https://github.com/poldracklab/fmriprep/pull/778
+    * https://github.com/poldracklab/fmriprep/issues/1268
+    * https://github.com/poldracklab/fmriprep/pull/1274
+    """
+
     # Use a tuple in case some object produces multiple transforms
     lta_outputs = ('out_lta_file',)
 
     def _post_run_hook(self, runtime):
 
-        outputs = self.aggregate_outputs(runtime)
-        print(outputs)
-        print(type(outputs))
-        # I will modify this as soon as I understand what is the right syntax to use it
         outputs = self._list_outputs()
 
         for lta_name in self.lta_outputs:
@@ -231,23 +239,26 @@ class TruncateLTA(object):
         return runtime
 
 
-class PatchedConcatenateLTA(ConcatenateLTA):
+class PatchedConcatenateLTA(TruncateLTA, ConcatenateLTA):
     """
     A temporarily patched version of ``fs.ConcatenateLTA`` to recover from
     `this bug <https://www.mail-archive.com/freesurfer@nmr.mgh.harvard.edu/msg55520.html>`_
     in FreeSurfer, that was
     `fixed here <https://github.com/freesurfer/freesurfer/pull/180>`__.
+
     The original FMRIPREP's issue is found
     `here <https://github.com/poldracklab/fmriprep/issues/768>`__.
 
-    this is no more useful, since we now change the lta file after their creation by
-    patching LTAconvert directly
-    we can remove it and the reference to it
+    the fix is now done through mixin with TruncateLTA
     """
-    pass
+    lta_outputs = ['out_file']
 
 
-class PatchedLTAconvert(TruncateLTA, LTAConvert):
+class PatchedLTAConvert(TruncateLTA, LTAConvert):
+    """
+    LTAconvert is producing a lta file refer as out_lta
+    truncate filename through mixin TruncateLTA
+    """
     lta_outputs = ('out_lta',)
 
 
@@ -257,6 +268,10 @@ class PatchedBBRegisterRPT(TruncateLTA, BBRegisterRPT):
 
 class PatchedMRICoregRPT(TruncateLTA, MRICoregRPT):
     pass
+
+
+class PatchedRobustRegister(TruncateLTA, RobustRegister):
+    lta_outputs = ('out_reg_file', 'half_source_xfm', 'half_targ_xfm')
 
 
 class RefineBrainMaskInputSpec(BaseInterfaceInputSpec):
--- a/fmriprep/interfaces/freesurfer.py
+++ b/fmriprep/interfaces/freesurfer.py
@@ -249,7 +249,6 @@ class PatchedConcatenateLTA(ConcatenateLTA):
 
 class PatchedLTAconvert(TruncateLTA, LTAConvert):
     lta_outputs = ('out_lta',)
-    pass
 
 
 class PatchedBBRegisterRPT(TruncateLTA, BBRegisterRPT):
--- a/fmriprep/interfaces/freesurfer.py
+++ b/fmriprep/interfaces/freesurfer.py
@@ -35,7 +35,7 @@ from nipype.interfaces.base import (
 from nipype.interfaces import freesurfer as fs
 from nipype.interfaces.base import SimpleInterface
 from nipype.interfaces.freesurfer.preprocess import ConcatenateLTA
-
+from nipype.interfaces.freesurfer.utils import LTAConvert
 from niworkflows.interfaces.registration import BBRegisterRPT, MRICoregRPT
 
 
@@ -198,6 +198,10 @@ class TruncateLTA(object):
 
     def _post_run_hook(self, runtime):
 
+        outputs = self.aggregate_outputs(runtime)
+        print(outputs)
+        print(type(outputs))
+        # I will modify this as soon as I understand what is the right syntax to use it
         outputs = self._list_outputs()
 
         for lta_name in self.lta_outputs:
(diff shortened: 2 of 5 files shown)


.zenodo.json
```diff
@@ -88,6 +88,10 @@
       "affiliation": "Department of Psychology, Stanford University",
       "orcid": "0000-0003-3321-7583"
     }
+    {
+      "name": "Romain, Valabregue",
+      "affiliation": "CENIR, ICM, Paris"
+    }
```
effigies: Please move this above Russ' entry (it and Chris' should be last). And you may want to consider registering for an [ORCID](https://orcid.org/) (or adding yours, if you already have one). It helps link your research contributions across a variety of media.

fmriprep/workflows/bold/registration.py
```diff
@@ -32,6 +33,57 @@
...
+            if line.startswith('filename = ') and len(line.strip("\n")) >= 255:
+                fixed = True
+                newfile.append('filename = path_too_long\n')
+            else:
+                newfile.append(line)
+    
+        if fixed:
+            with open(outputs['out_lta_file'], 'w') as f:
+                f.write(''.join(newfile))
+        return outputs
```
effigies: We keep interfaces and workflows separate.

These should go in [fmriprep/interfaces/freesurfer.py](https://github.com/poldracklab/fmriprep/blob/master/fmriprep/interfaces/freesurfer.py).

fmriprep/workflows/bold/registration.py
```diff
@@ -18,6 +18,7 @@
 from nipype.pipeline import engine as pe
 from nipype.interfaces import utility as niu, fsl, c3, freesurfer as fs
 from niworkflows.interfaces.registration import FLIRTRPT, BBRegisterRPT, MRICoregRPT
+#from niworkflows.interfaces.registration import PatchedMRICoregRPT as MRICoregRPT
```
effigies: This line is not needed.

fmriprep/workflows/bold/registration.py
```diff
@@ -18,6 +18,7 @@
 from nipype.pipeline import engine as pe
 from nipype.interfaces import utility as niu, fsl, c3, freesurfer as fs
 from niworkflows.interfaces.registration import FLIRTRPT, BBRegisterRPT, MRICoregRPT
```
effigies: For `BBRegisterRPT` and `MRICoregRPT`, we should import along with `PatchedConcatenateLTA`, on line 30, below:

```Python
from ...interfaces.freesurfer import (
    PatchedMRICoregRPT as MRICoregRPT,
    PatchedBBRegisterRPT as BBRegisterRPT,
    PatchedConcatenateLTA as ConcatenateLTA)
```

fmriprep/workflows/bold/registration.py
```diff
@@ -32,6 +33,57 @@
 
 DEFAULT_MEMORY_MIN_GB = 0.01
 
+class PatchedBBRegisterRPT(BBRegisterRPT):
+    def _list_outputs(self):
```
effigies: I would prefer to move this to `_post_run_hook`, rather than `_list_outputs`. (I know that `PatchedConcatenateLTA` has it there. We should change that.)

Please see https://github.com/poldracklab/fmriprep/issues/1268#issuecomment-419446891 for the general structure of a `_post_run_hook`.

effigies:
Also, please use [`flake8`](http://flake8.pycqa.org/en/latest/) to identify style issues. These will cause [Travis failures](https://travis-ci.org/poldracklab/fmriprep/builds/425800851).

fmriprep/workflows/bold/registration.py
romainVala: ok, 
is it not more logical then to put TuncateLTRA in interface/freesufer
but PatchedBBRegisterRTP and PatchedMRICoregRPT in the interface/registration ( juste after the definition of BBRegisterRPT nad MRICoregRPT

romainVala: oups I forget that BBRegisterRTP is in niworkflow,
so to stay in fmriprep, I add then as you said
_

romainVala:
Ok 
I made all the suggested changes (I hope)
Sorry for the mess:

note that I change the line 
runtime = super(TruncateLTA, self)._post_run_hook(runtime)
at the end of the function in the TruncateLTA classe 

so that it is executed after the truncation

This is quite unintuitive to realize that "super"  is calling BBRegisterRPT._post_run_hook just because we define

class PatchedBBRegisterRPT(TruncateLTA, BBRegisterRPT):
    lta_outputs = ['out_lta_file']

but ok it seems to works and indeed it is more efficient

romainVala:
since this is  very similar I can alos change the PatchedConcatenateLTA function to use the TruncateLTA. or do you prefer a different pull request ?

fmriprep/interfaces/freesurfer.py
```diff
@@ -190,6 +192,46 @@ def _run_interface(self, runtime):
...
+                    fixed = True
+                    newfile.append('filename = path_too_long\n')
+                else:
+                    newfile.append(line)
+
+            if fixed:
+                with open(outputs['out_lta_file'], 'w') as f:
+                    f.write(''.join(newfile))
+
+        runtime = super(TruncateLTA, self)._post_run_hook(runtime)
```
effigies: Given that we don't need to do this before other things, I would prefer to let other mixins take precedence, so I would place this as the first line of the method.

fmriprep/interfaces/freesurfer.py
```diff
@@ -190,6 +192,46 @@ def _run_interface(self, runtime):
         return runtime
 
 
+class TruncateLTA(object):
+    # Use a tuple in case some object produces multiple transforms
+    lta_outputs = ('out_lta_file',)
+
+    def _post_run_hook(self, runtime):
+        print('ENTERING post_run Truncatae RRR')
```
effigies: Let's remove the debug print statements.

fmriprep/interfaces/freesurfer.py
```diff
@@ -222,6 +264,14 @@ def _list_outputs(self):
         return outputs
 
 
+class PatchedBBRegisterRPT(TruncateLTA, BBRegisterRPT):
+    lta_outputs = ['out_lta_file']
```
effigies: We can just use `pass` here, since it's the default `lta_outputs` from `TruncateLTA`.

fmriprep/interfaces/freesurfer.py
romainVala: I am not used to mixing, so debugin this code, make me realize the following.

Just to be clear I resume the code I refer to :

class PatchedBBRegisterRPT(TruncateLTA, BBRegisterRPT):
   pass

both classs TruncateLTA and BBRegisterRPT have their own definition of _post_run_hook method. So when object PatchedBBRegisterRPT call this function the order of mixins make it use the first one that is the one define in TruncateLTA.

ok that is fine, but it unintuitive is the meaning of the call (define in TruncateLTA)

runtime = super(TruncateLTA, self)._post_run_hook(runtime)

At a first look it is not define since TruncateLTA is heriting from object where not method post_run_hook is define. In fact this method is calling the BBRegisterRPT._post_run_hook
I understand it like if the definition 
class PatchedBBRegisterRPT(TruncateLTA, BBRegisterRPT):
implyt that TruncateLTA herit from BBRegisterRPT

IS that correct ?

Any way I am sure that this "super._post_run_hook call in TruncateLTA is interpreted as BBRegisterRPT._post__run_hook
So if you place it at the begining, this will execute the mri_vol2vol function before we modify the filename path, and I get the segmentation fault again

So it has to be at the end

romainVala:
Ok
I removed the print and I change the PatchedConcatenateLTA to make it reuse the TruncatedLTA

May be I should now add  some doc ?

romainVala:
Sorry to push unfinished work ... but the current solution does not work for PatchedConcatenateLTA
This is because the previous version of it was re-defining the methos _list_outputs, but here we use (via mixin) the _post_run_hook method

so I guess it is then perform after the run, (and I get a segmentation fault )
whereas the _list_outputs method is called before the command run

So I am not sure of the best way to mix this 2 different use case ...

fmriprep/interfaces/freesurfer.py
```diff
@@ -190,7 +192,42 @@ def _run_interface(self, runtime):
...
 
 
-class PatchedConcatenateLTA(ConcatenateLTA):
+class TruncateLTA(object):
+    # Use a tuple in case some object produces multiple transforms
+    lta_outputs = ('out_lta_file',)
+
+    def _post_run_hook(self, runtime):
+
+        outputs = self._list_outputs()
```
effigies: This should be

```Python
outputs = self.aggregate_outputs(runtime)
```

`_list_outputs()` is the predicted outputs and may be called at any time, while `aggregate_outputs()` is the actual outputs, and must be called after `_run_interface()`.

romainVala:
Hi
The segemtation occur first on the freesurfer command mri_vol2vol when the lta file contains the filneame that is too long

This PR is now solving this issue, but I get a second segemtation fault with now the command mri_concatenate_lta (for the same reason)
	 [Node] Running "set_xfm_source" ("fmriprep.interfaces.freesurfer.PatchedConcatenateLTA"), a CommandLine Interface with command:

I first thought that this was due to my change of the PatchedConcatenateLTA but I then come back to the version on the git, and I get the exact same segmentation fault

It seems to me that the PatchedConcatenateLTA does not solve the problem. Is there changes between the commit of the PatchedConcatenateLTA and the current git version that could explain why the _list_outputs method is called after the run command ?

If I manually modify the lta file (to remove the long path) and I re-run fmriprep, the mri_concatenate_lta runs and I can see with a print statement that the _list_ouputs method defined in PatchedConcatenateLTA is call after the run of mri_concatenate_lta

to reprocude the bug you need to run fmriprep with long paht name (>250)

effigies:
I suspect the issue is actually that there's another place where these long paths are written:

https://github.com/poldracklab/fmriprep/blob/d1e2004b1366e783ec3063f29cea01cf2cda7de3/fmriprep/workflows/anatomical.py#L818-L819

This is one of the inputs to `ConcatenateLTA`:

https://github.com/poldracklab/fmriprep/blob/bbbb2a81734c4eb16331bbb08cb5f3b012e7a358/fmriprep/workflows/bold/resampling.py#L146

If it's not too much trouble, could you try patching `LTAConvert` as well?

romainVala:
Well the creation of the lta file is not the problem, the segmentation come when the lta is used with mri_concatenate or mri_vol2vol

the problem I have is when I patch the method that implement (mri_concatenate or mri_vol2vol) the patched (which modify the filename in the lta file) must be execute before the run of the command. 

But indeed I found a working solution where I Patched the method that create the lta file. in this case patching the post_run_hook is fine

I indeed solve the problem by patch 
LTAconvert which is use in workflows/blods/resampling 

If I do so I do not need any more the PatchedConcatenateLTA

But you wright there might be other place where lta file are creating like in the anatomical workflow. but for the anatomical file path are much shorter than the functional, so I did not comes with the same bug.

but I will also change them just to be safe

I clean the code, and do the commit

romainVala:
I made a simple test to reproduce the segmentation fault
just create a working dir in a location where the path is longer thant 255 character. Then I get additional error with tla_convert in the anatomical pipeline. Easy to correct but then I get a different error
mri_vol2vol is now crashing when the -lta command line argument is too long. with a relative path it works and with the absolute path I get segmentation fault. Note the lta has been corrected so that the 
fllename line contained a short string

I do not see a way to correct for this except to perform the command in the lta directories with a relative argument, but I am not sure how difficult it is .

may be we should stay with this partial correction and hopping that freesurfer will solve the issue
I just ask in the freesurfer mailing list

fmriprep/interfaces/freesurfer.py
```diff
@@ -190,36 +192,71 @@ def _run_interface(self, runtime):
...
+    # Use a tuple in case some object produces multiple transforms
+    lta_outputs = ('out_lta_file',)
+
+    def _post_run_hook(self, runtime):
+
+        outputs = self.aggregate_outputs(runtime)
+        print(outputs)
+        print(type(outputs))
+        # I will modify this as soon as I understand what is the right syntax to use it
+        outputs = self._list_outputs()
```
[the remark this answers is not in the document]
romainVala: I still do not understand how to change this one
if I use 
outputs = self.aggregate_outputs(runtime)
then I get an error with the line
    lta_file = outputs[lta_name]
TypeError: 'MRICoregOutputSpecRPT' object is not subscriptable

lta_name can be the string 'out_file' or 'out_lta_file'

if I understand from other example we should call
outputs.out_file or outputs.out_lta_file

but how to use the lta_name ?

romainVala:
Hello
I change d all call to  fs.utils.LTAConvert
I also patch RobustRegister

So now all class that produce lta file go through TruncateLTA

fmriprep/interfaces/freesurfer.py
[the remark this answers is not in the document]
effigies: Okay. This seems to be a bug, because the output of `Interface.aggregate_outputs(...)` should be a `dict`-like object, not an `OutputSpec`. I'll try to track that down.

In the meantime, using `_list_outputs()` should be fine. It's a little dissatisfying to me for architectural reasons (`aggregate_outputs` can take the `runtime` object and thus have access to the working directory if the directory has changed), but I think in practice it will be fine.

Sorry for dragging this out.

PR #1278, opened 2018-09-12
[RF] Split BOLD-T1w registration into calculation/application workflows

emdupre:

As discussed with @effigies, splits the existing `bold_reg_wf` into two separate, smaller workflows, one of which calculates the registrations and the other of which applies the result in a single-shot. This will make integration of ME-EPI workflow changes possible.  

Updated docstrings for both of the new functions.

### DIFF
--- a/fmriprep/workflows/bold/base.py
+++ b/fmriprep/workflows/bold/base.py
@@ -465,18 +465,18 @@ Non-gridded (surface) resamplings were performed using `mri_vol2surf`
             ('subject_id', 'inputnode.subject_id'),
             ('t1_2_fsnative_reverse_transform', 'inputnode.t1_2_fsnative_reverse_transform')]),
         (inputnode, bold_t1_trans_wf, [
+            ('bold_file', 'inputnode.name_source'),
             ('t1_brain', 'inputnode.t1_brain'),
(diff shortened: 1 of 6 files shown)


fmriprep/workflows/bold/base.py
```diff
@@ -211,7 +211,8 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
         * :py:func:`~fmriprep.workflows.bold.stc.init_bold_stc_wf`
         * :py:func:`~fmriprep.workflows.bold.hmc.init_bold_hmc_wf`
         * :py:func:`~fmriprep.workflows.bold.t2s.init_bold_t2s_wf`
-        * :py:func:`~fmriprep.workflows.bold.registration.init_bold_reg_wf`
+        * :py:func:`~fmriprep.workflows.bold.registration.init_bold_apply_reg_wf`
+        * :py:func:`~fmriprep.workflows.bold.registration.init_bold_calc_reg_wf`
```
effigies: For consistency, let's call the apply registration workflow and `init_bold_t1_trans_wf`. We could then go back to `init_bold_reg_wf` or `init_bold_calc_reg_wf`; either one.

PR #1283, opened 2018-09-13
[ENH] Replace FSL ``mcflirt`` with AFNI ``3dVolReg``

chrisfilo:
- Consistent with MRIQC
- Faster than MCFLIRT
- More accurate than MCFLIRT (https://www.ncbi.nlm.nih.gov/pubmed/16099178)


fmriprep/interfaces/images.py
```diff
@@ -342,6 +343,15 @@ class ValidateImage(SimpleInterface):
     output_spec = ValidateImageOutputSpec
 
     def _run_interface(self, runtime):
+
+        def deoblique(img):
+            import os
+            import nibabel as nb
+            import numpy as np
+            affine = img.affine
+            affine[:3, :3] = np.diag(np.diag(affine[:3, :3]))
```
effigies: This will simply zero out the off-diagonal, right? That will end up producing an affine with smaller zooms than the original. Probably the correct way to do this is to do a rotation/zoom/shear/translation decomposition, and then to remove the rotations and shears.

That said, this is less a de-oblique than a removal of information that describes obliqueness. Is this what you want, and what's the reason for that?

fmriprep/interfaces/images.py
```diff
@@ -342,6 +343,15 @@ class ValidateImage(SimpleInterface):
...
+            return nb.Nifti1Image(np.asanyarray(img.dataobj), affine, img.header)
```
effigies: I would prefer `img.__class__` to `nb.Nifti1Image`.

fmriprep/workflows/bold/hmc.py
```diff
@@ -73,24 +74,24 @@ def init_bold_hmc_wf(mem_gb, omp_nthreads, name='bold_hmc_wf'):
         name='outputnode')
 
     # Head motion correction (hmc)
-    mcflirt = pe.Node(fsl.MCFLIRT(save_mats=True, save_plots=True),
-                      name='mcflirt', mem_gb=mem_gb * 3)
+    mc = pe.Node(afni.Volreg(args='-prefix NULL -twopass',
+                             zpad=4, outputtype='NIFTI_GZ'), name="mc", mem_gb=mem_gb * 3)
```
effigies: Do we need to specify `outputtype` if `-prefix NULL`?

fmriprep/interfaces/images.py
chrisfilo: Yes the intention was to remove off diagonal values (because otherwise AFNI doesn't generate transforms that can be applied by ANTs). Is there a python code snippet for rotation/zoom/shear/translation decomposition I could use?

fmriprep/workflows/bold/hmc.py
chrisfilo: No, but we never use the output so there is no need for saving it (even if Nipype will delete it anyway).

chrisfilo:
> Is a 2005 comparison paper still a valid criterion for making such a move?
Probably not alone.

chrisfilo:
Tests are passing, but before I finish the PR by updating the documentation lets decide if we want to do this switch. @oesteban @effigies

effigies:
I'm okay with moving to 3dvolreg by default, primarily on the grounds of decreasing dependence on FSL-licensed code.

However, given that we've gone through the trouble of implementing a MCFLIRT pathway, does it make sense to leave it as an option? Particularly from the perspective of people eventually using fMRIPrep to compare the quality of different algorithms/implementations, it seems worth making this a swappable component.

oesteban:
I'm putting together the last release before RC1. My idea was to upgrade the version minor after that last release (i.e. start 1.2.x).

I think this is a somewhat large change that would fit better within the next 1.2.0, rather than 1.1.8.

WDYT?

chrisfilo:
I agree it's fairly substantial.

requirements.txt
```diff
@@ -1,4 +1,4 @@
-niworkflows>=0.4.3
+git+https://github.com/poldracklab/niworkflows.git@fdf20103b9dedd4632bb8e26a5eb3b896e349fd2#egg=niworkflows
```
oesteban: 0.4.3 has this commit already, please revert this change (I guess the niworkflows dependency has been updated since you sent the PR)

fmriprep/workflows/bold/hmc.py
```diff
@@ -73,24 +73,24 @@ def init_bold_hmc_wf(mem_gb, omp_nthreads, name='bold_hmc_wf'):
...
 
-    fsl2itk = pe.Node(MCFLIRT2ITK(), name='fsl2itk',
-                      mem_gb=0.05, n_procs=omp_nthreads)
+    afni2itk = pe.Node(niu.Function(function=afni2itk_func,
```
oesteban: If there's nothing against it, I'll convert this function node into a VOLREG2ITK fully-fledged nipype interface before merging in, similarly to MCFLIRT2ITK. I can remember that parallelizing that one was tricky but paid off with an important speed up on long datasets.

fmriprep/utils/misc.py
```diff
@@ -7,6 +7,28 @@
...
+    import numpy as np
+
+    T, _, Z, _ = decompose(img.affine)
+    affine = compose(T=T, R=np.diag([1, 1, 1]), Z=Z)
+    return img.__class__(np.asanyarray(img.dataobj), affine, img.header)
+
+
+def split_and_rm_rotshear_func(in_file):
+    import os
+    from nilearn.image import iter_img
```
oesteban: Is there any memory-related reason to use nilearn's `iter_img` which has quite a bit of overhead, instead of [`nibabel.func.four_to_three`](http://nipy.org/nibabel/reference/nibabel.funcs.html#four-to-three)?
```suggestion
    import nibabel as nb
```

fmriprep/utils/misc.py
```diff
@@ -7,6 +7,28 @@
...
+    from fmriprep.utils.misc import remove_rotation_and_shear
+    out_files = []
+    for i, img in enumerate(iter_img(in_file)):
```
oesteban:
```suggestion
    imgs = nb.four_to_three(nb.load(in_file))
    for i, img in enumerate(imgs):
```

oesteban:
First results out of the oven: https://580-66973489-gh.circle-artifacts.com/0/tmp/ds054/derivatives/fmriprep/sub-100185.html

effigies:
@oesteban Any results from those tests?

oesteban:
Not yet. I've been blocked with other things. I should have a decent run on ds005 with current master, so I just need to test out this branch.

oesteban:
I think this is a thumbs up, see below (cc @effigies):

Master:
mcflirt-sub-01_task-mixedgamblestask_run-01_carpetplot

3dVolReg:
volreg-sub-01_task-mixedgamblestask_run-01_carpetplot


Master:
mcflirt-sub-01_task-mixedgamblestask_run-02_carpetplot

3dVolReg:
volreg-sub-01_task-mixedgamblestask_run-02_carpetplot

Master:
mcflirt-sub-01_task-mixedgamblestask_run-03_carpetplot

3dVolReg:
volreg-sub-01_task-mixedgamblestask_run-03_carpetplot

fmriprep/utils/misc.py
```diff
@@ -7,6 +7,29 @@
...
+
+def split_and_rm_rotshear_func(in_file):
+    import os
+    import nibabel as nb
+    from fmriprep.utils.misc import remove_rotation_and_shear
+    out_files = []
+    imgs = nb.four_to_three(nb.load(in_file))
+    for i, img in enumerate(imgs):
+        out_file = os.path.abspath('vol%04d.nii.gz' % i)
+        img = remove_rotation_and_shear(img)
```
effigies: I think we need to do `img = nb.as_closest_canonical(img)` before this. Consider an image with the affine:

```Python
>>> img.affine
array([[  0.        ,   3.        ,   0.        , -88.4322052 ],
       [ -1.66738188,   0.        ,   2.6871357 , -47.71268463],
       [  3.35891914,   0.        ,   1.33390486, -77.85404205],
       [  0.        ,   0.        ,   0.        ,   1.        ]])
>>> img.shape
(33, 64, 64)
```

```Python
>>> remove_rotation_and_shear(img).affine
array([[  3.75000002,   0.        ,   0.        , -88.4322052 ],
       [  0.        ,   3.        ,   0.        , -47.71268463],
       [  0.        ,   0.        ,   3.00000007, -77.85404205],
       [  0.        ,   0.        ,   0.        ,   1.        ]])

>>> remove_rotation_and_shear(nb.as_closest_canonical(img)).affine
array([[  3.        ,   0.        ,   0.        , -88.4322052 ],
       [  0.        ,   3.00000007,   0.        , -47.71268463],
       [  0.        ,   0.        ,   3.75000002, -77.85404205],
       [  0.        ,   0.        ,   0.        ,   1.        ]])
```

Notice that both are RAS matrices, but with different zooms. And if you look at the shapes:

```Python
>>> remove_rotation_and_shear(img).shape
(33, 64, 64)

>>> remove_rotation_and_shear(nb.as_closest_canonical(img)).shape
(64, 64, 33)
```

We clearly have a rotated data matrix, which cannot be true if both correctly have an RAS orientation. So we end up destroying all orientation information, not just the "off-diagonal" information, unless we rotate the data matrix first.

oesteban: Good catch

effigies:
Also, there are docs to update: https://github.com/poldracklab/fmriprep/search?q=mcflirt&unscoped_q=mcflirt

oesteban:
Good job folks

oesteban:
Sorry I was too fast:

https://5616-53175327-gh.circle-artifacts.com/0/tmp/data/reports/ds001362/sub-01_task-taskname_run-01_bold_masks.svg

With the new bold references, it seems data are changed the orientation. I'm unsure whether we just need to update the masks or this regression reflects a deeper problem. But it seems clear that it was introduced here: https://circleci.com/gh/chrisfilo/fmriprep/616#artifacts/containers/0

Sorry about that :(

oesteban:
Yup

PR #1284, opened 2018-09-14
[MAINT] Pin grabbit and pybids in ``setup.py``

### LINKED ISSUE #1279: Pin grabbits and pybids to specific versions in Dockerfile
These two packages evolve quickly and often break backwards compatibility. We should pin them to be in control when our side needs updating when pybids API changes.

oesteban:
Continue #1281 and close #1279


effigies:
I thought we weren't worrying about this?

oesteban:
I'm putting together the last release before Derivatives-RC1. I think this should be in (basically, any dependency pinning should happen in the setup.py file, and we can propagate it to the requirements.txt when we want to save docker build time).

PR #1288, opened 2018-09-18
DOC: Note GIFTI surface alignment

effigies:
Following discussion in [this NeuroStars thread](https://neurostars.org/t/fmriprep-correction-of-surface-extracted-from-freesurfer-anat-only/2428/13), a clarifying note seems necessary.

Updated documentation.

* Add a note with the following text:

```RST
GIFTI surface outputs are aligned to the FreeSurfer T1.mgz image, which
may differ from the T1w space in some cases, to maintain compatibility
with the FreeSurfer directory.
Any measures sampled to the surface take into account any difference in
these images.
```


effigies:
[Rendered](https://1939-78558509-gh.circle-artifacts.com/0/home/circleci/out/docs/workflows.html#surface-preprocessing).

screen shot 2018-09-18 at 13 53 27

Issue #1135, opened 2018-05-17
Multiecho: plot one carpet per echo - show all ``preproc`` echoes


oesteban:

emdupre:
I think this is addressed with #1296, especially #1355 !

PR #1296, opened 2018-09-25
Omnibus multi-echo pull request

### LINKED ISSUE #1034: Multiecho: potential improvements to the T2* map calculation.
In https://github.com/poldracklab/fmriprep/pull/1021#issuecomment-375086376, I suggested the following improvements:
  * First remove any value below zero from all echoes.
  * Enable passing in a mask to the interface, so the T2* is calculated only within mask.
  * Crop out outliers: some very bright pixels force a compression of the histogram. Identify those and set them to a saturation value (a high percentile). Make sure that the output T2* fits in a int16 image.

effigies:
This PR tracks a series of updates to multi-echo preprocessing, an effort being led by @emdupre. The work has been broken into multiple PRs to ensure that each is independently reviewable.

New multi-echo related PRs should target the `multiecho` branch of the main repository, rather than `master`.

Any updates to `master` should be merged into `multiecho` as soon as possible to avoid getting out of sync.

Refactors that substantially affect the BOLD workflows should consider targeting the `multiecho` branch unless updates to `master` are needed before `multiecho` is likely to be merged.

Closes #1034, #1135 

* [x] #1263 (signed off by @effigies)
* [x] #1355 (signed off by @effigies)
* [x] #1383 (signed off by @oesteban)


effigies:
@emdupre I've given you write permissions. Please feel free to update the head post as new PRs are proposed.

emdupre:
I'm a little confused how merging in a pull request to merge in master resulted in new merge conflicts ! Do you want me to try and fix the conflict here, @effigies ?

effigies:
It was fine, but there've since been changes to master. I think if we wait for #1354, the conflicts will go away again.

emdupre:
@effigies Looking back at previous conversations from https://github.com/poldracklab/fmriprep/pull/1324#issuecomment-433405093, how would I go about modifying `bbregister` parameters ? Did you want me to loop in someone, or should I play with them myself ? Or should we instead follow previous suggestions from @oesteban (see: https://github.com/poldracklab/fmriprep/issues/1034) and modify the T2* map itself ?

emdupre:
Yes, happy to. It looks like merging in master broke something here, though, so maybe I'll try to fix that first !

effigies:
Ah, thanks for the heads up. I also apparently failed to read you saying you'd fix it, so I think I fixed it.

emdupre:
I had pushed this off until I had a little more time to sit down with it, but the solution was actually easier than I originally anticipated. 

Essentially, when calculating the T2* map we have the problem of what to do in voxels where only one echo has reliable signal. For the standard T2* map we simply exclude those voxels, as their values can be unpredictable. For the adaptive T2* map-- which I had originally set as the coregistration target in the case of `-t2s-coreg` we retain them and treat them at face value.

When I switched to the more stable, "standard" T2* map, the 10% cases vanished. The poorly behaving example I had shown previously (among others) now looks as expected, see:

screenshot_2018-11-11 screenshot

I've opened #1383 to address this change and to make sure that the reports and all other relevant documentation is up-to-date. But I'm feeling optimistic that this is quite close !

effigies:
Sounds good to me.

PR #1299, opened 2018-09-26
[FIX] Enhance T2 contrast ``enhance_t2`` in reference estimate

effigies:

Pass `enhance_t2` flag from `init_bold_reference_wf` to `init_enhance_and_skullstrip_wf`.

This parameter currently does nothing. It appears to have been an oversight in #966.

None.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -252,10 +252,12 @@ jobs:
           name: Run fMRIPrep tests
           no_output_timeout: 2h
           command: |
+            mkdir /tmp/data/reports && \
             docker run -ti --rm=false \
               -v /tmp/data:/tmp/data \
               -e FMRIPREP_REGRESSION_SOURCE=/tmp/data/fmriprep_bold_truncated \
               -e FMRIPREP_REGRESSION_TARGETS=/tmp/data/fmriprep_bold_mask \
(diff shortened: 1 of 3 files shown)


effigies:
The masks look very trivially different in the regression tests: https://circleci.com/gh/poldracklab/fmriprep/5048#artifacts/containers/0

ZhifangYe:
Sorry for the long radio silence. I’ve been quite busy the last few months.

I believe this change wouldn't fix the initial mask issue. I had applied the `_enhance_t2_contrast` function manually on my data and it yielded similar bad brain mask.

ZhifangYe:
I think maybe we need a new perspective on #1050 since the current implementation gives inferior results on some legacy data.

Also, I found the N4 correction step fails on some of the distortion corrected BOLD image during processing my dataset. As a result, even my PR gives good brain mask for the original reference image, it fails to generate good final mask as output. I had to replace the entire `init_enhance_and_skullstrip_wf` workflow to a single `Bet` step when processing my dataset.

I'm not sure whether the problem come from `N4BiasFieldCorrection` or my specific data since I didn't find any useful information in ANTs discussion forum. @effigies I'm happy to provide an example data to further investigate this problem and spend more time on #1050.

effigies:
Roping @oesteban into this conversation, since he's back from vacation and has spent a lot more time thinking about the problem than I have.

PR #1305, opened 2018-10-03
[FIX] Create template from one usable T1w image

effigies:
#925 switched from `StructuralReference` to `fs.RobustTemplate`, under the reasoning that a single T1w image could be caught at workflow construction time. However, the conformation process can remove images that would need to be upsampled too much, leaving only a single T1w image to produce a template from. In this case, we need to simply pass through the original image.

* Restore `t1_merge` to `StructuralReference`.

None.


effigies:
Tested this locally. It works. :-)

PR #1306, opened 2018-10-04
[ENH] Store BOLD reference images

oesteban:
- Stores BOLD reference images in the output folder (following new RC1 conventions).
- Updates ``DerivativesDataSink`` not to insert the ``_<bids_type>_`` in the middle of every derivative name.


effigies:
Just want to say: I think the renaming process should be all-or-nothing. Once it starts, we should finish before releasing. Do we want to make a release before starting to merge PRs like this?

oesteban:
> Do we want to make a release before starting to merge PRs like this?

Yesss

.circleci/ds005_outputs.txt
```diff
@@ -6,60 +6,64 @@ fmriprep/logs/CITATION.md
...
-fmriprep/sub-01/func/sub-01_task-mixedgamblestask_run-02_bold_space-T1w_preproc.nii.gz
+fmriprep/sub-01/func/sub-01_task-mixedgamblestask_run-01_AROMAnoiseICs.csv
+fmriprep/sub-01/func/sub-01_task-mixedgamblestask_run-01_confounds.tsv
+fmriprep/sub-01/func/sub-01_task-mixedgamblestask_run-01_MELODICmix.tsv
+fmriprep/sub-01/func/sub-01_task-mixedgamblestask_run-01_space-cifti_variant-space1_preproc.dtseries.json
+fmriprep/sub-01/func/sub-01_task-mixedgamblestask_run-01_space-cifti_variant-space1_preproc.dtseries.nii
+fmriprep/sub-01/func/sub-01_task-mixedgamblestask_run-01_space-fsaverage5.L.func.gii
+fmriprep/sub-01/func/sub-01_task-mixedgamblestask_run-01_space-fsaverage5.R.func.gii
+fmriprep/sub-01/func/sub-01_task-mixedgamblestask_run-01_space-MNI152NLin2009cAsym_brainmask.nii.gz
+fmriprep/sub-01/func/sub-01_task-mixedgamblestask_run-01_space-MNI152NLin2009cAsym_desc-3dref_bold.nii.gz
```
effigies: I don't think this qualifies as a `bold.nii.gz`, as it's not a 4D time series. We didn't try to get a `ref` suffix (or perhaps `boldref`, since there is no longer a `bold` entry) into RC1, but we can still use it as long as there isn't a more appropriate suffix in the spec.

Pinging BIDS-meister @chrisfilo for a second opinion, though.

chrisfilo:
You're right - I would use a new suffix and be ready to change it some time
in the future when this is standardized.

oesteban:
thumbs up?

effigies:
Sure. Let's move pretty quickly on the rest of the renamings, then. I listed the anat ones. I'll check for any functional ones that aren't covered by that in the morning.

PR #1324, opened 2018-10-16
MAINT: Merge master into multiecho

effigies:
Master is conflicting with multiecho.


emdupre:
Oof. Ok, this is going to get a little crazy to keep waiting on.

I wanted to write out where I am on this right now, and what I think are reasonable next steps. It'd be great to get your feedback on this !

**Background**: Right now, there's an active discussion on tedana about re-working our roadmap (https://github.com/ME-ICA/tedana/issues/131). This discussion will not impact any of the steps already incorporated into FMRIPREP (i.e., generating the T2* map and optimal combination) but it will impact those that are yet to be incorporated, such as the denoising selection criteria.

**Suggestion**: Personally, I am therefore leaning towards having a stable release of FMRIPREP that is not compatible with tedana denoising. I know that this is likely to be a less popular opinion, but I'd prefer to have tried and true methods in FMRIPREP, rather than something I know is likely to change substantially in a few weeks.

**Potential concerns**: For anyone interested in multi-echo denoising, for now I would point them to afni_proc.py or to original ME-ICA code. This also aligns with the range of denoising methods currently implemented, so users could chose that method with which they are most comfortable.

I do think we should integrate tedana denoising later, but I'd prefer this to be after it's a little more stable (once we get to a minor version release). And, I don't want to prevent merging the stable parts that we have now !

**WDYT ?** This is definitely a bigger picture question, so I'd appreciate any feedback !

If this all sounds ok, we can sketch out what PRs need to be added to this branch to have a stable multi-echo workflow (e.g., the boilerplate).

emdupre:
Do you want me to resolve these conflicts, @effigies @oesteban ? Let me know !

effigies:
Hi @emdupre, sorry for the delay. We're going to try to finish getting derivatives into RC1-compatible shape, and then have another look at this PR. I think the plan was for us to try to resolve these conflicts, checking in with you to keep from breaking your changes. If you'd rather take the lead on that, we can let you know when we're done with RC1-compatibility (Hopefully today or Monday...). Whichever you'd prefer is fine with us.

Regarding your earlier post, I'm comfortable with your logic, and so unless @oesteban has objections, I would suggest perhaps a good division of labor would be for us to get this branch back to a working state, and for you to propose a roadmap for what additional PRs are needed before merging this branch into master.

emdupre:
No worries on the delay ! I know there's a sprint right now, so I wasn't expecting any movement on this until next week-- I had just realized I wasn't sure if you were waiting on me !

Yes, I'm happy to propose a roadmap if @oesteban is ok with the above suggestions. Just let me know what's easiest from your end !

oesteban:
Yes, totally agree on your suggestions :)

emdupre:
Sorry for the lag on this ! Here's a sketch of where I think we are right now, and where I'd like this to go:

### Background

With #1263, we addressed a couple of points:
* Previously, we were duplicating `tedana` code for the creation of a T2* map; that is now  imported from `tedana` directly and wrapped in a `nipype` style interface.
* All echos were previously iterated over individually, with two exceptions. (1) at head motion correction, where only the first echo was used, and (2) at coregistration, where the T2* map was created jointly and used to coregister all echos. Now, all echos are only iterated in one sub-workflow. The T2* map *and* the optimally combined time series (a weighted average of the echos) are created, and the optimal combination is treated as the primary BOLD file and returned to the user.
* We made initial modifications to the documentation to discuss this update. 

There are a few things that I think still need to be done to finish this line of work, with a main aim being that _multi-echo series will return the optimal combination by default, in the future with optional TE-informed denoising._
 
### Roadmap

* In completing #1263, we noticed that there are a number of cases (I'd estimate 10%, from running a few subjects locally), where the generated T2* map shows significant distortions *only in surface space*. See below for an example:

screenshot_2018-10-26 screenshot

This needs to be addressed before merging. I know @effigies and I spent some time looking into this a few weeks back, but I feel like we decided against it being q- or s-form codes. Suggestions are greatly appreciated !

* The documentation and the boilerplate both need to be fully updated to reflect these changes. Should we  add in a multi-echo citation, and the `tedana` citation ? What is the recommendation, here ?

* I'd like to add an additional flag, like `--task-id`, for `--echo-number`. This would allow users to treat a multi-echo series as a single-echo acquisition by only selecting for processing those echos which they want back. A common use case would be selecting only the echo closest to 30ms for processing to approximate results from a single-echo acquisition. 

Happy to add or edit points as you think is appropriate ! Thanks so much as always for you support :sparkles:

emdupre:
From a read-through the added merge seems fine ! We'll see if any conflicts creep in ...
I'm at a conference so I'll be in and out for a few days. But, I'm looking forward to seeing this done ! @oesteban if you or @chrisfilo have feedback on the [next steps I outlined](https://github.com/poldracklab/fmriprep/pull/1324#issuecomment-433360946) please let me know !

PR #1328, opened 2018-10-17
[HOTFIX] Reportlets - bbregister vs flirtbbr (continues #1326)

oesteban:


oesteban:
Merging in because circle has gone bananas and the PR actually fixes the issue, finally.

PR #1330, opened 2018-10-18
[FIX] Quick update to new template structure

oesteban:
This should fix the test failed in #1283


oesteban:
`test_pytest` failed for reasons internal to CircleCI. Since this PR does not affect that build job, I'll go ahead and merge regardless that scary red cross.

Issue #1300, opened 2018-10-01
Crash on AROMA if all ICs are classified as motion


andersonwinkler:
The title says it all. For every functional run, out of hundreds, in which AROMA incidentally classified all ICs as being motion-related, FMRIPREP crashes.

It doesn't seem to be an issue with AROMA as fsl_regfilt works fine even in these cases. But somehow the workflow doesn't finish.

When I try to open the pklz crash reports with nipypecli, then nipypecli itself crashes too...

andersonwinkler:
The error message, btw, is this:

```
$ nipypecli crash result_ica_aroma.pklz
Traceback (most recent call last):
  File "/gpfs/gsfs6/users/EDB/Networks/Flanker2/code/env-hpc/bin/nipypecli", line 11, in <module>
    sys.exit(cli())
  File "/gpfs/gsfs6/users/EDB/Networks/Flanker2/code/env-hpc/lib/python3.6/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/gpfs/gsfs6/users/EDB/Networks/Flanker2/code/env-hpc/lib/python3.6/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/gpfs/gsfs6/users/EDB/Networks/Flanker2/code/env-hpc/lib/python3.6/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/gpfs/gsfs6/users/EDB/Networks/Flanker2/code/env-hpc/lib/python3.6/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/gpfs/gsfs6/users/EDB/Networks/Flanker2/code/env-hpc/lib/python3.6/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/gpfs/gsfs6/users/EDB/Networks/Flanker2/code/env-hpc/lib/python3.6/site-packages/nipype/scripts/cli.py", line 94, in crash
    display_crash_file(crashfile, rerun, debug, dir)
  File "/gpfs/gsfs6/users/EDB/Networks/Flanker2/code/env-hpc/lib/python3.6/site-packages/nipype/scripts/crash_files.py", line 53, in display_crash_file
    if 'node' in crash_data:
TypeError: argument of type 'InterfaceResult' is not iterable
```

[...]

Issue #1136, opened 2018-05-17
No clear way to ignore known dummy frames for ICA-AROMA, etc...


alexlicohen:
We are processing the GSP 1000 dataset, which does have dummy frames in the fMRI data, but we want to use the ICA-AROMA output. It would seem more logical to remove these frames up front before MELODIC is run. Informal comparison finds distinct differences in fcMaps between leaving them in vs truncating the file before fmriprep is run. What should be the appropriate sequence of steps here?

oesteban:
It seems to me like a very reasonable feature that we can add and set on by default. WDYT @chrisfilo @jdkent ?

alexlicohen:
My concern would be relying on the non-steady-state detector working
equally well for all subjects. I think having a truncation based on that as
a default would be good, but would also like a switch to explicitly (and
always) drop the first 4, etc...



----------------------------------------------------------------
Alexander Li Cohen, M.D., Ph.D.
E-mail: Alexander.Cohen2@childrens.harvard.edu (Medical/Science Email)
E-mail: alexcohen@gmail.com (Lifetime Email)
----------------------------------------------------------------

effigies:
That seems reasonable. My main concern in asking is whether we can avoid adding more options, when we don't need need to. Come to think of it, I wonder if there's anything in BIDS indicating the first non-dummy scan. It seems like the place to encode information like that is in the dataset, not in the command line arguments.

jdkent:
Interesting problem!

My intuition was that ICA-AROMA would generate noise component(s) that were associated with the dummy frames and wouldn't (significantly) impact the other signal components, but this wouldn't be the first (nor the last) time my intuition was incorrect. 

@alexlicohen When you compared the truncated versus non-truncated ICA-AROMA methods, you truncated the non-truncated data _after_ ICA-AROMA, correct? I assume so, but I want to be explicit. I think I have some data that can play with truncating a different number of volumes and see if the components change.

As for the general process of truncation (outside the context of MELODIC/ICA-AROMA), I understand that this can be accomplished with adding columns to the design matrix with a "1" in the row that represents the volume you wish to ignore and "0"s for every other row. Given that is correct, you can effectively remove that volume without manipulating the data.

My concern for changing the number of volumes that are output by fmriprep is that it may change the interpretation of the `events.tsv` file if the fMRI data was constructed for a task. I like @effigies suggestion to incorporate the # of dummy volumes in the dataset itself

[...]

PR #1335, opened 2018-10-20
[FIX] Remove non-steady-state volumes prior to ICA-AROMA

jdkent:
ref #1136
ref #1300

This pull request pulls the steady state detector into `bold/base.py` and uses it's output for both the confounds workflow and the aroma workflow.

I need to update the documentation for ica-aroma, more details to come...

### DIFF
--- a/fmriprep/workflows/bold/base.py
+++ b/fmriprep/workflows/bold/base.py
@@ -18,6 +18,7 @@ from nipype import logging
 from nipype.interfaces.fsl import Split as FSLSplit
 from nipype.pipeline import engine as pe
 from nipype.interfaces import utility as niu
+from nipype.algorithms import confounds as nac
 
 from ...interfaces import (
     DerivativesDataSink,
@@ -424,6 +425,9 @@ Non-gridded (surface) resamplings were performed using `mri_vol2surf`
                                              omp_nthreads=omp_nthreads,
                                              use_compression=False)
 
+    # get non steady-state volumes
+    non_steady_state = pe.Node(nac.NonSteadyStateDetector(), name='non_steady_state')
+
     # get confounds
     bold_confounds_wf = init_bold_confs_wf(
         mem_gb=mem_gb['largemem'],
@@ -525,6 +529,9 @@ Non-gridded (surface) resamplings were performed using `mri_vol2surf`
             ('outputnode.out_warp', 'inputnode.fieldwarp'),
             ('outputnode.bold_mask', 'inputnode.bold_mask')]),
         (bold_sdc_wf, summary, [('outputnode.method', 'distortion_correction')]),
+        # Calculate nonsteady state
+        (bold_bold_trans_wf, non_steady_state, [
+            ('outputnode.bold', 'in_file')]),
         # Connect bold_confounds_wf
         (inputnode, bold_confounds_wf, [('t1_tpms', 'inputnode.t1_tpms'),
                                         ('t1_mask', 'inputnode.t1_mask')]),
@@ -532,6 +539,8 @@ Non-gridded (surface) resamplings were performed using `mri_vol2surf`
             ('outputnode.movpar_file', 'inputnode.movpar_file')]),
         (bold_reg_wf, bold_confounds_wf, [
             ('outputnode.itk_t1_to_bold', 'inputnode.t1_bold_xform')]),
+        (non_steady_state, bold_confounds_wf, [
+            ('n_volumes_to_discard', 'inputnode.n_volumes_to_discard')]),
         (bold_confounds_wf, outputnode, [
             ('outputnode.confounds_file', 'confounds'),
         ]),
@@ -730,6 +739,8 @@ Non-gridded (surface) resamplings were performed using `mri_vol2surf`
                     ('outputnode.bold_mask', 'inputnode.bold_mask')]),
                 (bold_sdc_wf, ica_aroma_wf, [
                     ('outputnode.out_warp', 'inputnode.fieldwarp')]),
+                (non_steady_state, ica_aroma_wf, [
+                    ('n_volumes_to_discard', 'inputnode.n_volumes_to_discard')]),
                 (bold_confounds_wf, join, [
                     ('outputnode.confounds_file', 'in_file')]),
                 (ica_aroma_wf, join,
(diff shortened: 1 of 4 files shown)


fmriprep/workflows/bold/base.py
```diff
@@ -424,6 +425,9 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
                                              omp_nthreads=omp_nthreads,
                                              use_compression=False)
 
+    # get non steady-state volumes
+    non_steady_state = pe.Node(nac.NonSteadyStateDetector(), name='non_steady_state')
```
effigies: You should already be able to get this from `bold_reference_wf.outputnode.skip_vols`.

jdkent: nice, thanks!

fmriprep/workflows/bold/confounds.py
```diff
@@ -586,18 +612,67 @@ def _getbtthresh(medianval):
...
+
+        # modify header with new shape (fewer volumes)
+        data_shape = list(bold_img.header.get_data_shape())
+        data_shape[-1] -= skip_vols
+        bold_img.header.set_data_shape(tuple(data_shape))
+
+        # save the resulting bold file
+        out = fname_presuffix(bold_file, suffix='_cut')
+        bold_img.__class__(bold_data_cut, bold_img.affine, bold_img.header).to_filename(out)
+        return out
```
effigies: This could be much shorter:

```Python
import nibabel as nb
from nipype.utils.filemanip import fname_presuffix

out = fname_presuffix(bold_file, suffix='_cut')
bold_img = nb.load(bold_file)
bold_img.__class__(bold_img.dataobj[..., skip_vols:],
                   bold_img.affine, bold_img.header).to_filename(out)
return out
```

Note that the data shape will be set automatically, so there's no need to manually update it. And using the `dataobj` should save as much memory as you can, though I don't think there's much waste here, assuming `skip_vols` is small.

Or, with nibabel>=2.3:

```Python
import nibabel as nb
from nipype.utils.filemanip import fname_presuffix

out = fname_presuffix(bold_file, suffix='_cut')
nb.load(bold_file).slicer[..., skip_vols:].to_filename(out)
return out
```

jdkent:
If I wanted to write tests for the semi-private functions, where should I put them? Should I test them?

fmriprep/workflows/bold/confounds.py
```diff
@@ -586,18 +612,67 @@ def _getbtthresh(medianval):
...
+    from nipype.utils.filemanip import fname_presuffix
+
+    # load the data
+    bold_img = nb.load(bold_file)
+    bold_data = bold_img.get_data()
+    bold_cut_img = nb.load(bold_cut_file)
+    bold_cut_data = bold_cut_img.get_data()
+
+    # assign everything from skip_vols foward to bold_cut_data
+    bold_data[..., skip_vols:] = bold_cut_data
```
effigies: What about:

```Python
bold_img = nb.load(bold_file)
bold_cut_img = nb.load(bold_cut_file)

bold_data = np.concatenate((bold_img.dataobj[..., :skip_vols],
                            bold_cut_img.dataobj), axis=3)
```

This would reduce the memory usage by the uncompressed size of `bold_cut_img`.

Or, if we're willing to require `nibabel>=2.3`:

```Python
new_img = nb.concat_images((bold_img.slicer[..., :skip_vols], bold_cut_img), axis=3)
```

This last has the advantage of checking affines to ensure the images are in the same space.

fmriprep/workflows/bold/confounds.py
```diff
@@ -586,18 +612,67 @@ def _getbtthresh(medianval):
...
     ])
 
     return workflow
 
 
+def _remove_volumes(bold_file, skip_vols):
+        import nibabel as nb
+        from nipype.utils.filemanip import fname_presuffix
+
+        # load the bold file and get the 4d matrix
```
effigies: Before doing anything else, I would check:

```Python
if skip_vols == 0:
    return bold_file
```

fmriprep/workflows/bold/confounds.py
```diff
@@ -586,18 +612,67 @@ def _getbtthresh(medianval):
...
+        bold_img.__class__(bold_data_cut, bold_img.affine, bold_img.header).to_filename(out)
+        return out
+
+
+def _add_volumes(bold_file, bold_cut_file, skip_vols):
+    """prepend skip_vols from bold_file onto bold_cut_file"""
+    import nibabel as nb
+    from nipype.utils.filemanip import fname_presuffix
+
+    # load the data
```
effigies:
```Python
if skip_vols == 0:
    return bold_cut_file
```

effigies:
@jdkent We're trying to update to be largely conformant with BIDS Derivatives RC2 (due out Friday-ish). That will trigger fMRIPrep 1.2.0. Do you want to get this in by then, or do you need more time?

jdkent:
Hi @effigies, Thanks for the helpful review! I've integrated the changes you suggested in the last commit, but I have not written tests yet. I think I will be able generate tests by friday, but if not, I can make it a separate pull request, and have this one merged as is (if all the tests pass).

fmriprep/workflows/bold/confounds.py
```diff
@@ -428,40 +427,53 @@ def init_ica_aroma_wf(template, metadata, mem_gb, omp_nthreads,
...
         susan_fwhm : float
             Kernel width (FWHM in mm) for the smoothing step with
             FSL ``susan`` (default: 6.0mm)
         use_fieldwarp : bool
             Include SDC warp in single-shot transform from BOLD to MNI
-        ignore_aroma_err : bool
-            Do not fail on ICA-AROMA errors
-        aroma_melodic_dim: int or None
-            Set the dimensionality of the Melodic ICA decomposition
-            If None, MELODIC automatically estimates dimensionality.
```
effigies: It looks like you sorted these. I would keep these in the order that they are in the parameter list. If we want to reorder that, we can do that in another PR, but that's an API change, and we should avoid it unless there's a good reason.

effigies:
@jdkent I'm going to go ahead and merge. Please submit tests in a new PR.

PR #1336, opened 2018-10-21
[ENH]: ignore project settings files from popular python/code editors

jdkent:
Ignore project setting files from
- spyder
- pycharm
- vscode

If we ignore the settings files from these somewhat popular python/code editors, then contributors will not have to manually ignore them (I use vscode). 

Any more popular choices to add to the list?


effigies:
Can update as we encounter more. Thanks.

PR #1343, opened 2018-10-24
ENH: Conform confound regressor names to Derivatives RC2

effigies:

Updating confounds headers to match the RC2.

Known remaining tasks:
* [x] Update the `cosine_*` headers in nipype or update in a separate function in nipype.
* [x] Update DVARS and FD headers (might need to be in nipype)

Anywhere we discuss confounds. I believe we have example confounds that should be fixed, as well.

### DIFF
--- a/.circleci/ds005_outputs.txt
+++ b/.circleci/ds005_outputs.txt
@@ -6,6 +6,8 @@ fmriprep/logs/CITATION.md
 fmriprep/logs/CITATION.tex
 fmriprep/sub-01
 fmriprep/sub-01/anat
+fmriprep/sub-01/anat/sub-01_desc-aparcaseg_dseg.nii.gz
+fmriprep/sub-01/anat/sub-01_desc-aseg_dseg.nii.gz
 fmriprep/sub-01/anat/sub-01_desc-brain_mask.nii.gz
 fmriprep/sub-01/anat/sub-01_desc-preproc_T1w.nii.gz
 fmriprep/sub-01/anat/sub-01_dseg.nii.gz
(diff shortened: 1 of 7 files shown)


effigies:
For simplicity, I put a `camel_to_snake` function in the confound gathering interface. We can update things in Nipype to be more BIDS-compliant if we want, but that can be a separate thing.

At this point, is there anything besides metadata left to update?

fmriprep/interfaces/confounds.py
```diff
@@ -245,7 +255,7 @@ def _get_ica_confounds(ica_out_dir, skip_vols, newpath=None):
     # add one to motion_ic_indices to match melodic report.
     aroma_confounds = os.path.join(newpath, "AROMAAggrCompAROMAConfounds.tsv")
     pd.DataFrame(aggr_confounds.T,
-                 columns=['AROMAAggrComp%02d' % (x + 1) for x in motion_ic_indices]).to_csv(
+                 columns=['aroma_motion_%02d' % (x + 1) for x in motion_ic_indices]).to_csv(
```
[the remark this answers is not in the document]
jdkent: I think that sounds good :+1:, that's how I read it too.

effigies:
@oesteban I'm going to rebase this. I'm not sure what fixes you were going for, but these last merges look messed up.

oesteban:
Yes, I did a bad merge. Please go ahead :+1:

effigies:
Assuming this passes, I think we should be good to go.

effigies:
@oesteban I'm seeing desc-aseg and label-aseg. I don't think we output label- anymore. Is this a thing where clearing the anatomical cache is required, or am I missing something?

oesteban:
Looking into this right now

.circleci/ds005_outputs.txt
```diff
@@ -6,6 +6,8 @@ fmriprep/logs/CITATION.md
 fmriprep/logs/CITATION.tex
 fmriprep/sub-01
 fmriprep/sub-01/anat
+fmriprep/sub-01/anat/sub-01_desc-aparcaseg_dseg.nii.gz
+fmriprep/sub-01/anat/sub-01_desc-aseg_dseg.nii.gz
 fmriprep/sub-01/anat/sub-01_desc-brain_mask.nii.gz
 fmriprep/sub-01/anat/sub-01_desc-preproc_T1w.nii.gz
 fmriprep/sub-01/anat/sub-01_dseg.nii.gz
```
effigies: Thought for another PR: Since we do sometimes have other dseg files, do we want to specify `desc-fast` or similar?

oesteban: That would be a good idea. I'd rather go with `desc-3tissues` and leave the provenance for the json file.

PR #1347, opened 2018-10-25
ENH: Deprecate --debug with --sloppy

effigies:

It's time to kill `--debug`. This makes it do nothing, and replaces it with `--sloppy`, with a better docstring.

Do we want to adjust anything else? A single flag to run `stop_on_first_crash` and set `--nthreads` to 1? I think what would be really nice is something like `--create-issue` to construct a pre-filled-in issue with run parameters and tracebacks, but that's almost certainly a job for another day.

* [ ] Make sure that the usage docs are updated.


effigies:
Docs look good: https://2122-78558509-gh.circle-artifacts.com/0/home/circleci/out/docs/usage.html

Merging.

PR #1350, opened 2018-10-25
[ENH] Respect SliceEncodingDirection metadata

### LINKED ISSUE #1345: STC ignoring SliceEncodingDirection
Hi.

It seems that stc is not taking SliceEncodingDirection into account. 

I was under the impression that this parameter should alter how SliceTiming is handled.
If my nifti orientation is LPS and the slices where acquired in z direction in descending order (from superior to inferior), SliceEncodingDirection should be 'k-' and SliceTiming something like [0, .05, .1 (...)] correct?
Thanks.

fliem:
closes #1345 
adds slice encoding direction to stc

depends on [nipype PR#2753](https://github.com/nipy/nipype/pull/2753)

Not sure...


fliem:
@effigies thanks for merging the nipype PR. I can't figure out how to get the newest nipype version into fmriprep. Via niworkflows?

effigies:
Once the new nipype is released, we'll incorporate the fix here. We're waiting on one fix in nipype before release (scheduled for Monday anyway).

effigies:
Although, you can pin to the latest nipype commit in requirements.txt, for testing purposes:

```
git+https://github.com/nipy/nipype.git@d83b06d7f8ddb76d4e1033c149eecc966eddb514#egg=nipype
```

effigies:
If you update the nipype dependency to `>=1.1.4`, I think this should be nearly ready.

https://github.com/poldracklab/fmriprep/blob/92786d1456c4098197f6a61e46fde7275f891777/fmriprep/__about__.py#L87-L108

fmriprep/workflows/bold/stc.py
```diff
@@ -71,6 +71,8 @@ def init_bold_stc_wf(metadata, name='bold_stc_wf'):
                     tr='{}s'.format(metadata["RepetitionTime"]),
                     slice_timing=metadata['SliceTiming']),
         name='slice_timing_correction')
+    if 'SliceEncodingDirection' in metadata:
+        slice_timing_correction.inputs.slice_encoding_direction = metadata['SliceEncodingDirection']
```
effigies: Maybe we could be as clear (and not hit line length limits) by just adding `slice_encoding_direction=metadata.get('SliceEncodingDirection', 'k')` to the `TShift()` call.

fliem: OK. I have tested the previous version and this looks good
* same list order, different direction gives different results (as expected)
* flipped list, flipped direction gives same results (as expected)
* no direction vs "k" gives same results (as expected)

effigies:
Thanks!

PR #1354, opened 2018-10-27
[FIX] Replace FSL ``mcflirt`` with AFNI ``3dVolReg``

### LINKED ISSUE #1535: Replace `mcflirt` with `3dVolreg`
- Consistent with MRIQC
- Faster than MCFLIRT
- More accurate than MCFLIRT (https://www.ncbi.nlm.nih.gov/pubmed/16099178)

oesteban:
Closes #1535.

A second attempt to get #1283 finally merged in.


oesteban:
Testing this here https://github.com/oesteban/fmriprep/tree/tests/fix-1354 to avoid excessive Circle usage.

However, I'm not positive the removal of rotation and shear does not have any additional side effects we haven't envisaged.

fmriprep/interfaces/images.py
```diff
@@ -370,7 +373,14 @@ def _run_interface(self, runtime):
 
         # Both match, qform valid (implicit with match), codes okay -> do nothing, empty report
         if matching_affines and qform_code > 0 and sform_code > 0:
-            self._results['out_file'] = self.inputs.in_file
+            if self.inputs.remove_rotation_and_shear:
+                out_fname = fname_presuffix(self.inputs.in_file, suffix='_valid',
+                                            newpath=runtime.cwd)
+                img = remove_rotation_and_shear(img)
```
effigies: Could it be that we're not correcting to closest canonical before removing rotation and shear in this interface?

oesteban: https://circleci.com/workflow-run/a814ff84-9ddd-4d8b-9961-81348b8fa202

oesteban:
Status report:

- Smoke tests on ds005, ds054 and ds210 look good
- Regression tests of masks don't work unless the `remove_rotation_and_shear` is disabled. I see two options for these tests to pass: keep `remove_rotation_and_shear` off, or update both test datasets and masks with the orientation fixed.

There seems to be no side effects to the `remove_rotation_and_shear` in the workflow. I think we can go ahead and merge this (currently, with `remove_rotation_and_shear` disabled in the regression tests).

Opinions? @chrisfilo @effigies

fmriprep/interfaces/images.py
effigies: So I think if we're removing rotation and shear, we need to be updating to RAS. Perhaps it makes sense to actually put that step directly into `remove_rotation_and_shear`, to prevent the case where we destroy orientation information.

If the output masks are valid, just in a different space, then we should update the targets.

fmriprep/utils/misc.py
```diff
@@ -7,6 +7,31 @@
 """
 
 
+def remove_rotation_and_shear(img):
+    from transforms3d.affines import decompose, compose
+    import numpy as np
+    import nibabel as nb
+
+    img = nb.as_closest_canonical(img)
```
effigies: Oh, I see you did put this in here. Ignore the above.

fmriprep/workflows/bold/hmc.py
```diff
@@ -10,16 +10,16 @@
...
+from ...interfaces import Volreg2ITK, MCFLIRT2ITK
 from ...engine import Workflow
-from ...interfaces import MCFLIRT2ITK
 
 DEFAULT_MEMORY_MIN_GB = 0.01
 
 
 # pylint: disable=R0914
-def init_bold_hmc_wf(mem_gb, omp_nthreads, name='bold_hmc_wf'):
+def init_bold_hmc_wf(use_mcflirt, mem_gb, omp_nthreads, name='bold_hmc_wf'):
```
effigies: If we eventually want to be able to swap in any MC algorithm, `use_mcflirt` is not a great name for the parameter. What about `motion_correction` that takes a string such as `'3dvolreg'` or `'mcflirt'` (and I would accept `'afni'` and `'fsl'` as synonyms, at least at the command line).

fmriprep/workflows/bold/tests/test_util.py
```diff
@@ -58,7 +58,8 @@ def symmetric_overlap(img1, img2):
     )
 ])
 def test_masking(input_fname, expected_fname):
-    bold_reference_wf = init_bold_reference_wf(omp_nthreads=1)
+    bold_reference_wf = init_bold_reference_wf(
+        omp_nthreads=1, remove_rotation_and_shear=False)
```
effigies: Since you're not setting this to `False` in the actual pipeline, I would not set it `False` here. It's not obvious that a change to the algorithm that causes a 2% shift without this step will cause the same 2% shift with the step.

chrisfilo: AFNI ignores those when calculating transformations. If you keep them in the reference and use ANTs to apply the transformations estimated using AFNI the results will be misaligned.

oesteban: That makes sense. Do you think it would be possible to include rotation and shear [here in the `Volreg2ITK` conversion](https://github.com/poldracklab/fmriprep/pull/1354/files#diff-e5bf7060b352a39b7fdadceab77399d6R91)?

chrisfilo: Potentially, but that would be harder to implement.

effigies: > What if we decide to fall back into mcflirt? Then this test would require rolling data back again.

As long as we always apply this cleaning to the entire BOLD series and , MCFLIRT will also be working from the cardinal RAS+ orientation. This would actually end up finishing up #260.

effigies:
Should re-raise the following point from https://github.com/poldracklab/fmriprep/issues/260#issuecomment-316137647:

> If we change the orientation of input files we need to adjust the phase encoding direction as well.

I would also add slice-encoding direction (#1345).

I think we need a quick interface (here as a function) to translate incoming directions to reoriented directions:

```Python
def update_encoding_axes(img, pe_dir, slice_dir):
    import nibabel as nb

    ornt = nb.orientations.io_orientation(img)

    axes = ornt[:, 0]
    flips = ornt[:, 1] # -1 if flip, 1 if no flip

    if pe_dir is not None:
        pe_idx = 'ijk'.index(pe_dir[0])
        pe_sign = -1 if pe_dir[1:] else 1
        pe_sign *= flips[pe_idx]
        pe_idx = axes[pe_idx]
        pe_dir = 'ijk'[pe_idx] + '-'[pe_sign:]
        
    if slice_dir is not None:
        ...

    return pe_dir, slice_dir
```

I can flesh out the interface and write tests, if you agree this is a good idea.

oesteban:
Isn't safer/easier to calculate the analytic inverse of `remove_rotation_and_shear` and apply it when building the ITK transforms?

oesteban:
I agree that adjusting the affine without reslicing is really fninicky.

Another concern to take into account. Merging #1334 (or any derivative of it) will make it very easy to implement an `orig` keyword to generate BOLD series in the original space. However, these outputs will not match the original orientation somewhat defeating the purpose of the reference. Does it make sense?

This is less of a problem if we set a `boldref` as reference. Since that reference would be produced by fMRIPrep, it is not a problem that the shear and rotation have been removed.

effigies:
I guess the question is whether we ever really care about the "original space" so much as "absolutely minimal resampling". I'm inclined to say it's actually the latter, in which case `boldref` is the better choice, and we should just be explicit about it, but perhaps there is a use-case for `orig`?

effigies:
Is this okay to push off to the next release?

oesteban:
:+1:

dangom:
Hi guys,

related to #1928, I would love to see this PR merged in order to start experimenting with fMRIPREP for the robust preprocessing of high-res/partial FOV data.

I noticed, and correct me if I'm wrong, that one of the obstacles towards getting this PR merged is thatAFNI messes up the nifti headers and therefore may cause potential issues in downstream resampling/registration steps - in particular when interacting with ANTs.

Interestingly enough, a solution to this problem seems to have already been implemented by Freesurfer. If we look at the code of freesurfer's 3dVolReg wrapper, `mc-afni2`, we see that they use `mri_convert` to [fix the 3dVolReg output](https://github.com/freesurfer/freesurfer/blob/8b88b346483ab97daeee45623953cdcf137f18ca/fsfast/bin/mc-afni2#L172). Would it not simplify the implementation of this PR simply swapping mcflirt for `mc-afni2` or porting freesurfer's solution by piping the output of 3dVolReg through a call to `mri_convert` as well?


```
# Always convert the output. AFNI mucks with the goemetry,
# so we have to run mri_convert with --like to keep it.
# 3dvolreg will remove off-diagonal components (but keeps
# the basic geometry correct).
echo "#@# --------------------------------" |& tee -a $LF
set cmd = (mri_convert $newoutvol $outvol --in_like $tempvol)
pwd |& tee -a $LF
echo $cmd |& tee -a $LF
$cmd |& tee -a $LF
if($status) then
  echo "ERROR: converting output " |& tee -a $LF
  if($cleanup) then
    echo "... cleaning up ...."; |& tee -a $LF
    rm -r $tmpdir
  endif
  exit 1;
endif
```

oesteban:
We are addressing the issue in poldracklab/nitransforms#9 without tricky hacks - just interpreting the "deobliquing" transforms as ``3dVolreg`` does. An extra pair of hands will be more than very much appreciated there :)

dangom:
Thanks for your answer.

I can't be of much help at the moment in interpreting AFNI transforms, since I don't have experience with AFNI. That said, I've been comparing 3dVolReg (via mc-afni2) and antsMotionCorr for high-res /partial FOV motion correction. Both work very well, and up to minor interpolation differences, results are almost identical.

3dVolReg has the disadvantage that it doesn't accept reference images with a different matrix size / resolution. antsMotionCorr does the right thing out of the box, so replacing an sbref for a whole-brain ref would work without further tweaks. Motion-correcting directly towards a whole brain (or large FOV) reference makes intuitively more sense for cases of extreme partial FOV ( < 1 cm coverage), where concepts such as a "mean volume" start to break down ( under motion each time-point may cover different, slightly non-overlapping anatomy).
The downside is that whole-brain references are generally separate acquisitions, so motion parameter estimates would potentially have to take that into account.

antsMotionCorr has the disadvantage that it's not currently supported by Nipype. An orphaned old-PR to add support exists, though.

oesteban:
I'd be 👍 with bringing back the idea of antsMotionCorr - that stub in nipype was created when we were first playing with it.

oesteban:
This PR has gone stale and almost impossible to merge right now. Also, we will hopefully go around the _oblique_ x-forms issue in NiTransforms proper, and this migration (to AFNI's tooling) will become pretty trivial.

It is important to say that #1535 stands as relevant as when Chris initially posted it if not more (given the unreliability of FD Power traces we get from mcflirt).

PR #1355, opened 2018-10-29
[ENH] Add echo-idx flag

emdupre:

Addition to #1263. Following up on discussions in #1324, adds an `echo-idx` flag to the fmriprep cli so that users can elect to preprocess only one echo in a mutliecho BOLD series.  

Currently failing due to merge conflicts in #1263.

Usage documentation is automatically built from the CLI, so I don't think any additional documentation needs to be added ? Please let me know f there's elsewhere you'd like me to update !

### DIFF
--- a/fmriprep/cli/run.py
+++ b/fmriprep/cli/run.py
@@ -77,6 +77,8 @@ def get_parser():
     #                     help='select a specific run to be processed')
     g_bids.add_argument('-t', '--task-id', action='store',
                         help='select a specific task to be processed')
+    g_bids.add_argument('--echo-idx', action='store',
+                        help='select a specific echo to be processed in a multiecho series')
 
     g_perfm = parser.add_argument_group('Options to handle performance')
     g_perfm.add_argument('--nthreads', '--n_cpus', '-n-cpus', action='store', type=int,
@@ -529,6 +531,7 @@ def build_workflow(opts, retval):
     retval['workflow'] = init_fmriprep_wf(
         subject_list=subject_list,
         task_id=opts.task_id,
+        echo_idx=opts.echo_idx,
         run_uuid=run_uuid,
         ignore=opts.ignore,
         debug=opts.sloppy,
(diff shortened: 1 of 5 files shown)


fmriprep/cli/run.py
```diff
@@ -77,6 +77,8 @@ def get_parser():
     #                     help='select a specific run to be processed')
     g_bids.add_argument('-t', '--task-id', action='store',
                         help='select a specific task to be processed')
+    g_bids.add_argument('--echo-idx', action='store',
```
effigies: `type=int,`?

And maybe let's add the short option `'-e'`.

fmriprep/workflows/base.py
```diff
@@ -56,6 +56,7 @@ def init_fmriprep_wf(subject_list, task_id, run_uuid, work_dir, output_dir, bids
         from fmriprep.workflows.base import init_fmriprep_wf
         wf = init_fmriprep_wf(subject_list=['fmripreptest'],
                               task_id='',
+                              echo_idx='',
```
effigies: `echo_idx=None`

effigies:
LGTM. Minor suggestions.

@chrisfilo Is `--<ent>-idx` BIDS-Apps compliant? I see we used to have `--run-id`, but we also had `--session-id` and `--participant-label`, so I'm not sure whether there's a convention to follow here or not.

emdupre:
Sorry about that ! Thanks for catching it, @effigies ! Updated both places.

fmriprep/cli/run.py
emdupre: The `-e` option is conflicting with the `-e` option for setting the env in `fmriprep-docker`. I'm reverting to only the long-form command, but let me know if you'd prefer a different approach !

effigies: That seems fine. We can always reconsider later, but I agree, let's not cause a conflict.

Issue #1169, opened 2018-06-14
Error occurs when using two phase images and two magnitude images as fieldmap inputs


wangboc:
Hi, according to BIDS naming convention, I use two magnitude and phase images as the fieldmap inputs. While, fmriprep failed to run at the beginning. The error infos  :

![image](https://user-images.githubusercontent.com/12145052/41398930-8ba57866-6feb-11e8-9290-71c60dcc8b88.png)

And my inputs structure is:
![image](https://user-images.githubusercontent.com/12145052/41398972-a8cd46da-6feb-11e8-9521-41dbe409cac4.png)

effigies:
Hi, what version of fMRIPrep are you running?

wangboc:
should be the newest version: 1.1.1
![image](https://user-images.githubusercontent.com/12145052/41401271-fde686da-6ff1-11e8-84ce-71885826a174.png)

chrisfilo:
@oesteban does FMRIPREP support  phase1/phase2 type of SDC input data? I could not find traces of it in the codebase.

wangboc:
How can I calculate the phase difference image by using the two phase image? I find the fslcomplex is so complex to run...  :)

mattcieslak:
So does this mean that BIDS section 8.9.2 isn't currently supported?

effigies:
@mattcieslak I believe that's correct. We would be happy to review a patch, however.

mattcieslak:
Happy to write a workflow for this. Would make sense to calculate a phasediff image from the two phases and send that as input to the workflow for 8.9.3-style fieldmaps? Or is there something fundamentally different between derived phasediffs and those that come from scanners?

effigies:
I'm not really sure, myself. I have dug into the fieldmap math much less than @oesteban, who might have thoughts here.

mattcieslak:
It looks like there is support for phase1/phase2 in the dipy workflow (https://github.com/nipy/nipype/blob/master/nipype/workflows/dmri/fsl/utils.py#L645).  Would it make sense to check if the `type` is `"phase"` and if so send the two phase images to `pha2rads`? 

Maybe it would be good to change the `type` from `"phase"` to `"phases"` or something more descriptive?

oesteban:
It's been a long way @wangboc, but this has been finally addressed and released with version 1.5.1 (although you should go with the newest 1.5.2).

Could you check whether this is working out for you now? Please reopen if you find any issues.

PR #1359, opened 2018-11-01
[ENH] Add phase1/phase2 SDC support

mattcieslak:

Add support for `phase1` and `phase2` fieldmaps, related to #1169. This is just one idea for how to support them, another possibility would be to make a separate workflow. 

Documentation probably doesn't need to be changed.


fmriprep/interfaces/fmap.py
```diff
@@ -530,3 +584,30 @@ def _delta_te(in_values, te1=None, te2=None):
...
+        raise RuntimeError('EchoTime metadata fields not found. '
+                           'Please consult the BIDS specification.')
+    if te1 is None:
+        raise RuntimeError(
+            'EchoTime metadata field not found for phase1. Please consult the BIDS specification.')
+    if te2 is None:
+        raise RuntimeError(
+            'EchoTime metadata field not found for phase2. Please consult the BIDS specification.')
+
+    return abs(float(te2) - float(te1))
```
effigies: I suspect you want to preserve the sign, so that you can subtract phase1 from phase2, and let the negative delta_te correct the sign if te1 is longer than te2.

mattcieslak: Sorry for the mess in these, I was mostly wondering if the handling in `init_sdc_wf` looked ok. I'm still filling out the details in the other functions

effigies: Oh, okay. I'll tag this as `[WIP]`. Just ping us when you want a review. (Or remove the `[WIP]` tag from the title to let us know if you think it's basically ready.)

oesteban:
Please remove the [WIP] tag to indicate this is almost there, and I'll gladly review it :)

Thanks for the contribution!!

mattcieslak:
@oesteban have you ever worked with phase1, phase2 images?  I tried running some directly through the siemens2rads function, which looks like it handled these at some point, but the output looks wrong. 

image

There is a script used here internally that handles these images, but it has two different equations depending on whether the readout was bipolar or monopolar (which isn't specified in the json sidecars). I'm wondering if supporting this kind of fieldmap is worth the trouble.

fmriprep/interfaces/fmap.py
```diff
@@ -497,6 +539,56 @@ def phdiff2fmap(in_file, delta_te, newpath=None):
...
+    image1 = nb.load(long_phase_image)
+    phase1 = image1.get_data().astype(np.float)
+
+    # Calculate fieldmaps
+    rad0 = np.pi * phase0 / 2048 - np.pi
+    rad1 = np.pi * phase1 / 2048 - np.pi
+    a = mag * np.cos(rad0)
+    b = mag * np.sin(rad0)
+    c = mag * np.cos(rad1)
+    d = mag * np.sin(rad1)
```
effigies: Since the same `mag` is applied to all angles, I think you can drop it entirely. Or am I missing something? Should there be a `mag1` and `mag2`?

fmriprep/interfaces/fmap.py
```diff
@@ -497,6 +539,56 @@ def phdiff2fmap(in_file, delta_te, newpath=None):
...
+    magnitude_image = magnitude_files[short_echo_index]
+    short_phase_image = phase_files[short_echo_index]
+    long_phase_image = phase_files[long_echo_index]
+
+    mag_img = nb.load(magnitude_image)
+    mag = mag_img.get_data().astype(np.float)
+    image0 = nb.load(short_phase_image)
+    phase0 = image0.get_data().astype(np.float)
+    image1 = nb.load(long_phase_image)
+    phase1 = image1.get_data().astype(np.float)
```
effigies: You can use

```Python
phase1 = image1.get_fdata()
```

Or to be more concise, you can just replace `phase1` later with `image1.get_fdata()`.

Same comments apply to `phase0` and `mag` (`get_fdata()` is caching, so calling multiple times is cheap), though I also have a question about whether `mag` is needed at all.

fmriprep/interfaces/fmap.py
```diff
@@ -497,6 +539,56 @@ def phdiff2fmap(in_file, delta_te, newpath=None):
...
+
+    # Calculate fieldmaps
+    rad0 = np.pi * phase0 / 2048 - np.pi
+    rad1 = np.pi * phase1 / 2048 - np.pi
```
effigies: Will phases always range from 0 to 4096?

mattcieslak: We're trying to recreate these scanning sequences to see if this is still true for the current Siemens product. This has been true for all the older versions of the sequence I've run into.

effigies: Thanks for the diligence on this. Do you know if any other scanners produce phase1/phase2 fieldmaps? It would be good to see if this convention extends beyond Siemens.

Perhaps @neurolabusc might have the experience or know someone who does.

.zenodo.json
```diff
@@ -84,6 +84,10 @@
       "orcid": "0000-0002-8508-9088"
     },
     {
+      "name": "Cieslak, Matthew",
+      "affiliation": "Department of Neuropsychiatry, University of Pennsylvania",
+      "orcid": "0000-0002-1931-4734"
+    },
       "name": "Liem, Franz",
```
effigies:
```suggestion
    {
      "name": "Liem, Franz",
```

effigies: Just realizing, your entry should also be moved down to immediately before Russ. At least for now, we're using the order of first contribution, with Russ and Chris as co-last authors.

mattcieslak: Aha, I was wondering how this worked. In my copy there is someone between Chris and Russ, should I move Romain elsewhere?

effigies: I just fixed it on master.

fmriprep/interfaces/fmap.py
neurolabusc: I have a web page that is [relevant](https://www.mccauslandcenter.sc.edu/crnl/tools/fieldmap). For Siemens images I think the scaled range is -4096..4092 (scl_slope = 2.000000; scl_inter = -4096.000000). For Philips I think the scaled range is ~-500..496 for some systems and -pi..+pi for others. For GE I think the convention is to save real/imaginary images so you would use fslcomplex to get the phase map. As a good estimate, you can assume min..max should be scaled very close to -pi..+pi.

effigies: Thanks for the rundown. If the slope and intercept are 2, -4096, and the original range is 0..4096, then this should be:

```Python
rad0 = np.pi * phase0 / 4096
rad1 = np.pi * phase1 / 4096
```

This is because `img.get_fdata()` already applies the scaling, so we should work from the scaled values, not the raw values.

I think we can ignore the GE case for now; that seems like it should probably not be considered "phase1"/"phase2" fieldmaps, and it's a job for BIDS to establish what the correct naming scheme is to indicate those files.

I think in the short term, we could have some a heuristic where we check for values above 500 and pi, to establish the correct denominator.

In the long term, I'm leery of putting too many heuristics for interpreting data into fMRIPrep that are not in the BIDS spec. We may want to consider a metadata field that unambiguously indicates the correct interpretation. (Though the heuristic would probably need to stay, to keep processing datasets that conform with BIDS 1.0.)

mattcieslak: I changed so the scaling is more general in 09316a9, but in my test data this causes a very small error because the maximum observed value is 4094 instead of 4096.

effigies: Maybe we could leave that as a last-resort option, and see if we can figure out a simple rescaling (no shift) in most cases? For example, a range of -4096.0 to 4092.0 exactly fits Chris's description of Siemens, so a `np.pi / 4096` scaling seems like a better choice.

mattcieslak: The minimum value that came out of `.get_fdata()` was 0, though.

effigies: I see. So I guess we need to watch for cases where the scale and intercept are missing.

fmriprep/interfaces/fmap.py
```diff
@@ -201,9 +198,34 @@ class Phasediff2Fieldmap(SimpleInterface):
 
     def _run_interface(self, runtime):
         self._results['out_file'] = phdiff2fmap(
-            self.inputs.in_file,
-            _delta_te(self.inputs.metadata),
-            newpath=runtime.cwd)
+            self.inputs.in_file, _delta_te(self.inputs.metadata), newpath=runtime.cwd)
+        return runtime
```
effigies: Revert.

fmriprep/interfaces/fmap.py
```diff
@@ -233,8 +255,7 @@ def _despike2d(data, thres, neigh=None):
                 patch_range = vals.max() - vals.min()
                 patch_med = np.median(vals)
 
-                if (patch_range > 1e-6 and
-                        (abs(thisval - patch_med) / patch_range) > thres):
+                if (patch_range > 1e-6 and (abs(thisval - patch_med) / patch_range) > thres):
```
effigies: Revert.

fmriprep/interfaces/fmap.py
```diff
@@ -255,9 +276,10 @@ def _unwrap(fmap_data, mag_file, mask=None):
     nb.Nifti1Image(magnii.get_data(), magnii.affine).to_filename('fmap_mag.nii.gz')
 
     # Run prelude
-    res = PRELUDE(phase_file='fmap_rad.nii.gz',
-                  magnitude_file='fmap_mag.nii.gz',
-                  mask_file='fmap_mask.nii.gz').run()
+    res = PRELUDE(
+        phase_file='fmap_rad.nii.gz',
+        magnitude_file='fmap_mag.nii.gz',
+        mask_file='fmap_mask.nii.gz').run()
```
effigies: Revert.

fmriprep/interfaces/fmap.py
```diff
@@ -290,8 +312,9 @@ def get_ees(in_meta, in_file=None):
 
            =  T_\\text{ro} \\,  (N_\\text{PE} / f_\\text{acc} - 1)^{-1}
 
-    where :math:`N_y` is the number of pixels along the phase-encoding direction
-    :math:`y`, and :math:`f_\\text{acc}` is the parallel imaging acceleration factor
+    where :math:`N_y` is the number of pixels along the phase-encoding
+    direction :math:`y`, and :math:`f_\\text{acc}` is the parallel imaging
+    acceleration factor
```
effigies: Revert.

fmriprep/interfaces/fmap.py
```diff
@@ -471,7 +494,8 @@ def phdiff2fmap(in_file, delta_te, newpath=None):
 
     .. math::
 
-        \Delta B_0 (\text{T}^{-1}) = \frac{\Delta \Theta}{2\pi\gamma \Delta\text{TE}}
+        \Delta B_0 (\text{T}^{-1}) = \frac{\Delta\Theta}{\
+                                           2\pi\gamma\Delta\text{TE}}
```
effigies: Revert.

fmriprep/interfaces/fmap.py
```diff
@@ -520,13 +595,13 @@ def _delta_te(in_values, te1=None, te2=None):
...
     if te1 is None:
-        raise RuntimeError(
-            'EchoTime1 metadata field not found. Please consult the BIDS specification.')
+        raise RuntimeError('EchoTime1 metadata field not found. Please consult the BIDS '
+                           'specification.')
     if te2 is None:
-        raise RuntimeError(
-            'EchoTime2 metadata field not found. Please consult the BIDS specification.')
+        raise RuntimeError('EchoTime2 metadata field not found. Please consult the BIDS '
+                           'specification.')
```
effigies: Please revert the above three lines, as no semantic changes were made.

fmriprep/workflows/fieldmap/base.py
```diff
@@ -9,8 +9,8 @@
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 If the dataset metadata indicate tha more than one field map acquisition is
-``IntendedFor`` (see BIDS Specification section 8.9) the following priority will
-be used:
+``IntendedFor`` (see BIDS Specification section 8.9) the following priority
+will be used:
```
effigies: Revert.

fmriprep/workflows/fieldmap/base.py
```diff
@@ -80,13 +81,17 @@ def init_sdc_wf(fmaps, bold_meta, omp_nthreads=1,
...
                 'type': 'phasediff',
-                'phasediff': 'sub-03/ses-2/fmap/sub-03_ses-2_run-1_phasediff.nii.gz',
-                'magnitude1': 'sub-03/ses-2/fmap/sub-03_ses-2_run-1_magnitude1.nii.gz',
-                'magnitude2': 'sub-03/ses-2/fmap/sub-03_ses-2_run-1_magnitude2.nii.gz',
+                'phasediff': \
+                    'sub-03/ses-2/fmap/sub-03_ses-2_run-1_phasediff.nii.gz',
+                'magnitude1': \
+                    'sub-03/ses-2/fmap/sub-03_ses-2_run-1_magnitude1.nii.gz',
+                'magnitude2': \
+                    'sub-03/ses-2/fmap/sub-03_ses-2_run-1_magnitude2.nii.gz',
```
effigies: Revert.

fmriprep/workflows/fieldmap/base.py
```diff
@@ -80,13 +81,17 @@ def init_sdc_wf(fmaps, bold_meta, omp_nthreads=1,
...
             }],
             bold_meta={
                 'RepetitionTime': 2.0,
-                'SliceTiming': [0.0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9],
+                'SliceTiming': [0.0, 0.1, 0.2, 0.3, 0.4, 0.5,
+                                0.6, 0.7, 0.8, 0.9],
```
effigies: Revert.

fmriprep/workflows/fieldmap/base.py
```diff
@@ -175,10 +180,12 @@ def init_sdc_wf(fmaps, bold_meta, omp_nthreads=1,
 
     # PEPOLAR path
     if fmap['type'] == 'epi':
-        outputnode.inputs.method = 'PEB/PEPOLAR (phase-encoding based / PE-POLARity)'
+        outputnode.inputs.method = \
+            'PEB/PEPOLAR (phase-encoding based / PE-POLARity)'
```
effigies: Revert.

fmriprep/workflows/fieldmap/base.py
```diff
@@ -175,10 +180,12 @@ def init_sdc_wf(fmaps, bold_meta, omp_nthreads=1,
...
         # Get EPI polarities and their metadata
-        epi_fmaps = [(fmap_['epi'], fmap_['metadata']["PhaseEncodingDirection"])
-                     for fmap_ in fmaps if fmap_['type'] == 'epi']
+        epi_fmaps = [
+            (fmap_['epi'], fmap_['metadata']["PhaseEncodingDirection"])
+            for fmap_ in fmaps if fmap_['type'] == 'epi']
```
effigies: Revert.

fmriprep/workflows/fieldmap/base.py
```diff
@@ -243,7 +257,8 @@ def init_sdc_wf(fmaps, bold_meta, omp_nthreads=1,
         workflow.connect([
             (inputnode, syn_sdc_wf, [
                 ('t1_brain', 'inputnode.t1_brain'),
-                ('t1_2_mni_reverse_transform', 'inputnode.t1_2_mni_reverse_transform'),
+                ('t1_2_mni_reverse_transform',
+                    'inputnode.t1_2_mni_reverse_transform'),
```
effigies: Revert.

fmriprep/workflows/fieldmap/phdiff.py
```diff
@@ -27,14 +28,15 @@
...
 )
 
 
-def init_phdiff_wf(omp_nthreads, name='phdiff_wf'):
+def init_phdiff_wf(omp_nthreads, phasetype="phasediff", name='phdiff_wf'):
     """
     Estimates the fieldmap using a phase-difference image and one or more
-    magnitude images corresponding to two or more :abbr:`GRE (Gradient Echo sequence)`
+    magnitude images corresponding to two or more
+    :abbr:`GRE (Gradient Echo sequence)`
```
effigies: Revert.

fmriprep/workflows/fieldmap/phdiff.py
```diff
@@ -64,18 +66,13 @@ def init_phdiff_wf(omp_nthreads, name='phdiff_wf'):
 further improvements of HCP Pipelines [@hcppipelines].
 """
 
-    inputnode = pe.Node(niu.IdentityInterface(fields=['magnitude', 'phasediff']),
+    inputnode = pe.Node(niu.IdentityInterface(
+                        fields=['magnitude', 'phasediff']),
```
effigies: Revert.

fmriprep/workflows/fieldmap/phdiff.py
```diff
@@ -84,19 +81,18 @@ def _pick1st(inlist):
                  name='n4', n_procs=omp_nthreads)
     bet = pe.Node(BETRPT(generate_report=True, frac=0.6, mask=True),
                   name='bet')
-    ds_fmap_mask = pe.Node(DerivativesDataSink(suffix='fmap_mask'), name='ds_report_fmap_mask',
+    ds_fmap_mask = pe.Node(DerivativesDataSink(suffix='fmap_mask'),
+                           name='ds_report_fmap_mask',
```
effigies: Revert.

fmriprep/workflows/fieldmap/phdiff.py
```diff
@@ -84,19 +81,18 @@ def _pick1st(inlist):
...
 
-    # phase diff -> radians
-    pha2rads = pe.Node(niu.Function(function=siemens2rads), name='pha2rads')
-
     # FSL PRELUDE will perform phase-unwrapping
     prelude = pe.Node(fsl.PRELUDE(), name='prelude')
 
-    denoise = pe.Node(fsl.SpatialFilter(operation='median', kernel_shape='sphere',
+    denoise = pe.Node(fsl.SpatialFilter(operation='median',
+                                        kernel_shape='sphere',
```
effigies: Revert.

fmriprep/workflows/fieldmap/phdiff.py
```diff
@@ -107,19 +103,48 @@ def _pick1st(inlist):
 
     # The phdiff2fmap interface is equivalent to:
     # rad2rsec (using rads2radsec from nipype.workflows.dmri.fsl.utils)
-    # pre_fugue = pe.Node(fsl.FUGUE(save_fmap=True), name='ComputeFieldmapFUGUE')
+    # pre_fugue = pe.Node(fsl.FUGUE(save_fmap=True),
+    #                     name='ComputeFieldmapFUGUE')
```
effigies: Revert.

.zenodo.json
effigies: This still needs updating. Please move your entry to just before Russ.

oesteban:
Hi, sorry for being missing all this time. I agree with @effigies that reverting stylistic changes will help review this PR substantially.

That said, why the phase maps cannot be just subtracted and passed over to the phasediff routines?

neurolabusc:
Here is the formula, assuming GE data with 12-bit ADC... Might want to check with 16-bit ADC

diffTE := abs(TE1sec - lTE2sec); // difference in phase, sec
scale := (2*pi/4096)/diffTE;
rescaleImg(-2048, scale); //intercept, slope

fmriprep/interfaces/fmap.py
```diff
@@ -116,11 +116,11 @@ def _run_interface(self, runtime):
             if nslices > 1:
                 diffmapmsk = mask[..., errorslice[0]:errorslice[-1]]
                 diffmapnii = nb.Nifti1Image(
-                    diffmap[..., errorslice[0]:errorslice[-1]] * diffmapmsk,
-                    datanii.affine, datanii.header)
+                    diffmap[..., errorslice[0]:errorslice[-1]] * diffmapmsk, datanii.affine,
```
oesteban: revert

fmriprep/interfaces/fmap.py
```diff
@@ -116,11 +116,11 @@ def _run_interface(self, runtime):
...
+                    datanii.header)
 
-                bspobj2 = fbsp.BSplineFieldmap(diffmapnii, knots_zooms=[24., 24., 4.],
```
oesteban: revert

fmriprep/interfaces/fmap.py
```diff
@@ -129,8 +129,8 @@ def _run_interface(self, runtime):
             else:
                 final = smoothed1.get_data()
 
-            nb.Nifti1Image(final, datanii.affine, datanii.header).to_filename(
```
oesteban: revert

fmriprep/interfaces/fmap.py
```diff
@@ -201,12 +201,38 @@ class Phasediff2Fieldmap(SimpleInterface):
 
     def _run_interface(self, runtime):
         self._results['out_file'] = phdiff2fmap(
-            self.inputs.in_file,
```
oesteban: revert

fmriprep/interfaces/fmap.py
```diff
@@ -471,7 +496,7 @@ def phdiff2fmap(in_file, delta_te, newpath=None):
 
     .. math::
 
-        \Delta B_0 (\text{T}^{-1}) = \frac{\Delta \Theta}{2\pi\gamma \Delta\text{TE}}
+        \Delta B_0 (\text{T}^{-1}) = \frac{\Delta \Theta}{\2\pi\gamma \Delta\text{TE}}
```
oesteban: Is this intended?

fmriprep/workflows/fieldmap/phdiff.py
```diff
@@ -110,16 +102,44 @@ def _pick1st(inlist):
...
+by Mark Elliott [@pncprocessing].
+    """
+        # Special case for phase1, phase2 images
+        meta = pe.MapNode(ReadSidecarJSON(), name='meta', mem_gb=0.01,
+                          run_without_submitting=True, iterfield=['in_file'])
+        phases2fmap = pe.Node(Phases2Fieldmap(), name='phases2fmap')
+        workflow.connect([
+            (meta, phases2fmap, [('out_dict', 'metadatas')]),
+            (inputnode, phases2fmap, [('phasediff', 'phase_files')]),
+            (phases2fmap, prelude, [('out_file', 'phase_file')]),
```
oesteban: the difference map needs to be phase unwrapped? aren't the phase maps unwrapped individually?

effigies: Consider:

p1 = pi - eps
p2 = -pi + eps

p1 - p2 = 2pi - 2eps = -2eps

Phase differences should be between -pi and pi, as well.

oesteban: That means you are clipping it again, correct?

effigies: Oh sorry, I think I misunderstood the context here. You're saying that `prelude` just does the clipping between -pi and pi, and we've already done that in `phases2fmap`?

oesteban: Yes, I understood before that each phase map needs to be unwrapped (which makes sense to me). If they are phase-unwrapped, there's no point on running [PRELUDE](https://fsl.fmrib.ox.ac.uk/fsl/fslwiki/FUGUE/Guide#PRELUDE_.28phase_unwrapping.29) (Phase Region Expanding Labeller for Unwrapping Discrete Estimates).

Also, I am very curious as to how the individual phase maps are unwrapped, if not done with PRELUDE. I know there are better methods out there (e.g., using graph cuts) but they are rarely used in general.

fmriprep/interfaces/fmap.py
mattcieslak: We scanned all the available GRE sequences on a prisma and uploaded them on openneuro in case you want to take a look. Here's the link: https://openneuro.org/datasets/ds001600. These were converted with a recent `dcm2niix`. 

I'm open to suggestions on how to handle these robustly. Also, sorry about the formatting mess. I ran a misconfigured yapf and am still putting things back.

neurolabusc: @mattcieslak can you send me the raw DICOM data and permission to distribute the DICOMs on the [dcm2niix](https://www.nitrc.org/plugins/mwiki/index.php/dcm2nii:MainPage) wiki? This would provide a nice reference set for all DICOM-to-BIDS conversion tools to test, and it would provide nice validation for any future changes or features.

neurolabusc: Despite being acquired on a E11C Prisma that supports high dynamic range (16-bit DICOM), all the samples you acquired appear to use 12-bit precision (0..4095). My guess is 12-bit is more than sufficient for the SNR provided, but I would make sure that any code generates an error if image intensities are outside this range. This will detect any sequences that provide more than 12-bit precision.

- sub-1_acq-v1_magnitude1	0..1821
- sub-1_acq-v1_magnitude2	0..1705
- sub-1_acq-v1_phase1	0..4095
- sub-1_acq-v1_phase2	0..4095
- sub-1_acq-v2_magnitude1	0..1835
- sub-1_acq-v2_magnitude2	0..1738
- sub-1_acq-v2_phase1	0..4095
- sub-1_acq-v2_phase2	0..4095
- sub-1_acq-v4_magnitude1	0..1598
- sub-1_acq-v4_magnitude2	0..1566
- sub-1_acq-v4_phasediff	0..4095
- sub-1_dir-AP_epi	0..2341
- sub-1_dir-PA_epi	0..1796

mattcieslak: @neurolabusc sent a link to your mailbox.sc.edu from my pennmedicine email

chrisfilo:
How close is it from being merged? There is at least one more user that would benefit from this work: https://neurostars.org/t/fmriprep-fieldmaps-keyerror-phase/3047

effigies:
I missed some of the conversation in here... I think the main thing remaining is the exact scaling algorithm, as in [this thread](https://github.com/poldracklab/fmriprep/pull/1359#discussion_r231272638). I'm not very happy with the current `rescale_image` approach:

https://github.com/poldracklab/fmriprep/blob/f2e861d61782d54bd31331bb701af7e029b6fbbf/fmriprep/interfaces/fmap.py#L546-L559

as it ignores zero-points in known conventions, ending with a slightly positive shift in phase. Additionally, if the minimum and maximum values aren't present for some reason, it causes a stretch. For the former concern, given that we're dealing with phase differences, it should be fine as long as the same shift occurs; for the latter, I don't know whether that's a plausible situation, given the number of voxels and range of values. So perhaps it's fine, but I'm a little inclined to try to interpret the values exactly as intended by the scanner, if we can.

effigies:
I think it would be good to support, even if people end up preferring to ignore fieldmaps and use SyN-SDC.

chrisfilo:
+1

oesteban:
I'll have a look ASAP

effigies:
Please merge master to fix the doc build.

mattcieslak:
Any idea what's happening with Tcompcor?

oesteban:
Happened the same with my previous pr and the docs build. Two is a pattern
(not particularly good this one)

effigies:
I'll revert #1443.

neurolabusc:
@mattcieslak: the developmental branch of dcm2niix (v1.0.20181218) now inserts a "DiffusionScheme" tag into your BIDS sidecar. This requires the tag `sDiffusion.dsScheme` in the Siemens CSA header - I think this is in both the CMRR and Siemens product DICOMs with E11. I am not sure how to detect this for  other vendors or older Siemens data (though with earlier versions of the Siemens product, I think one could only select the `Bipolar` scheme). Regardless, if this tag is detected, the BIDS sidecar should include a tag that looks like this `"DiffusionScheme": "Monopolar"` or this `"DiffusionScheme": "Bipolar"`.

mattcieslak:
Thanks @neurolabusc this will be very useful for the diffusion pipeline I'm working on. For these scans, whether it was bipolar or monopolar was specified in a tag called "ucReadOutMode" that was 2 if the scan bipolar or 1 if monopolar. I'm suspicious that the bipolar readoutmode was a non-standard variation that was some sort of a work-around at the time. Have you ever seen this vary on fieldmaps?

neurolabusc:
@mattcieslak please check the latest dcm2niix development branch commit. It will attempt to read the CSA tag sDiffusion.dsScheme (where 1=bipolar, 2=monopolar) and if that fails it will attempt to read ucReadOutMode (where 0x1=monopolar and 0x2=bipolar). If either is found, the BIDS file will report either `"DiffusionScheme": "Monopolar"` or `"DiffusionScheme": "Bipolar"`.

I believe sDiffusion.dsScheme is used on E11 for both CMRR and Siemens product. I think the ucReadOutMode was used by CMRR prior to E11. I am pretty sure that prior to E11, Siemens product diffusion sequences were always bipolar (e.g. used a bipolar twice-refocused spin echo sequence).

neurolabusc:
@mattcieslak I think your method will work with all the Siemens images I have seen. I do not know how to detect these properties for GE and Philips. My sense is that monopolar is the default for Philips. @drmclem might know the answer.

neurolabusc:
@mattcieslak if it is Siemens and not detected by my software, it seems that the it is most likely bipolar. I think the Siemens sequences used to be only bipolar, and I think my new version detects bipolar/monopolar settings for CMRR sequences as well as the new Siemens sequences that allow one to select between bipolar and monopolar. For many DWI scans, one should be able to reliably infer bipolar or monopolar once knows the matrix size, partial Fourier and iPAT. For most DTI sequences users tend to select  the minimum possible echo time to maximize SNR. While bipolar sequences show less spatial distortion and are less sensitive to background gradients, they pay a large penalty with regards to TE. The one excpetion might be scans where users want to measure kurtosis, as these groups tend to prefer longer echo times, on the other hand they are also concerned with background gradients so the kurtosis sequences I have seen follow the trend of long TE and bipolar sequence.

effigies:
I'm okay in either direction. Whether we reject anything that isn't explicitly monopolar or only reject things that are explicitly bipolar both seem like arguable choices. If we run into the problem that a lot of people with unlabeled bipolar fieldmaps are having issues, we can revisit it or perhaps encourage users to better label their data.

mattcieslak:
That makes sense for diffusion-weighted image series, but in this case the difference between monopolar and bipolar phase images doesn't just change the SNR/TE but the biploar readout also changes the values in the phase images. Bipolar phase images require that you scale the non-phase-encoding direction by its coordinate. This was previously corrected using

```bash
NP=`fslval ${outfroot}_rad2raw dim1`
3dcalc -a ${outfroot}_rad2raw.nii -expr "(i-$NP/2-1)*3.14159/$NP/2" -prefix ${outfroot}_rad2err.nii
```

which was checked against the monopolar sequence using a phantom.  I have no idea how general this solution would be, which is why I don't think it should go into fmriprep. It's also probably very rare to see a bipolar GRE scan.  

The phasediff calculation in this patch works easily on the monopolar sequences, but the displacements that come out of the bipolar sequence greatly shift one hemisphere (you can see the shearing here: https://github.com/mattcieslak/example_fieldmaps/blob/master/figures/sub-1_task-rest_acq-v1_sdc_phase.svg).

neurolabusc:
OK, thanks for the clarification.

mattcieslak:
I'm hoping to finish this up. I merged in the latest fmriprep and am running into an issue with fmriprep-docker where --patch-fmriprep results in smriprep being missing. I'm using version 1.3.2. It this expected?

effigies:
@mattcieslak We should probably add a `--patch-smriprep` option to `fmriprep-docker`. But you can use `-i poldracklab/fmriprep:1.4.0a1` to use the latest pre-release, which should have a compatible version of `smriprep`.

mattcieslak:
Ok, this is working now on all my test datasets. Anything else I should add?

oesteban:
Thanks a lot. I'll need some time to go through this PR with enough depth. Since I'm at ISMRM right now, and we are about to release a new version with big changes, the review may take a little while.

fmriprep/workflows/fieldmap/phdiff.py
oesteban: Okay, I have confirmed this point: as per [the FUGUE guide](https://fsl.fmrib.ox.ac.uk/fsl/fslwiki/FUGUE/Guide), each phase map should be unwrapped with PRELUDE separately (step 3) of the guide. Then they can be subtracted as they are not wrapped within [-pi, pi) anymore.

After that point, the pipeline should be exactly the same as for the `phasediff` type of fieldmap.

mattcieslak:
This might also solve some of the issues we've been running into with these fieldmaps when they come from our 7T scanner. Also, I'd be happy to help with a refactor of the fieldmap workflows. I'm maintaining a diffusion pipeline that shamelessly copies them and would like to keep them up to date. Should we close this PR then?

oesteban:
This will be continued on poldracklab/sdcflows

Issue #1365, opened 2018-11-05
Fmriprep 1.2.1: ica_aroma_wf.rm_nonsteady: skip_vols argument missing


markushs:
Hi!

We're testing out 1.2.1 and get the following errors for all datasets during ICA_AROMA:

<img width="1636" alt="image" src="https://user-images.githubusercontent.com/44242647/47993432-b7c8dd00-e0ef-11e8-88c1-c4f1ce6f01b8.png">

The error log says:
<img width="1609" alt="image" src="https://user-images.githubusercontent.com/44242647/47993467-cb744380-e0ef-11e8-85fb-4a1c1a20f721.png">

I guess this is likely due to the changes following #1335, but looking through the code I couldn't see any obvious problems (my python-skills are limited, however..). For information, we're running the following fmriprep-command on (BIDS-conformed) MB-data with single-band refs:

<img width="725" alt="image" src="https://user-images.githubusercontent.com/44242647/47993767-a3391480-e0f0-11e8-8573-e35f97aa6556.png">

(sorry about the screenshots - our data is stored in a secure environment that won't even let us copy-paste text).

Any ideas?

Thanks!
Markus

effigies:
Hi, thanks for the report. I think the issue is that you have single-band references, which means that we don't attempt to use non-steady-state volumes to create a reference image, so we skip discovery and counting altogether. I'll push a change to fix that.

PR #1368, opened 2018-11-06
TEST: Verify LegacyMultiProc functionality

### LINKED ISSUE #1363: Plugin-fix for memory issues fails in 1.2.1
Hi!

We've been running large MB-datasets successfully using fmriprep 1.1.8 and the the "--use-plugin" argument (yaml file configured as in #1259). 

However, when testing out version 1.2.1 we get the following error (on the same data with the same  input parameters as in 1.1.8):

image

The error occurs immediately after the citation dump (i.e., before initiation of any workflows)

Here's the command:

image

And the content of the yaml-file:

image

We do not see the error when running fmriprep without the --use-plugin argument (however other issues arises, see #1365).

Any ideas?

Thanks!
Markus

(sorry about the screenshots - our data is stored in a secure environment that won't even let us copy-paste text)

effigies:

* Add a `legacy.yml` plugin configuration to the Circle directory
* Run full ds005 with `LegacyMultiProc`
* Upgrade to nipype>=1.1.5

Closes #1363.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -397,7 +397,7 @@ jobs:
             - /tmp/ds005/derivatives/fmriprep
 
       - run:
-          name: Run full fMRIPrep on ds005
+          name: Run full fMRIPrep on ds005 (LegacyMultiProc plugin)
           no_output_timeout: 2h
           command: |
             sudo setfacl -d -m group:$(id -gn):rwx /tmp/ds005/derivatives && \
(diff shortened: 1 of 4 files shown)


effigies:
Reproducing #1363. cc @markushs

markushs:
Nice! I reproduced the error locally using docker (as you asked in #1363) ~1 minute ago, but seems like you solved it without that info.

effigies:
Yup. I hope to have a fixed release out today or tomorrow. For your other issue (#1365), we need to upgrade niworkflows, which might have a few kinks to work out.

PR #1378, opened 2018-11-09
[FIX] Correctly populate right-hemisphere time series in CIFTI derivatives

sebnaze:

The indexing of the cortical structure name after the split was taking the first letter of 'LEFT' (L) or 'RIGHT' (R) and then comparing it to 'RIGHT' a few lines below to select the proper hemisphere (which was then always False). This resulted in LEFT and RIGHT cortical hemisphere BOLD time series being identical, it now compares the whole word and the output is consistent.

No need for changes in doc.


effigies:
@sebnaze Thanks for the fix. Realizing I forgot to ask if you would like to add yourself to the Authors list in the .zenodo file.

PR #1394, opened 2018-11-15
Sentry refinement

chrisfilo:
Fixes two issues:
- If you build a container from master `__version__` is set to `''` - this was tripping Environment assignement
- `fMRIPrep finished without errors` was being sent if errors without crash files (KeyboardInterrupt for example) occurred


oesteban:
Okay, I think we can merge now and release after tests pass on master.

chrisfilo:
Thanks!

PR #1395, opened 2018-11-17
[MAINT] Import from ``bids.layout`` instead of ``grabbids``

oesteban:
Fixes a warning that is showing up lately, derived from this change in
pybids.

### DIFF
--- a/fmriprep/utils/bids.py
+++ b/fmriprep/utils/bids.py
@@ -20,7 +20,7 @@ import os.path as op
 import json
 import warnings
 from itertools import groupby
-from bids.grabbids import BIDSLayout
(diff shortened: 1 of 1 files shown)


fmriprep/utils/bids.py
```diff
@@ -20,7 +20,7 @@
 import json
 import warnings
 from itertools import groupby
-from bids.grabbids import BIDSLayout
+from bids.layout import BIDSLayout
```
effigies: I would just use `from bids import BIDSLayout`.

oesteban: Done with merge.

PR #1397, opened 2018-11-18
[FIX] Always run FreeSurfer interfaces that sink outside working directory

### LINKED ISSUE #1342: Mark FreeSurfer tools that sink to external directories to always run
Any tool that produces outputs in an external directory (data sinks, FreeSurfer manipulation tools) can produce an internally consistent node working directory that will nonetheless fail to populate the `OutputSpec` with valid locations, if the output directories have been removed. We should mark such tools to always run, and ensure they're written such that they will not re-perform expensive or incorrect computation if the outputs do already exist.

See discussion in https://neurostars.org/t/fmriprep-node-hash-trait-errors/2711/6

effigies:

Fixes #1342.


effigies:
This should be good to go.

PR #1399, opened 2018-11-18
FIX: Calculate template transforms explicitly as RAS2RAS

### LINKED ISSUE #1371: "template_transforms "-output of init_anat_template_wf seems to be off
Hi there,

I am trying to make a version of fmriprep that can gracefully deal with MP2RAGEs. I seem to fail quite ungracefully.

 I need the `template_transforms`-output of the `init_anat_template_wf` workflow, so that I can get the 2nd inversions into the same template space.

However, when I use those transforms to transform the T1w-template to the original T1ws with FSLeyes, FSL applyXFM or ANTS's ApplyTransforms, the transformation is completely off. The template ends up translated a couple of cms.

I have the suspicion that there is something wrong with the `fmriprep.interfaces.image.Conform`-interface. If I apply the output matrix fo this interface to the input of the interface, I get a very different orientation from what the interface outputs itself.

Anyone a clue what happens here?

effigies:

@gilles86 This is a first pass at fixing the template transform calculation. I think it probably needs to be inverted, but the structure should be there. I'll give this a shot on the data you sent me on Monday, but if you have some time to try it out before then, you might be able to see if I'm moving us in the right direction.

Fixes #1371.

### DIFF
--- a/fmriprep/interfaces/images.py
+++ b/fmriprep/interfaces/images.py
@@ -208,7 +208,7 @@ class ConformInputSpec(BaseInterfaceInputSpec):
 
 class ConformOutputSpec(TraitedSpec):
     out_file = File(exists=True, desc='Conformed image')
-    transform = File(exists=True, desc='Conformation transform')
+    transform = File(exists=True, desc='Conformation transform (voxel-to-voxel)')
 
 
 class Conform(SimpleInterface):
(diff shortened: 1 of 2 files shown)


Gilles86:
Hey!

Yes! This works like I would expect, out-of-the-box. Great work.

effigies:
Great! Thanks for the report... I'm not sure how soon I would've caught that. I haven't tried to work back in the original T1w image spaces.

@satra @mgxd I think we put that feature in at your request. If your input files weren't RAS, you might have bad transforms in your derivative directories.

satra:
thanks @effigies

PR #1414, opened 2018-11-20
[MAINT] Replace ``img.get_affine()`` -> ``img.affine``

oesteban:


fmriprep/interfaces/images.py
```diff
@@ -69,7 +69,7 @@ def _run_interface(self, runtime):
                 else:
                     in_files = [fname_presuffix(in_files[0], suffix='_squeezed',
                                                 newpath=runtime.cwd)]
-                    nb.Nifti1Image(sqdata, filenii.get_affine(),
+                    nb.Nifti1Image(sqdata, filenii.affine,
                                    filenii.get_header()).to_filename(in_files[0])
```
effigies: `filenii.header`, while we're at it. (And `hmcnii.header` below.)

PR #1418, opened 2018-11-21
[ENH] Group common issues with fingerprints

### LINKED ISSUE #1407: Tag memory and disk space issues

chrisfilo:
Closes #1407. This should reduce the amount of noise on sentry.

### DIFF
--- a/fmriprep/viz/reports.py
+++ b/fmriprep/viz/reports.py
@@ -165,7 +165,8 @@ class Report(object):
                                         'no-disk-space': ["OSError: [Errno 28] No space left on "
                                                           "device", "[Errno 122] Disk quota "
                                                                     "exceeded"],
-                                        'sigkill': ["Return code: 137"]}
+                                        'sigkill': ["Return code: 137"],
+                                        'keyboard-interrupt': ["KeyboardInterrupt"]}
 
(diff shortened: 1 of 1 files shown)


fmriprep/viz/reports.py
```diff
@@ -147,18 +147,42 @@ def index_error_dir(self, error_dir):
...
+                                        'no-disk-space': ["OSError: [Errno 28] No space left on "
+                                                          "device", "[Errno 122] Disk quota "
+                                                                    "exceeded"],
+                                        'sigkill': ["Return code: 137"]}
+
+                    fingerprint = ''
+                    issue_title = node_name + ': ' + gist
+                    for new_fingerprint, error_snippets in fingerprint_dict.items():
+                        for error_snippet in error_snippets:
+                            if error_snippet in crash_info['traceback']:
```
effigies: Are these all `str.startswith` matches? Because `in` will be a more extensive search in a potentially large traceback string. Normally wouldn't mind, but this is in a loop, and the dictionary is likely to grow, so this could start becoming a non-trivial.

effigies: Okay. Since it's not a quick replacement, I'm okay leaving it as is.

PR #1419, opened 2018-11-21
[ENH] Make sure inputs are BIDS compliant before running fmriprep

### LINKED ISSUE #1408: Make sure inputs are BIDS compatible prior to running fmriprep
Add a flag allowing people to disable the check.  Consider elevating some warnings (for example missing phase encoding direction) from warnings to errors via a config flag.

Alternatively, use strict mode from pybids.

chrisfilo:
Closes #1408. Ignores issues not relevant to FMRIPREP and data from participants not listed in `--participant_label` to limit the amount of noise.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -87,6 +87,9 @@ RUN curl -sL https://deb.nodesource.com/setup_10.x | bash -
 RUN apt-get install -y nodejs
 RUN npm install -g svgo
 
+# Installing bids-validator
+RUN npm isntall -g bids-validator@1.1.0
+
 # Installing and setting up ICA_AROMA
 RUN mkdir -p /opt/ICA-AROMA && \
   curl -sSL "https://github.com/maartenmennes/ICA-AROMA/archive/v0.4.4-beta.tar.gz" \
--- a/Dockerfile
+++ b/Dockerfile
@@ -88,7 +88,7 @@ RUN apt-get install -y nodejs
 RUN npm install -g svgo
 
 # Installing bids-validator
-RUN npm isntall -g bids-validator@1.1.0
+RUN npm install -g bids-validator@1.1.0
 
 # Installing and setting up ICA_AROMA
 RUN mkdir -p /opt/ICA-AROMA && \
--- a/fmriprep/cli/run.py
+++ b/fmriprep/cli/run.py
@@ -503,7 +503,6 @@ def validate_input_dir(exec_env, bids_dir, participant_label):
             logger.error("bids-validator does not appear to be installed")
 
 
-
 def build_workflow(opts, retval):
     """
     Create the Nipype Workflow that supports the whole execution
--- a/fmriprep/cli/run.py
+++ b/fmriprep/cli/run.py
@@ -455,7 +455,7 @@ def validate_input_dir(exec_env, bids_dir, participant_label):
             "MISSING_SESSION"
         ],
         "error": ["NO_T1W"],
-        "ignoredFiles": ['dataset_description.json', 'participants.tsv']
+        "ignoredFiles": ['/dataset_description.json', '/participants.tsv']
     }
     # Limit validation only to data from requested participants
     if participant_label:
--- a/fmriprep/cli/run.py
+++ b/fmriprep/cli/run.py
@@ -260,6 +260,16 @@ def main():
     warnings.showwarning = _warn_redirect
     opts = get_parser().parse_args()
 
+    exec_env = os.name
+
+    # special variable set in the container
+    if os.getenv('IS_DOCKER_8395080871'):
+        exec_env = 'singularity'
+        if 'docker' in Path('/proc/1/cgroup').read_text():
+            exec_env = 'docker'
+            if os.getenv('DOCKER_VERSION_8395080871'):
+                exec_env = 'fmriprep-docker'
+
     sentry_sdk = None
     if not opts.notrack:
         import sentry_sdk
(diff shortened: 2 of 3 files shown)


Dockerfile
```diff
@@ -87,6 +87,9 @@ RUN curl -sL https://deb.nodesource.com/setup_10.x | bash -
 RUN apt-get install -y nodejs
 RUN npm install -g svgo
 
+# Installing bids-validator
+RUN npm isntall -g bids-validator@1.1.0
```
effigies:
```suggestion
RUN npm install -g bids-validator@1.1.0
```

fmriprep/cli/run.py
```diff
@@ -399,6 +412,95 @@ def before_send(event, hints):
...
+
+        ignored_subs = all_subs.difference(selected_subs)
+        if ignored_subs:
+            for sub in ignored_subs:
+                validator_config_dict["ignoredFiles"].append("/sub-%s/**" % sub)
+    with tempfile.NamedTemporaryFile('w+') as temp:
+        temp.write(json.dumps(validator_config_dict))
+        temp.flush()
+        subprocess.check_call('bids-validator %s -c %s' % (bids_dir, temp.name),
+                              shell=True)
```
effigies: For the rare native installs or custom builds, we can't assume that `bids-validator` exists.

We may want to wrap this in a `try`/`except` block, with an error message indicating how to install bids-validator or skip validation.

fmriprep/cli/run.py
```diff
@@ -399,6 +412,95 @@ def before_send(event, hints):
...
+            "ACQTIME_FMT",
+            "SUSPICIOUSLY_LONG_EVENT_DESIGN",
+            "SUSPICIOUSLY_SHORT_EVENT_DESIGN",
+            "MALFORMED_BVEC",
+            "MALFORMED_BVAL",
+            "MISSING_TSV_COLUMN_EEG_ELECTRODES",
+            "MISSING_SESSION"
+        ],
+        "error": ["NO_T1W"],
+        "ignoredFiles": ['/dataset_description.json', '/participants.tsv']
```
effigies: Does this indicate that we don't care if these files are missing, or we don't validate their contents? If we want to fingerprint the dataset, the contents of `dataset_description.json` could be useful. And to correctly produce a derivative `dataset_description.json`, I think we'll require the information from this file, as well.

fmriprep/cli/run.py
```diff
@@ -399,6 +412,95 @@ def before_send(event, hints):
...
+    }
+    # Limit validation only to data from requested participants
+    if participant_label:
+        all_subs = set([os.path.split(i)[1][4:] for i in glob(os.path.join(bids_dir,
```
effigies: Clearer, IMO:

```suggestion
        all_subs = set([os.path.basename(i)[4:] for i in glob(os.path.join(bids_dir,
```

fmriprep/cli/run.py
```diff
@@ -399,6 +412,95 @@ def before_send(event, hints):
...
+            raise RuntimeError(error_msg % ','.join(bad_labels))
+
+        ignored_subs = all_subs.difference(selected_subs)
+        if ignored_subs:
+            for sub in ignored_subs:
+                validator_config_dict["ignoredFiles"].append("/sub-%s/**" % sub)
+    with tempfile.NamedTemporaryFile('w+') as temp:
+        temp.write(json.dumps(validator_config_dict))
+        temp.flush()
+        subprocess.check_call('bids-validator %s -c %s' % (bids_dir, temp.name),
```
effigies: Is there any advantage to `shell=True`, here? This will cause problems if `bids_dir` has a space in it.

```suggestion
        subprocess.check_call(['bids-validator', bids_dir, '-c', temp.name])
```

Dockerfile
effigies: Looks like `npm` is forgiving of misspellings?

https://circleci.com/gh/chrisfilo/fmriprep/782

fmriprep/cli/run.py
chrisfilo: I added this exception because one of our test datasets is missing `dataset_description.json` (which should be fixed eventually but is not critical). In general, I wanted to avoid being too strict about checks that are irrelevant to FMRI. When we add dataset fingerprinting we can change this behavior.

fmriprep/cli/run.py
chrisfilo: Well, I added bids-validator to the list of dependencies. I would prefer people to still use it even if they use native installs.

effigies: Sure. But the error rising from `check_call` will be somewhat hard to interpret. Perhaps just:

```Python
try:
    subprocess.check_call(...)
except FileNotFoundError:
    logger.error("bids-validator does not appear to be installed")
```

effigies:
Just a quick ping: @adelavega @tyarkoni How close is `BIDSLayout` to having full validation implemented?

chrisfilo:
I don't think pybids validation would be a suitable replacement. It does not check nifti headers or validate the contents of JSON files.

tyarkoni:
It wasn't on my near-term agenda to work on... are there specific things you want validated? File content-checking is probably out of scope at this point, at least from my side of things (I'll accept PRs, obviously).

tyarkoni:
I guess if the JSON schema is stable and added to the central validation repo, it would probably be fairly easy to add JSON validation using a third-party library (looks like `jsonschema` is the standard).

PR #1422, opened 2018-11-23
[ENH] Sentry event categorization propagation

chrisfilo:
This update builds on top of #1418 and adds fingerprint propagation for certain events. In other words, if say memory error was raised during a run all events raised for that run will be marked as memory errors. This should limit noise and group side effects with common causes.

### DIFF
--- a/fmriprep/cli/run.py
+++ b/fmriprep/cli/run.py
@@ -279,14 +279,10 @@ def main():
             if 'breadcrumbs' in event and isinstance(event['breadcrumbs'], list):
                 fingerprints_to_propagate = ['no-disk-space', 'memory-error', 'permission-denied',
                                              'keyboard-interrupt']
-                for bc in [b for b in event['breadcrumbs'] if 'message' in b]:
-                    break_outer = False
-                    for fp in fingerprints_to_propagate:
-                        if bc['message'] == fp:
-                            event['fingerprint'] = [fp]
-                            break_outer = True
-                            break
-                    if break_outer:
+                for bc in event['breadcrumbs']:
+                    msg = bc.get('message', 'empty-msg')
+                    if msg in fingerprints_to_propagate:
+                        event['fingerprint'] = [msg]
                         break
 
             return event
--- a/fmriprep/cli/run.py
+++ b/fmriprep/cli/run.py
@@ -275,8 +275,21 @@ def main():
                     return None
                 elif re.match("Node .+ failed to run on host .+", msg):
                     return None
-            else:
-                return event
+
+            if 'breadcrumbs' in event and isinstance(event['breadcrumbs'], list):
+                fingerprints_to_propagate = ['no-disk-space', 'memory-error', 'permission-denied',
+                                             'keyboard-interrupt']
+                for bc in [b for b in event['breadcrumbs'] if 'message' in b]:
+                    break_outer = False
+                    for fp in fingerprints_to_propagate:
+                        if bc['message'] == fp:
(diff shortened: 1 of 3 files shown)


fmriprep/cli/run.py
```diff
@@ -275,8 +275,21 @@ def before_send(event, hints):
                     return None
                 elif re.match("Node .+ failed to run on host .+", msg):
                     return None
-            else:
-                return event
+
+            if 'breadcrumbs' in event and isinstance(event['breadcrumbs'], list):
```
oesteban: I think this does the same but seems more readable.
```Python
            if 'breadcrumbs' in event and isinstance(event['breadcrumbs'], list):
                fingerprints_to_propagate = ['no-disk-space', 'memory-error', 'permission-denied',
                                             'keyboard-interrupt']
                for bc in event['breadcrumbs']:
                    msg = bc.get('message', 'empty-msg')
                    if msg in fingerprints_to_propagate:
                        event['fingerprint'] = [msg]
```

fmriprep/viz/reports.py
```diff
@@ -154,20 +154,46 @@ def index_error_dir(self, error_dir):
...
+                    # Group common events with pre specified fingerprints
+                    fingerprint_dict = {'permission-denied': ["PermissionError: [Errno 13] "
+                                                              "Permission denied"],
+                                        'memory-error': ["MemoryError", "Cannot allocate memory"],
+                                        'reconall-already-running': ["ERROR: it appears that "
+                                                                     "recon-all is already "
+                                                                     "running"],
+                                        'no-disk-space': ["OSError: [Errno 28] No space left on "
+                                                          "device", "[Errno 122] Disk quota "
+                                                                    "exceeded"],
```
oesteban: indentation

fmriprep/viz/reports.py
```diff
@@ -154,20 +154,46 @@ def index_error_dir(self, error_dir):
...
+                                issue_title = new_fingerprint
+                                break
+                        if fingerprint:
                             break
 
+                    message = issue_title + '\n\n'
+                    message += exception_text[-(8192-len(message)):]
+                    if fingerprint:
+                        self.sentry_sdk.add_breadcrumb(fingerprint, 'fatal')
+                    else:
```
oesteban: is it intended that fatal fingerprints do not seem to be stripped off of paths and words containing numbers?

chrisfilo: fatal fingerprints consist only of a single string (such as `sigkill`) and therefore  do not include any paths or numbers

fmriprep/viz/reports.py
chrisfilo: what's wrong with it? (mind that `"exceeeded" is part of the list item from the previous line)

fmriprep/cli/run.py
chrisfilo: nice!

fmriprep/viz/reports.py
oesteban: I'd swear that it should be aligned right after the opening bracket, with `"device"`. That is taking into account that "exceeded" is part of the previous line. However, this would be more readable and use the same number of lines with:

```Python
                                        'no-disk-space': [
                                            "OSError: [Errno 28] No space left on device",
                                            "[Errno 122] Disk quota exceeded"],
```

effigies:
@oesteban Are you happy with this?

oesteban:
Sure, let's go :)

PR #1443, opened 2018-12-12
FIX: Output NaN columns for CompCor on failure

effigies:

* Use `failure_mode` parameter added nipy/nipype#2819 to produce columns of `NaN`s when SVD fails to converge.


oesteban:
Should we add a warning to the reports when the column is all NaNs?

effigies:
I think that's probably a good idea, but haven't gotten around to it. I think a small warning box just below the BOLD masks figure would make the most sense.

Perhaps text like:

```
Warning: [at]CompCor components could not be estimated, due to a linear algebra error.
While not definitive, this may be an indication of a poor mask. Please inspect the (magenta|blue)
contours above to ensure that they are located in the white matter/CSF.
```

oesteban:
Ref poldracklab/niworkflows#268

effigies:
@oesteban If you could review the logic of the new patched interfaces, that's the most likely place to need fixes. I added a snippet that will only show up as a comment so we can verify this in the CircleCI artifacts.

fmriprep/interfaces/patches.py
```diff
@@ -12,49 +12,110 @@
...
+        components = np.loadtxt(outputs['components_file'], skiprows=1)
+        return np.isnan(components).all()
+
+    def _generate_report(self):
+        snippet = '<!-- {} completed without error -->'.format(self._header)
+        if self._check_nans():
+            snippet = '''\
+<p class="elem-desc">
+    Warning: {} components could not be estimated, due to a linear algebra error.
+    While not definitive, this may be an indication of a poor mask.
```
oesteban: Should we add a strict check here that `mask.sum() > 1` and raise if not?: https://github.com/nipy/nipype/blob/fbed1f3c432e80b1e199f93d17e4d8f1da232897/nipype/algorithms/confounds.py#L1155

fmriprep/interfaces/patches.py
```diff
@@ -12,49 +12,110 @@
...
-class RobustACompCor(nac.ACompCor):
-    """
-    Runs aCompCor several times if it suddenly fails with
-    https://github.com/poldracklab/fmriprep/issues/776
+class RetryCompCorInputSpecMixin(reporting.ReportCapableInputSpec):
+    out_report = File('report.html', usedefault=True, hash_files=False,
+                      desc='filename for warning HTML snippet')
+    # 'NaN` by default
+    failure_mode = traits.Enum(
+        'NaN', 'error',
```
oesteban: Is there any reason not to make error the default?

```suggestion
        'error', 'NaN',  # 'error' by default
```

fmriprep/interfaces/patches.py
```diff
@@ -12,49 +12,110 @@
...
 
 
-class RobustACompCor(nac.ACompCor):
-    """
-    Runs aCompCor several times if it suddenly fails with
-    https://github.com/poldracklab/fmriprep/issues/776
+class RetryCompCorInputSpecMixin(reporting.ReportCapableInputSpec):
+    out_report = File('report.html', usedefault=True, hash_files=False,
+                      desc='filename for warning HTML snippet')
+    # 'NaN` by default
```
oesteban: included in line 24.
```suggestion

```

effigies:
> I wonder whether after 10 failures we should call it an error (i.e. always raise in that case)

We already do that. The point is to give it a few chances to work and gracefully fail.

fmriprep/interfaces/patches.py
effigies: It doesn't really matter, since I'm already specifying `NaN` in the workflow. I'll just drop this trait altogether, as it defaults to error in the `CompCorInputSpec`.

fmriprep/interfaces/patches.py
effigies: We could do that, but why raise an error instead of issue a warning? The whole point of this is not to crash. People who care about CompCor can worry about why, but if they don't, the masks don't matter at all.

fmriprep/interfaces/patches.py
```diff
@@ -12,49 +12,104 @@
...
-        return runtime
+class RobustTCompCorInputSpec(RetryCompCorInputSpecMixin, nac.TCompCorInputSpec):
+    pass
+
+
+class RobustTCompCorOutputSpec(reporting.ReportCapableOutputSpec, nac.TCompCorOutputSpec):
+    pass
+
+
+class RobustTCompCor(RetryCompCorMixin, nac.ACompCor):
```
effigies:
```suggestion
class RobustTCompCor(RetryCompCorMixin, nac.TCompCor):
```

PR #1455, opened 2018-12-18
Revert "[MRG] FIX: Output NaN columns for CompCor on failure"

effigies:
Reverts poldracklab/fmriprep#1443


effigies:
I guess might as well make sure this passes before merging, but it's almost certainly this...

Issue #1319, opened 2018-10-12
This document format requires a nonempty <title> element. 


hasanasim:
I get the error
"This document format requires a nonempty <title> element. Please specify either 'title' or 'pagetitle' in the metadata."

How do I go about specifying a title exactly?

oesteban:
You can dismiss that warning. It is generated by the generator of the new citation boilerplate, in one of the conversion from markdown to html/latex.

I'll see whether we can just not show it.

oesteban:
I should have said ignore it. 👍

oesteban:
https://neurostars.org/t/help-with-fmriprep/2654/2?u=oesteban

oesteban:
Closed via #1464

PR #1464, opened 2019-01-08
[FIX] Avoid warning when generating boilerplate

oesteban:
This warning would still happen even if we do not print the boilerplate out (which was planned to be done).


oesteban:
Closes #1319

PR #1476, opened 2019-01-20
[FIX] Bad ``fsnative`` replacement in CIfTI workflow

### LINKED ISSUE #1439: gen_cifti: value: ['fsaverage6', 'fsaverage5', 'sub-50061']
Sentry Issue: [FMRIPREP-QC](https://sentry.io/poldracklab/fmriprep/issues/796836328/?referrer=github_integration)

```
gen_cifti: value: ['fsaverage6', 'fsaverage5', 'sub-50061']

traits.trait_errors.TraitError: Each element of the 'surface_target' trait of a DynamicTraitedSpec instance must be 'fsaverage5' or 'fsaverage6', but a value of 'sub-50061' <class 'str'> was specified.
Error setting node input:
Node: gen_cifti
input: surface_target
results_file: /lscratch/oesteban/work/fmriprep_wf/single_subject_50061_wf/func_preproc_task_pamret_wf/bold_surf_wf/targets/result_targets.pklz
value: ['fsaverage6', 'fsaverage5', 'sub-50061']
```

oesteban:
The cifti workflow was picking up target names from the wrong source node because fsnative was being replaced by the subject's id.

Fixes #1439

### DIFF
--- a/fmriprep/workflows/bold/base.py
+++ b/fmriprep/workflows/bold/base.py
@@ -805,7 +805,8 @@ data and volume-sampled data, were also generated.
             gen_cifti = pe.MapNode(GenerateCifti(), iterfield=["surface_target", "gifti_files"],
                                    name="gen_cifti")
             gen_cifti.inputs.TR = metadata.get("RepetitionTime")
-            gen_cifti.inputs.surface_target = surface_spaces
+            gen_cifti.inputs.surface_target = [s for s in surface_spaces
+                                               if s.startswith('fsaverage')]
 
             workflow.connect([
                 (bold_surf_wf, gen_cifti, [
--- a/fmriprep/workflows/bold/base.py
+++ b/fmriprep/workflows/bold/base.py
@@ -779,10 +779,11 @@ Non-gridded (surface) resamplings were performed using `mri_vol2surf`
             ])
 
     # SURFACES ##################################################################################
-    if freesurfer and any(space.startswith('fs') for space in output_spaces):
+    surface_spaces = [space for space in output_spaces if space.startswith('fs')]
+    if freesurfer and surface_spaces:
         LOGGER.log(25, 'Creating BOLD surface-sampling workflow.')
         bold_surf_wf = init_bold_surf_wf(mem_gb=mem_gb['resampled'],
-                                         output_spaces=output_spaces,
+                                         output_spaces=surface_spaces,
                                          medial_surface_nan=medial_surface_nan,
                                          name='bold_surf_wf')
         workflow.connect([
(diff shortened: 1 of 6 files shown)


oesteban:
`fs.SampleToSurface` wants the subject id instead of "fsnative", but that part of the surface workflow remains untouched.

However, `gen_cifti` (a `GenerateCifti` interface) only accepts "fsaverages". We were getting the names from a middle node in the surface resampling workflow, including subjects ids (after replacement of "fsnative"). With this PR the target surface spaces are set directly from the workflow creation arguments, instead of connecting a node. I think this is a lot easier to read and more reliable (there is one less node dependency).

fmriprep/workflows/bold/base.py
```diff
@@ -796,18 +797,19 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
         ])
 
         # CIFTI output
-        if cifti_output and 'template' in output_spaces:
+        if cifti_output and surface_spaces:
```
effigies: We still need there to be a `'template'` (or whatever it's changing to) output space for the volumetric portion, correct?

effigies:
Thanks for clarifying. I agree with your approach. Made a minor comment, but you may have already considered it in another PR.

fmriprep/workflows/bold/base.py
oesteban: What about we add the volumetric reference to the output_spaces list at parameter parsing (like we do with AROMA)?

For the current implementation, we would add `'template'`. With poldracklab/smriprep#37 this will probably mean adding something like `'fsLR'` instead.

WDYT?

effigies: I don't really understand the proposal.

oesteban: Sorry.

The reason we checked `'template' in output_spaces` here is that we needed the volumetric segmentation to extract the BOLD signal. The check ensures that the registration to MNI was performed and the aseg resampled into native space.

What I'm proposing is that we do as with AROMA: https://github.com/poldracklab/fmriprep/blob/84672ee6e171d84f06bb271dbc7bb46d2b20cc2f/fmriprep/cli/run.py#L575-L582 If the command line argument does not include `'template'` but `--cifti-output` is present, then we add `'template'` to the `output_spaces` right after argument parsing.

In the future (referenced PR to smriprep above) we will likely have a new `'fsLR`' option for templates, and we probably want to bind the `--cifti-output` option to the HCP template directly. But that is a different story I shouldn't have introduced here, probably.

effigies: Oh, sure.

PR #1482, opened 2019-01-25
[RF] Use anatomical workflows from sMRIPrep

oesteban:
Migrates structural processing to sMRIPrep


oesteban:
Working on fixing the docs build under `oesteban:docs/smriprep` - https://circleci.com/workflow-run/45b654ad-b4ae-4f20-86c9-b8bb9efadada

If that works I'll merge that branch into this and remove the WIP tag. Fingers crossed.

oesteban:
This is looking good: https://3032-53608443-gh.circle-artifacts.com/0/home/circleci/out/docs/index.html

oesteban:
Please do not merge before changing smriprep pin to 0.0.4

effigies:
I'll try to have a look ASAP... Doing FitLins/pybids work, though, so it might not be until Monday.

oesteban:
We have docs built on the merging branch: https://3036-53608443-gh.circle-artifacts.com/0/home/circleci/out/docs/index.html

I think there is no other pressing PR to fmriprep that may conflict, so we can totally wait for you to have more availability :+1:

Dockerfile
```diff
@@ -189,9 +202,9 @@ ARG BUILD_DATE
 ARG VCS_REF
 ARG VERSION
 LABEL org.label-schema.build-date=$BUILD_DATE \
-      org.label-schema.name="FMRIPREP" \
-      org.label-schema.description="FMRIPREP - robust fMRI preprocessing tool" \
-      org.label-schema.url="http://fmriprep.org" \
+      org.label-schema.name="fMRIPrep" \
+      org.label-schema.description="fMRIPrep - robust fMRI preprocessing tool" \
+      org.label-schema.url="http://fMRIPrep.org" \
```
effigies: Just aesthetically, I don't like caps in URLs. You can take this or leave it though.

```suggestion
      org.label-schema.url="http://fmriprep.org" \
```

Dockerfile
```diff
@@ -134,49 +149,47 @@ RUN conda install -y python=3.7.1 \
...
+    git config --global user.email "email@domain.com" && \
+    datalad install -r https://github.com/templateflow/templateflow.git
+RUN datalad get $TEMPLATEFLOW_HOME/tpl-MNI152NLin2009cAsym/*_T1w.nii.gz \
+                $TEMPLATEFLOW_HOME/tpl-MNI152NLin2009cAsym/*_desc-brain_mask.nii.gz \
+                $TEMPLATEFLOW_HOME/tpl-MNI152Lin/*_T1w.nii.gz \
+                $TEMPLATEFLOW_HOME/tpl-MNI152Lin/*_desc-brain_mask.nii.gz \
+                $TEMPLATEFLOW_HOME/tpl-OASIS30ANTs/*_T1w.nii.gz \
+                $TEMPLATEFLOW_HOME/tpl-OASIS30ANTs/tpl-OASIS30ANTs_res-01_desc-brain_mask.nii.gz \
+                $TEMPLATEFLOW_HOME/tpl-OASIS30ANTs/tpl-OASIS30ANTs_res-01_label-brain_probseg.nii.gz \
+                $TEMPLATEFLOW_HOME/tpl-OASIS30ANTs/tpl-OASIS30ANTs_res-01_desc-BrainCerebellumExtraction_mask.nii.gz
```
effigies: Just to make sure, is it still the case that if we *didn't* pre-fetch these, s/fMRIPrep would still download them as needed?

docs/environment.yml
```diff
@@ -29,4 +29,5 @@ dependencies:
     - nitime
     - nilearn
     - niworkflows>=0.5.4,<0.6
+    - smriprep>=0.0.4,<1.0.0
```
effigies: I think this will ensure that we get everything in the 0.0.x series and no pre-releases of 0.1.x or later:

```suggestion
    - smriprep>=0.0.4,<0.1.0a0
```

Or are you thinking sMRIPrep 1.0 is coming very soon?

docs/workflows.rst
```diff
@@ -68,9 +68,9 @@ T1w/T2w preprocessing
                               reportlets_dir='.',
                               output_dir='.',
                               template='MNI152NLin2009cAsym',
-                              output_spaces=['T1w', 'fsnative',
-                                             'template', 'fsaverage5'],
-                              skull_strip_template='OASIS',
+                              fs_spaces=['T1w', 'fsnative',
```
effigies: I assume `fs_spaces` means FreeSurfer? `T1w` and `template` don't seem appropriate entries...

fmriprep/__about__.py
```diff
@@ -90,15 +90,17 @@
     'nilearn',
     'nipype>=1.1.6',
     'nitime',
-    'niworkflows>=0.5.2.post7,<0.5.3',
+    'niworkflows>=0.6.1,<0.7',
```
effigies:
```suggestion
    'niworkflows>=0.6.1,<0.7.0a0',
```

fmriprep/__about__.py
```diff
@@ -90,15 +90,17 @@
...
     'numpy',
     'pandas',
     'psutil>=5.4',
     'pybids==0.6.5',
     'pyyaml',
     'scikit-image',
+    'smriprep>=0.0.4,<1.0.0',
```
effigies:
```suggestion
    'smriprep>=0.0.4,<0.1.0a0',
```

requirements.txt
```diff
@@ -1,3 +1,5 @@
 grabbit==0.2.3
 pybids==0.6.5
-niworkflows>=0.5.2.post7,<0.5.3
+niworkflows>=0.6.1,<0.7
```
effigies:
```suggestion
niworkflows>=0.6.1,<0.7.0a0
```

requirements.txt
```diff
@@ -1,3 +1,5 @@
...
+smriprep>=0.0.4,<1.0.0
```
effigies:
```suggestion
smriprep>=0.0.4,<0.1.0a0
```

Dockerfile
oesteban: Yes.

Although there are two known issues:
- Race conditions - when running in parallel (I think this happened in niworkflows) datalad chokes. This does not refer to datalad's capabilities of parallel downloads, this is when you run datalad.get on the same file twice. This was one reason to pre-cache everything we probably need. But running, e.g., with the skull-stripping with NKI as reference (so not cached), would work (download, and then keep cached).
- This pre-fetching forced me to stop running the containers as user. People who want to run them as user will need to define a `$TEMPLATEFLOW_HOME` variable. I haven't actually tried this, and would also require revising permissions on `/templateflow` to allow reuse by the user.

But yeah, the central functionality is working well (or better said, datalad works well)

docs/environment.yml
oesteban: I was actually thinking whether sMRIPrep should keep up versions with fMRIPrep - when we are sure about this, push both to 1.3.0.

docs/workflows.rst
oesteban: I renamed `output_spaces` in sMRIPrep because I still believe we need to handle that differently, but at this point it is just a renaming. I could roll it back in sMRIPrep, but honestly I don't think it is worth the pain.

oesteban:
Only the `fs_spaces` if it is so annoying as not to merge.

oesteban:
I've opened poldracklab/smriprep#50 to make sure we don't forget about that.

Issue #1029, opened 2018-03-09
Number of voxels in phase-encoding direction is not always equal to number of phase-encoding lines on Philips systems


Gilles86:
Heya,

I was looking into the code that calculates the effective echo spacing for Philips scans. This uses the number of voxels in the phase-encoding dimension as a proxy for (EPI train length * acceleration factor). 

https://github.com/poldracklab/fmriprep/blob/260872273a1f4ef02de2cae20dd7d6948b531c4b/fmriprep/interfaces/fmap.py#L327

However, especially on Philips systems, the "Acquisition matrix size" (corresponding to actual number of k-space lines), and the "Reconstruction matrix size" can sometimes be different, because Philips systems automatically select a reconstruction matrix size that is easy to reconstruct using the fast fourier transform (FFT; powers of 2). This is something we also sometimes do deliberately in our data because it might somewhat alleviate partial voluming.

Would it be an idea to first try to use the "epi_factor"-field (standard in Philips parrecs) to calculate the number of k-space lines before using the nifti dimensions?

(this is also reccomended here:
http://dbic.dartmouth.edu/pipermail/mrusers/attachments/20141112/eb1d20e6/attachment.pdf )

Gilles86:
Yeah apparently they do not "trust" this field?
https://github.com/rordenlab/dcm2niix/blob/8fc0c514a7595d4afb8fac09091f5c54ecbf28ee/console/nii_dicom_batch.cpp#L829

In any case, if you have a different reconstruction vs acquisition matrix size, the DICOM will already have a different resolution from what was acquired. This should somehow be encoded in BIDS metadata I would argue.

oesteban:
Hi @Gilles86, would you like to submit a PR from your sketch?

oesteban:
I'll re-up this. @Gilles86, any possibility that you would send us a PR? That'd be really sweet - otherwise, I would try to cherry-pick your commits to maintain your authorship and that is prone to get pretty hairy.

Gilles86:
I made a [pull request](https://github.com/poldracklab/fmriprep/pull/1489).

I was wondering: to what extent have you tested B0-map correction on open datasets? It seems that most of the stuff on openfmri does not provide B0 maps.

[...]

PR #1489, opened 2019-01-30
[ENH] More advanced heuristics for finding out echo train length

Gilles86:
This pull request slightly changes the heuristics for finding the echo train length, taking into account that the length of the phase-encoding dimension in a nifti-file (or DICOM) is not necessarily the same as that of the acquired data. There can be a difference between the acquired and reconstructed matrix.

See also issue https://github.com/poldracklab/fmriprep/issues/1029

None

### DIFF
--- a/fmriprep/interfaces/fmap.py
+++ b/fmriprep/interfaces/fmap.py
@@ -322,8 +322,19 @@ def get_ees(in_meta, in_file=None):
     if ees is not None:
         return ees
 
-    # All other cases require the parallel acc and npe (N vox in PE dir)
-    etl = _get_etl(in_meta, in_file)
+    acc = float(in_meta.get('ParallelReductionFactorInPlane', 1.0))
+    npe = nb.load(in_file).shape[_get_pe_index(in_meta)]
+    etl = npe // acc
+
+    # Prefer metadata over size of the nifti file (Philips data can have
+    # different number of acquired versus reconstructed k-space lines
+    # see https://github.com/poldracklab/fmriprep/issues/1029 )
+
+    if 'EPIFactor' in in_meta:
+        etl = in_meta['EPIFactor'] + 1  # ETL is EPIFactor + 1 for Philips
+
+    # Not part of the BIDS spec now
+    etl = int(in_meta.get('EchoTrainLength', etl))
 
     # Use case 2: TRT is defined
     trt = in_meta.get('TotalReadoutTime', None)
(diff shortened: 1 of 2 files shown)


fmriprep/interfaces/fmap.py
```diff
@@ -322,11 +322,20 @@ def get_ees(in_meta, in_file=None):
...
     etl = npe // acc
 
+    # Prefer metadata over size of the nifti file (Philips data can have
+    # different number of acquired versus reconstructed k-space lines
+    # see https://github.com/poldracklab/fmriprep/issues/1029 )
+
+    if 'EPIFactor' in in_meta:
+        etl = in_meta['EPIFactor'] + 1  # ETL is EPIFactor + 1 for Philips
+
+    # Not part of the BIDS spec now
```
oesteban: is this key under revision? I can remember suggesting it at some point, but I don't know where it stands at this point in time.

fmriprep/interfaces/fmap.py
```diff
@@ -322,11 +322,20 @@ def get_ees(in_meta, in_file=None):
...
     acc = float(in_meta.get('ParallelReductionFactorInPlane', 1.0))
     npe = nb.load(in_file).shape[_get_pe_index(in_meta)]
     etl = npe // acc
 
+    # Prefer metadata over size of the nifti file (Philips data can have
+    # different number of acquired versus reconstructed k-space lines
+    # see https://github.com/poldracklab/fmriprep/issues/1029 )
+
+    if 'EPIFactor' in in_meta:
+        etl = in_meta['EPIFactor'] + 1  # ETL is EPIFactor + 1 for Philips
```
oesteban: What would we expect for etl here, if the scanner is not Phillips?

fmriprep/interfaces/fmap.py
oesteban: And another question: does `EchoTrainLength` make sense if no `EPIFactor` is defined? (i.e., should next line be under this if branch?)

oesteban:
Hi @Gilles86, sorry for pestering you with this PR, but I'm doubtful to hit merge before having some explicit answer to the questions above. With that, we'll be able to update the documentation accordingly and resolve related issues more promptly. Thanks a lot for your contribution!

Gilles86:
Hey @oesteban,

After discussing this a bit more with some colleagues and reading up, I would like to revoke the pull request for now. I think I misunderstood. The `effective echo spacing` is the 'virtual' amount of time it takes between k-space lines in the *reconstructed* image (and not the acquired one). So we should take into account the number of _reconstructed_ voxels rather than the number of _acquired_ k-space lines.
See also the BIDS specification ("RECOMMENDED. The "effective" sampling interval, ..., defined based on the size of the _reconstructed_ image in the phase direction.") and this thread on the [FSL mailing list](https://www.jiscmail.ac.uk/cgi-bin/webadmin?A2=ind1403&L=fsl&D=0&P=340901).

This suggests that the current code works well, even though in a way it 'overestimates' the number of phase-encoding lines, because this is what FUGUE expects you to do.

I think that the current fmriprep code probably behaves like it should and my pull request would break things. I find it very hard to understand what exactly goes on though and whether FUGUE indeed behaves correctly when you upsample the image by factor `x` in the phase-encodign direction, but then also reduce the effective echo spacing by the same factor `x`. 

What is especially strange to me is that the TotalReadOutTime is defined 'effectively' as follows: "... specified in seconds, that would have generated data with the given level of distortion. It is NOT the actual, physical duration of the readout train". 

In that case, I think this line is off, since it does not take into account the acceleration.
https://github.com/poldracklab/fmriprep/blob/a255683b2f2a0d643accfb13b341cbd9311cda55/fmriprep/interfaces/fmap.py#L431

Is there anyone else in the fmriprpe/BIDS community who is a true expert on this?

oesteban:
Thanks for all these details. I've opened a new issue to note your request for information and double check that line you believe is off. Please reopen if you feel we'd need to take further actions.

PR #1496, opened 2019-02-06
[MAINT] Update to keep up with poldracklab/niworkflows#299

oesteban:
And poldracklab/smriprep#51 as well.


fmriprep/cli/run.py
```diff
@@ -606,8 +607,9 @@ def build_workflow(opts, retval):
 
     # First check that bids_dir looks like a BIDS folder
     bids_dir = os.path.abspath(opts.bids_dir)
+    layout = BIDSLayout(bids_dir, validate=False)
```
effigies: Just out of curiosity, why are we using `validate=False`?

fmriprep/workflows/fieldmap/phdiff.py
```diff
@@ -73,7 +73,8 @@ def _pick1st(inlist):
         return inlist[0]
 
     # Read phasediff echo times
-    meta = pe.Node(ReadSidecarJSON(), name='meta', mem_gb=0.01, run_without_submitting=True)
+    meta = pe.Node(ReadSidecarJSON(bids_validate=False), name='meta',
+                   mem_gb=0.01, run_without_submitting=True)
```
effigies: For nodes that construct BIDS layouts, we may want to consider whether to use `run_without_submitting`. Large datasets can take a long time, and `run_without_submitting` means we block the master thread.

Maybe we can time the first creation, and if it's >2sec (or something) go ahead and submit?

fmriprep/cli/run.py
oesteban: Assuming that `--skip-bids-validator` was not set. We could actually tail it to that and force going through some validation.

oesteban: ping @chrisfilo 

What about:
```Python
    layout = BIDSLayout(bids_dir, validate=not opts.skip_bids_validator)
```

fmriprep/workflows/fieldmap/phdiff.py
oesteban: Let's submit it, I feel like timing the first creation is somewhat splitting hairs.

fmriprep/cli/run.py
chrisfilo: I think `validate=False` is the way to go. `bids-validator` would provide better validation and if it's turned off we should not validate anyway.

oesteban: Oh, okay

PR #1514, opened 2019-02-16
FIX: Make sure ``--cifti-output`` requires at least one of ``fsaverage{5,6}``

oesteban:
Fixes poldracklab/niworkflows#309

### DIFF
--- a/fmriprep/cli/run.py
+++ b/fmriprep/cli/run.py
@@ -580,13 +580,6 @@ def build_workflow(opts, retval):
     if not opts.run_reconall:
         output_spaces = [item for item in output_spaces if not item.startswith('fs')]
 
-    # Map fsaverage to fsaverage6
-    if 'fsaverage' in output_spaces:
-        logger.warning('Selected "fsaverage" output space has been mapped to "fsaverage6"')
-        output_spaces.remove('fsaverage')
-        if 'fsaverage6' not in output_spaces:
-            output_spaces = sorted(output_spaces + ['fsaverage6'])
-
     # Validity of some inputs
     # ERROR check if use_aroma was specified, but the correct template was not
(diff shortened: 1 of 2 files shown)


fmriprep/cli/run.py
```diff
@@ -573,7 +573,19 @@ def build_workflow(opts, retval):
...
+    # Reduce to unique space identifiers
+    output_spaces = sorted(set(opts.output_space))
+
+    # If FS is not run, drop all fs* output spaces
+    if not opts.run_reconall:
+        output_spaces = [item for item in output_spaces if not item.startswith('fs')]
+
+    # Map fsaverage to fsaverage6
+    if 'fsaverage' in output_spaces:
+        logger.warning('Selected "fsaverage" output space has been mapped to "fsaverage6"')
```
effigies: I don't think we should override user selected output spaces.

oesteban: fsaverage is an alias for fsaverage6, isn't that correct?

oesteban: The reason for this is that our cifti interface fails if it is passed "fsaverage". We could do the mapping silently as FS does, but I prefer being explicit.

effigies: No. fsaverage6 is a deformed order 6 icosahedron. fsaverage is order 7.

effigies: The reason we require fsaverage6 for cifti is because fsaverage would be huge.

oesteban: Gotcha

oesteban:
Sorry for the history rewrite, I forgot that ``--no-fs-reconall`` and ``--cifti-output`` are mutually exclusive. After all that mess, the corresponding tests are here: https://circleci.com/workflow-run/c6b97b23-e35c-4d63-a9a7-eb7807299c7f

PR #1529, opened 2019-03-07
[FIX] Avoid using ``$HOME`` for storing templates

chrisfilo:


chrisfilo:
Confirmed to solve the issue for a user https://neurostars.org/t/filenotfounderror-errno-2-no-such-file-or-directory-home-fmriprep-cache/3770/8. I recommend merging and publishing new release on Monday.

PR #1559, opened 2019-03-29
[ENH] add --skip-vols-num option

jdkent:
There has been a request to manually specify the number of volumes to classify as "non-steady-state" by the user to enforce consistency between runs/sessions/participants. This was referenced most recently in a [neurostars conversation](https://neurostars.org/t/fmriprep-nonsteady-states-added-back-into-aroma-non-aggressively-denoised-outputs/3945/6?u=jdkent), although I recall conversations like this popping up before I could not find links to those conversations.

- [x] add option `--skip-vols-num` to fmriprep
- [x] give warning if non-steady-state number calculated algorithmically is higher than what was manually specified. (maybe?)
- [x] document the option more clearly

The usage documentation since I added a new option


effigies:
I think we discussed this before, and my preferred option would be to specify in the BIDS metadata the number of dummy scans in a BOLD series. Probably around [Modality-specific files - Magnetic Resonance Imaging - Task imaging data - Other recommended metadata](https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/01-magnetic-resonance-imaging-data.html#other-recommended-metadata).

jdkent:
ah, that conversation is coming back to me (although I still can't remember where), and I agree, having that information encoded in the metadata would be a better solution.

effigies:
I wouldn't rewrite this just yet. I'd start a issue over on https://github.com/bids-standard/bids-specification/ to see if there's any chance of getting this piece of metadata added.

effigies:
Okay, it looks like I'm overruled over in bids-standard/bids-specification#195, or at least that there's no enthusiasm on the metadata side to match the pushback.

So the main questions then are:

1) What should the parameter be called? I'd suggest something like `--dummy-scans N`.
2) Reporting. I think a new line in the functional summary:

https://github.com/poldracklab/fmriprep/blob/4470ffb5125b665fbfe57d08f4b2f790b4c581b9/fmriprep/interfaces/reports.py#L35-L45

Here I would have the following cases:

| `--dummy-scans` value | Heuristic | Text |
|-|-|-|
| `0`/unset | N | "Non-steady-state volumes: N detected" |
| N | N | "Non-steady-state volumes: N (Confirmed: N automatically detected)" |
| M | N | "Non-steady-state volumes: M (Warning: N automatically detected)" |

Or similar.

@oesteban Any opinions on reporting?

oesteban:
I think your proposal is the best we can do, after the pushback from the spec :+1:

jdkent:
[ds210 html output](https://567-92465059-gh.circle-artifacts.com/0/tmp/ds210/derivatives/fmriprep/sub-02.html#Functional), here is some example output of the new reported line.

oesteban:
I'll try to push to get this merged next week. Sorry for the delays.

jdkent:
Thanks for the update. No need to apologize, I have a hard enough time keeping up with my own projects. There is more to life than pull requests!

fmriprep/workflows/bold/util.py
```diff
@@ -34,7 +34,8 @@
 
 
 def init_bold_reference_wf(omp_nthreads, bold_file=None, pre_mask=False,
-                           name='bold_reference_wf', gen_report=False):
+                           name='bold_reference_wf', gen_report=False,
+                           dummy_scans=None):
```
oesteban: Missing argument in docstring. BTW, please remove the `enhance_t2` which is outdated.

fmriprep/workflows/bold/util.py
```diff
@@ -412,3 +429,34 @@ def init_skullstrip_bold_wf(name='skullstrip_bold_wf'):
...
+    if dummy_scans:
+        if algo_dummy_scans > dummy_scans:
+            LOGGER.warning("The non steady state algorithm detected "
+                           "more volumes than what was manually specified")
+
+        skip_vols_num = dummy_scans
+    else:
+        skip_vols_num = algo_dummy_scans
+
+    return dummy_scans, algo_dummy_scans, skip_vols_num
```
oesteban: I think returning unmodified values can be prone to error.
```suggestion
    return skip_vols_num
```

jdkent: good to know, my next set of commits should work around this.

fmriprep/workflows/bold/util.py
jdkent: Thanks! I will address this in my next round of commits

jdkent:
I should have something up by the end of the day, particular attention should be paid to the conflicts I merged to make sure I didn't accidentally remove any functionality.

docs/workflows.rst
```diff
@@ -57,6 +57,15 @@ is presented below:
         use_aroma=False,
         use_bbr=True,
         use_syn=True,
+<<<<<<< HEAD
```
effigies: Merge artifacts.

fmriprep/interfaces/reports.py
```diff
@@ -184,9 +187,26 @@ def _generate_segment(self):
         if isdefined(self.inputs.confounds_file):
             with open(self.inputs.confounds_file) as cfh:
                 conflist = cfh.readline().strip('\n').strip()
+
+        dummy_scan_tmp = "{n_dum}"
+        if self.inputs.dummy_scans:
```
oesteban: Can we flatten these nested choices?
```Python
        # Default case: dummy_scans was not manually set
        dummy_scan_msg = dummy_scan_tmp.format(n_dum=self.inputs.algo_dummy_scans)
        if self.inputs.dummy_scans == self.inputs.algo_dummy_scans:  # No need to test for None
            dummy_scan_msg = (
                ' '.join([dummy_scan_tmp, "(Confirmed: {n_alg} automatically detected)"])
                .format(n_dum=self.inputs.dummy_scans, n_alg=self.inputs.algo_dummy_scans)
            )
        elif self.inputs.dummy_scans:  # Not need to test for manual != algo
            dummy_scan_msg = (
                ' '.join([dummy_scan_tmp, "(Warning: {n_alg} automatically detected)"])
                .format(n_dum=self.inputs.dummy_scans, n_alg=self.inputs.algo_dummy_scans)
            )
```

fmriprep/workflows/bold/base.py
```diff
@@ -453,6 +458,7 @@ def init_func_preproc_wf(
 
     # Generate a tentative boldref
     bold_reference_wf = init_bold_reference_wf(omp_nthreads=omp_nthreads)
+    bold_reference_wf.inputs.inputnode.dummy_scans = dummy_scans
```
oesteban: I like this much better :+1:

fmriprep/workflows/bold/util.py
```diff
@@ -413,3 +426,29 @@ def init_skullstrip_bold_wf(name='skullstrip_bold_wf'):
...
+    **Returns**
+    dummy_scans : int or None
+        number of volumes user specified to skip
+    algo_dummy_scans : int
+        number of volumes algorithm detected to skip
+    skip_vols_num : int
+        number of volumes to skip
+    """
+
+    if dummy_scans:
```
oesteban: Maybe:
```Python
    if dummy_scans:
        return dummy_scans
    return algo_dummy_scans
```

oesteban: Or
```Python
    return dummy_scans if dummy_scans else algo_dummy_scans
```

oesteban: or:
```Python
    return dummy_scans or algo_dummy_scans
```

fmriprep/workflows/bold/util.py
```diff
@@ -413,3 +426,29 @@ def init_skullstrip_bold_wf(name='skullstrip_bold_wf'):
...
+    """
+    **Parameters**
+
+    algo_dummy_scans : int
+        number of volumes to skip determined by an algorithm
+    dummy_scans : int or None
+        number of volumes to skip determined by the user
+
+    **Returns**
+    dummy_scans : int or None
```
oesteban: Please remove dummy_scans and algo_dummy_scans from the Returns list

fmriprep/workflows/bold/util.py
```diff
@@ -118,16 +123,24 @@ def init_bold_reference_wf(omp_nthreads, bold_file=None, pre_mask=False,
     enhance_and_skullstrip_bold_wf = init_enhance_and_skullstrip_bold_wf(
         omp_nthreads=omp_nthreads, pre_mask=pre_mask)
 
+    calc_dummy_scans = pe.Node(niu.Function(function=_pass_dummy_scans,
+                                            output_names=['skip_vols_num']),
+                               name='calc_dummy_scans',
```
oesteban:
```suggestion
                               name='calc_dummy_scans',
                               run_without_submitting=True,
```

oesteban:
@jdkent do you think this could be ready for next release? We would like to release ASAP.

jdkent:
@oesteban, I am working on edits now, should be done shortly.

jdkent:
I want to say thank you @oesteban and @effigies. Every time I contribute I feel like I learn more about python and how to improve my own code, which wouldn't be possible without your guidance, so thanks! 

:smile: :smile:

Issue #1348, opened 2018-10-25
`bold_mni_ref` connection failing when specifying --output-space T1w


effigies:
CLI:

```
RUNNING: docker run --rm -it \
    -v /data/cjmarkie/Projects/crn/fmriprep/fmriprep:/usr/local/miniconda/lib/python3.6/site-packages/fmriprep:ro \
    -v /data/cjmarkie/Projects/crn/niworkflows/niworkflows:/usr/local/miniconda/lib/python3.6/site-packages/niworkflows:ro \
    -v /opt/freesurfer/license.txt:/opt/freesurfer/license.txt:ro \
    -v /data/bids/ds000114:/data:ro \
    -v /data/out/ds000114/derivatives:/out \
    -v /data/scratch/ds000114:/scratch \
    poldracklab/fmriprep:1.1.8 /data /out participant \
    --output-space T1w --participant-label 07 -t fingerfootlips --notrack -w /scratch
```

Using poldracklab/niworkflows/master and #1343, and a fresh scratch directory.

```
Traceback (most recent call last):
  File "/usr/local/miniconda/bin/fmriprep", line 11, in <module>
    sys.exit(main())
  File "/usr/local/miniconda/lib/python3.6/site-packages/fmriprep/cli/run.py", line 343, in main
    fmriprep_wf.run(**plugin_settings)
  File "/usr/local/miniconda/lib/python3.6/site-packages/nipype/pipeline/engine/workflows.py", line 585, in run
    execgraph = generate_expanded_graph(deepcopy(flatgraph))
  File "/usr/local/miniconda/lib/python3.6/site-packages/nipype/pipeline/engine/utils.py", line 995, in generate_expanded_graph
    graph_in = _remove_nonjoin_identity_nodes(graph_in, keep_iterables=True)
  File "/usr/local/miniconda/lib/python3.6/site-packages/nipype/pipeline/engine/utils.py", line 880, in _remove_nonjoin_identity_nodes
    _remove_identity_node(graph, node)
  File "/usr/local/miniconda/lib/python3.6/site-packages/nipype/pipeline/engine/utils.py", line 905, in _remove_identity_node
    portinputs)
  File "/usr/local/miniconda/lib/python3.6/site-packages/nipype/pipeline/engine/utils.py", line 980, in _propagate_internal_output
    value = getattr(node.inputs, field)
AttributeError: 'DynamicTraitedSpec' object has no attribute 'bold_mni_ref'
```

[...]

PR #1560, opened 2019-03-30
FIX: Not having ``template`` as one ``--output-space`` crashes fMRIPrep

oesteban:
This PR checks that ``template`` is in the output spaces list, skipping the connection to datasinks of workflows resampling the BOLD in MNI.

In particular, we were getting (see #1348):
```
AttributeError: 'DynamicTraitedSpec' object has no attribute 'bold_mni_ref'
```

When forwarding the IdentityInterface of the derivatives data sink because it wasn't included within the input fields.


oesteban:
Tested locally without `template` amongst `--output-spaces`

Issue #1526, opened 2019-03-01
Spit a big WARNING when version does not match that of an existing ``dataset_description.json`` output file


oesteban:

oesteban:
Hey @adelavega what is the status of this?

adelavega:
Never got around to it.

PR #1567, opened 2019-04-03
ENH: remove all prior existing citation files first

yarikoptic:
- makes it friendly to the cases where all output is under git-annex.
  .write_text() just opens a file in "w" mode so it would fail on
  annex protected file.

- if previous run did succeed generating e.g. .html file, if subsequent run
  fails, updated version of .html would not be generated and provided
  CITATION files could get out of sync.

Removing all  relevant `CITATION*` files prior generation resolves both of the issues.

No pressure -- if you don't like it just close it.

Comes not tested whatsoever ;-) (are there unittests to run for it?)

### DIFF
--- a/fmriprep/cli/run.py
+++ b/fmriprep/cli/run.py
@@ -786,7 +786,18 @@ def build_workflow(opts, retval):
     boilerplate = retval['workflow'].visit_desc()
 
     if boilerplate:
-        (logs_path / 'CITATION.md').write_text(boilerplate)
+        # To please git-annex users and also to guarantee consistency
+        # among different renderings of the same file, first remove any
+        # existing one
+        citation_files = {
+            ext: str(logs_path / 'CITATION.' + ext)
+            for ext in ('bib', 'tex', 'md', 'html')
+        }
+        for citation_file in citation_files.values():
+            if op.lexists(citation_file):
+                citation_file.unlink()
+
+        Path(citation_files['md']).write_text(boilerplate)
         logger.log(25, 'Works derived from this fMRIPrep execution should '
                    'include the following boilerplate:\n\n%s', boilerplate)
 
@@ -795,8 +806,8 @@ def build_workflow(opts, retval):
                pkgrf('fmriprep', 'data/boilerplate.bib'),
                '--filter', 'pandoc-citeproc',
                '--metadata', 'pagetitle="fMRIPrep citation boilerplate"',
-               str(logs_path / 'CITATION.md'),
-               '-o', str(logs_path / 'CITATION.html')]
+               citation_files['md'],
+               '-o', citation_files['html']]
         try:
             check_call(cmd, timeout=10)
         except (FileNotFoundError, CalledProcessError, TimeoutExpired):
@@ -806,8 +817,8 @@ def build_workflow(opts, retval):
         # Generate LaTex file resolving citations
         cmd = ['pandoc', '-s', '--bibliography',
                pkgrf('fmriprep', 'data/boilerplate.bib'),
-               '--natbib', str(logs_path / 'CITATION.md'),
-               '-o', str(logs_path / 'CITATION.tex')]
+               '--natbib', citation_files['md'],
+               '-o', citation_files['tex']]
         try:
             check_call(cmd, timeout=10)
         except (FileNotFoundError, CalledProcessError, TimeoutExpired):
@@ -815,7 +826,7 @@ def build_workflow(opts, retval):
                            ' '.join(cmd))
         else:
             copyfile(pkgrf('fmriprep', 'data/boilerplate.bib'),
-                     (logs_path / 'CITATION.bib'))
+                     citation_files['bib'])
 
     return retval
 
--- a/fmriprep/cli/run.py
+++ b/fmriprep/cli/run.py
@@ -790,14 +790,14 @@ def build_workflow(opts, retval):
         # among different renderings of the same file, first remove any
         # existing one
         citation_files = {
-            ext: str(logs_path / 'CITATION.' + ext)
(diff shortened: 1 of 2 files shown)


fmriprep/cli/run.py
```diff
@@ -786,7 +786,18 @@ def build_workflow(opts, retval):
     boilerplate = retval['workflow'].visit_desc()
 
     if boilerplate:
-        (logs_path / 'CITATION.md').write_text(boilerplate)
+        # To please git-annex users and also to guarantee consistency
+        # among different renderings of the same file, first remove any
+        # existing one
+        citation_files = {
+            ext: str(logs_path / 'CITATION.' + ext)
```
oesteban: Ordering of operations, better keep pathlib.Path

```suggestion
            ext: logs_path / ('CITATION.%s' % ext)
```

fmriprep/cli/run.py
```diff
@@ -786,7 +786,18 @@ def build_workflow(opts, retval):
...
+            for ext in ('bib', 'tex', 'md', 'html')
+        }
+        for citation_file in citation_files.values():
+            if op.lexists(citation_file):
```
oesteban:
```suggestion
            if citation_file.exists() or citation_file.is_symlink():
```

fmriprep/cli/run.py
```diff
@@ -786,7 +786,18 @@ def build_workflow(opts, retval):
...
+                citation_file.unlink()
```
oesteban: This would not work if `citation_file` is a string (should be addressed by my above comment)

fmriprep/cli/run.py
```diff
@@ -786,7 +786,18 @@ def build_workflow(opts, retval):
...
+
+        Path(citation_files['md']).write_text(boilerplate)
```
oesteban:
```suggestion
        citation_files['md'].write_text(boilerplate)
```

fmriprep/cli/run.py
```diff
@@ -795,8 +806,8 @@ def build_workflow(opts, retval):
                pkgrf('fmriprep', 'data/boilerplate.bib'),
                '--filter', 'pandoc-citeproc',
                '--metadata', 'pagetitle="fMRIPrep citation boilerplate"',
-               str(logs_path / 'CITATION.md'),
-               '-o', str(logs_path / 'CITATION.html')]
+               citation_files['md'],
```
oesteban:
```suggestion
               str(citation_files['md']),
```

fmriprep/cli/run.py
```diff
@@ -795,8 +806,8 @@ def build_workflow(opts, retval):
...
+               '-o', citation_files['html']]
```
oesteban:
```suggestion
               '-o', str(citation_files['html'])]
```

fmriprep/cli/run.py
```diff
@@ -806,16 +817,16 @@ def build_workflow(opts, retval):
         # Generate LaTex file resolving citations
         cmd = ['pandoc', '-s', '--bibliography',
                pkgrf('fmriprep', 'data/boilerplate.bib'),
-               '--natbib', str(logs_path / 'CITATION.md'),
-               '-o', str(logs_path / 'CITATION.tex')]
+               '--natbib', citation_files['md'],
```
oesteban:
```suggestion
               '--natbib', str(citation_files['md']),
```

fmriprep/cli/run.py
```diff
@@ -806,16 +817,16 @@ def build_workflow(opts, retval):
...
+               '-o', citation_files['tex']]
```
oesteban:
```suggestion
               '-o', str(citation_files['tex'])]
```

fmriprep/cli/run.py
yarikoptic: rright... first I followed the scheme of using pathlib and then felt that this function seems to not really need it - majority of "operations" is just stringification of pathlib objects and then a single `write_text`... so I regressed to using strings.  So is this not desired and fmriprep in general should prefer pathlib over regular strings for clarity/uniformity?

fmriprep/cli/run.py
yarikoptic: why bother with two instead of a single one?

oesteban: I'm trying to get rid of all `os.path` calls. Additionally, I think this is more explicit - I'm sure I'll read `op.exists` instead of `op.lexists` in just two months from now.

fmriprep/cli/run.py
oesteban: I generally prefer pathlib because it is pretty neat. Some functions expect `Path`s down the line. If you don't mind, please go ahead with Path.

yarikoptic: done. let's see if tests pass alright ;)

oesteban:
@yarikoptic, do you want to add yourself to the `.zenodo.json` file?

yarikoptic:
done... wasn't sure where to place myself so 38cfd773470b27c002d6ad4ba89e8d905840aee9

oesteban:
That is correct, will merge in soon. Thanks a lot!

oesteban:
BTW, this made me think whether we should also delete other top-level files (e.g. ``pipeline_description.json``, and this subject's html report and sub-XYZ folder)

yarikoptic:
well, if you keep thinking then may be you will even converge on my thinking that those files should not be just removed but loaded/compared to and if they are different - at least a warning or info message should be displayed like "you are now using different preprocessing settings from what you have used before - you might arrive to an incongruent collection of results (may be across or even within subjects!)" ;-)  or that is not possible? ;-)

oesteban:
Well, that would apply to the boilerplate too, since it is dynamic and reflects the methodological choices. 
(related to #1526).

Issue #1579, opened 2019-04-10
Error with N4BiasFieldCorrection


IanEisenberg:
Silent error with N4BiasFieldCorrection. I reran with the "verbose" flag and found that the input images do not occupy the same physical space. Same  as #1463, but potentially still a problem for sbref files. Besides fixing the error, I suggest adding the verbose flag to the N4BiasFieldCorrection command.

```
singularity exec poldracklab_fmriprep_1.3.1-2019-03-06-9db26dbbab52.simg N4BiasFieldCorrection -d 3 --input-image /scratch/users/ieisenbe/fmriprep_wf/single_subject_s130_wf/func_preproc_ses_1_task_motorSelectiveStop_run_1_wf/bold_reference_wf/validate_ref/sub-s130_ses-1_task-motorSelectiveStop_run-1_sbref_valid.nii.gz --mask-image /scratch/users/ieisenbe/fmriprep_wf/single_subject_s130_wf/func_preproc_ses_1_task_motorSelectiveStop_run_1_wf/bold_reference_wf/enhance_and_skullstrip_bold_wf/check_hdr/tpl-MNI152NLin2009cAsym_res-02_desc-brain_mask_trans_dil_hdr.nii.gz --output sub-s130_ses-1_task-motorSelectiveStop_run-1_sbref_valid_corrected.nii.gz -v
```
```
Running N4 for 3-dimensional images.

Exception caught: 
itk::ExceptionObject (0x203c6e0)
Location: "unknown" 
/tmp/ANTs/build/ITKv4-install/include/ITK-4.11/itkImageToImageFilter.hxx
Line: 250
Description: itk::ERROR: N4BiasFieldCorrectionImageFilter(0x202c780): Inputs do not occupy the same physical space! 
InputImage Origin: [-1.0183700e+02, -1.1194000e+02, -3.2479801e+01], InputImage_1 Origin: [-1.0183700e+02, -1.1163189e+02, -3.2819803e+01]
        Tolerance: 8.8000002e-06
InputImage Direction: 1.0000000e+00 0.0000000e+00 0.0000000e+00
0.0000000e+00 1.0000000e+00 0.0000000e+00
0.0000000e+00 0.0000000e+00 1.0000000e+00
, InputImage_1 Direction: 1.0000000e+00 0.0000000e+00 0.0000000e+00
0.0000000e+00 9.9516690e-01 9.8197960e-02
0.0000000e+00 -9.8197958e-02 9.9516690e-01

[...]

Issue #1511, opened 2019-02-15
Revise SBRef handling


oesteban:
- [ ] Correctly handle multiple 3D and 4D SBRefs (see #1463). That implies agreeing on a policy for SBRefs (relates to #450).
- [ ] Address differences in resolution (initiated in poldracklab/niworkflows#253).
- [ ] Better boilerplate (#1282)

oesteban:
I'm pushing the priority of this up. In working through #1803, I've realized the workflow for SBRefs needs an urgent revision.

effigies:
Use BEAST and UH2 to test.

effigies:
Per meeting: This mostly affects Poldrack lab-internal datasets, so shouldn't hold up the LTS.

oesteban:
Highly-related: #2367

Issue #1463, opened 2019-01-07
N4BiasFieldCorrection error in sbref image


heikestein:
Hi,

first of all, thank you for your great effort in standardizing preprocessing pipelines! 
For now, I am trying to run a workflow with
- 2 bold series with MB factor 3 (bad contrast)
- 1 sbref series collected before run-01 for registration with T1 (high contrast)
- 1 T1w
- 1 phasediff+ 2 magnitude fmap images.

I am not quite sure why the workflow doesn't like my data - it fails on the bold_reference_wf.skullstrip_and_enhance_wf, specifically in the sbref N4 correction wf step. Error message below:

```
[Node] Error on "fmriprep_wf.single_subject_C19_wf.func_preproc_ses_01_task_wm_run_01_wf.bold_reference_wf.enhance_and_skullstrip_bold_wf.n4_correct" (/scratch/fmriprep_wf/single_subject_C19_wf/func_preproc_ses_01_task_wm_run_01_wf/bold_reference_wf/enhance_and_skullstrip_bold_wf/n4_correct)

RuntimeError: Command:
N4BiasFieldCorrection -d 3 --input-image /scratch/fmriprep_wf/single_subject_C19_wf/func_preproc_ses_01_task_wm_run_01_wf/bold_reference_wf/validate_ref/sub-C19_ses-01_task-wm_run-01_sbref_valid.nii.gz --mask-image /scratch/fmriprep_wf/single_subject_C19_wf/func_preproc_ses_01_task_wm_run_01_wf/bold_reference_wf/enhance_and_skullstrip_bold_wf/check_hdr/tpl-MNI152NLin2009cAsym_space-MNI_res-02_brainmask_trans_dil_hdr.nii.gz --output sub-C19_ses-01_task-wm_run-01_sbref_valid_corrected.nii.gz
Standard output:

Standard error:

Return code: 1

```

I am also not sure I conceptually understand what is failing here - my best guess that this is a problem of correcting the reference singleband before being registered the template?
Edit: I tried to run the pipeline without providing sbref images, and it looks like this way, bold images are N4 corrected and the workflow doesn't fail. However, direct registration of MB bold to MNI looks pretty imprecise.

I would appreciate any help, as I really want to promote your pipeline as a new standard in our lab :)

Btw, this is the exact command I ran:

`
[...]

PR #1581, opened 2019-04-11
FIX: Support 4D SBRefs when generating the bold reference

oesteban:
Part of #1511. Addresses #1579 and #1463.

### DIFF
--- a/fmriprep/workflows/bold/base.py
+++ b/fmriprep/workflows/bold/base.py
@@ -338,7 +338,7 @@ Non-gridded (surface) resamplings were performed using `mri_vol2surf`
 """
 
     inputnode = pe.Node(niu.IdentityInterface(
-        fields=['bold_file', 'sbref_file', 'subjects_dir', 'subject_id',
+        fields=['bold_file', 'subjects_dir', 'subject_id',
                 't1_preproc', 't1_brain', 't1_mask', 't1_seg', 't1_tpms',
                 't1_aseg', 't1_aparc',
                 't1_2_mni_forward_transform', 't1_2_mni_reverse_transform',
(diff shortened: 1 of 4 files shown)


fmriprep/workflows/bold/base.py
```diff
@@ -346,7 +346,12 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
         name='inputnode')
     inputnode.inputs.bold_file = bold_file
     if sbref_file is not None:
-        inputnode.inputs.sbref_file = sbref_file
+        from niworkflows.interfaces.images import ValidateImage
+        val_sbref = pe.Node(ValidateImage(), name='val_sbref')
+        val_sbref.inputs.in_file = sbref_file
+        workflow.connect([
+            (val_sbref, inputnode, [('out_file', 'sbref_file')]),
+        ])
```
effigies: Weren't we already doing a post-estimation validation to catch the sbref case? If we're going to do a pre- for both sbref and bold, then we should drop the post.

effigies: Also, just from a conceptual workflows-as-functiony-things perspective, I'm not a huge fan of setting a new node to input to `inputnode`. I'm not sure if it's more maintainable to insert `val_sbref` between `inputnode` and `bold_reference_wf` or not.

oesteban: I'm not a fan of that either, but I couldn't see other alternatives. Only passing the sbref_file as an argument to the workflow, and I very much disliked the idea. I was tempted to fix that for the bold_file itself but understood that fix should go on a different PR.

effigies: What about removing `sbref_file` from `inputnode`, and just connecting `val_sbref` to `bold_reference_wf`? Since we only populate the inputnode from the argument anyway, there's no real advantage to it being there.

oesteban: I have no issues with that.

oesteban:
We can ask Ian to run it. Can you give him some tips on how to build a new
singularity image? Otherwise, it'll need to wait until Tuesday

effigies:
I've pushed this as `docker/4dsbref`, so when the build finishes we can build a singularity image from `poldracklab/fmriprep:4dsbref`.

effigies:
@ianeisenberg Are you able to build a new singularity image from a Docker image? Try:

```
singularity build fmriprep_4dsbref.simg docker://podracklab/fmriprep:4dsbref
```

I don't know if other options are needed on your environment, but if you use other options usually, go ahead.

oesteban:
I just confirmed this patch works out.

PR #1583, opened 2019-04-17
RF: progress bids-validator 1.2.3

yarikoptic:


oesteban:
Asking review for @rwblair to weigh in

rwblair:
Looking at the change logs there hasn't been any major changes. the 1.1 to 1.2 rollover was for keeping revision numbers low.

@oesteban @effigies From now on when I publish a new version of the validator you want me to go ahead and open a PR for fmriprep to bump the version?

effigies:
@rwblair I think that would be good. Does the validator output its version, by the way? That might be a useful thing when getting reports to quickly tell if the problem is likely just an out-of-date validator.

yarikoptic:
> Does the validator output its version, by the way?

possibly related: https://github.com/bids-standard/bids-validator/issues/602

Issue #1584, opened 2019-04-17
Master is broken (reportlet generation of mgz images)


oesteban:
As first reported in https://github.com/poldracklab/fmriprep/pull/1487#issuecomment-483847558, and shown by [this crashfile](https://8105-53175327-gh.circle-artifacts.com/0/tmp/ds005/derivatives/fmriprep/sub-01/log/20190417-053056_750c5ea2-5b99-4618-9414-300c88578a75/crash-20190417-055151-UID1001-recon_report-079497dd-26ea-487e-8ece-2b5fdb02638d.txt), master is currently broken.

The installed version of nilearn is 0.5.1, as reflected by the installation summary of the `pip install --no-cache-dir -r requirements.txt` line:

```
Successfully installed MarkupSafe-1.1.1 PyWavelets-1.0.3 click-7.0 decorator-4.4.0 funcsigs-1.0.2 future-0.17.1 imageio-2.5.0 indexed-gzip-0.8.8 isodate-0.6.0 jinja2-2.10.1 lockfile-0.12.2 lxml-4.3.3 networkx-2.3 neurdflib-5.0.0.post1 nilearn-0.5.1 nipype-1.1.9 niworkflows-0.8.2+8.gb7d111c packaging-19.0 pillow-6.0.0 prov-1.5.3 pydot-1.4.1 pydotplus-2.0.2 pyyaml-5.1 rdflib-4.2.2 scikit-image-0.15.0 seaborn-0.9.0 simplejson-3.16.0 smriprep-0.1.0+18.g423bcc4 svgutils-0.3.1
```

The version is not further changed:
```
Requirement already satisfied: nilearn in /usr/local/miniconda/lib/python3.7/site-packages (from fmriprep==1.3.2+28.gba4ebeb3) (0.5.1)
```

effigies:
I'd just add `!=0.5.0,!=0.5.1` constraints to nilearn.

PR #1585, opened 2019-04-17
PIN: ``nilearn!=0.5.0,!=0.5.1`` to avoid problems plotting mgz images

oesteban:


oesteban:
I just saw https://github.com/poldracklab/fmriprep/issues/1584#issuecomment-484137250

effigies:
Should be fine as long as we don't do pre-installs. But perhaps `<0.5.0a0` is a better move.

emdupre:
There's a brand new release that fixes this problem ! You could pin to `nilearn >= 0.5.2`

effigies:
Not sure I want to commit to >=0.5.2. There are other issues in the 0.5 series that haven't been addressed and we may need to go with <0.5.0a in the end.

oesteban:
Banning 0.5.0 and 0.5.1 seems just fine for now. Let's not trigger an update for all installed systems as fmriprep is indeed compatible with older versions.

PR #1589, opened 2019-04-18
ENH: Do not allow writing derivatives directly into the BIDS root folder

oesteban:
This PR:
- [x] Sets `type=Path` to path arguments
- [x] Modifies the output folder to `<bids_root>/derivatives/fmriprep-<version>` when output folder and `<bids_root>` are the same, issuing a warning.

### DIFF
--- a/fmriprep/cli/run.py
+++ b/fmriprep/cli/run.py
@@ -7,7 +7,6 @@ fMRI preprocessing workflow
 """
 
 import os
-import os.path as op
 from pathlib import Path
 import logging
 import sys
--- a/fmriprep/cli/run.py
+++ b/fmriprep/cli/run.py
@@ -393,28 +393,15 @@ def main():
         p.start()
         p.join()
 
-        if p.exitcode != 0:
-            sys.exit(p.exitcode)
+        retcode = p.exitcode or retval.get('return_code', 0)
 
-        fmriprep_wf = retval['workflow']
-        plugin_settings = retval['plugin_settings']
-        bids_dir = retval['bids_dir']
-        output_dir = retval['output_dir']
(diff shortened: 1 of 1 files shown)


fmriprep/cli/run.py
```diff
@@ -681,26 +680,33 @@ def build_workflow(opts, retval):
...
-    output_dir = op.abspath(opts.output_dir)
-    log_dir = op.join(output_dir, 'fmriprep', 'logs')
-    work_dir = op.abspath(opts.work_dir or 'work')  # Set work/ as default
+    output_dir = opts.output_dir.resolve()
+    if output_dir == bids_dir:
+        output_dir = bids_dir / 'derivatives' / ('fmriprep-%s' % __version__.split('+')[0])
+        logger.warning(
+            'The selected output folder is the same as the input BIDS folder. '
+            'Cowardly redirecting outputs to %s', output_dir
+        )
```
effigies: I have a slight preference for crashing and letting users know to change the output directory.

oesteban: I agree in general, but for this case, there is a recommendation of a default from the BIDS specification. Moreover, people implementing BIDS-Apps are making the output folder optional (which feels intuitive) setting `derivatives/workflow-version` (or similar) as default. It also writes the version in the folder name, which can be life-saving.

I think it is a good call to go on with this override of the output folder until we can make the output folder optional because the suggestion has been accepted for the BIDS-Apps CLI spec.

That said, I'm not too strong about this opinion and would be very happy with just crashing.

oesteban: Since this is not going to play out well with fmriprep-docker, I'll take the crash route.

PR #1596, opened 2019-04-30
ENH: Allow multiple {non,}standard spaces (``--output-spaces``)

### LINKED ISSUE #1558: Support templates other than MNI152NLin2009cAsym
Now that TemplateFlow is up and seemingly running okay, it is time to open up the `--template` option and allow additional template keywords.

This is related to poldracklab/smriprep#50 where the CLI interface will be reconsidered and where we will come up with a solution for people who want to run normalization to two or more templates.

### LINKED ISSUE #1588: Revise command line interface of spaces/references/outputs
### Currently supported

- [x] Allow the user to change the atlas used in brain extraction (as long as the selected atlas has a binary brain mask, although it would be better to use only templates that have a probabilistic brain mask and Registration/Extraction masks).
- [x] Allow the user to change the Spatial Normalization target with the `--template` argument (although some paths may not be fully implemented, e.g. with `--cifti-outputs` or `--ica-aroma`).
- [x] Allow the user to write preprocessed data, resampled to several alternative spaces (namely `T1w`, `template`, and `fsXXXX`) with the `--output-space` argument.
- [x] Allow the user to specify an output resolution with `--template-resampling-grid`.

### Problems

- The use of templates other than `MNI152NLin2009cAsym` is quite unstable. 
- Inconsistency of methods relying on other templates (ICA-AROMA and CIFTI generation)
- Only support freesurfer's surface spaces, and the extension to other surface spaces is difficult for the poor integration with TemplateFlow of the surface world.
- Spatial normalization restricted to only one template (although for long we've allowed supplying several arguments to `--template` it didn't actually turn out in a different workflow).
- Lack of flexibility to satisfy requests for new outputs resampled in new spaces (e.g. native bold).

### Proposal

- *Templates*: Deprecate `--template`, rename it to `--std-spaces`, allow it to accept a list of template identifiers from TemplateFlow (which can be either surface, volume or both). Volumetric registration will then run for all the volumetric templates in the list. The first template defines the "default" standard space for preprocessing. If `--ica-aroma` is set, and the keyword `MNI152NLin6Asym` is not in the list, an internal nonlinear registration will be added. Same for `--cifti-outputs` with `fsLR`. An special keyword `custom` (or maybe `user-provided`, please mention what you like better) can be used to provide a path with a custom template. In that case, `--custom-template-path` should be provided and pointing to a TemplateFlow-like organized template, and having at least two resolutions (high and low) with indexes res-1 and res-2.
- *Nonstandard spatial references*: a `--nonstd-spaces` flag that allows values like `anat`, `run`, and `sbref` (will need to think this through). We could add `T1w` for backwards compatibility, but that is the same meaning as `anat`. Key `run` indicates the individual, native space of this particular BOLD run, `sbref` is the space of the SBRef used as reference (which may or may not be in the same resolution or even session). That would open the door to, for instance, have other keywords such as `individual` or `participant` that would calculate a reference for all tasks and runs of a participant and similarly, `session`.
- *Resolution/density*: `--template-sampling-grid` is deprecated, and the `--standard-spaces` accepts comma separated values indicating resolutions and or densities for resampling. Special key `orig` is the default, and would generate data with the same resolution as the original volume (in other words, the current default of fMRIPrep). For instance `MNI152NLin2009cAsym,res-2` or `fsLR,den-92k`, `MNI152NLin2009cAsym,orig`. I'd be happy to change the default resolution to `res-2`. Additionally, one can set the special key `dismiss` if the user does not want the data to be resampled into the template space (e.g. `MNI152NLin2009cAsym,dismiss` will prevent the generation of `space-MNI152NLin2009cAsym` derivatives).

### Example

Old:
```
fmriprep ... --output-space T1w template fsnative fsaverage5 --template-resampling-grid 2mm
```
New:
```
fmriprep --std-spaces MNI152NLin2009cAsym,res-2 fsaverage,den-10k --nonstd-spaces anat,orig fsnative
```

WDYT?

Paging @edickie @satra for opinions (as this may interact with the discussions about volspace/reference/etc in BIDS Derivatives).

oesteban:
This is a large refactor using two major inter-related features:
  - Uses poldracklab/smriprep#72 and its continuation within
    poldracklab/smriprep#75 to permit the specification of several
    spatial normalization targets (close #1558).
  - Uses the refactor of Reports generation of
    poldracklab/niworkflows#344, which enables rendering reports with
    several normalization spaces. Correspondingly, report generation
    code and config files have been removed.

Correspondingly, the new interface for ``--output-spaces`` is also
incorporated (close #1588).


oesteban:
This is working seemingly fine with ds005 locally, including ``--use-aroma`` and ``--use-syn-sdc`` along with ``--output-spaces anat MNI152NLin2009cAsym:res-2 MNI152NLin6Asym:res-2 fsaverage:den-10k MNI152Lin:res-2``.

I expect many issues building docs and running ds054 though.

effigies:
If you want a review before the CI is fixed, let me know.

oesteban:
Let's wait for Circle to be all green :+1:

oesteban:
Okay, this is looking much better. Last commit is a good candidate for getting all green lights.

Documentation is not failing anymore and it looks fine: https://3557-53608443-gh.circle-artifacts.com/0/home/circleci/out/docs/usage.html#defining-standard-and-nonstandard-spaces-where-data-will-be-resampled
Also interesting to check the description of [the ``--output-spaces`` argument](https://3557-53608443-gh.circle-artifacts.com/0/home/circleci/out/docs/usage.html#Workflow%20configuration) expanded by the sphinx argparse extension.

wrapper/fmriprep_docker.py
```diff
@@ -173,7 +173,8 @@ def merge_help(wrapper_help, target_help):
     # Make sure we're not clobbering options we don't mean to
     overlap = set(w_flags).intersection(t_flags)
     expected_overlap = set(['h', 'version', 'w', 'template-resampling-grid',
-                            'output-grid-reference', 'fs-license-file', 'use-plugin'])
+                            'fs-license-file', 'use-plugin'])
+    print(overlap, expected_overlap)
```
oesteban: Will push a commit removing this leftover ASAP.

oesteban:
Ref. ds210, it seems the reports generator has been confused with the echos since forever. In principle, there's only one run in our test version of ds210.

emdupre:
I can check and see if things have been updated, but it should be showing only one "echo" because that image is actually the optimal combination. We implemented that as the default in the last big ME PR !

oesteban:
I think that big PR is working alright, my suspicion is that this is a little bug of the reports generator.

oesteban:
@effigies this is finally ready for you

fmriprep/workflows/base.py
```diff
@@ -129,18 +157,14 @@ def init_fmriprep_wf(layout, subject_list, task_id, echo_idx, run_uuid, work_dir
...
-
-            Valid spaces:
-
-             - T1w
-             - template
-             - fsnative
-             - fsaverage (or other pre-existing FreeSurfer templates)
-        template : str
-            Name of template targeted by ``template`` output space
+        output_spaces : OrderedDict
```
effigies: Why is the ordering important?

fmriprep/workflows/bold/base.py
```diff
@@ -370,7 +401,10 @@ def init_func_preproc_wf(bold_file, ignore, freesurfer,
         name='summary', mem_gb=DEFAULT_MEMORY_MIN_GB, run_without_submitting=True)
 
     # CIfTI output: currently, we only support fsaverage{5,6}
-    cifti_spaces = [s for s in output_spaces if s in ('fsaverage5', 'fsaverage6')]
+    cifti_spaces = set([s for s in output_spaces.keys() if s in ('fsaverage5', 'fsaverage6')])
```
effigies:
```suggestion
    cifti_spaces = set(s for s in output_spaces.keys() if s in ('fsaverage5', 'fsaverage6'))
```

fmriprep/workflows/bold/resampling.py
```diff
@@ -167,128 +168,156 @@ def select_target(subject_id, space):
...
-        hmc_xforms
-            List of affine transforms aligning each volume to ``ref_image`` in ITK format
-        fieldwarp
-            a :abbr:`DFM (displacements field map)` in ITK format
+        templates
+            List of templates that were applied as targets during
+            spatial normalization.
 
-    **Outputs**
+    **Outputs** - Two outputnodes are available. One output node will be parameterized
```
effigies: I don't think anybody will know what "parameterized" means.

docs/usage.rst
```diff
@@ -38,6 +38,110 @@ Command-Line Arguments
    :nodefaultconst:
 
 
+.. _output-spaces:
```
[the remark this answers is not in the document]
oesteban: I'll open a subsequent PR with branch name `docs/<>` for all these suggestions. Thanks!

fmriprep/cli/run.py
```diff
@@ -853,6 +635,79 @@ def build_workflow(opts, retval):
...
+        print("""\
+Option "--use-aroma" requires functional images to be resampled to MNI152NLin6Asym space. \
+The argument "MNI152NLin6Asym:res-2" has been automatically added to the list of output spaces \
+(option ``--output-spaces``).""", file=stderr)
+
+    if opts.cifti_output and 'MNI152NLin2009cAsym' not in output_spaces:
+        if 'MNI152NLin2009cAsym' not in output_spaces:
+            output_spaces['MNI152NLin2009cAsym'] = {'res': 2}
+            print("""Option ``--cifti-output`` requires functional images to be resampled to \
+``MNI152NLin6Asym`` space. Such template identifier has been automatically added to the list \
```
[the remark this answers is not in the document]
oesteban: Oh man, I thought I had changed the three instances. It should read `MNI152NLin2009cAsym`. It will change though when we support real HCP (to `MNI152NLin6Asym`). Sorry about that.

fmriprep/workflows/base.py
oesteban: Because the first volumetric standard space is taken as "primary". I agree we will get rid of both aspects of an ordered dict (ordering and unique keys). One may want to generate all possible resamplings for a template (for example).
I would leave that for future iterations, though.

fmriprep/workflows/bold/resampling.py
oesteban: Will revise in the `docs/` branch

fmriprep/cli/run.py
```diff
@@ -209,7 +209,7 @@ def get_parser():
     # FreeSurfer options
     g_fs = parser.add_argument_group('Specific options for FreeSurfer preprocessing')
     g_fs.add_argument(
-        '--fs-license-file', metavar='PATH', type=os.path.abspath,
```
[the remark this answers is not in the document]
oesteban: not really. will re-add

docs/usage.rst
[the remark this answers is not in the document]
oesteban: We actually have an "outputs" page. Would you rather open a new page for this or move it there?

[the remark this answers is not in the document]
effigies: I think in-depth discussion of specific command line options fits more in their own pages than in outputs.

PR #1602, opened 2019-05-02
FIX: Mount Nipype config file under new ``$HOME``

oesteban:
When ``$HOME`` was changed to ``/home/fmriprep`` for Docker images I
missed updating the mount point of the wrapper.

This PR addresses that issue.


oesteban:
Quick thumbs up?

PR #1603, opened 2019-05-03
FIX: Update spline fitting distance for BOLD bias-field correction

markushs:
## Changes proposed in this pull request
Adding `-b [200]` as input to ants.N4BiasFieldCorrection in `bold_bold_trans_wf/bold_reference_wf/enhance_and_skullstrip_bold_wf/n4_correct`.
Previously, default setting (i.e., "single mesh element over the entire domain, `-b [1x1x1,3]`) was used. This seems to fix issues reported in https://github.com/poldracklab/fmriprep/issues/1460

## Documentation that should be reviewed
Not sure, but don't think docs need to be updated.

PR #1606, opened 2019-05-03
CI: Build docs outside of container

effigies:

This should speed up builds somewhat by avoiding using the container in the documentation build and avoiding the container build in `doc(s)/*` branches.


.circleci/config.yml
```diff
@@ -294,47 +294,36 @@ jobs:
...
-            cat $PWD/builddocs.log
-            grep -qv "ERROR" $PWD/builddocs.log
+              apt-get update
+              apt-get install -y graphviz
+      - run:
+          name: Install fMRIPrep
+          command: python -m pip install ".[doc]"
+      - run:
+          name: Build documentation
+          command: make -C docs html
```
oesteban: would this treat warnings as errors?

effigies:
Locally I'm getting this failure:

```
Warning, treated as error:
/Users/markiewicz/Projects/crn/fmriprep/venv/lib/python3.6/site-packages/fmriprep/workflows/base.py:docstring of fmriprep.workflows.base.init_fmriprep_wf:7:Exception occurred in plotting index-1
 from /Users/markiewicz/Projects/crn/fmriprep/docs/api/index.rst:
  File "/Users/markiewicz/Projects/crn/fmriprep/venv/lib/python3.6/site-packages/nipype/sphinxext/plot_workflow.py", line 485, in run_code
    exec(code, ns)
  File "<string>", line 38, in <module>
  File "/Users/markiewicz/Projects/crn/fmriprep/venv/lib/python3.6/site-packages/fmriprep/workflows/base.py", line 218, in init_fmriprep_wf
    err_on_aroma_warn=err_on_aroma_warn,
  File "/Users/markiewicz/Projects/crn/fmriprep/venv/lib/python3.6/site-packages/fmriprep/workflows/base.py", line 462, in init_single_subject_wf
    skull_strip_template=skull_strip_template,
TypeError: init_anat_preproc_wf() got an unexpected keyword argument 'output_spaces'
make: *** [html] Error 2
```

Looks like what @markushs is seeing over in #1603.

oesteban:
smriprep version?

effigies:
> smriprep version?

`v0.1.1`. Whatever builds with `pip install ".[doc]"`.

effigies:
It's very unclear to me what `LINKS_REQUIRES` is supposed to be doing. Does anything respect it?

fmriprep/__about__.py
```diff
@@ -89,14 +89,14 @@
...
-    'niworkflows',
+    'niworkflows>=0.9',
     'numpy',
     'pandas',
     'psutil>=5.4',
     'pybids<0.8.0a0,>=0.7.1',
     'pyyaml',
     'scikit-image',
-    'smriprep',
+    'smriprep>=0.2.0a',
```
oesteban: version pinning here disables reading dependency links.

effigies: Those dependency links do nothing anyway. Have you ever found them to work?

oesteban: They are definitely finicky. But the version here will likely remove the version you've previously installed via `requirements.txt`.

Then, when it gets to process the dependency links, I think setuptools has a hard time figuring out the version of the commit (because that is returned by versioneer, and that needs to have access to tags in the repo).

But yeah, I can't definitely say yes to your question.

pyproject.toml
```diff
@@ -0,0 +1,2 @@
+[build-system]
+requires = ["pip>=18.1", "setuptools>=40.8", "wheel", "numpy"]
```
effigies: @oesteban This should allow us to install purely with `pip install .`, as `pip` will parse this first to make sure it has the build dependencies.

pip 18.1 allows us to use a dependency that looks like `"package @ URL"`, so we can handle all dependencies directly in `install_requires`.

effigies: Relevant discussion: https://stackoverflow.com/a/53412651

Possibly the reason this is failing for me is that pip removed dependency_links in version 19.

effigies: Hmm. Looks like you can't actually set a minimum version on pip in this way. https://github.com/pypa/pip/issues/4145

So this would mean we need to prepare for users with old pip, which is likely on most HPCs.

oesteban:
smriprep 0.2.0 has just posted. I would not spend a lot of time fighting pip.

fmriprep/__about__.py
```diff
@@ -89,14 +89,16 @@
...
+    'niworkflows @ '
+    'git+https://github.com/effigies/niworkflows.git@5e9daf0b08fcad2b17350a71ecaf71fa9ce4eff7',
     'numpy',
     'pandas',
     'psutil>=5.4',
     'pybids<0.8.0a0,>=0.7.1',
     'pyyaml',
     'scikit-image',
-    'smriprep @ git+https://github.com/poldracklab/smriprep.git@f1cfc37bcdc346549dbf1d037cdade3a3b32d5de',
+    'smriprep @ ',
```
oesteban: this was potentially working. please note that comma breaking up the string concatenation.

PR #1607, opened 2019-05-04
RFC/MAINT: Consolidate build configuration in setup.cfg

effigies:
We currently have a kind of messy build system, with multiple places to specify dependencies (`requirements.txt`, `fmriprep/__about__.py` and `docs/environment.yml`), and different build strategies for Docker, ReadTheDocs and Circle for PyPI distribution.

This PR aims to reduce the dependency list to one place (`setup.cfg`) and produce one preferred way to build source distributions (`python setup.py sdist`), binary distributions (`pip wheel --no-deps .`) and install from the current repository (`pip install .`). These commands are to be used, as appropriate, in Docker, ReadTheDocs and Circle. I will start a RTD build for this branch to verify that it works.

#1606 has moved us to requiring `pip>=18.1`, which should support all of this. I can add Travis jobs to verify that each of these commands works as expected.

An additional thing to note is that the [current PyPI metadata](https://pypi.org/project/fmriprep/) is a bit messed up, I suspect because we had a multi-line `description`, which is a one-line field.

This appears to be where the Python ecosystem is heading, with PEPs [508](https://www.python.org/dev/peps/pep-0508/), [517](https://www.python.org/dev/peps/pep-0517/), [518](https://www.python.org/dev/peps/pep-0518/) and [566](https://www.python.org/dev/peps/pep-0566/), and it's the approach I'm using on FitLins. If y'all agree, I'd be inclined to make this the standard organizational scheme for our packages, though there's not a great rush to move everything over.


effigies:
Docs are successfully building in [RTD](https://fmriprep.readthedocs.io/en/mnt-dist/) and on [Circle](https://circleci.com/gh/poldracklab/fmriprep/8295#artifacts/containers/0), and the Dockerfile is [building](https://circleci.com/gh/poldracklab/fmriprep/8296).

effigies:
LOL, @oesteban watch out I just pushed some new stuff.

.circleci/config.yml
```diff
@@ -764,26 +755,131 @@ jobs:
...
+          command: |
+            pyenv local 3.5.2
+            pip install twine  # For use in checking distributions
+            THISVERSION=$( python get_version.py )
+            virtualenv --python=python build
+            source build/bin/activate
+            pip install --upgrade "pip>=19.1" numpy
+            echo "${CIRCLE_TAG:-$THISVERSION}" > fmriprep/VERSION
+            python setup.py sdist
+            pip wheel --no-deps -w dist/ .
```
effigies: Note that because of the CPython bits, this produces a wheel like `fmriprep-1.3.2+166.gbe8452ab-cp35-cp35m-linux_x86_64.whl`, which will only be installable with Python 3.5. We may want to consider distributing 3.6 and 3.7 wheels as well.

oesteban: Let's get this merged first :+1:

effigies: Cool. Just stumbled down a little manylinux rabbit hole and wonder if it might make sense to pull the cpython bits out into a very small package that can be pre-compiled and that can let us drop the numpy/cpython build requirements for fMRIPrep.

effigies:
This does what I want at this point. I have a follow-up that fiddles with caching to try to reduce test time, but it's small turnips compared to this, and I don't want to muddy an already big PR with another goal.

oesteban:
Sounds sensible :+1:

PR #1609, opened 2019-05-05
ENH: Reduce BSplineFieldmap memory usage

effigies:

Was looking at the `maths.pyx` thing to see about splitting it out into its own package. Noticed some `get_data()` uses in `BSplineFieldmap`. `get_data()` is supposed to be being deprecated any day now... and the two calls to it are sub-optimal.

The first can just be replaced by the image shape, which is just read from the header and doesn't require loading the data.

The second can use `get_fdata`, and if we use `caching='unchanged'`, then it won't store a copy of the array internally, so it can get garbage-collected after filling `_data`.

I don't expect this is going to have a significant effect, but just while I was there.

### DIFF
--- a/fmriprep/utils/bspline.py
+++ b/fmriprep/utils/bspline.py
@@ -38,8 +38,7 @@ class BSplineFieldmap(object):
         self._padding = padding
 
         # Pad data with zeros
-        self._data = np.zeros(tuple(np.array(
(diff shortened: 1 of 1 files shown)


oesteban:
When I revise the implementation of fieldmaps I'll rid fmriprep of Cython and this buggy implementation. I know there is a pure python implementation on scipy that fits the needs. Meanwhile, this is a great idea.

PR #1621, opened 2019-05-07
CI: Check all versions with better diagnostic messages

### LINKED ISSUE #1619: Running fMRIPREP version 0+unknown
Versioneer has stopped working with the revision of the packaging.

effigies:
Closes #1619.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -174,10 +174,13 @@ RUN pip install --no-cache-dir "templateflow>=0.1.3,<0.2.0a0" && \
 # Installing FMRIPREP
 COPY . /src/fmriprep
 ARG VERSION
-# Force static versioning within container
-RUN echo "${VERSION}" > /src/fmriprep/fmriprep/VERSION && \
-    echo "include fmriprep/VERSION" >> /src/fmriprep/MANIFEST.in && \
-    pip install --no-cache-dir "/src/fmriprep[all]"
+# Simulate a git repository with one commit and a version
+RUN git config --global user.name null && \
+    git config --global user.email null && \
+    git -C /src/fmriprep init && \
+    git -C /src/fmriprep commit --allow-empty -m noop && \
+    git -C /src/fmriprep tag "${VERSION}"
+RUN pip install --no-cache-dir "/src/fmriprep[all]"
 
 RUN install -m 0755 \
     /src/fmriprep/scripts/generate_reference_mask.py \
--- a/Dockerfile
+++ b/Dockerfile
@@ -108,9 +108,6 @@ RUN curl -sL https://deb.nodesource.com/setup_10.x | bash -
 RUN apt-get install -y nodejs
 RUN npm install -g svgo
 
-# Installing bids-validator
-RUN npm install -g bids-validator@1.2.3
-
 # Installing and setting up ICA_AROMA
 RUN mkdir -p /opt/ICA-AROMA && \
   curl -sSL "https://github.com/maartenmennes/ICA-AROMA/archive/v0.4.4-beta.tar.gz" \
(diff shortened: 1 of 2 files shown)


effigies:
While I'm at it, I added the PyPI packages to the artifacts, to help with diagnostics when something goes wrong there.

effigies:
I believe what's happening is that `versioneer` is writing a new copy of `_version.py` every time it's built.

So I think there are two options:

1. Update versioneer to write `_version.py` as you want with `VERSION`.
2. ```
    git init && git commit --allow-empty -m noop && git tag ${VERSION}
    pip install .
    ```

effigies:
cc @oesteban

oesteban:
I'm looking into option 1. However, it messes up with the system-wide installation of versioneer :S

effigies:
Okay. I'm understanding what's going on now. I think 1 is going to be a pain in the ass. 2 is a bit hacky, but the only alternative is to revert the bits of `setup.py` that look for `VERSION`. So I guess make that option 3.

Dockerfile
```diff
@@ -171,13 +168,19 @@ RUN pip install --no-cache-dir "templateflow>=0.1.3,<0.2.0a0" && \
     find $TEMPLATEFLOW_HOME -type d -exec chmod go=u {} + && \
     find $TEMPLATEFLOW_HOME -type f -exec chmod go=u {} +
 
+# Installing bids-validator
+RUN npm install -g bids-validator@1.2.3
```
effigies: I did this because this is prone to relatively frequent updates. For small things like this, AROMA, TemplateFlow, I'm inclined to move them to just before fMRIPrep as they get upgrades, which will ensure that the most frequently changed tend to be latest.

effigies:
Tested with:

```
docker build --rm=false -t poldracklab/fmriprep:test_version \
    --build-arg BUILD_DATE=`date -u +"%Y-%m-%dT%H:%M:%SZ"` \
    --build-arg VCS_REF=`git rev-parse --short HEAD` \
    --build-arg VERSION="1.4.0" .
```

```
$ docker run --rm -it poldracklab/fmriprep:test_version --version
fmriprep v1.4.0
```

Dockerfile
```diff
@@ -171,13 +168,19 @@ RUN pip install --no-cache-dir "templateflow>=0.1.3,<0.2.0a0" && \
...
-# Force static versioning within container
-RUN echo "${VERSION}" > /src/fmriprep/fmriprep/VERSION && \
-    echo "include fmriprep/VERSION" >> /src/fmriprep/MANIFEST.in && \
-    pip install --no-cache-dir "/src/fmriprep[all]"
+# Simulate a git repository with one commit and a version
+RUN git config --global user.name null && \
+    git config --global user.email null && \
+    git -C /src/fmriprep init && \
+    git -C /src/fmriprep commit --allow-empty -m noop && \
+    git -C /src/fmriprep tag "${VERSION}"
```
effigies: I'm honestly not sure why it doesn't mark these as dirty, but whatever. If it works, it works.

oesteban:
Adding this at line 1406 of `/versioneer.py` works for me:

```Python
    from pathlib import Path
    ver_file = Path(__file__).parent / 'fmriprep' / 'VERSION'

    if ver_file.is_file():
        ver = ver_file.read_text().splitlines()[0].strip()
        return {"version": ver, 'full': ver}
```

effigies:
@oesteban That doesn't look right to me. Do you want to suggest a commit, instead?

oesteban:
@effigies https://github.com/poldracklab/fmriprep/compare/docker/versioneer / https://circleci.com/workflow-run/c1907b14-318a-4f97-8a79-489b0b489f35

effigies:
Ah, that was L1405 before adding the import up top. Seems fine. Feel free to pop that directly on master and close this if it builds correctly.

oesteban:
I've stolen your edits to `.circle/config.yml` without cherry picking your commit. If you don't mind that, I'll merge when tests pass.

effigies:
Yeah, I don't care.

oesteban:
hahaha the dirty biting back :D

effigies:
It's not just dirty... it's null bytes.

oesteban:
It's dirty because of all those `sed` in the previous build step

effigies:
I meant the `^@`s are null bytes. I suspect those are going to screw things up as well. But maybe `test` will ignore them.

oesteban:
Oh, yes, they were also breaking the test.

PR #1643, opened 2019-05-20
FIX: FreeSurfer without adding some ``fs*`` to ``--output-spaces``

### LINKED ISSUE #1641: Module bold_std_trans_wf has no output called poutputnode.bold_aseg_std
Dear fmriprep experts,
When I am trying to run the latest version fmriprep (1.4.0) with the --output-space fsnative, I received the following error.  How I can solve the issue? With many thanks!

Exception: Some connections were not found
Module bold_std_trans_wf has no output called poutputnode.bold_aseg_std

Module bold_std_trans_wf has no output called poutputnode.bold_aparc_std

Module bold_std_trans_wf has no output called outputnode.bold_aseg_std

Module bold_std_trans_wf has no output called outputnode.bold_aparc_std

oesteban:
Tested locally, closes #1641.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,6 +9,15 @@ jobs:
       - image: docker:18.01.0-ce-git
     working_directory: /tmp/src/fmriprep
     steps:
+      - checkout
+      - run:
+          name: Check whether build should be skipped
+          command: |
+            if [[ "$( git log --format='format:%s' -n 1 $CIRCLE_SHA1 | grep -i -E '^docs?(\(\w+\))?:' )" != "" ]]; then
+              echo "Only docs build"
+              circleci step halt
+            fi
+
       - run:
           name: Install parallel gzip and python3
           command: |
@@ -21,7 +30,6 @@ jobs:
             - docker-v3-
           paths:
             - /tmp/cache/docker.tar.gz
-      - checkout
       - setup_remote_docker
       - run:
           name: Load Docker image layer cache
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -9,6 +9,15 @@ jobs:
       - image: docker:18.01.0-ce-git
     working_directory: /tmp/src/fmriprep
     steps:
+      - checkout
+      - run:
+          name: Check whether build should be skipped
+          command: |
+            if [[ "$( git log --format='format:%s' -n 1 $CIRCLE_SHA1 | grep -i -E '^docs?(\(\w+\))?:' )" != "" ]]; then
```
effigies: I see you've created a new committing convention; is it documented somewhere?

Also, is it a good idea to skip tests if the most recent commit is a doc update? It seems likely to trip people up.

oesteban: > is it documented somewhere?

It's not. I can add documentation if we finally decide to keep it.

> It seems likely to trip people up.

Why? we should be careful though and always request for a full build before merging, but this adds nice flexibility when you are in the middle of a big PR (or situations like Franklin's).

effigies: > > It seems likely to trip people up.
> 
> Why?

Just because if I were putting together a batch of commits to submit, docs would often be the last thing before submitting.

oesteban: For users: at that point we would walk them through forcing one more commit

For us, we can just use `dox` to indicate that the commit is documentation but you want to trigger a full build.

Honestly, I think the risks/value tradeoff is clear here. On the bright side, it is a lot easier to ask people to commit with `docs:` than `[skip ds005][skip ds054][skip ds210][skip tests]` which would be equivalent (although it would waste time building the docker image).

oesteban: Would you be okay merging and keeping the discussion (here or elsewhere)? I think the discussion is worth having and merging this would not be harmful.

effigies: Yeah, this is fine.

PR #1695, opened 2019-07-09
MAINT: Use PyBIDS 0.9.x via niworkflows/smriprep PRs

### LINKED ISSUE #1694: MAINT: Bump PyBIDS to 0.9.x
Time to get our projects generally up-to-date.

Related: poldracklab/smriprep#93

effigies:
Closes #1694.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,2 +1,2 @@
 [build-system]
-requires = ["setuptools", "wheel", "numpy", "cython"]
+requires = ["setuptools >= 30.3.0", "wheel", "numpy", "cython"]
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -1,2 +1,2 @@
 [build-system]
-requires = ["setuptools", "wheel", "numpy", "cython"]
+requires = ["setuptools >= 40.8.0", "wheel", "numpy", "cython"]
```
effigies: I honestly don't remember why 40.8.0 was the settled-on version, but I'm not sure that I care to titrate it back to see if we can get to 30.3.0 or some earlier one. At least it's not a regression to leave it as it is.

@oesteban Good to merge as-is?

oesteban: Very good!

PR #1703, opened 2019-07-17
FIX: Use MNI152NLin2009cAsym for SDC if no templates are specified

### LINKED ISSUE #1675: fMRIPrep stops when no volumetric standard space is set via ``--output-spaces``
See https://neurostars.org/t/fmriprep-quits-after-a-few-minutes-of-running/4446

effigies:

This PR will resolve #1675, which fails to identify a template to be used for SDC normalization when no template is specified as an output space.

I'm beginning with the portion that raises the exception, which is simply failure to find an object in an empty container. It may be that doing this is insufficient to ensure that a normalization to the default template is found, and further work will be needed.


effigies:
Hopefully that fixes things...

PR #1710, opened 2019-07-21
FIX: Postpone ``pandoc`` conversion of boilerplate after workflow has fully run

### LINKED ISSUE #1667: Allow subjects to be run in parallel without causing errors
While running fmriprep (--anat-only) on 13 subjects in parallel, I ran into the following error on 4 of the subjects due to subjects running simultaneously:

FileNotFoundError: [Errno 2] No such file or directory: '/scratch/03201/jbwexler/openneuro_fmriprep/data/ds000003_work/ds000003-download/derivatives/fmriprep-1.4.0/fmriprep/logs/CITATION.md'

I'm attaching the full log here.
[ds003_anatonly.txt](https://github.com/poldracklab/fmriprep/files/3259486/ds003_anatonly.txt)

### LINKED ISSUE #1709: Postpone ``pandoc`` execution to the end of all processing
Soichi found that [the memory peak at the beginning of fMRIPrep is caused by pandoc](https://neurostars.org/t/large-virtual-memory-allocations-in-fmriprep-on-pbs-system/4665/10?u=oesteban), which is aligned to [this open issue in pandoc](https://github.com/jgm/pandoc/issues/3169).

Since that issue has been open for a while now, but we are not going to think of an alternative just for the conversion of markdown to html to generate the boilerplate, a good solution is to postpone pandoc to be the latest process run, and also offer a flag to totally skip the conversion.

oesteban:
This PR addresses the memory peak caused by pandoc, running the process
at the very end of fMRIPrep.

We probably want to explicitly clear up as many objects as we can from
memory and force gc before the subprocess opens pandoc, but that's left
for a future PR.

This PR just closes #1709 with the delayed pandoc run. I'm not
implementing the flag to avoid the boilerplate generation because I'm
not sure we want to introduce an additioinal argument for this.

Coincidentally, it also closes #1667 - now the deletion of citation
files is done inside a try..catch where ``FileNotFoundError`` is
dismissed. That should avert the issue, while keeping the functionality
of #1567.

Individual reports should keep rendering exactly the same after this PR.

### DIFF
--- a/fmriprep/cli/run.py
+++ b/fmriprep/cli/run.py
@@ -664,7 +664,7 @@ def build_workflow(opts, retval):
 
         citation_files['md'].write_text(boilerplate)
         build_log.log(25, 'Works derived from this fMRIPrep execution should '
-                   'include the following boilerplate:\n\n%s', boilerplate)
+                      'include the following boilerplate:\n\n%s', boilerplate)
     return retval
 
 
--- a/fmriprep/cli/run.py
+++ b/fmriprep/cli/run.py
@@ -411,6 +411,43 @@ license file at several paths, in this order: 1) command line argument ``--fs-li
                                        level='info')
     finally:
         from niworkflows.reports import generate_reports
+        from subprocess import check_call, CalledProcessError, TimeoutExpired
+        from pkg_resources import resource_filename as pkgrf
+        from shutil import copyfile
+
+        citation_files = {
+            ext: output_dir / 'fmriprep' / 'logs' / ('CITATION.%s' % ext)
+            for ext in ('bib', 'tex', 'md', 'html')
+        }
+
+        if citation_files['md'].exists():
+            # Generate HTML file resolving citations
+            cmd = ['pandoc', '-s', '--bibliography',
+                   pkgrf('fmriprep', 'data/boilerplate.bib'),
+                   '--filter', 'pandoc-citeproc',
+                   '--metadata', 'pagetitle="fMRIPrep citation boilerplate"',
+                   str(citation_files['md']),
+                   '-o', str(citation_files['html'])]
+            try:
+                check_call(cmd, timeout=10)
+            except (FileNotFoundError, CalledProcessError, TimeoutExpired):
+                logger.warning('Could not generate CITATION.html file:\n%s',
+                               ' '.join(cmd))
+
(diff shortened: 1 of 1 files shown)


oesteban:
This should be fairly non-controversial. I forgot to mention during the meeting.

fmriprep/cli/run.py
```diff
@@ -411,6 +411,43 @@ def main():
...
+        from subprocess import check_call, CalledProcessError, TimeoutExpired
+        from pkg_resources import resource_filename as pkgrf
+        from shutil import copyfile
+
+        citation_files = {
+            ext: output_dir / 'fmriprep' / 'logs' / ('CITATION.%s' % ext)
+            for ext in ('bib', 'tex', 'md', 'html')
+        }
+
+        if citation_files['md'].exists():
```
effigies: `else` should we print an error?

fmriprep/cli/run.py
```diff
@@ -411,6 +411,43 @@ def main():
...
+            # Generate HTML file resolving citations
+            cmd = ['pandoc', '-s', '--bibliography',
+                   pkgrf('fmriprep', 'data/boilerplate.bib'),
+                   '--filter', 'pandoc-citeproc',
+                   '--metadata', 'pagetitle="fMRIPrep citation boilerplate"',
+                   str(citation_files['md']),
+                   '-o', str(citation_files['html'])]
+            try:
+                check_call(cmd, timeout=10)
```
effigies: Perhaps log that we're generating HTML boilerplate.

fmriprep/cli/run.py
```diff
@@ -411,6 +411,43 @@ def main():
...
+                logger.warning('Could not generate CITATION.html file:\n%s',
+                               ' '.join(cmd))
+
+            # Generate LaTex file resolving citations
+            cmd = ['pandoc', '-s', '--bibliography',
+                   pkgrf('fmriprep', 'data/boilerplate.bib'),
+                   '--natbib', str(citation_files['md']),
+                   '-o', str(citation_files['tex'])]
+            try:
+                check_call(cmd, timeout=10)
```
effigies: And again with LaTeX.

Since these are clearly less critical bits, it might make people less worried if a crash follows this...

PR #1715, opened 2019-07-25
ENH: Add check for updates and check whether version has been flagged.

### LINKED ISSUE #1645: Version check & release blacklisting
Enable a version check to suggest upgrading when a new full release has been posted.

This mechanism would also make sure that the current release has not been blacklisted (because of some critical issue). In that case, fmriprep would be runnable only forcing with a specific flag.

oesteban:
Adds two new functions to assess the current version of fMRIPrep before
operating.

Close #1645.


oesteban:
Any outstanding issue that you can foresee?

I'm thinking we should extend the ArgumentParser object and hook all the extra tests we have enough info to check into the `parse_args()` - WDYT?

effigies:
Oh, sorry. I saw this but forgot to comment and then it fell off my radar. Let me have another look through.

effigies:
So I remember AFNI's version check being a royal pain when their server went down. I see that we're wrapping things in `try`/`except` blocks, but if the result is hanging instead of a quick 500 error, then we're degrading UX significantly.

I would give an out. I know we've got `--no-track`, but perhaps we could make a more general `FMRIPREP_NETWORK` environment variable that can disable sentry and any calls out when set to 0. Maybe there's a better approach, but I think if we don't build in a way to turn it off, we're going to annoy people.

oesteban:
Good catch, what about just a very short timeout?
https://2.python-requests.org/en/master/user/quickstart/

fmriprep/cli/_version.py
```diff
@@ -0,0 +1,78 @@
...
+            if abs((datetime.now() - date).days) > RELEASE_EXPIRY_DAYS:
+                outdated = True
+
+    if latest is None or outdated is True:
+        try:
+            response = requests.get(url='https://pypi.org/pypi/fmriprep/json', timeout=1.0)
+        except Exception:
+            pass
+
+        if response and response.status_code == 200:
```
effigies: If `requests.get()` raises an exception, `response` will be undefined.

fmriprep/cli/_version.py
```diff
@@ -0,0 +1,78 @@
...
+    """Check whether current version is flagged."""
+    # https://raw.githubusercontent.com/poldracklab/fmriprep/master/.versions.json
+    flagged = tuple()
+    try:
+        response = requests.get(url="""\
+https://raw.githubusercontent.com/poldracklab/fmriprep/master/.versions.json""", timeout=1.0)
+    except Exception:
+        pass
+
+    if response and response.status_code == 200:
```
effigies: If requests.get() fails, response will be undefined.

fmriprep/cli/_version.py
effigies: And in tests, you should mock a `requests.exceptions.Timeout`.

fmriprep/cli/run.py
```diff
@@ -274,6 +276,26 @@ def get_parser():
...
+    currentv = Version(__version__)
+    latest = check_latest()
+
+    if latest is not None and currentv < latest:
+        print("""\
+WARNING: The current version of fMRIPrep (%s) is outdated.
+Please consider upgrading to the latest version %s.
+Before upgrading, please consider that mixing fMRIPrep versions
+within a single study is strongly discouraged.""" % (
+            __version__, latest), file=sys.stderr)
```
effigies: This is sending some mixed messages, by making using an old version a WARNING but then also saying not to mix versions.

What about making it more informational:

```
Newer version of fMRIPrep available: {latest}
https://fmriprep.readthedocs.io/en/latest/faq.html#upgrading
```

And we could add a note on upgrading that talks about whether to upgrade, which should be in our docs anyway.

fmriprep/cli/_version.py
```diff
@@ -0,0 +1,78 @@
+# emacs: -*- mode: python; py-indent-offset: 4; indent-tabs-mode: nil -*-
+# vi: set ft=python sts=4 ts=4 sw=4 et:
+"""Version CLI helpers."""
```
effigies: I don't like giving this a name like `_version.py`. Why not just `version` or `version_check`?

fmriprep/cli/_version.py
```diff
@@ -0,0 +1,78 @@
...
+
+from pathlib import Path
+from datetime import datetime
+import requests
+from .. import __version__
+
+
```
effigies:
```suggestion
```

docs/faq_tips_tricks.rst
```diff
@@ -126,3 +126,23 @@ A symptomatic output looks like: ::
 
 If you would like to run *fMRIPrep* in parallel on multiple subjects please use
 `this method <https://neurostars.org/t/fmriprep-workaround-for-running-subjects-in-parallel/4449>`__.
+
+
+.. _upgrading:
```
effigies: I think we need to rename this file to `faq` and update any links to `faq_tips_tricks`. This is to make the `faq.html` link work, but I think it's good to move to a guessable file name anyway.

oesteban:
Sounds sensible

PR #1735, opened 2019-08-20
ENH: Make fmriprep print defaults for arguments with help

gdevenyi:
Use ArgumentDefaultsHelpFormatter so that defaults are printed with --help

I don't consider this work complete, because a substantial number of the options listed in the argument parser don't provide a default value which should.


gdevenyi:
CI failure appears unrelated to this change.

effigies:
Re-ran the test. Seems like a one-off. Thanks for this.

Issue #1587, opened 2019-04-17
OSError: Handle is closed


effigies:
This is now appearing in the outputs on CircleCI, which makes it potentially amenable to fixing. It may be related to the forkserver.

```
Error in atexit._run_exitfuncs:
Traceback (most recent call last):
  File "/usr/local/miniconda/lib/python3.7/concurrent/futures/process.py", line 101, in _python_exit
    thread_wakeup.wakeup()
  File "/usr/local/miniconda/lib/python3.7/concurrent/futures/process.py", line 89, in wakeup
    self._writer.send_bytes(b"")
  File "/usr/local/miniconda/lib/python3.7/multiprocessing/connection.py", line 183, in send_bytes
    self._check_closed()
  File "/usr/local/miniconda/lib/python3.7/multiprocessing/connection.py", line 136, in _check_closed
    raise OSError("handle is closed")
OSError: handle is closed
```

It doesn't mark any particular problem, but it has caused users to worry that something's gone wrong.

gdevenyi:
I can reproduce this with 1.5.0rc1 and ds000003 with a singularity container, let me know if I can dump any logs.

effigies:
Not sure that there's much to be gleaned from our output. But thanks for bumping this issue. I've opened #1743, which should hopefully help us narrow down what is registering the cleanup call that's failing.

oesteban:
Is this issue stale? Should we just close? - I don't see it happening anymore.

effigies:
Yeah, I don't see it happening either, anymore. Hopefully it was a bug in Python that's been fixed.

PR #1743, opened 2019-08-22
DIAGNOSTIC: Print traceback when atexit calls are registered

effigies:
Starting on #1587.


oesteban:
The original issue is closed now, so I guess this can also be.

PR #1770, opened 2019-09-12
ENH: Enable users to pass JSON filters to select subsets of data

bpinsard:

reference issue: https://github.com/poldracklab/smriprep/issues/104

Adding a `--bids_filters` option to further specify the options to select input files to the pipeline.
This points to a json files containing a dict with pybids entities filters, that is to be passed to nipype.interfaces.io.BIDSDataGrabber output_query option.
I created the same branch in smriprep  with similar changes in run.py and workflow.base.py

While I was coding that I realized that f/smriprep uses in fact niworkflows.interfaces.bids.BIDSDataGrabber (this is a mistaking name btw).
Is there a specific reason to recode that part, could it be possible to use the original nipype one, or it is  missing a  feature?
I need you insight there.
Thanks

TODO/WIP


fmriprep/cli/run.py
```diff
@@ -81,6 +82,8 @@ def get_parser():
     # Re-enable when option is actually implemented
     # g_bids.add_argument('-r', '--run-id', action='store', default='single_run',
     #                     help='select a specific run to be processed')
+    g_bids.add_argument('--bids_filters', action='store', type=Path,
```
effigies: We've generally been moving to hyphen-separated words in arguments. It will still show up as `opts.bids_filters` later.

```suggestion
    g_bids.add_argument('--bids-filters', action='store', type=Path,
```

emdupre:
I wonder if this work wouldn't make sense to upstream to BIDS Apps more generally ? I'm thinking in particular of the spec @effigies was co-leading. It seems like a general question of how BIDS Apps should handle multiple eligible inputs. 

@bpinsard and I discussed this in person, and I would have thought the best "BIDS" way to do this for the moment is just to modify a study's `.bidsignore` when generating the derivatives. The work here is a much more rigorous solution, but it would seem like something that should exist beyond f/smriprep and/or niworkflows ?

bpinsard:
@emdupre 
This is definitely a recurring issues when using BIDSApps on complex datasets. It would make sense to factor it, but for now BIDS grabbing is coded in each run.py of each BIDSApps. 
For fMRIPrep, as the BIDS grabbing is not at the top level, but goes through multiple layers of fmriprep/smriprep/niworkflows, this is more difficult to factor it with other BIDSApps.

emdupre:
> For fMRIPrep, as the BIDS grabbing is not at the top level, but goes through multiple layers of fmriprep/smriprep/niworkflows, this is more difficult to factor it with other BIDSApps.

I completely understand that :smile_cat: 

I just mean that this seems like something that we would want to upstream to the [BIDS Execution specification](https://github.com/bids-standard/bids-specification/issues/313), since it strikes me as a very general issue.

bpinsard:
Tests fails due to dependency on niworkflow changes, same for PR on smriprep.
Otherwise I think these changes are ready.

effigies:
I think this is in good shape. How do we feel about squeezing it in to 20.0.0? It adds a CLI option but does not change anything when the option is not passed.

Need to merge poldracklab/smriprep#143 and release niworkflows and smriprep, but we can do that tomorrow if everybody's happy.

@bpinsard Have you been actually testing this? I can try it with a multi-session OpenNeuro dataset tomorrow if not. Or we can push a `docker/` branch to get a test image up on Docker Hub.

fmriprep/cli/run.py
```diff
@@ -76,6 +77,11 @@ def get_parser():
     # Re-enable when option is actually implemented
     # g_bids.add_argument('-r', '--run-id', action='store', default='single_run',
     #                     help='select a specific run to be processed')
+    g_bids.add_argument(
+        '--bids-filters', action='store', type=Path,
```
mgxd: I think making it explicit this is a file will improve clarity and be more in line with the other file cli arguments
```suggestion
        '--bids-filter-file', action='store', type=Path,
```

fmriprep/cli/run.py
```diff
@@ -499,6 +505,8 @@ def build_workflow(opts, retval):
     bids_dir = opts.bids_dir.resolve()
     output_dir = opts.output_dir.resolve()
     work_dir = opts.work_dir.resolve()
+    bids_filters = json.loads(opts.bids_filters.read_text()) \
```
mgxd:
```suggestion
    bids_filters = json.loads(opts.bids_filter_file.read_text()) \
```

fmriprep/cli/run.py
```diff
@@ -499,6 +505,8 @@ def build_workflow(opts, retval):
...
+        if opts.bids_filters else None
```
mgxd:
```suggestion
        if opts.bids_filter_file else None
```

setup.cfg
```diff
@@ -24,14 +24,14 @@ install_requires =
     nibabel >= 3.0.1
     nipype >=1.3.1
     nitime
-    niworkflows ~= 1.1.6
+    niworkflows @ git+https://github.com/poldracklab/niworkflows.git@edeea81352f3f2b276f11e42e7829e1c5a03770f
```
mgxd:
```suggestion
    niworkflows ~= 1.1.7
```

setup.cfg
```diff
@@ -24,14 +24,14 @@ install_requires =
...
     numpy
     pandas
     psutil >=5.4
     pybids ~= 0.9.4
     pyyaml
     sdcflows ~=1.1.0
-    smriprep ~= 0.5.1
+    smriprep @ git+https://github.com/bpinsard/smriprep.git@ef552bbad2418b558a2f639be502bcf98ec1cf3b
```
mgxd:
```suggestion
    smriprep ~= 0.5.2
```

setup.cfg
effigies: Let's do this on release prep and get things merged now.

setup.cfg
effigies: Pin to poldracklab master to ensure we don't break things. Re-pin on release prep.

```suggestion
    smriprep @ git+https://github.com/poldracklab/smriprep.git@07a7ba3fce8a9953923bcd0fa6ffbffac401208d
```

PR #1790, opened 2019-09-25
CI: Use recent Python image to build packages

effigies:

Switches to `circleci/python:3.7.4` for building packages on, as this should have a long lifetime before we need to upgrade it.

The latest twine breaks Python 3.5 but still installs on it, so our current strategy fails. We could pin an older twine, but there doesn't seem much value.

@oesteban I see we're still using manylinux to build wheels. I believe we no longer need that?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -260,7 +260,7 @@ jobs:
 
   build_docs:
     docker:
-        - image: python:latest
+      - image: python:latest
     working_directory: /tmp/src/fmriprep
     environment:
       - FSLOUTPUTTYPE: 'NIFTI'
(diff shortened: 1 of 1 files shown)


oesteban:
> @oesteban I see we're still using manylinux to build wheels. I believe we no longer need that?

Probably not after this change :+1:

effigies:
I looked, and it looks like we still have a cython extension. For some reason I thought that had gone with sdcflows.

oesteban:
mmm yeah, that should be gone.

effigies:
Okay. I'll open a PR to kill it shortly.

PR #1792, opened 2019-09-25
MAINT: Purge Cython and depend on SDCflows

effigies:
We've moved all the fieldmap tools into [sdcflows](https://github.com/poldracklab/sdcflows/), so we don't need the Cython extensions in fMRIPrep. This removes `maths.pyx` and all code that depends on it, and updates `fmriprep.interfaces.fmap` to import its contents from `sdcflows.interfaces.fmap`.

Cython and numpy have been removed from the build requirements, and the wheel built on CircleCI will be pure Python, so we can drop manylinux.


effigies:
@oesteban I don't know what's up with the ds005 and ds210 errors. Looks like some other nipype regression.

oesteban:
I'm pretty positive those errors will go away if we clean the cache. We should use this as an opportunity to debug https://github.com/nipy/nipype/issues/3014 (this is an instance of this issue). Can you think of a way to retrieve Circle's cache and take it to a controlled environment?

effigies:
Just create a job that loads the cache and saves it as an artifact?

oesteban:
> Just create a job that loads the cache and saves it as an artifact? Could tarball it for easy download.

I think Circle tells you the identifier of the cache it found and retrieved. We may want to have a build that finds a cache identifier on the commit message and tarballs the cache. Nothing otherwise. Could be helpful in the future. WDYT?

oesteban:
Since I rebased to upstream/master, these are the links to the broken builds:

https://circleci.com/gh/poldracklab/fmriprep/9753?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link
https://circleci.com/gh/poldracklab/fmriprep/9756?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Issue #1798, opened 2019-09-27
Problems during processing with 1.5.0


JoffJones:
Hi there,

I am trying to re-run some preprocessing using the 1.5.0 stable release but encountered an error only when using the -use-aroma flag. Any help would be appreciated! 

FYI this is run using singularity on a HPC. 

The error reported in the html report is: 
```
Node Name: _autorecon31
```

The command with: 
```
/usr/local/miniconda/bin/fmriprep /CALM_BIDS_New_2 /CALM_BIDS_New_2/derivatives/fmriprep-1.5.0-test-aroma participant --participant_label 004 -v -w /CALM_BIDS_New_2/derivatives/fmriprepwork-1.5.0-test-aroma --skull-strip-template MNIPediatricAsym:cohort-1 --output-spaces MNIPediatricAsym:cohort-1:res-2 T1w MNI152NLin6Asym:res-2 MNI152NLin2009cAsym fsaverage --use-aroma --fs-license-file /CALM/license.txt --write-graph --fd-spike-threshold 0.5 --dvars-spike-threshold 0.5 --notrack --resource-monitor
```
The output with:
[use-aroma.txt](https://github.com/poldracklab/fmriprep/files/3663700/use-aroma.txt)

The command without:
```
/usr/local/miniconda/bin/fmriprep /CALM_BIDS_New_2 /CALM_BIDS_New_2/derivatives/fmriprep-1.5.0-test participant --participant_label 001 -v -w /CALM_BIDS_New_2/derivatives/fmriprepwork-1.5.0-test --skull-strip-template MNIPediatricAsym:cohort-1 --output-spaces MNIPediatricAsym:cohort-1:res-2 T1w MNI152NLin6Asym:res-2 MNI152NLin2009cAsym fsaverage --fs-license-file /CALM/license.txt --write-graph --fd-spike-threshold 0.5 --dvars-spike-threshold 0.5 --notrack --resource-monitor 
```
The output without:
[no-aroma.txt](https://github.com/poldracklab/fmriprep/files/3663708/no-aroma.txt)

oesteban:
Hi @JoffJones please note that these are two different participants. Looking at your logs, participant 004 (with `--use-aroma`) crashed during FreeSurfer operation. You might want to try it again or make sure you use an existing FreeSurfer folder.

Regarding participant 001 (without `--use-aroma`), it indeed contains an error that is likely to go away if you do not reuse the work directory anymore.

[...]

PR #1802, opened 2019-10-02
DOC: Improve documentation about TemplateFlow and Containers

### LINKED ISSUE #1801: TemplateFlow API and operation pitfalls (esp. Singularity containers)
Since we included it in [1.3.0](https://github.com/poldracklab/fmriprep/releases/tag/1.3.0), TemplateFlow has become a common friction point. That is especially true for users running fMRIPrep via Singularity containers, since the containers are read-only when running unprivileged.

This issue is an attempt to gather up all the existing knowledge segregated across several GitHub issues and NeuroStars questions, mostly.

Based on the collection of issues, we will then write a corresponding new section of the documentation, condensing the solutions and tips (https://github.com/poldracklab/fmriprep/issues/1778#issuecomment-532752241).

The list starts here:

* #1500, #1510, [NS#3747](https://neurostars.org/t/fmriprep-1-3-x-on-singularity-git-annex-permission-errors/3747), [NS#3507](https://neurostars.org/t/fmri-prep-v1-3-0-command-error/3507/7) Templateflow with DataLad + Singularity (DataLad is disabled by default since 1.3.1)
* [NS#3809](https://neurostars.org/t/fmriprep-templateflow-unable-to-access-these-remotes/3809) TF + DataLad "Unable to access these remotes" (fixed in 1.3.2).
* [NS#3619](https://neurostars.org/t/datalad-error-in-1-3-0-post2-libc-kernel-mismatch/3619) [NS#3639](https://neurostars.org/t/fmriprep-1-3-0-post2-crash/3639) TemplateFlow via DataLad is not supported by very old kernels.
* [NS#3585](https://neurostars.org/t/disable-datalad-in-fmriprep-or-a-templateflow-something-issue/3585) TemplateFlow and fMRIPrep installed 'bare-metal" - datalad not available.
* #1525, [NS#3693](https://neurostars.org/t/docker-fmriprep-permissionerror-errno-13-permission-denied-cache/3693) Docker (as user) + TemplateFlow = Permission Denied.
* [NS#2249](https://neurostars.org/t/issue-running-fmriprep-with-singularity-image/2249/8?u=oesteban) - Binding the TemplateFlow folder into the singularity container.
* [NS#3770](https://neurostars.org/t/filenotfounderror-errno-2-no-such-file-or-directory-home-fmriprep-cache/3770) Default TemplateFlow folder was set to `$HOME/.cache/templateflow` after redefining `$HOME=/home/fmriprep` in containers triggered a burst of `FileNotFoundError`. Related #1534, #1669.
* [NS#4174](https://neurostars.org/t/availability-of-the-harvard-oxford-atlas-in-mni152nlin2009casym-template/4174/8?u=oesteban) Problems accessing TemplateFlow before the automated update of the TF stub was included in the installation script.
* #1635, #1644, [NS#4588](https://neurostars.org/t/fmriprep-templateflow-fails-with-oserror-errno-30-read-only-file-system/4588), [NS#3949](https://neurostars.org/t/antsbrainextraction-failure-freesurfer-does-fine/3949/15?u=oesteban),  TF produces `OSError: [Errno 30] Read-only filesystem`. Problems with Singularity systems that automatically bind `/home` but do not overwrite `$HOME` and it remains pointing to `/home/fmriprep`.
* [NS#4336](https://neurostars.org/t/questions-about-adding-a-custom-standard-template/4336) Similar to the one above, but in this case asking for custom templates.
* #1640 [NS#4646](https://neurostars.org/t/using-custom-template-output-spaces/4646) Custom templates (UNRESOLVED).
* [NS#4527](https://neurostars.org/t/unable-to-fetch-template-from-templateflow/4527) Accessing TF directly.
* [NS#4566](https://neurostars.org/t/problems-using-pediatric-template-from-templateflow/4566/15?u=oesteban) HTTPS unsuported by system - redefine `templateflow.api.TF_S3_ROOT` to  `http://templateflow.s3.amazonaws.com`
* [NS#4842](https://neurostars.org/t/failed-to-establish-a-new-connection-templateflow-s3-amazonaws-com/4842) Other connection problems (UNRESOLVED).
* #1699, #1778 [NS#5027](https://neurostars.org/t/fmriprep-1-5-0-fails-in-singularity-container-with-read-only-file-system/5027) Connection errors, problems binding into `/home/fmriprep`.

A great deal of documentation was generated here: https://github.com/poldracklab/fmriprep/issues/1778#issuecomment-531884251 and https://github.com/poldracklab/fmriprep/issues/1778#issuecomment-532295876

oesteban:
This PR addresses #1801. The intent is to provide clearer documentation about how to run fMRIPrep via Singularity containers. In particular, the interaction between TemplateFlow and Singularity should be better described.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,5 +1,5 @@
-fmriprep: A Robust Preprocessing Pipeline for fMRI Data
-=======================================================
+*fMRIPrep*: A Robust Preprocessing Pipeline for fMRI Data
+=========================================================
 
 This pipeline is developed by the `Poldrack lab at Stanford University
 <https://poldracklab.stanford.edu/>`_ for use at the `Center for Reproducible
@@ -43,7 +43,7 @@ About
 .. image:: https://github.com/oesteban/fmriprep/raw/38a63e9504ab67812b63813c5fe9af882109408e/docs/_static/fmriprep-workflow-all.png
 
 
-``fmriprep`` is a functional magnetic resonance imaging (fMRI) data
+*fMRIPrep* is a functional magnetic resonance imaging (fMRI) data
 preprocessing pipeline that is designed to provide an easily accessible,
 state-of-the-art interface that is robust to variations in scan acquisition
 protocols and that requires minimal user input, while providing easily
@@ -56,12 +56,12 @@ volume-based statistics, etc.
 
 .. note::
 
-   fmriprep performs minimal preprocessing.
+   *fMRIPrep* performs minimal preprocessing.
    Here we define 'minimal preprocessing'  as motion correction, field
    unwarping, normalization, bias field correction, and brain extraction.
    See the workflows_ for more details.
 
-The ``fmriprep`` pipeline uses a combination of tools from well-known software
+The *fMRIPrep* pipeline uses a combination of tools from well-known software
 packages, including FSL_, ANTs_, FreeSurfer_ and AFNI_.
 This pipeline was designed to provide the best software implementation for each
 state of preprocessing, and will be updated as newer and better neuroimaging
@@ -86,7 +86,7 @@ https://fmriprep.readthedocs.io/
 Principles
 ----------
 
-``fmriprep`` is built around three principles:
+*fMRIPrep* is built around three principles:
 
 1. **Robustness** - The pipeline adapts the preprocessing steps depending on
    the input dataset and should provide results as good as possible
@@ -97,22 +97,22 @@ Principles
    automatic fashion.
 3. **"Glass box"** philosophy - Automation should not mean that one should not
    visually inspect the results or understand the methods.
-   Thus, ``fmriprep`` provides visual reports for each subject, detailing the
+   Thus, *fMRIPrep* provides visual reports for each subject, detailing the
    accuracy of the most important processing steps.
    This, combined with the documentation, can help researchers to understand
    the process and decide which subjects should be kept for the group level
    analysis.
 
 
-Limitations and reasons not to use ``fmriprep``
------------------------------------------------
+Limitations and reasons not to use *fMRIPrep*
+---------------------------------------------
 
 1. Very narrow :abbr:`FoV (field-of-view)` images oftentimes do not contain
    enough information for standard image registration methods to work correctly.
    Also, problems may arise when extracting the brain from these data.
    Supporting these particular images is already a future line of the development
    road-map.
-2. ``fmriprep`` may also underperform for particular populations (e.g., infants) and
+2. *fMRIPrep* may also underperform for particular populations (e.g., infants) and
    non-human brains, although appropriate templates can be provided to overcome the
    issue.
 3. The "EPInorm" approach is currently not supported, although we plan to implement
@@ -131,6 +131,6 @@ Acknowledgements
 ----------------
 
 Please acknowledge this work by mentioning explicitly the name of this software
-(fmriprep) and the version, along with a link to the `GitHub repository
+(*fMRIPrep*) and the version, along with a link to the `GitHub repository
 <https://github.com/poldracklab/fmriprep>`_ or the Zenodo reference.
 For more details, please see :ref:`citation`.
--- a/docs/_static/sbatch.slurm
+++ b/docs/_static/sbatch.slurm
@@ -1,0 +1,60 @@
+#!/bin/bash
+#
+#SBATCH -J fmriprep
+#SBATCH --time=48:00:00
+#SBATCH -n 1
+#SBATCH --cpus-per-task=16
+#SBATCH --mem-per-cpu=4G
+#SBATCH -p normal,mygroup  # Queue names you can submit to
+# Outputs ----------------------------------
+#SBATCH -o log/%x-%A-%a.out
+#SBATCH -e log/%x-%A-%a.err
+#SBATCH --mail-user=%u@domain.tld
+#SBATCH --mail-type=ALL
+# ------------------------------------------
+
+BIDS_DIR="$STUDY/data"
+DERIVS_DIR="derivatives/fmriprep-1.5.0"
+
+# Prepare some writeable bind-mount points.
+TEMPLATEFLOW_HOST_HOME=$HOME/.cache/templateflow
+FMRIPREP_HOST_CACHE=$HOME/.cache/fmriprep
+mkdir -p ${TEMPLATEFLOW_HOST_HOME}
+mkdir -p ${FMRIPREP_HOST_CACHE}
+
+# Prepare derivatives folder
+mkdir -p ${BIDS_DIR}/${DERIVS_DIR}
+
+# This trick will help you reuse freesurfer results across pipelines and fMRIPrep versions
+mkdir -p ${BIDS_DIR}/derivatives/freesurfer-6.0.1
+if [ ! -d ${BIDS_DIR}/${DERIVS_DIR}/freesurfer ]; then
+    ln -s ${BIDS_DIR}/derivatives/freesurfer-6.0.1 ${BIDS_DIR}/${DERIVS_DIR}/freesurfer
+fi
+
+# Make sure FS_LICENSE is defined in the container.
+export SINGULARITYENV_FS_LICENSE=$HOME/.freesurfer.txt
+
+# Designate a templateflow bind-mount point
+export SINGULARITYENV_TEMPLATEFLOW_HOME="/templateflow"
+SINGULARITY_CMD="singularity run --cleanenv -B $BIDS_DIR:/data -B ${TEMPLATEFLOW_HOST_HOME}:${SINGULARITYENV_TEMPLATEFLOW_HOME} -B $L_SCRATCH:/work $STUDY/images/poldracklab_fmriprep_1.5.0.simg"
+
+# Parse the participants.tsv file and extract one subject ID from the line corresponding to this SLURM task.
+subject=$( sed -n -E "$((${SLURM_ARRAY_TASK_ID} + 1))s/sub-(\S*)\>.*/\1/gp" ${BIDS_DIR}/participants.tsv )
+
+# Remove IsRunning files from FreeSurfer
+find ${BIDS_DIR}/derivatives/freesurfer-6.0.1/sub-$subject/ -name "*IsRunning*" -type f -delete
+
+# Compose the command line
+cmd="${SINGULARITY_CMD} /data /data/${DERIVS_DIR} participant --participant-label $subject -w /work/ -vv --omp-nthreads 8 --nthreads 12 --mem_mb 30000 --output-spaces MNI152NLin2009cAsym:res-2 anat fsnative fsaverage5 --use-aroma"
+
+# Setup done, run the command
+echo Running task ${SLURM_ARRAY_TASK_ID}
+echo Commandline: $cmd
+eval $cmd
+exitcode=$?
+
+# Output results to a table
+echo "sub-$subject   ${SLURM_ARRAY_TASK_ID}    $exitcode" \
+      >> ${SLURM_JOB_NAME}.${SLURM_ARRAY_JOB_ID}.tsv
+echo Finished tasks ${SLURM_ARRAY_TASK_ID} with exit code $exitcode
+exit $exitcode
--- a/docs/citing.rst
+++ b/docs/citing.rst
@@ -2,11 +2,11 @@
 
 .. _citation:
 
-===============
-Citing FMRIPREP
-===============
+=================
+Citing *fMRIPrep*
+=================
 
-Select which options you have run FMRIPREP with to generate custom language
+Select which options you have run *fMRIPrep* with to generate custom language
 we recommend to include in your paper.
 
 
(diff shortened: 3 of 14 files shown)


oesteban:
The most delicate section of the new documentation is:
https://4189-53608443-gh.circle-artifacts.com/0/tmp/src/fmriprep/docs/_build/html/singularity.html#templateflow-and-singularity

oesteban:
Tagging @dmd @JoffJones @utooley @oricon @a3sha2 @marcelzwiers @jooh, who have experienced issues, and proposed solutions to problems arising from Singularity + TemplateFlow. They are perfect to have a quick look and make sure the documentation now contains what they had to figure out the hard way.

oesteban:
For convenience:

- Access to the latest build - https://4192-53608443-gh.circle-artifacts.com/0/tmp/src/fmriprep/docs/_build/html/index.html
- Quick link to singularity: https://4192-53608443-gh.circle-artifacts.com/0/tmp/src/fmriprep/docs/_build/html/singularity.html
- Quick link to TemplateFlow: https://4192-53608443-gh.circle-artifacts.com/0/tmp/src/fmriprep/docs/_build/html/spaces.html#templateflow

utooley:
Two thoughts on this, would it be useful to lay out the error message you might get if 

a) your Singularity config doesn't allow you to bind non-existent bind points, i.e. has not enabled overlay?
or b) doesn't auto-bind a directory such as `\scratch` that you reference in the documentation? I believe at some point I got errors trying to follow paths that I thought existed inside the container but don't by default unless your configuration enabled them.

oesteban:
> Two thoughts on this, would it be useful to lay out the error message you might get (i.e. `WARNING: Skipping user bind, non existent bind point (directory) in container`) if
> 
> a) your Singularity config doesn't allow you to bind non-existent bind points, i.e. has not enabled overlay?
> or b) doesn't auto-bind a directory such as `\scratch` that you reference in the documentation? I believe at some point I got errors trying to follow paths that I thought existed inside the container but don't by default unless your configuration enabled them.

You're right, this is not yet covered in the documentation. Your cluster is one example of these right? Do you mind if I ask you to try a few things over the next week?

JoffJones:
Hi All, 
Thank you for the documentation and examples of running fmriprep in singularity on HPCs. We are using a slightly different implementation of sbatch through the subprocess module in Jupyter Lab. Some differences include using the -C contain all flag, not using -clean-env, and I have also found that I have to bind mount /tmp:/tmp otherwise there is a persistent error with surface reconstruction (see log below). Weirdly, when I do mount this the log returns that /tmp will not be mounted as it is already mounted. 

[notmp2-sub-001.out.txt](https://github.com/poldracklab/fmriprep/files/3685841/notmp2-sub-001.out.txt)

Happy to share the 'subprocess' based sbatch submission - although this isn't yet running perfectly (https://github.com/poldracklab/fmriprep/issues/1798).
We could also try to get your template bash script working on our HPC as a validation if that would be useful?

Lastly just some tiny typos under 'Accessing the host’s filesystem' 
$ singularity run --cleanenv -B /work:/work fmriprep.**smig** \

And under 'Internet access problems'
requests.exceptions.SSLError: HTTPSConnectionPool .... In this case, **you** container seems to be able to

utooley:
> > Two thoughts on this, would it be useful to lay out the error message you might get (i.e. `WARNING: Skipping user bind, non existent bind point (directory) in container`) if
> > a) your Singularity config doesn't allow you to bind non-existent bind points, i.e. has not enabled overlay?
> > or b) doesn't auto-bind a directory such as `\scratch` that you reference in the documentation? I believe at some point I got errors trying to follow paths that I thought existed inside the container but don't by default unless your configuration enabled them.
> 
> You're right, this is not yet covered in the documentation. Your cluster is one example of these right? Do you mind if I ask you to try a few things over the next week?

@oesteban Not at all! I will say that I have no sense of how common these misconfigurations or weird default settings are across different cluster setups, so I don't know how useful (or not) it is to document them, but happy to try things out! As a side note, everything I know, I knew from running 1-4-1; haven't tested 1-5-0 myself to see if anything has resolved, but a few days ago someone at Penn ran 1-5-0 and we had similar problems with TemplateFlow.

oesteban:
@JoffJones 

> Some differences include using the -C contain all flag

Could you elaborate more? I would like to see whether this could be of help to others.

> not using ``--cleanenv``

I have to admit I've been pretty resistant to this flag but finally surrendered to its usefulness. What is preventing you from using it?

> I have to bind mount /tmp:/tmp otherwise there is a persistent error with surface reconstruction (see log below). Weirdly, when I do mount this the log returns that /tmp will not be mounted as it is already mounted.

It seems to me that the FreeSurfer error is unrelated to mounting /tmp...

> Happy to share the 'subprocess' based sbatch submission - although this isn't yet running perfectly (#1798).

I wonder how you deal with the environment given that you are not using ``--cleanenv`` - the python environment to run the submission is very likely to leak into the containers, isn't it?

> We could also try to get your template bash script working on our HPC as a validation if that would be useful?

That would be really sweet if you can.

oesteban:
I'll also ping @yarikoptic here - I'm sure he will find missing points in the proposed documentation, esp. about Singularity.

effigies:
I'll try to review the content in detail this afternoon, but to give you the feedback I've already got, I think this overshoots the prominence of these things by promoting them to the top level of the table of contents. I would suggest the following TOC:

* Installing and Running 
  * [Installation](https://4192-53608443-gh.circle-artifacts.com/0/tmp/src/fmriprep/docs/_build/html/installation.html) (subpage)
    * pip install fmriprep
    * pip install fmriprep-docker
    * docker pull
    * docker build (might just link out to contributors)
    * Fetch singularity (I guess from docker hub)
    * Build singularity (`singularity build` and `docker2singularity`)
  * Running fMRIPrep (subpage - mostly just the current [usage notes](https://4192-53608443-gh.circle-artifacts.com/0/tmp/src/fmriprep/docs/_build/html/usage.html))
  * [Running fMRIPrep via Docker containers](https://4192-53608443-gh.circle-artifacts.com/0/tmp/src/fmriprep/docs/_build/html/docker.html) (subpage)
  * [Running fMRIPrep via Singularity containers](https://4192-53608443-gh.circle-artifacts.com/0/tmp/src/fmriprep/docs/_build/html/singularity.html) (subpage)
* Processing pipeline details
* ...

The top page doesn't need to be very elaborate, but we can add the context of what we generally recommend for different use cases, e.g., `fmriprep-docker` for local systems, `singularity` for HPC, and make clear that all approaches accept the same set of arguments, and the distinctions lie in how to make sure the right resources are available to fmriprep.

oesteban:
Okay, I've tried to cover two of your points, @effigies, one moving the fmriprep-docker installation to the installation page (cross-referencing execution) and the second trying to make it more clear the command line structure and how it is modified for the case of containers.

There's perhaps some need to mention that ``fmriprep-docker`` is a bit special, but things should be getting closer to how they will look in the end.

Regarding the restructure, I agree with the feeling that giving new sections top-level relevance is an overshoot. But looking at how many of the questions in NeuroStars go around Singularity issues, I think it makes sense to cast them to top-level sections.

There's plenty of room on the left menu, and we will make it a lot more accessible.

effigies:
I think we harm overall accessibility by adding to the top level, as now there are more choices to consider as you try to find what you need (and this isn't always going to be the top thing that people are looking for). I think having a clean hierarchy will be much easier to follow, even if it means one extra click.

In fact, I would probably reduce some of what is already there, such as moving SDC under processing details, and usage notes under installing and running. And I feel like the standard/non-standard spaces discussion is also a bit high up, though I'm not sure what I'd put it under. Maybe an "advanced configuration" section, or possibly we could rename "processing pipeline details" to indicate a broader consideration of processing options. But that's out of scope for this PR, certainly.

oesteban:
Would you agree with deferring the conversation about structure (including the new sections) to the docusprint?

effigies:
Sure.

yarikoptic:
My problem with any suggestion of exposing system `$HOME` within container environment as HOME is possible side-effects for other tools.  And exposing `$HOME` as some other directory inside the container for templateflow to use, is making it impossible to capture the state for any specific run to be able to reproduce it later on.
If reproducibility is in mind, I would recommend having per-project/study specific `$HOME` to reside under that study directory (lets call it `$STUDY`), and only directories under that study directory to be bind mounted inside.  Then if $STUDY is also version controlled, e.g. with datalad, you have a chance to capture all necessary details to be able later on reproduce that specific result.
E.g. that is what we do in https://github.com/ReproNim/containers/
- [scripts/singularity_cmd](https://github.com/ReproNim/containers/blob/master/scripts/singularity_cmd) is our shim to run singularity to 
   - sanitize execution
   - bind mount top level directory
   - bind mount [binds/HOME](https://github.com/ReproNim/containers/tree/master/binds/HOME) which could carry anything desired to be tracked/carried around (or ignored)

Or may be I am misunderstanding the situation and bind mounting $HOME is solely for truly just caching (so exact version(s) of templates are guaranteed to be the correct ones the other way)?  Then may be it would be best to bind mount `~/.cache/templateflow`?  But then, since I trust nobody, I might still prefer to just track that cache directory within the study folder.

oesteban:
> Or may be I am misunderstanding the situation and bind mounting $HOME is solely for truly just caching (so exact version(s) of templates are guaranteed to be the correct ones the other way)?

You got it right. There is a way of using TemplateFlow with DataLad (i.e., version-controlled) but we ended up disabling it by default since it created a pile of problems. `$HOME` is used just for caching, although exact versions of templates are not guaranteed. Since 1.5.0, `$HOME` is also used to cache the latest version of fMRIPrep in a text file under `$HOME/.cache/fmriprep` - but I guess that wouldn't worry you (it's there to shave off just one HTTP GET). 

I'll think about this - I like the idea of the `$STUDY` environment variable - that is something we might want to propagate to the revision of BIDS-Apps in general (wdyt @effigies?).

If you come up with the markdown explaining your `$STUDY` idea to be included within the docs (in this PR or sending a separate one after this), it would be fantastic (and probably much more efficient than me trying to draft something first) - but I guess you might not have the time. Please let me know if this is too much to ask :).

effigies:
I'm a little fuzzy on what exactly is being proposed. I think my brain is kind of shot...

So we should move things out of `/home/fmriprep`, to head off people accidentally dropping their home directory into it, and I guess make a `/study` directory? That seems fine. It's not entirely clear how a `$STUDY` environment variable comes into play.

Is there a proposal on how we're actually supposed to capture in a datalad-consumable way what specifically is being injected into the container? Or is this mostly just making it easier for people who do have a well-organized setup to capture these things with datalad?

oesteban:
@effigies, @yarikoptic in reference to the `$STUDY` environment variable I think it could make sense in a revised BIDS-App document/spec/protocol/recommendation to act as a default entrypoint for sharing resources with the host in containerized apps, and also a good default for the interface. This way, we could say that, by default, data are expected to be in `$STUDY/data`, derivatives in `$STUDY/data/derivatives`, configs/settings under `$STUDY/settings`, etc. The idea works on bare-metal installations (perhaps with DataLad control) and also with containers (although the recommendation of having derivatives under the dataset breaks potential read-only mounts of only `$STUDY/data`). For singularity users without overlay enabled (@yarikoptic should correct me if I'm wrong here) having a bind-mount point available in the image (e.g., `$STUDY` is `/study` unless otherwise defined) could be very useful to avoid the `WARNING: Skipping user bind, non existent bind point (directory) in container` @utooley mentioned above.

@utooley - made some changes following your suggestions in https://github.com/poldracklab/fmriprep/pull/1802/commits/2ef69801b79fe8bfd9a1bff9c38e80fdaf9caa21 - can you check?

@JoffJones can you comment on some questions I left in https://github.com/poldracklab/fmriprep/pull/1802#issuecomment-538020046? - thanks!

oesteban:
May I ask some quick eyeballing from @rwblair and @franklin-feingold 

I'd like to get this merged soon. Please let me know if there are any blocking points.

JoffJones:
Hi @oesteban,

Sorry for the delay in responding. We have been using the -C flag as standard for any singularity images, @jooh would be able to elaborate but he is away at the moment. I think that we don't use --cleanenv because we are running things in a neuroimaging environment on the HPC, but I haven't tried it without.

Apologies I can't elaborate any more!

oesteban:
I'll merge - I think it is better to test this with real users and revise during the docusprint.

utooley:
Also sorry for the late reply @oesteban ! I looked over your initial changes after I commented and they made sense to me (although I didn't know about docker2singularity), happy to go over again and revise down the line. Sorry again for the delay!

JoffJones:
Hi @oesteban,

Thank you very much for the updated slurm script, I have been doing some testing with it and the only trouble so far is with creating the freesurfer-6.0.1 directory for future re-use. I'm currently testing it without this, but would be grateful for your input.

bash script:
[fmriprep1.5.0-new.txt](https://github.com/poldracklab/fmriprep/files/3741103/fmriprep1.5.0-new.txt)

output:
[fmriprep-27211-2.out.txt](https://github.com/poldracklab/fmriprep/files/3741078/fmriprep-27211-2.out.txt)

Issue #1745, opened 2019-08-23
Fmriprep handling of multi-echo data with single-band reference scans 


markushs:
Hi, I posted this on neurostars a while ago:
(https://neurostars.org/t/fmriprep-handling-of-multi-echo-data-with-single-band-reference-scans/4822). 
Registration of Multi-echo multi-band (MEMB) data would likely benefit from using single-band reference scans (when available). Right now I believe registration is done based on some kind of average of the volumes representing the first echo, and with multiband the contrast in these images gets very low.

As we cannot openly share participant data from the relevant project, we've collected MEMB data on two volunteers in our lab for the purpose of sharing. If someone thinks it's worthwhile looking into this, let me know and I will provide you with the data (or put it on openneuro, if that's better).

Thanks!
Markus

effigies:
If your participants are willing to have their scans put on OpenNeuro (note that OpenNeuro releases all data under [CC0](https://creativecommons.org/publicdomain/zero/1.0/)), that would probably be simplest (we can just grab it with [DataLad](https://www.datalad.org/)), and make it easier for others to benefit from the effort you went to.

markushs:
Here it is:
https://openneuro.org/datasets/ds002147/versions/1.0.0

tsalo:
How should single-band reference images fit into the multi-echo workflow?

I assume that the goal here is to use the single-band reference images (probably the one for the first echo) as BOLD reference images for motion correction and brain-mask generation. Then, if `t2s_coreg` is not enabled, use the single-band reference image for coregistration?

Is that roughly the extent to which the single-band reference images would be used with multi-echo data?

markushs:
In addition to the steps mentioned above I would ensure that the echo-1 sbref is used for all co-registration steps not involving the T2*-map (e.g. with fieldmaps). We get sub-optimal results for quite a few subjects at this step. 

An example.:

<img width="2686" alt="image" src="https://user-images.githubusercontent.com/44242647/65032225-061b4f00-d943-11e9-8ca7-bbe863e04b9d.png">

tsalo:
Would it make sense to use the mean SBRef (averaged across echoes) instead of just echo 1?

markushs:
Dropout typically becomes quite prominent at later echoes, and in my experience there is sufficient contrast in the echo-1 sbref. So my immediate intuition would be to just use echo 1 rather than the mean, at least when sbrefs are available.

tsalo:
[...]

PR #1803, opened 2019-10-02
FIX: Revise multi-echo reference generation, permitting using SBRefs too

tsalo:
References #1745, https://github.com/nipreps/niworkflows/issues/399, and https://github.com/nipreps/niworkflows/pull/408.

- Currently, just tags a specific version of niworkflows to ensure backward compatibility, by request of @effigies.

None

### DIFF
--- a/fmriprep/workflows/bold/base.py
+++ b/fmriprep/workflows/bold/base.py
@@ -227,11 +227,6 @@ Non-gridded (surface) resamplings were performed using `mri_vol2surf`
                 't1w2fsnative_xfm', 'fsnative2t1w_xfm']),
         name='inputnode')
     inputnode.inputs.bold_file = bold_file
-    if sbref_files is not None:
-        from niworkflows.interfaces.images import ValidateImage
-        val_sbref = pe.MapNode(ValidateImage(), name='val_sbref',
-                               iterfield=['in_file'])
-        val_sbref.inputs.in_file = sbref_files
 
     outputnode = pe.Node(niu.IdentityInterface(
         fields=['bold_t1', 'bold_t1_ref', 'bold_mask_t1', 'bold_aseg_t1', 'bold_aparc_t1',
@@ -294,6 +289,10 @@ Non-gridded (surface) resamplings were performed using `mri_vol2surf`
     bold_reference_wf = init_bold_reference_wf(omp_nthreads=omp_nthreads)
     bold_reference_wf.inputs.inputnode.dummy_scans = config.workflow.dummy_scans
     if sbref_files is not None:
+        from niworkflows.interfaces.images import ValidateImage
+        val_sbref = pe.MapNode(ValidateImage(), name='val_sbref',
+                               iterfield=['in_file'])
+        val_sbref.inputs.in_file = sbref_files
         workflow.connect([
             (val_sbref, bold_reference_wf, [('out_file', 'inputnode.sbref_file')]),
         ])
--- a/fmriprep/workflows/bold/base.py
+++ b/fmriprep/workflows/bold/base.py
@@ -166,29 +166,22 @@ def init_func_preproc_wf(bold_file):
         'Memory resampled/largemem=%.2f/%.2f GB.',
         ref_file, mem_gb['filesize'], bold_tlen, mem_gb['resampled'], mem_gb['largemem'])
 
-    sbref_file = None
     # Find associated sbref, if possible
     entities = layout.parse_file_entities(ref_file)
+    if multiecho:
+        # Remove echo entity to collect all echoes
+        _ = entities.pop('echo', None)
     entities['suffix'] = 'sbref'
     entities['extension'] = ['nii', 'nii.gz']  # Overwrite extensions
-    files = layout.get(return_type='file', **entities)
+
+    sbref_files = layout.get(return_type='file', **entities)
     refbase = os.path.basename(ref_file)
     if 'sbref' in config.workflow.ignore:
         config.loggers.workflow.info("Single-band reference files ignored.")
-    elif files and multiecho:
-        config.loggers.workflow.warning(
-            "Single-band reference found, but not supported in "
-            "multi-echo workflows at this time. Ignoring.")
-    elif files:
-        sbref_file = files[0]
-        sbbase = os.path.basename(sbref_file)
-        if len(files) > 1:
-            config.loggers.workflow.warning(
-                "Multiple single-band reference files found for {}; using "
-                "{}".format(refbase, sbbase))
-        else:
-            config.loggers.workflow.info("Using single-band reference file %s.",
-                                         sbbase)
+    elif sbref_files:
+        sbbase = ','.join([os.path.basename(sbf) for sbf in sbref_files])
+        config.loggers.workflow.info("Using single-band reference file(s) %s.",
+                                     sbbase)
     else:
         config.loggers.workflow.info("No single-band-reference found for %s.",
                                      refbase)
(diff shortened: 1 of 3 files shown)


effigies:
Depends on: poldracklab/niworkflows#408

setup.cfg
```diff
@@ -20,7 +20,7 @@ classifiers =
 [options]
 python_requires = >=3.5
 install_requires =
-    niworkflows @ git+https://github.com/poldracklab/niworkflows.git@6118af2d8ef59abfa79f665da7caad646c40bc5c
+    niworkflows @ git+https://github.com/tsalo/niworkflows.git@14d33d560643f462dc80d4de8fbcca15a7713d25
```
effigies: Can you also update nipype with:

```
nipype @ git+https://github.com/effigies/nipype.git@e206f3d53198c683e959fe1e73426ce5cf362b93

File: setup.cfg [hunk 2]
```diff
@@ -20,13 +20,13 @@ classifiers =
...
     smriprep ~= 0.3.2
     templateflow ~= 0.4.1
     nibabel >=2.2.1
     indexed_gzip >=0.8.8
     nilearn !=0.5.0, !=0.5.1
-    nipype >=1.2.2
+    nipype @ git+https://github.com/effigies/nipype.git@e206f3d53198c683e959fe1e73426ce5cf362b93
```
effigies: Sorry, minor fix upstream.

```suggestion
    nipype @ git+https://github.com/effigies/nipype.git@26b662f0ff48886524577c83da22a0bc68189a08
```

tsalo:
@effigies test_pytest failed due to what looks like a random connection error, unless I'm missing something and the failure is actually meaningful. Would you mind retriggering that test?

effigies:
Done. Thanks for the heads up.

effigies:
@tsalo I merged master, so you'll want to `pull` before doing anything else. Feel free to rebase, if you'd rather not have the pin and merge in the history,

pvelasco:
Hi @tsalo,
Do you have a working Docker image of your `multi-input-reference` branch? I'm trying to test it on some ME-EPI data and I need a Singularity image.
I downloaded your branch, built a Docker image, tagged it as `cbinyu/tsalo_fmriprep`, uploaded it to DockerHub and finally did a `singularity pull` of that image.
When I ran it on some data, I got the following error:
```
  File "/usr/local/miniconda/bin/fmriprep", line 10, in <module>
    sys.exit(main())
  File "/usr/local/miniconda/lib/python3.7/site-packages/fmriprep/cli/run.py", line 312, in main
    opts = get_parser().parse_args()
  File "/usr/local/miniconda/lib/python3.7/site-packages/fmriprep/cli/run.py", line 50, in get_parser
    currentv = Version(__version__)
  File "/usr/local/miniconda/lib/python3.7/site-packages/packaging/version.py", line 277, in __init__
    raise InvalidVersion("Invalid version: '{0}'".format(version))
packaging.version.InvalidVersion: Invalid version: ''
```
So I was wondering if you had a working Docker image that I can `singularity pull`.
Thanks!

tsalo:
@pvelasco I don't have a Docker image for it since I haven't touched this branch or the corresponding niworkflows branch in a while. However, I just resolved the current merge conflicts in https://github.com/nipreps/niworkflows/pull/408 so if you switch the pinned niworkflows version in a current fMRIPrep version's (e.g., master's) setup.cfg to `niworkflows @ git+https://github.com/tsalo/niworkflows.git@7a000cc7ebebffac50f1069ba0ecf8d005cced88` and build from that you should be good.

I'm reticent to do that here because I don't want to spam folks with a dummy commit.

pvelasco:
Hi @tsalo,
I managed to get my a Docker image of your branch. I didn't notice that I had to use the `VERSION` argument in `docker build`.
Now the dataset is being processed. I'll report if the `boldref` image for multi-echo improves.

tsalo:
@pvelasco Nothing should improve yet. I was waiting on nipreps/niworkflows#408 to be merged before trying to implement things in fMRIPrep, so the only change in this PR is the pinned version of nipreps in the setup file. That said, I can probably work on getting those changes implemented in the next few days.

fmriprep/workflows/bold/base.py
```diff
@@ -295,7 +285,11 @@ def init_func_preproc_wf(bold_file):
     # Generate a tentative boldref
     bold_reference_wf = init_bold_reference_wf(omp_nthreads=omp_nthreads)
     bold_reference_wf.inputs.inputnode.dummy_scans = config.workflow.dummy_scans
-    if sbref_file is not None:
+    if sbref_files is not None:
+        from niworkflows.interfaces.images import ValidateImage
+        val_sbref = pe.MapNode(ValidateImage(), name='val_sbref',
+                               iterfield=['in_file'])
+        val_sbref.inputs.in_file = sbref_files
```
oesteban, line 292: shouldn't this be done within the `bold_reference_wf` (as it is done for the bold datasets)?

oesteban: (if @tsalo agrees on this, I'll take care of it)

tsalo: That sounds good. Thank you!

oesteban: just pushed a commit to nipreps/niworkflows#408

fmriprep/workflows/bold/base.py
```diff
@@ -141,52 +143,59 @@ def init_func_preproc_wf(bold_file):
...
+    metadata = layout.get_metadata(ref_file)
+
+    # Drop echo entity for future queries, have a boolean shorthand
+    multiecho = bool(entities.pop("echo", None))
+    if multiecho and (isinstance(bold_file, str) or len(bold_file) == 1):
+        multiecho = False
+        config.loggers.warning(
+            f"Only one echo <{ref_file}> found in a seemingly multi-echo dataset. "
+            "Falling back to single-echo mode."
+        )
```
tsalo, line 168: This would definitely be the case if `echo-idx` was passed, right? Should it raise a warning in that case?

oesteban: You mean ``--echo <idx>``? Yeah, although it seems like a tricky case - do we really want to allow people to process just one echo?

I am happy to account for that, but let's first think about whether selecting echoes should be allowed.

tsalo: The argument is `--echo-idx <idx>`.

If you want to analyze multi-echo like single-echo, it makes sense to just preprocess the echo time closest to a typical single-echo echo time (e.g., 30ms for 3T). That said, I'm not sure how common that is.

oesteban: Well, it's a use-case. Probably not very common but reasonable. Okay, I'll change the message for that situation.

There is another anomalous use case: two echoes. How should fMRIPrep proceed in that case?

tsalo: @emdupre can correct me, but I think the best move would be to process both echoes in parallel without trying to combine them. That way, folks could apply the Bright & Murphy denoising method if they want to, after preprocessing. In that method, you just regress the shorter echo out of the longer echo, so it shouldn't be that big of a deal to wait until after preprocessing to do it.

oesteban: Okay, that will require some refinement here: https://github.com/nipreps/niworkflows/blob/master/niworkflows/utils/bids.py#L199-L201 so that two echoes are returned separately as if they were just BOLD runs.

oesteban: If you all agree, let's address this on another PR and get this one finished.

oesteban:
Okay, a side-by-side comparison of ROI plots seems reassuring:

- Unnoticeable changes for ds005
- Minimal changes for ds054 (none problematic). I attribute these to the fact that the SBRef is actually used now. Then, because subsequent runs of the workflow do not use the SBRef (#1511), the final ROI plot does not really change.
- Finally, ds210 shows some changes but those are also minimal.

I think we can push this one in and keep working on #1511, #2182, #2174, and #2162.

PR #1806, opened 2019-10-04
FIX: Bad results files loads; PIN: nipy/nipype master

effigies:
##  Changes proposed in this pull request

Pinning nipype with nipy/nipype#3061 merged.

Trying to identify whether all of our test failures are related to this.


effigies:
Looking good. Will merge this so @tsalo and @emdupre can drop their pins on my branch.

PR #1820, opened 2019-10-12
FIX: Treat missing field maps as empty list instead of None

### LINKED ISSUE #1818: Boilerplate says it used SDC transforms even with ``--ignore fieldmaps``
As reported in https://neurostars.org/t/can-fmriprep-be-run-without-susceptibility-distortion-correction/5217/3

tsalo:
When field maps are ignored or unavailable, the associated variable is an empty list, but most bools associated with the field maps (i.e., `use_fieldwarp`) assume that they're `None`. 
- Closes #1818. 
- Replaces all uses of `fmaps is not None` with `bool(fmaps)` in workflows/bold/base.py.

The generated boilerplate should now have "the transforms to correct for head-motion" instead of "a single, composite transform to correct for head-motion and susceptibility distortions" when field maps are ignored or not present in the dataset.


tsalo:
Tests appear to be failing because of the bump of `sdcflows` to 0.1.2, where `init_sdc_wf` dropped the `bold_meta` argument.

tsalo:
Yes, when field maps are ignored the generated boilerplate now has

> The BOLD time-series (including slice-timing correction when applied)
were resampled onto their original, native space by applying
**the transforms to correct for head-motion**.

instead of 

> The BOLD time-series (including slice-timing correction when applied)
were resampled onto their original, native space by applying
**a single, composite transform to correct for head-motion and
susceptibility distortions**.

I don't know the downstream effects in `init_bold_std_trans_wf`, `init_ica_aroma_wf`, `init_bold_t1_trans_wf`, but they can only fix possible bugs, right?

Issue #1828, opened 2019-10-15
Add failsafe version detection, to prevent CLI crash


adelavega:
When building an `fmriprep` image directly from github, the version is not set correctly, as defaults to `''`.

As a consequence the CLI crashes, because it can't parse the version:
```
Traceback (most recent call last):
  File "/usr/local/miniconda/bin/fmriprep", line 10, in <module>
    sys.exit(main())
  File "/usr/local/miniconda/lib/python3.7/site-packages/fmriprep/cli/run.py", line 312, in main
    opts = get_parser().parse_args()
  File "/usr/local/miniconda/lib/python3.7/site-packages/fmriprep/cli/run.py", line 50, in get_parser
    currentv = Version(__version__)
  File "/usr/local/miniconda/lib/python3.7/site-packages/packaging/version.py", line 221, in __init__
    raise InvalidVersion("Invalid version: '{0}'".format(version))
packaging.version.InvalidVersion: Invalid version: ''
```

I suggest a failsafe `unknown` version is set if the version is an empty string.

oesteban:
To consider in conjunction with #1779

effigies:
What about 'stable'?

oesteban:
> When building an `fmriprep` image directly from github, the version is not set correctly, as defaults to `''`.

What did you exactly try? I'll need to replicate to understand what is going on. Only now I understand why you proposed this https://github.com/poldracklab/fmriprep/pull/1829/files#diff-b93ab0122f4b2ac6738b896651cac6ccR49-R50

This apparently simple issue might be masking a deeper problem with versioneer and our version tracking infrastructure.

adelavega:
I believe that I git cloned this repository, `cd fmriprep` and `docker build . -t fmriprep`.

Then when I did `docker run` there version was an empty string.

oesteban:
Okay, then I guess you should've passed the VERSION --build-arg:

https://github.com/poldracklab/fmriprep/blob/2e5a95fc3643f3507209bad18407757989f63c5e/.circleci/config.yml#L61-L67

Can you try the following?
```Bash
docker build --rm -t fmriprep \
    --build-arg BUILD_DATE=`date -u +"%Y-%m-%dT%H:%M:%SZ"` \
    --build-arg VCS_REF=`git rev-parse --short HEAD` \
    --build-arg VERSION=$( python get_version.py ) .
```

We probably want to have a Makefile with a build target that includes the build args.

effigies:
[...]

PR #1827, opened 2019-10-15
Update connection names from sMRIPrep's ``anat_preproc_wf``

adelavega:
Fixes #1825

I also bumped smriprep to the current master commit.

PR #1829, opened 2019-10-15
Add --skip-brain-extraction flag

adelavega:
This PR depends on https://github.com/poldracklab/smriprep/pull/116

I also merged https://github.com/poldracklab/fmriprep/pull/1827 into this so I can test it.

Let me know if you rather implement brain extraction detection instead of a flag (or in addition, could change flag to `--force-skip-brain-extraction`

I also added a fix for #1828 so I could test this


oesteban:
Please follow up conversations on #2039

PR #1832, opened 2019-10-16
MAINT: Cleaning up dependencies

oesteban:
- [x] Remove unused MaskEPI interface (moved to niworkflows).
- [x] Remove unused dependencies
- [x] Alphabetical sort of dependencies


effigies:
scikit-image is causing problems in `build_docs` and `from ...interfaces.nilearn import Merge` in `test_pytest`.

oesteban:
Yes, I've realized I can't just remove the module. I'll first migrate stuff to niworkflows and then come back to this.

effigies:
Is it worth the effort to move more things to niworkflows right now? We've had a lot of API-breaking changes in our tangle of sub-packages that's already making it hard to keep everything in sync.

oesteban:
> Is it worth the effort to move more things to niworkflows right now? We've had a lot of API-breaking changes in our tangle of sub-packages that's already making it hard to keep everything in sync.

I've been thinking about this and:
- the migration seems pretty low-risk, we are porting a small section of code to niworkflows where these interfaces will be more exposed to additional uses
- I'm planning to release a niworkflows 1.0.0 sometime soon, I think this migration works well with the major release.
- If there is anything we test more thoroughly that is integration, so hopefully, that also plays in our favor making this little migration painless.

I hope I don't have to regret being writing this...

effigies:
Okay. Just want to say that any change that doesn't happen directly in fMRIPrep seems to lead to cascading re-pins and compatibility updates, and hours of waiting on CI and bugfix releases delayed by days if not weeks. The more things we move outside of fMRIPrep, the more that condition will be hit, so I would strongly discourage moving things before there's a real need elsewhere.

I would like to find some point at which we dramatically slow down these refactors. We might need to put a dependency matrix somewhere so we know what minor versions each project depends on, and keep a maintenance branch for any minor series that is still depended on by a master branch somewhere downstream.

If you have a plan for niworkflows 1.0, that's fine. Just feeling like it's hard to keep up with all the moving parts, especially when trying to divide attention to other projects.

oesteban:
Looking good - I'll merge this in and cut another rc soon to test more widely.

Issue #1835, opened 2019-10-17
Circle's ``build_docs`` is broken


oesteban:
Possibly because of the python 3.8 recent publication:

```
    Complete output (70 lines):
    /tmp/easy_install-zqzjbf9e/scipy-1.3.1/setup.py:386: UserWarning: Unrecognized setuptools command ('-q bdist_egg --dist-dir /tmp/easy_install-zqzjbf9e/scipy-1.3.1/egg-dist-tmp-ec3h9llm'), proceeding with generating Cython sources and expanding templates
      warnings.warn("Unrecognized setuptools command ('{}'), proceeding with "
    Traceback (most recent call last):
      File "/usr/local/lib/python3.8/site-packages/setuptools/sandbox.py", line 154, in save_modules
        yield saved
      File "/usr/local/lib/python3.8/site-packages/setuptools/sandbox.py", line 195, in setup_context
        yield
      File "/usr/local/lib/python3.8/site-packages/setuptools/sandbox.py", line 250, in run_setup
        _execfile(setup_script, ns)
      File "/usr/local/lib/python3.8/site-packages/setuptools/sandbox.py", line 45, in _execfile
        exec(code, globals, locals)
      File "/tmp/easy_install-zqzjbf9e/scipy-1.3.1/setup.py", line 505, in <module>
      File "/tmp/easy_install-zqzjbf9e/scipy-1.3.1/setup.py", line 481, in setup_package
    ModuleNotFoundError: No module named 'numpy'
    
    During handling of the above exception, another exception occurred:
    
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-install-jfexlpb1/statsmodels/setup.py", line 324, in <module>
        setup(name=DISTNAME,
      File "/usr/local/lib/python3.8/site-packages/setuptools/__init__.py", line 144, in setup
        _install_setup_requires(attrs)
      File "/usr/local/lib/python3.8/site-packages/setuptools/__init__.py", line 139, in _install_setup_requires
        dist.fetch_build_eggs(dist.setup_requires)
      File "/usr/local/lib/python3.8/site-packages/setuptools/dist.py", line 717, in fetch_build_eggs
        resolved_dists = pkg_resources.working_set.resolve(
      File "/usr/local/lib/python3.8/site-packages/pkg_resources/__init__.py", line 780, in resolve
        dist = best[req.key] = env.best_match(
[...]

PR #1836, opened 2019-10-17
MAINT: pin python 3.7.4 in circle's build_docs

oesteban:
Addresses #1835 trying to get lucky (as python version is the most significant change w.r.t. to the latest successful build).


oesteban:
Okay, this works. Merging.

Issue #1766, opened 2019-09-11
fMRIprep MNI normalization issue


arnaudleroy59:
Hi, 

I use fMRIprep which is working well with most of my MRI data. But I have an issue with a few patients when using fMRIprep. Coregistration is ok and no issue reported when looking at the html file. However, when I'm looking at the functional coverage, here is the output and there is nearly a flat time course. Native data seem ok. Have you got any explanation and tips to fix this issue?

Best

Arnaud Leroy 

![Pb MNI](https://user-images.githubusercontent.com/55145845/64695537-b0572a80-d49c-11e9-8b49-e12ed2f3ff04.png)

![VTC pb](https://user-images.githubusercontent.com/55145845/64695570-c1a03700-d49c-11e9-99a2-6de49c4f6a76.png)

oesteban:
What you mean by _functional coverage_?

Can you share the reports with us?

jdkent:
Answering @oesteban's questions will provide necessary context, but as a hail mary, if @arnaudleroy59 is talking about an ICA-AROMA denoised brain, has ICA-AROMA specified, and is using fmriprep version < v1.2.0, then this reminds of some of the aroma masks before #1332. 

I'm likely wrong, but I wanted to throw this out there while I was thinking of it.

arnaudleroy59:
On addition we are suing fMRIprep v1.4

arnaudleroy59:
Hi, 

What do you think about these outputs?

Do you need more to answer? If yes, how can I send you?

Thank you in advance for your help

Best regards

Arnaud Leroy

oesteban:
Besides the full html report (which I requested via private email), could you write here the exact name of the file represented in the figure of your first post?

oesteban:
Hi @arnaudleroy59, I've checked on your HTML report and it indeed looks good. Can you provide me the exact name of the file you used for the plot on top?

oesteban:
Okay, I just realized that you added it to the shared file. So, if I understand correctly, the file represented above is `sub-3064SEAHS1_task-rest_space-MNI152NLin2009cAsym_desc-smoothAROMAnonaggr_bold.nii.gz`, correct?

[...]

PR #1838, opened 2019-10-18
FIX: Do not generate ``desc-smoothAROMAnonaggr_bold`` conversions on standard spaces

oesteban:
This PR removes the connection of the ``template`` identifier (which is an _iterable_ and is propagated as such by nipype's default behavior) to the ``space`` input of the data sink.

Following nipype's execution rules, the same input was copied over each time for each input at the ``template`` iterable. So, the same data were written over and over again with different space value. In other words, this PR *does not* address #1766. It just improves the clarity as only one conversion is generated.

Since ``desc-smoothAROMAnonaggr_bold`` is calculated always on ``MNI152NLin6Asym`` the keyword is not necessary anymore for disambiguation. I have concerns about whether removing it completely is a good idea (although BIDS-Derivatives mandated, I guess). WDYT?


oesteban:
Okay, the test is passing and there's no more space- keywords on the denoised files. I'm merging. If we want to bring the space key back to the name, it is as simple as setting it here:

https://github.com/oesteban/fmriprep/blob/ddbf63728b20099fc70d5fc3dd7abcda2a518d35/fmriprep/workflows/bold/outputs.py#L241

effigies:
I think it's a good idea to restore `space-`. It feels like like a good reminder that will make it harder to make mistakes downstream.

PR #1854, opened 2019-11-05
FIX: Do not run STC if SliceTiming metadata is set but empty

oesteban:
Currently, the BIDS-Validator would just raise a Warning when metadata contains:

```
{
  "SliceTiming": []
}
```

### DIFF
--- a/fmriprep/workflows/bold/base.py
+++ b/fmriprep/workflows/bold/base.py
@@ -357,7 +357,7 @@ https://github.com/poldracklab/fmriprep/issues/1655.""")
             fmaps.append({'suffix': 'syn'})
 
         # Short circuits: (True and True and (False or 'TooShort')) == 'TooShort'
-        run_stc = ("SliceTiming" in metadata and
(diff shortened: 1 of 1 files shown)


fmriprep/workflows/bold/base.py
```diff
@@ -357,7 +357,7 @@ def init_func_preproc_wf(
             fmaps.append({'suffix': 'syn'})
 
         # Short circuits: (True and True and (False or 'TooShort')) == 'TooShort'
-        run_stc = ("SliceTiming" in metadata and
+        run_stc = (metadata.get("SliceTiming", False) and
```
effigies: If `None` doesn't work, `[]` isn't going to work.
```suggestion
        run_stc = (bool(metadata.get("SliceTiming")) and
```

effigies:
Have you tested this on a dataset?

oesteban:
Just submitted a job on lonestar5 to run ds001242, which has this problem. Will report back when done.

oesteban:
The fix is working out. Merging :)

Issue #813, opened 2017-11-03
Read the docs always serves the latest Zenodo DOI


chrisfilo:
Instead of the one corresponding to the currently viewed version. Not a big issue, but might be worth looking into one day. It's due to the fact that docs after this injection https://github.com/poldracklab/fmriprep/blob/master/circle.yml#L38 are not pushed to RTD, but built by RTD independently.

effigies:
Maybe we could hack `conf.py` to do the substitution?

oesteban:
related to #1229

effigies:
Was looking at the Zenodo [API](http://developers.zenodo.org/?python#representation), and there's a `prereserve_doi` option, which could allow us to prepare the docs prior to tagging.

It might make sense to create a `Makefile` where `make release` does:

1) Reserves a DOI
2) Adds the DOI to documentation, changelog and zenodo config
3) Tags a release
4) Creates the release text from the changelog (the `hub release` command would be useful here)

erramuzpe:
Still not working, will review it

erramuzpe:
When calling https://fmriprep.readthedocs.io/en/latest/citing.html from the browser I get the latest changes I made to citing.rst and the page works correctly (not showing the annoying "2. 1." issue and with the adequate Zenodo DOI).

If I request https://fmriprep.readthedocs.io/en/**stable**/citing.html or any other version, I receive the old citing.rst (without the changes). 

Any idea of why this behavior?

PR #1855, opened 2019-11-06
FIX: Read the docs always serves the latest Zenodo DOI

erramuzpe:

Trying to solve #813 

There are a pair of issues:

- Annoying 1 (one) after the 2 (two). The first "one" comes from the citation itself. 
`2. *1.* Esteban, Oscar, Blair, Ross(...)`

Solved with `citation.substring(2)`

- Read the docs always serves the latest Zenodo DOI

In [this](https://github.com/poldracklab/fmriprep/blame/278fb84c5407520cdbcc7228c8bcbe8f82d19c6c/docs/citing.rst#L16) line, there's this assignment `var version = 'latest';`. If assigned to the desired version, it changes and works adequately. I guess this can be obtained from the HTML? I've no experience with javascript, I'll give it a try though.

### DIFF
--- a/docs/citing.rst
+++ b/docs/citing.rst
@@ -21,7 +21,7 @@ we recommend to include in your paper.
 
       function cb(err, zenodoID) {
          getCitation(zenodoID, 'vancouver-brackets-no-et-al', function(err, citation) {
-            $('#fmriprep_citation').text(citation);
+            $('#fmriprep_citation').text(citation.substring(2));
          });
          getDOI(zenodoID, function(err, DOI) {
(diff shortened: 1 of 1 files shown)


erramuzpe:
I got the version from the URL :information_desk_person: , IDK if that works for you. I couldn't test it locally.

oesteban:
I'll go ahead and merge this - that's going to be the real test and I feel the risk of a catastrophe is inexistent.

oesteban:
(which means I'll probably break master for a totally unrelated reason :P)

PR #1856, opened 2019-11-06
DOC: docker db size warning

erramuzpe:
### Changes proposed in this pull request

Warning in docs about this satanic issue poldracklab/mriqc#666 :smiling_imp: 

It's a matter of if you like the chunk of code inside the warning or not

### DIFF
--- a/docs/docker.rst
+++ b/docs/docker.rst
@@ -113,3 +113,11 @@ executable in a bare-metal installation: ::
         /data /out/fmriprep-<latest-version> \   | These lines
         participant \                            | correspond to
         -w /work                                 | fmriprep arguments.
+
+.. warning::
+
+   When using docker with big datasets (+10GB) docker might fail. 
+   Changing the maximum size of the container will solve it: ::
+
+    $ service docker stop
+    $ dockerd --storage-opt dm.basesize=30G


erramuzpe:
I will also send a PR for mriQC

docs/docker.rst
```diff
@@ -113,3 +113,11 @@ executable in a bare-metal installation: ::
...
         participant \                            | correspond to
         -w /work                                 | fmriprep arguments.
+
+.. warning::
+
+   When using docker with big datasets (+10GB) docker might fail. 
+   Changing the maximum size of the container will solve it: ::
+
+    $ service docker stop
+    $ dockerd --storage-opt dm.basesize=30G
```
effigies: You should specify what OS/service managers this works with. I'm guessing it's systemd?

erramuzpe: If this is an issue of the working directory not being mounted, shall we just discard it? Or change the warning to "make sure you mounted the working dir"?

emdupre: t might be worth mentioning something about the memory allocated here ? Could optionally link out:
https://docs.docker.com/config/containers/resource_constraints/

emdupre: (if we're adding warnings, but +1 on clarifying re working directory being mounted !)

oesteban:
@erramuzpe may I ask you to:
- add the warning first that we recommend mounting a work directory
- then, if the user does not want to mount the work directory for whatever reason, keep the modifications you already have included in the pr, but indicating that we do not advise this route.
- Add a link to the documentation bit that @emdupre suggested above.

erramuzpe:
Sorry @oesteban I was completely out for the last month, will work on this in the following weeks

oesteban:
The general documentation about containers has been moved over to nipreps.org (in particular, Docker here - https://www.nipreps.org/apps/docker/). The end goal is to identify commonalities and unify the documentation across different projects (fMRIPrep, MRIQC, dMRIPrep...).

This particular fix is relevant, since it has already been pushed in MRIQC's documentation. Please send a PR from https://github.com/nipreps/nipreps.github.io/blob/mkdocs/docs/apps/docker.md .

I'll close this PR in favor of the new documentation site. Please move there all the conversation about what is the best advice we can give to go around this issue.

PR #1860, opened 2019-11-06
FIX: Raise error if work_dir is a child of bids_dir

### LINKED ISSUE #1760: Raise warning/error if working directory is a subdirectory of BIDS root
This is almost guaranteed to cause problems, so perhaps it should be made an error.

Related #1738, #1758.

adelavega:
Closes #1760

### DIFF
--- a/fmriprep/cli/run.py
+++ b/fmriprep/cli/run.py
@@ -538,6 +538,13 @@ def build_workflow(opts, retval):
         retval['return_code'] = 1
         return retval
 
+    if bids_dir in work_dir.parents:
+        build_log.error(
+            'The selected working directory is a subdirectory of the input BIDS folder. '
+            'Please modify the output path.')
+        retval['return_code'] = 1
+        return retval
+
     output_spaces = parse_spaces(opts)
 
     # Set up some instrumental utilities
--- a/wrapper/fmriprep_docker.py
+++ b/wrapper/fmriprep_docker.py
@@ -19,6 +19,7 @@ import os
 import re
 import subprocess
 from warnings import warn
+from pathlib import Path
 
 __version__ = '99.99.99'
 __copyright__ = 'Copyright 2019, Center for Reproducible Neuroscience, Stanford University'
@@ -365,6 +366,12 @@ def main():
         command.extend(['-v', ':'.join((opts.work_dir, '/scratch'))])
         unknown_args.extend(['-w', '/scratch'])
 
+    # Check that work_dir is not a child of bids_dir
+    if Path(opts.bids_dir).resolve() in Path(opts.work_dir).resolve().parents:
+        print(
+            'The selected working directory is a subdirectory of the input BIDS folder. '
+            'Please modify the output path.')
+        return 1
     if opts.config:
         command.extend(['-v', ':'.join((
             opts.config, '/home/fmriprep/.nipype/nipype.cfg', 'ro'))])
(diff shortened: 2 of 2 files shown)


fmriprep/cli/run.py
```diff
@@ -538,6 +538,13 @@ def build_workflow(opts, retval):
         retval['return_code'] = 1
         return retval
 
+    if bids_dir in work_dir.parents:
+        build_log.error(
+            'The selected working directory is a subdirectory of the input BIDS folder. '
+            'Please modify the output path.')
+        retval['return_code'] = 1
+        return retval
```
effigies: Should probably do a similar thing in `fmriprep-docker`, since we could do:

```
fmriprep-docker /a /b participant -w /a/work
```

And it would translate to:

```Shell
docker run --rm -it -v /a:/data:ro -v /b:/out -v /a/work:/scratch \
    poldracklab/fmriprep:<version> /data /out participant -w /scratch
```

And not hit this condition.

adelavega: Nice catch.

wrapper/fmriprep_docker.py
```diff
@@ -19,6 +19,7 @@
 import re
 import subprocess
 from warnings import warn
+from pathlib import Path
```
effigies: This needs to be installable with zero dependencies on Python 2.

adelavega: oh. i forgot about that antiquated technology.

adelavega: Turns out this is a PITA without pathlib. See: https://stackoverflow.com/questions/3812849/how-to-check-whether-a-directory-is-a-sub-directory-of-another-directory/17624617#17624617

From what I can tell the only foolproof solutions are:
```
def os_path_split_asunder(path, debug=False):
    parts = []
    while True:
        newpath, tail = os.path.split(path)
        if debug: print repr(path), (newpath, tail)
        if newpath == path:
            assert not tail
            if path: parts.append(path)
            break
        parts.append(tail)
        path = newpath
    parts.reverse()
    return parts


def is_subdirectory(potential_subdirectory, expected_parent_directory):
    """
    Is the first argument a sub-directory of the second argument?
    """

    def _get_normalized_parts(path):
        return os_path_split_asunder(os.path.realpath(os.path.abspath(os.path.normpath(path))))

    # make absolute and handle symbolic links, split into components
    sub_parts = _get_normalized_parts(potential_subdirectory)
    parent_parts = _get_normalized_parts(expected_parent_directory)

    if len(parent_parts) > len(sub_parts):
        # a parent directory never has more path segments than its child
        return False

    # we expect the zip to end with the short path, which we know to be the parent
    return all(part1==part2 for part1, part2 in zip(sub_parts, parent_parts))
```

or maybe the following if we're okay with not being as thorough:

```
def is_in_directory(filepath, directory):
    return os.path.realpath(filepath).startswith(
        os.path.realpath(directory) + os.sep)
```

adelavega: I tested the latter a bit, and it seems to work even if the directory ends with a sep (e.g. : `/bids/dir/`). That was my main concern with that approach.

effigies: Yeah, quick and dirty is fine. If someone wants to figure out a way to hack around it, they will, and I wish them well with their life choices.

Issue #1573, opened 2019-04-05
Improve description of FreeSurfer reuse in docs


oesteban:
Particularly, write the trick described at the end of https://neurostars.org/t/fmriprep-confounds-error-copied-columns-for-multiple-subjects/4016/6?u=oesteban

Update with the new ``--fs-subjects-dir`` flag from sMRIPrep (when we propagate it into fMRIPrep)

PR #1863, opened 2019-11-06
DOC: Skipping FreeSurfer reconall 

franklin-feingold:
- Adding brief text to help users skip FreeSurfer's reconall skip. Bring more attention to the flag. Partially addresses https://github.com/poldracklab/fmriprep/issues/1573 - waiting for the sMRIPrep flag to propagate into fMRIPrep

- faq


emdupre:
Could we make it clear that users still need an FS license file in this docstring ?

franklin-feingold:
@emdupre added clarity that the FS license is still required

effigies:
This doesn't seem very clear to me. "Skipping" `recon-all` can mean two things to me:

1) Don't do any surface processing (what the flag does)
2) Don't re-run FreeSurfer but use pre-existing surfaces (the flag will prevent this)

If we're going to have this, it needs to be clear what we're talking about. We can link to re-using FreeSurfer directories for case 2.

More minor comments:

* "We have added" - This flag (or another, equivalent) has been around as long as FreeSurfer functionality. Feels weird to say we added it.
* What is passing through what?
* The tool is called `recon-all` (with a hyphen), and the operation is surface reconstruction. I would generally prefer the English, as `recon-all` feels a bit jargony, but your call.
* RST uses double-backticks to indicate code, not single. Single backticks are used for links.
* Please add line breaks for new sentences, and add line breaks within sentences at <99 characters.
* Please start new branches from `master`. These commits are messy:

Screen Shot 2019-11-06 at 16 36 06

You can just:

```Shell
git fetch upstream
git rebase upstream/master
```

franklin-feingold:
this may serve better waiting until the more clear case of reusing FreeSurfer than an ambiguous "skip". It may be the more common use that a user wants to re-use their FreeSurfer outputs rather than not doing surface processing.

Issue #1665, opened 2019-05-31
SyN-SDC fails for non-default template spaces


effigies:
See [Neurostars #4381](https://neurostars.org/t/fmriprep1-4-0-does-not-work-using-syn-distortion-correction/4381).

```
traits.trait_errors.TraitError: The ‘transforms’ trait of an ApplyTransformsInputSpec instance must be a list of items which are an existing file name or ‘identity’, but a value of [’/scratch/fmriprep_wf/single_subject_01_wf/func_preproc_task_facelocalizer_wf/sdc_wf/syn_sdc_wf/ref_2_t1/transform0GenericAffine.mat’, ‘/scratch/fmriprep_wf/single_subject_01_wf/anat_preproc_wf/anat_norm_wf/_template_MNI152NLin6Asym/registration/ants_t1_to_mniInverseComposite.h5’, ‘/usr/local/miniconda/lib/python3.7/site-packages/fmriprep/data/fmap_atlas_2_MNI152NLin6Asym_affine.mat’] <class ‘list’> was specified.

Error setting node input:
Node: atlas_2_ref
input: transforms
results_file: /scratch/fmriprep_wf/single_subject_01_wf/func_preproc_task_facelocalizer_wf/sdc_wf/syn_sdc_wf/transform_list/result_transform_list.pklz
value: [’/scratch/fmriprep_wf/single_subject_01_wf/func_preproc_task_facelocalizer_wf/sdc_wf/syn_sdc_wf/ref_2_t1/transform0GenericAffine.mat’, ‘/scratch/fmriprep_wf/single_subject_01_wf/anat_preproc_wf/anat_norm_wf/_template_MNI152NLin6Asym/registration/ants_t1_to_mniInverseComposite.h5’, ‘/usr/local/miniconda/lib/python3.7/site-packages/fmriprep/data/fmap_atlas_2_MNI152NLin6Asym_affine.mat’]
```

effigies:
We're missing affine transformations from the fieldmap atlas to other templates.

oesteban:
This should go into TemplateFlow (it is one of its inception principles and requirements)

alexlicohen:
Hi all,
I just realized this is still not working; is there a way around this for now?

>Node Name: fmriprep_wf.single_subject_NDARAU530GLJ_wf.func_preproc_task_rest_run_2_wf.sdc_wf.syn_sdc_wf.atlas_2_ref
[...]

PR #1886, opened 2019-11-26
ENH: Upgrade SDCFlows to new API (1.0.0rc1)

### LINKED ISSUE #1884: PIN: SDCFlows 1.0.0rc1 / compatibility with refactored API

oesteban:
Closes #1884.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -30,7 +30,7 @@ install_requires =
     psutil >=5.4
     pybids ~= 0.9.4
     pyyaml
-    sdcflows @ git+https://github.com/oesteban/sdcflows.git@cf28b970aeac9bf2611b7330f53725c8ed81edb1
+    sdcflows ~= 1.0.0
     smriprep ~= 0.4.0
     tedana >=0.0.5
     templateflow ~= 0.4.1
--- a/setup.cfg
+++ b/setup.cfg
@@ -22,7 +22,7 @@ python_requires = >=3.5
 install_requires =
     indexed_gzip >=0.8.8
     nibabel >=2.2.1
-    nipype ~= 1.3.1
+    nipype >=1.3.1
     nitime
     niworkflows ~= 1.0.1
     numpy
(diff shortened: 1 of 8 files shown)


oesteban:
Please note the new condensed view for phase1/2-phasediff-fieldmap kind of fieldmaps (ds054). This PR merges together the magnitude segmentation view, the magnitude registration view and the fieldmap registration view (hopefully you like the new one better).

I think we still need to add a colorbar in Hz to the fieldmap overlay, and make it symmetric bc right now it is a bit confusing. But overall I believe it is much better.

oesteban:
Another focus point (esp. for @effigies) is poldracklab/sdcflows#63, which is necessary for this PR to work.

oesteban:
Last comment - this PR should be then followed by addressing #1892, ensuring we add MNI152NLin2009cAsym to the standard spaces list if SyN is to run.

oesteban:
Should also fix #1665.

AKSoo:
The actual fieldmaps are fine, but the fieldmap report figures have these values spilling over the y-axis.

yspill

oesteban:
That extrapolation is added by FUGUE. I agree we should open an issue in SDCFlows about it, to double-check that the extrapolation is done once the fieldmap has been mapped into the target EPI.

The alternative to this, once again, is implementing poldracklab/sdcflows#14 so that extrapolation is less of a blanket across the PE-axis solution.

effigies:
Sorry, missed this in a flood of notifications... I'll review now.

setup.cfg
```diff
@@ -79,6 +79,7 @@ fmriprep =
     data/boilerplate.bib
     data/itkIdentityTransform.txt
     data/flirtsch/bbr.sch
+    data/reports-spec.yml
```
effigies: For this to correctly work, we might need to remove `include_package_data = True`. I've recently discovered (https://github.com/niflows/nipype1-workflows/pull/3, https://github.com/nipy/nibabel/commit/7120347111d0f0a2af29bb9ba2d133f61fdb17b8) that that line seems to mean *only* use `MANIFEST.in`.

oesteban: If I understand correctly, we either set `include_package_data = False` or add all these objects in the `MANIFEST.in`?

oesteban:
Okay, updated the pin to a released version since the patch is already in SDCFlows/master. That was breaking builds.

setup.cfg
effigies: I think so.

OTOH I think these must have worked previously...

oesteban: Moving the conversation to #1896

Issue #1905, opened 2019-12-06
Unfold AROMA variants of confounds with ``--ica-aroma`` flag


oesteban:
Comes from https://github.com/poldracklab/fmriprep/issues/1049#issuecomment-391382809

When the user requests running ICA-AROMA, most of the regressors in the confounds file should be recalculated if they plan to use the nonaggressively denoised outputs (is this completely correct?).

/cc @kfinc @jdkent @alexlicohen @satra

jdkent:
terms that should be recalculated:
- csf
- white_matter
- global_signal
- std_dvars
- dvars
- t_comp_cor
- a_comp_cor   

terms not necessary to recalculate:
- non_steady_state_outlier 
- framewise_displacement
- trans_x
- trans_y
- trans_z
- rot_x
- rot_y
- rot_z

PR #1877, opened 2019-11-22
DOC: Improving accessibility of confounds description

kfinc:
## Changes proposed in this pull request

This PR addresses issues #489 #1726 #1049 related to improving the accessibility and educational value of fMRIPrep documentation, in particular, description of _Confounds_.

The work was done during fMRIPrep docusprint and is a result of collaborative effort of multiple people, including Pierre Bellec, Elizabeth DuPre, Karolina Finc, James Kent, Chris Markiewicz, Saren Seeley, Ursula Tooley, Hao-Ting Wang. 

You can track our discussion here:
<https://docs.google.com/document/d/14WyhCvEn0G5fbLY-Rk7BQa9ISzYEB2ll3YCjyd-OjB8/edit>  

## Documentation that should be reviewed

We made major edits to the _Confounds_ subsection in `outputs.rst`:

- we added a general explanation of sources of noise in fMRI signal and introduction to confound regression - that may help users to grasp the idea and advantage of multiple regressors returned by fMRIPrep;
- we warned a user of possible misuse of `confounds_regressors.tsv` table, such as (1) regressing out all confounds during denoising procedure, (2) using ICA-AROMA confounds on AROMA-denoised data, and (3) not selecting a subset of CompCor confounds;
- we expanded the list of references by some fresh and informative papers; 
- we proposed a structure for describing confounds in sections: (1) Basic confounds, (2) Parameter expansion of basic confounds, (3) Confounds for outlier detection, (4) ICA-AROMA confounds, (5) CompCor confounds [I will be grateful for the discussion on the proposed structure as it is an alternative to the structure discussed during docusprint];
- we added more description to almost all confound regressors.

There is still much to improve and clarify, but I think that this is a good start.

Issue #1521, opened 2019-02-28
Explain what are the highpass filter parameters


chrisfilo:
1. For the aroma outputs
2. For the cosine confounds
3. Emphasize that other outputs are not bandpass filtered

Issue #1049, opened 2018-04-11
add warning to confounds description page


satra:
1. if i'm already using aroma cleaned data, it does not make sense to further regress the aroma regressors. 
2. also it may not make sense to use global signal regression and a/t compcor 
3. if i'm doing my own high pass filtering, using the cosine filters could interfere

since it's so easy to add the entire confounds.tsv to a feat or other analysis, it may be useful to add a warning that people should selectively add confounds relevant to their specific analysis.

chrisfilo:
We should add this to documentation. Maybe in red.

alexlicohen:
Per the AROMA paper, WM and CSF confounds were regressed after AROMA. How does one do this with fMRIprep? I'm presuming we'll have to extract new timecourses using WM and CSF masks from the AROMA residual nii.gz?

> We validated ICA-AROMA by applying it to rfMRI and stop-signal task fMRI data (SST-fMRI). We compared results obtained with ICA-AROMA to two commonly applied motion correction strategies: extensive nuisance regression including 24 RPs, and spike regression. All three preprocessing strategies were complemented by **standard nuisance regression including time-series for WM and CSF signal** as well as a linear trend to remove residual (non-motion related) structured noise. 

> Therefore, motion artifacts are preferably removed prior to temporal filtering. To accommodate this, we designed ICA-AROMA to be applied prior to temporal filtering. Moreover, applying secondary motion artifact removal prior to temporal filtering not only prevents ringing artifacts but additionally prevents removal of low-frequency motion-related signal variance which can aid in identification of motion artifacts. **Similarly, as signal from WM and CSF time- series might also contain variance related to motion artifacts, we applied ICA-AROMA prior to nuisance regression**. Additionally, we applied ICA-AROMA after spatial smoothing. Spatial smoothing will blur motion-related artifacts to neighboring voxels. Yet, as spatial smoothing will increase signal to noise ratio by reducing speckled noise, it will increase the ability of ICA to detect coherent signal patterns, including better separation of structured artifacts. Note that we did not include slice-timing correction in our preprocessing pipeline. Although slice-timing correction might interact with motion-induced artifacts, it is not expected to substantially alter the characteristics of the motion components and therefore impact the ICA-AROMA classification accuracy.

[...]

PR #1906, opened 2019-12-06
DOC: Refinement of confounds documentation after #1877

### LINKED ISSUE #1167: Add description of CosineX regressors to the documentation
Explain why the number of cosines can change depending on TR, length of the scan and number of outliers.

oesteban:
This PR addresses further issues after #1877:

- Address warnings 2) and 3) of #1049, and I've opened #1905 to follow up conversations
- Add documentation about DCT regressors, close #1167.
- Address items 2) and 3) of #1521.

### DIFF
--- a/docs/outputs.rst
+++ b/docs/outputs.rst
@@ -420,8 +420,16 @@ ICA-AROMA can be enabled with the flag ``--use-aroma``.
 - ``aroma_motion_XX`` - the motion-related components identified by ICA-AROMA.
 
 .. danger::
-    If you are already using ICA-AROMA cleaned data (``~desc-smoothAROMAnonaggr_bold.nii.gz``),
+    If you are already using ICA-AROMA's cleaned-data (``~desc-smoothAROMAnonaggr_bold.nii.gz``),
     do not include ICA-AROMA confounds during your design specification or denoising procedure.
+
+    Additionally, as per [Hallquist2013]_, when using ICA-AROMA's cleaned-data most of the
+    confound regressors should be recalculated (this feature is a work-in-progress, follow up on
+    `#1905 <https://github.com/poldracklab/fmriprep/issues/1905>`__).
+    Surprisingly, `our simulations
+    <https://github.com/poldracklab/fmriprep-notebooks/blob/9933a628dfb759dc73e61701c144d67898b92de0/05%20-%20Discussion%20AROMA%20confounds%20-%20issue-817%20%5BJ.%20Kent%5D.ipynb>`__
+    (with thanks to JD. Kent) suggest that using the confounds as currently calculated by
+    *fMRIPrep* --before denoising-- would be just fine.
 
 .. caution::
     *Nonsteady-states* (or *dummy scans*) in the beginning of every run
@@ -508,6 +516,10 @@ See implementation on :mod:`~fmriprep.workflows.bold.confounds.init_bold_confs_w
   .. [Friston1996] Friston KJ1, Williams S, Howard R, Frackowiak RS, Turner R,
      Movement‐Related effects in fMRI time‐series. Magnetic Resonance in Medicine. 1996.
      doi:`10.1002/mrm.191035031 <https://doi.org/10.1002/mrm.1910350312>`_
+
+  .. [Hallquist2013] Hallquist MN, Hwang K, Luna B. The Nuisance of Nuisance Regression.
+     NeuroImage. 2013. doi:`10.1016/j.neuroimage.2013.05.116
+     <https://doi.org/10.1016/j.neuroimage.2013.05.116>`_
 
   .. [Muschelli2014] Muschelli J, Nebel MB, Caffo BS, Barber AD, Pekar JJ, Mostofsky SH,
      Reduction of motion-related artifacts in resting state fMRI using aCompCor. NeuroImage. 2014.
--- a/docs/outputs.rst
+++ b/docs/outputs.rst
@@ -420,12 +420,12 @@ ICA-AROMA can be enabled with the flag ``--use-aroma``.
 - ``aroma_motion_XX`` - the motion-related components identified by ICA-AROMA.
 
 .. danger::
-    If you are already using ICA-AROMA's cleaned-data (``~desc-smoothAROMAnonaggr_bold.nii.gz``),
+    If you are already using AROMA-cleaned data (``~desc-smoothAROMAnonaggr_bold.nii.gz``),
     do not include ICA-AROMA confounds during your design specification or denoising procedure.
 
-    Additionally, as per [Hallquist2013]_, when using ICA-AROMA's cleaned-data most of the
-    confound regressors should be recalculated (this feature is a work-in-progress, follow up on
-    `#1905 <https://github.com/poldracklab/fmriprep/issues/1905>`__).
+    Additionally, as per [Hallquist2013]_ and [Lindquist2019]_, when using AROMA-cleaned data
+    most of the confound regressors should be recalculated (this feature is a work-in-progress,
+    follow up on `#1905 <https://github.com/poldracklab/fmriprep/issues/1905>`__).
     Surprisingly, `our simulations
     <https://github.com/poldracklab/fmriprep-notebooks/blob/9933a628dfb759dc73e61701c144d67898b92de0/05%20-%20Discussion%20AROMA%20confounds%20-%20issue-817%20%5BJ.%20Kent%5D.ipynb>`__
     (with thanks to JD. Kent) suggest that using the confounds as currently calculated by
@@ -520,6 +520,11 @@ See implementation on :mod:`~fmriprep.workflows.bold.confounds.init_bold_confs_w
   .. [Hallquist2013] Hallquist MN, Hwang K, Luna B. The Nuisance of Nuisance Regression.
      NeuroImage. 2013. doi:`10.1016/j.neuroimage.2013.05.116
      <https://doi.org/10.1016/j.neuroimage.2013.05.116>`_
+
+  .. [Lindquist2019] Lindquist, MA, Geuter, S, and Wager, TD, Caffo, BS,
+     Modular preprocessing pipelines can reintroduce artifacts into fMRI data.
+     Human Brain Mapping. 2019. doi: `10.1002/hbm.24528
+     <https://doi.org/10.1002/hbm.24528>`_
 
   .. [Muschelli2014] Muschelli J, Nebel MB, Caffo BS, Barber AD, Pekar JJ, Mostofsky SH,
      Reduction of motion-related artifacts in resting state fMRI using aCompCor. NeuroImage. 2014.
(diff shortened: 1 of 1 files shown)


oesteban:
Quick link: https://10794-53175327-gh.circle-artifacts.com//0/tmp/src/fmriprep/docs/_build/html/outputs.html

docs/outputs.rst
```diff
@@ -285,16 +288,36 @@ the command line options ``--fd-spike-threshold`` and ``--dvars-spike-threshold`
...
+One characteristic of the cosine regressors is that they are identical for two different
+datasets with the same :abbr:`TR (repetition time)` and the same *effective* number of
+time points (*effective* length).
+It is relevant to mention *effective* because initial time points identified as *nonsteady
+states* are removed before generating the cosine regressors.
+
+.. caution::
+    If the analysis includes your own high-pass filtering step, the ``cosine_XX``
+    regressors may interfere (meaning, they may re-introduce drifts) if included
+    in the design matrix.
```
effigies: I'm a little skeptical of this claim of reintroducing drifts.

What about:

"If your analysis includes separate high-pass filtering, do not also include ``cosine_XX`` regressors in your design matrix."

docs/outputs.rst
```diff
@@ -285,16 +288,36 @@ the command line options ``--fd-spike-threshold`` and ``--dvars-spike-threshold`
...
+
+.. admonition:: See also
+
+    - A detailed explanation about temporal high-pass filtering is provided with
+      the `BrainVoyager User Guide
+      <http://www.brainvoyager.com/bvqx/doc/UsersGuide/Preprocessing/TemporalHighPassFiltering.html>`_.
```
effigies:
```suggestion
      <https://www.brainvoyager.com/bvqx/doc/UsersGuide/Preprocessing/TemporalHighPassFiltering.html>`_.
```

docs/outputs.rst
```diff
@@ -366,6 +377,42 @@ For CompCor decompositions, entries include:
...
+
+.. caution::
+    Similarly, if you are using anatomical or temporal CompCor it may not make sense
+    to use global signal regressors (``csf``, ``white_matter``, or ``global_signal``) -
+    see `#1049 <https://github.com/poldracklab/fmriprep/issues/1049>`_.
+
+.. danger::
+    *fMRIPrep* does high-pass filtering before running anatomical or temporal CompCor.
+    Therefore, when using CompCor regressors, the corresponding ``cosine_XX`` regressors
+    should be included too in the design matrix.
```
effigies:
```suggestion
    should also be included in the design matrix.
```

docs/outputs.rst
```diff
@@ -366,6 +377,42 @@ For CompCor decompositions, entries include:
...
+**AROMA confounds**.
+:abbr:`AROMA (Automatic Removal Of Motion Artifacts)` is an :abbr:`ICA (independent components analysis)`
+based procedure to identify confounding time series related to head-motion [Prium2015]_.
+ICA-AROMA can be enabled with the flag ``--use-aroma``.
+
+- ``aroma_motion_XX`` - the motion-related components identified by ICA-AROMA.
+
+.. danger::
+    If you are already using ICA-AROMA cleaned data (``~desc-smoothAROMAnonaggr_bold.nii.gz``),
+    do not include ICA-AROMA confounds during your design specification or denoising procedure.
```
effigies: Let's mention that dummy scans are dropped before AROMA is performed, and therefore any analysis of these must drop the same dummy scans.

oesteban: Done

docs/outputs.rst
```diff
@@ -366,6 +376,48 @@ For CompCor decompositions, entries include:
...
+    The original Behzadi aCompCor implementation [Behzadi2007]_ can be applied using
+    components from the combined masks, while the more recent Muschelli implementation
+    [Muschelli2014]_ can be applied using
+    the :abbr:`WM (white matter)` and :abbr:`CSF (cerebro-spinal fluid)` masks.
+    To determine the provenance of each component, consult the metadata file (described above).
+
+.. caution::
+    Similarly, if you are using anatomical or temporal CompCor it may not make sense
+    to use global signal regressors (``csf``, ``white_matter``, or ``global_signal``) -
+    see `#1049 <https://github.com/poldracklab/fmriprep/issues/1049>`_.
```
rciric: Although I'd probably not recommend including both compartmental aCompCor time series and mean compartment (CSF, WM) time series in a single model, I don't agree that it doesn't make sense to use the overall global signal. In particular, @lindenmp and colleagues found in their evaluation that there can be an added benefit of GSR+aCompCor over and above aCompCor alone.

See the green aCompCor bars, particularly in B and C (fig. 1 from Parkes et al.):

docs/outputs.rst
```diff
@@ -366,6 +376,48 @@ For CompCor decompositions, entries include:
     Entries that are not saved in the data file for denoising are still stored in metadata with the
     ``dropped`` prefix.
 
+.. caution::
+    Only a subset of these CompCor decompositions should be used for further denoising.
+    The original Behzadi aCompCor implementation [Behzadi2007]_ can be applied using
+    components from the combined masks, while the more recent Muschelli implementation
+    [Muschelli2014]_ can be applied using
+    the :abbr:`WM (white matter)` and :abbr:`CSF (cerebro-spinal fluid)` masks.
+    To determine the provenance of each component, consult the metadata file (described above).
```
rciric: It might be worth mentioning the option to include either a fixed number of regressors or sufficient regressors to explain some minimum fraction of variance (e.g., 50%), both of which are possible using the confounds generated by fMRIPrep.

For instance:
```
.. caution::
    There are many valid ways of selecting CompCor components for further denoising.
    In general, the components with the largest singular values (i.e., those that
    explain the largest fraction of variance in the data) should be selected.
    *fMRIPrep* outputs components in descending order of singular value.
    Common approaches include selecting a fixed number of components (e.g., the
    first 5 or 6), using a quantitative or qualitative criterion (e.g., elbow, broken
    stick, or condition number), or using sufficiently many components that a minimum
    cumulative fraction of variance is explained (e.g., 50%)
```

docs/outputs.rst
oesteban: /cc @satra

oesteban:
I'd really love to have a green tick from @kfinc and/or @jdkent before merging this...

docs/outputs.rst
```diff
@@ -366,6 +376,59 @@ For CompCor decompositions, entries include:
...
+    to CompCor in general and *fMRIPrep*'s implementation in particular.
+
+**AROMA confounds**.
+:abbr:`AROMA (Automatic Removal Of Motion Artifacts)` is an :abbr:`ICA (independent components analysis)`
+based procedure to identify confounding time series related to head-motion [Prium2015]_.
+ICA-AROMA can be enabled with the flag ``--use-aroma``.
+
+- ``aroma_motion_XX`` - the motion-related components identified by ICA-AROMA.
+
+.. danger::
```
jdkent: Time to bring up the debate again.
If we abide by [Hallquist](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3759585/pdf/nihms-492938.pdf), then we should mention that we should recalculate the confounds listed in [#1905](https://github.com/poldracklab/fmriprep/issues/1905#issuecomment-562906047) following the strategies listed on [neurostars](https://neurostars.org/t/fmriprep-ica-aroma-filtering-including-wm-csf-etc-confounds-in-fsl-regfilt/3137/3?u=jdkent). However, if you trust the [simulations I ran](https://github.com/poldracklab/fmriprep-notebooks/blob/9933a628dfb759dc73e61701c144d67898b92de0/05%20-%20Discussion%20AROMA%20confounds%20-%20issue-817%20%5BJ.%20Kent%5D.ipynb), then using the original components would be fine. This problem will eventually need to be settled, but I think presenting both sides currently is okay.

docs/outputs.rst
```diff
@@ -366,6 +376,67 @@ For CompCor decompositions, entries include:
...
+    If you are already using ICA-AROMA's cleaned-data (``~desc-smoothAROMAnonaggr_bold.nii.gz``),
```
effigies: "ICA-AROMA's cleaned-data" is weird. Both the `'s` and the hyphen between cleaned and data. You could do

```suggestion
    If you are already using ICA-AROMA-cleaned data (``~desc-smoothAROMAnonaggr_bold.nii.gz``),
```

Or perhaps

```suggestion
    If you are already using AROMA-cleaned data (``~desc-smoothAROMAnonaggr_bold.nii.gz``),
```

docs/outputs.rst
jdkent: In additional Hallquist, This other paper by [Linquist](https://onlinelibrary.wiley.com/doi/full/10.1002/hbm.24528) is important to add. They show the detrimental impact of using "modular pipelines", which in this case means denoising in multiple steps.

docs/outputs.rst
```diff
@@ -446,6 +517,10 @@ See implementation on :mod:`~fmriprep.workflows.bold.confounds.init_bold_confs_w
      Movement‐Related effects in fMRI time‐series. Magnetic Resonance in Medicine. 1996.
      doi:`10.1002/mrm.191035031 <https://doi.org/10.1002/mrm.1910350312>`_
 
+  .. [Hallquist2013] Hallquist MN, Hwang K, Luna B. The Nuisance of Nuisance Regression.
+     NeuroImage. 2013. doi:`10.1016/j.neuroimage.2013.05.116
+     <https://doi.org/10.1016/j.neuroimage.2013.05.116>`_
+
```
jdkent:
```suggestion

.. [Lindquist2019] Lindquist, MA, Geuter, S, and Wager, TD, Caffo, BS,
Modular preprocessing pipelines can reintroduce artifacts into fMRI data.
Human Brain Mapping. 2019. doi: `10.1002/hbm.24528
<https://doi.org/10.1002/hbm.24528>`_

```

docs/outputs.rst
```diff
@@ -366,6 +376,67 @@ For CompCor decompositions, entries include:
...
+based procedure to identify confounding time series related to head-motion [Prium2015]_.
+ICA-AROMA can be enabled with the flag ``--use-aroma``.
+
+- ``aroma_motion_XX`` - the motion-related components identified by ICA-AROMA.
+
+.. danger::
+    If you are already using ICA-AROMA's cleaned-data (``~desc-smoothAROMAnonaggr_bold.nii.gz``),
+    do not include ICA-AROMA confounds during your design specification or denoising procedure.
+
+    Additionally, as per [Hallquist2013]_, when using ICA-AROMA's cleaned-data most of the
```
jdkent:
```suggestion
    Additionally, as per [Hallquist2013]_ and [Lindquist2019]_, when using ICA-AROMA's cleaned-data most of the
```

oesteban:
```suggestion
    Additionally, as per [Hallquist2013]_ and [Lindquist2019]_, when using AROMA-cleaned data most of the
```

kfinc:
Hi! Sorry for the delay, I was out of the office for 1 week. Confounds page looks great! Many thanks for incorporating all the changes.

PR #1908, opened 2019-12-07
[FIX] Respect ``--dummy-scans 0``

### LINKED ISSUE #1904: fmriprep ignores --dummy-scans 0 when calculating aCompCor
I think this is a bug, but if it's the desired behavior perhaps it's worth mentioning in the documentation. When `fmriprep` detects non-steady-state volumes, those volumes seem to be excluded from the aCompCor calculations, regardless of whether the `--dummy-scans` flag specifies that there are no non-steady-state volumes. For example, one component will be missing from a run in which the html report says "Non-steady-state volumes: 0 (Warning: 1 automatically detected)".

```
import pandas as pd
import numpy as np

json_file = 'run-with-non-stead-state-detected_desc-confounds_regressors.json'
regressjson = pd.read_json(json_file)
n_comps = regressjson.loc[:,np.asarray(regressjson.loc['Mask'] == 'combined')].columns.size
# in a run with 300 TRs, n_comps is less than 300 when non-steady-state 
# volumes were automatically detected
```

Ran with `fmriprep-docker`
fMRIPrep version: 1.4.0
fMRIPrep command: `/usr/local/miniconda/bin/fmriprep /data /out participant --output-space T1w fsaverage fsnative --ignore slicetiming --participant_label 05 06 --longitudinal --n_cpus 6 --dummy-scans 0 -w /scratch`

jdkent:
fixes #1904 
this pull request only updates pin to niworkflows where poldracklab/niworkflows#438 was merged

N/A

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -24,7 +24,7 @@ install_requires =
     nibabel >=2.2.1
     nipype >=1.3.1
     nitime
-    niworkflows ~= 1.0.1
+    niworkflows @ git+https://github.com/poldracklab/niworkflows.git@fe3eaea17cab1d8afb4c053e4f2f4d5053c45747
     numpy
     pandas
     psutil >=5.4
--- a/setup.cfg
+++ b/setup.cfg
@@ -24,7 +24,7 @@ install_requires =
     nibabel >=2.2.1
     nipype >=1.3.1
     nitime
-    niworkflows @ git+https://github.com/poldracklab/niworkflows.git@89e46ccdc5d2aadc14e6a18b22ea51000e456a3b
+    niworkflows ~= 1.0.2
     numpy
     pandas
     psutil >=5.4
--- a/setup.cfg
+++ b/setup.cfg
@@ -24,7 +24,7 @@ install_requires =
     nibabel >=2.2.1
     nipype >=1.3.1
     nitime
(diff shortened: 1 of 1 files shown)


setup.cfg
```diff
@@ -24,7 +24,7 @@ install_requires =
     nibabel >=2.2.1
     nipype >=1.3.1
     nitime
-    niworkflows ~= 1.0.1
+    niworkflows @ git+https://github.com/poldracklab/niworkflows.git@fe3eaea17cab1d8afb4c053e4f2f4d5053c45747
```
effigies:
```suggestion
    niworkflows @ git+https://github.com/poldracklab/niworkflows.git@19584f3030302e1b8b9d7cb4ae96a040f10e53d6
```

setup.cfg
```diff
@@ -24,7 +24,7 @@ install_requires =
     nibabel >=2.2.1
     nipype >=1.3.1
     nitime
-    niworkflows ~= 1.0.1
+    niworkflows @ git+https://github.com/poldracklab/niworkflows.git@89e46ccdc5d2aadc14e6a18b22ea51000e456a3b
```
effigies:
```suggestion
    niworkflows ~= 1.0.2
```

PR #1915, opened 2019-12-17
ENH: Rework output spaces to be explicit

### LINKED ISSUE #1892: Separate between ``--output-spaces`` and internal standard spaces
The concept of ``--output-spaces`` should be separate of the internal standard spaces we use as target for normalizations.

Right now, we write some warnings along the lines "--ica-aroma was enabled but the MNI152NLin6Asym space was not found in the --output-spaces list", which is not informative and very confusing.

We should add normalization templates independently of what --output-spaces the user wants. Of course if one space is in the --output-spaces, normalization to that space needs to be included.

This issue is also to remind ourselves to drop the deprecated CLI arguments regarding spaces.

mgxd:
Closes #1892 

- Distinguish user defined `--output-spaces` from internal intermediary steps. This PR ensures that spaces not requested (i.e. `fsaverage5` when running with freesurfer, `fsLR` surfaces when using `--cifti-output`, etc) are not included in the final output.
  - the `no-output` keyword is used to distinguish handling these intermediary steps.

- Change of various `init_wf`s function signature to better reflect the current spaces:
   - `output_spaces` --> `target_spaces`, since extraneous spaces are also included.


fmriprep/cli/run.py
```diff
@@ -697,48 +697,41 @@ def build_workflow(opts, retval):
...
+
+    Certain options require normalization to a space not explicitly defined by users.
+    These spaces are labeled with a unique key ``no-output`` set to ``True``,
+    and will not be included in the final outputs.
+    """
     from collections import OrderedDict
     # Set the default template to 'MNI152NLin2009cAsym'
-    output_spaces = opts.output_spaces or OrderedDict([('MNI152NLin2009cAsym', {})])
+    target_spaces = opts.output_spaces or OrderedDict(
+        [('MNI152NLin2009cAsym', {'no-output': True})]
```
mgxd: I'm still a bit iffy on this change, so would like to draw attention here for any future reviews

mgxd:
Hm, what do you think of this rough idea?

```python
from collections import namedtuple

Space = namedtuple("Space", "template specs output")

def _create_space(template, specs=None, output=True):
    if specs is None:
        specs = {}
    return Space(template=template, specs=specs, output=output)
```

then we could filter based on `Space` fields, i.e.

```python
spaces = []
spaces.append(_create_space('fsLR'))
spaces.append(_create_space('MNI152', specs={'res': '2'}))
spaces.append(_create_space('fsaverage', output=False))

spaces
[Space(template='fsLR', specs={}, output=True), Space(template='MNI152', specs={'res': '2'}, output=True), Space(template='fsaverage', specs={}, output=False)]

output_spaces = [space for space in spaces if space.output]

output_spaces
[Space(template='fsLR', specs={}, output=True), Space(template='MNI152', specs={'res': '2'}, output=True)]
```

oesteban:
It feels inflexible to try to keep both concepts together. I agree there is some overlap, but it feels an overkill to try to come up with a data structure when we can deal with a vanila lists, tuples and dicts.

For the internal spaces you don't need to specify modifiers (with the exception of `cohort` which is considered a new, different space) - we use the best option (e.g. res-1 in general and res-2 if running with ``--sloppy``, or the `fsaverage7` with `--cifti-outputs`).

For the output spaces, we should keep it as closest as possible to the input ``--output-spaces`` argument (including under this umbrella, e.g., ``--cifti-outputs den-170k``). Here we do want to be able to keep the extra selectors/resampling options (e.g., `:native:res-1:res-2`). Otherwise, we'll not be able to generate all the outputs prescribed by the user.

Example:

```
fmriprep blah blah --output-spaces MNI152Lin:res-1:res-2:bold \
                   --cifti-outputs den-91k \
                   --ica-aroma
```
requires us to have the following:
```Python
output_std_spaces = [
    ('MNI152Lin', {'resolution': 1}),
    ('MNI152Lin', {'resolution': 2}),
    ('MNI152Lin', {'resolution': 'bold'}),
    ('fsLR', {'density': '91k'})
]
```
because this is what the user wants to see under derivatives/

and, simultaneously:
```Python
internal_std_spaces = [
    'MNI152Lin',  # Added by the user
    'MNI152NLin6Asym',  # Added with `--ica-aroma`
    'MNI152NLin2009cAsym',  # Added by fMRIPrep, needed for, e.g., carpetplot
    ('fsaverage', {'density': '168k'}),  # Added with `--ica-aroma`
    ('fsLR', {'density': '91k'}),  # Not sure this is necessary, actually.
]
```

.circleci/config.yml
```diff
@@ -657,6 +658,7 @@ jobs:
                 --config $PWD/nipype.cfg -w /tmp/ds210/work \
                 /tmp/data/ds210 /tmp/ds210/derivatives participant \
                 --fs-no-reconall --sloppy --write-graph \
+                --output-spaces MNI152NLin2009cAsym \
```
effigies: Why on for anatomical, off for full run? And isn't this one of the default output spaces?

.circleci/ds210_outputs.txt
```diff
@@ -28,10 +28,5 @@ fmriprep/sub-02/anat/sub-02_space-MNI152NLin2009cAsym_label-WM_probseg.nii.gz
 fmriprep/sub-02/func
 fmriprep/sub-02/func/sub-02_task-cuedSGT_run-01_desc-confounds_regressors.json
 fmriprep/sub-02/func/sub-02_task-cuedSGT_run-01_desc-confounds_regressors.tsv
-fmriprep/sub-02/func/sub-02_task-cuedSGT_run-01_space-MNI152NLin2009cAsym_boldref.nii.gz
-fmriprep/sub-02/func/sub-02_task-cuedSGT_run-01_space-MNI152NLin2009cAsym_desc-brain_mask.json
-fmriprep/sub-02/func/sub-02_task-cuedSGT_run-01_space-MNI152NLin2009cAsym_desc-brain_mask.nii.gz
-fmriprep/sub-02/func/sub-02_task-cuedSGT_run-01_space-MNI152NLin2009cAsym_desc-preproc_bold.json
-fmriprep/sub-02/func/sub-02_task-cuedSGT_run-01_space-MNI152NLin2009cAsym_desc-preproc_bold.nii.gz
```
effigies: It seems really weird to produce regressors but not BOLD series.

fmriprep/cli/run.py
```diff
@@ -195,7 +190,7 @@ def get_parser():
     #  ANTs options
     g_ants = parser.add_argument_group('Specific options for ANTs registrations')
     g_ants.add_argument(
-        '--skull-strip-template', action='store', default='OASIS30ANTs', type=_output_space,
+        '--skull-strip-template', action='store', default='OASIS30ANTs', type=Space.from_string,
```
effigies: Does the error message look reasonable if an unexpected input is received?

fmriprep/cli/run.py
```diff
@@ -134,7 +129,7 @@ def get_parser():
              'option is not enabled, standard EPI-T1 coregistration is performed '
              'using the middle echo.')
     g_conf.add_argument(
-        '--output-spaces', nargs='+', action=ParseTemplates,
+        '--output-spaces', nargs='+', action=OutputSpaceAction, default=SpatialReferences(),
```
effigies: Also here.

fmriprep/workflows/bold/outputs.py
```diff
@@ -149,65 +156,92 @@ def init_func_derivatives_wf(
...
 
         extract_surf_info = pe.MapNode(niu.Function(function=_extract_surf_info),
                                        iterfield=['in_file'], name='extract_surf_info',
                                        mem_gb=DEFAULT_MEMORY_MIN_GB, run_without_submitting=True)
-        extract_surf_info.inputs.density = fslr_density
 
         name_surfs = pe.MapNode(GiftiNameSource(
-            pattern=r'(?P<LR>[lr])h.(?P<space>\w+).gii',
+            pattern=r'(?P<LR>[lr])h.\w+',
             template='space-{space}{den}_hemi-{LR}.func'),
```
effigies: Where is the `{space}` variable coming from, now?

Issue #1921, opened 2020-01-02
Upgrading from 1.5.2 to 1.5.4


soichih:
I am trying to upgrade from fmriprep 1.5.2 to 1.5.4. 

If I run fmrprep 1.5.4 with the same command line argument that worked for 1.5.2, the 1.5.4 now complains that I am missing an option.

```
    Running fMRIPREP version 1.5.4:
      * BIDS dataset path: /mnt/scratch/hayashis/syncthing/git/app-fmriprep/fmripinput.
      * Participant list: ['TTTEMPSUB'].
      * Run identifier: 20200102-042619_5fdc5d07-5cca-4785-9028-ad271e9d524f.
    
Process Process-2:
Traceback (most recent call last):
  File "/usr/local/miniconda/lib/python3.7/multiprocessing/process.py", line 297, in _bootstrap
    self.run()
  File "/usr/local/miniconda/lib/python3.7/multiprocessing/process.py", line 99, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/local/miniconda/lib/python3.7/site-packages/fmriprep/cli/run.py", line 687, in build_workflow
    work_dir=str(work_dir),
TypeError: init_fmriprep_wf() missing 1 required positional argument: 'fs_subjects_dir'

```

So I've added the following command line option (fs_subjects_dir) to point to where my freesurfer output is.

>  --fs-subjects-dir="$outdir/freesurfer/$bidsSub" \

However, now fmrprep is complaining that there is no such option.

> fmriprep: error: unrecognized arguments: --fs-subjects-dir=fmripout/freesurfer/sub-TTTEMPSUB

I've tried without "=" and with fs_subjects_dir instead of fs-subjects-dir, but none of that seems to fix the problem. Could you help me troubleshoot?

mgxd:
hi @soichih - can you share the full command?

soichih:
@mgxd 

I was able to find the cause of the fs-subjects-dir issue (I think), but now I am seeing a different error message that I didn't see when I was running 1.5.2 version.

```
200106-17:19:07,412 nipype.workflow WARNING:
	 Storing result file without outputs
200106-17:19:07,413 nipype.workflow WARNING:
[...]

PR #1932, opened 2020-01-13
FIX: Only pass volumetric spaces to carpetplot

effigies:

We had a situation where `['fsaverage', 'MNI152NLin2009c']` was being passed as the set of spaces to `carpetplot_wf`, but only the transforms for MNI were being passed, so the indexing was getting screwed up. It appears that the resolution is simply to pass the volumetric spaces.

This resolves the specific failure documented in https://github.com/poldracklab/fmriprep/issues/1921#issuecomment-572206926.


effigies:
@mgxd @oesteban Any concerns?

fmriprep/workflows/bold/base.py
```diff
@@ -817,7 +825,7 @@ def init_func_preproc_wf(
 
         if 'MNI152NLin2009cAsym' in std_spaces:
             carpetplot_wf = init_carpetplot_wf(
-                standard_spaces=std_spaces,
+                standard_spaces=volume_std_spaces,
```
oesteban: I believe this line needs to go away, and pass the list of spaces through a connection to the collated list of templates.
```suggestion
```
That means adding a line right after:
https://github.com/poldracklab/fmriprep/pull/1932/files#diff-e705526667f071ca0c3b04c2bd96272bR834

Like
```Python
                    ('joint_std2anat_xfm', 'inputnode.std2anat_xfm'),
                    ('joint_template', 'inputnode.templates')]),
```

Unfortunately, the carpetplot workflow would not support this directly:

- we need to remove the function argument
- we need to add a new inputnode field
- we need to then connect the new field to `select_std.inputs.keys`

effigies:
Bearing in mind that this is a bug-fix, and the final fix will be #1915, is there a use-case that your suggestion fixes that this doesn't? I would like to keep this as small as possible to reduce the potential for new bugs.

Will wait on response before diving into effigies#2.

effigies:
Merged effigies#2. Needed small fixes that I added.

@mgxd If we're replacing `KeySelect` anyway, this just moves it from one subworkflow to another. Apologies if it causes merge conflicts.

Any objections to merging on passing tests?

mgxd:
Nope

Issue #1842, opened 2019-10-25
Strategies for alignment to T1 for small EPI slabs


JWinawer:
Dear All,

We use fMRIprep for a number of our studies.  While it usually works quite well, we have found a problem in one study. This study uses a small FOV for EPIs (in the slice dimension). From the same session, some of the EPIs, all of which are coronal, get correctly aligned to the T1, but many are way off (eg rotated 150 to 200 deg around an R-L axis).

In contrast, our [home-brewed pre-processing tool ](https://github.com/WinawerLab/MRI_tools/blob/master/preprocessing/prisma_preproc.py)(written by @sfavila and @billbrod), which takes a different alignment strategy, works fine. The strategy in that tool is:

In addition to unwarping the EPIs using distortion scans, we also unwarp the distortion scan itself. We then compute the alignment between the distortion scan and the T1, rather than the EPIs and the T1. This works better than aligning the EPIs to the T1 because the distortion scan is spin echo and has  better quality. We then compute the alignment of each EPI to the distortion scan. This works better than aligning the EPI to the T1 because the EPIs and the distortion scans have the identical acquisition volume. (We save out the alignment matrices, and then only apply a single resampling).

So the question is, is there any way to implement this alignment strategy in fMRIprep? Specifically, to align each EPI to the (unwarped) distortion scan, and to then align the (unwarped) distortion scan to  the whole-brain T1? 

Thanks
Jon

effigies:
Hi, thanks a lot for this question. Limited-FoV is definitely an issue that fMRIPrep has not found a good general solution for, so I'm glad to see you've also been thinking about this.

This doesn't sound very far out of line with what we do. We currently find some reference image to use as stand-in for the BOLD series, whether that's an sbref, dummy scans at the start of the BOLD series, or motion-corrected and averaged first 20 (or so) scans. This is the target of HMC, the object of SDC correction, and the SDC-corrected boldref is the object of BOLD-T1w registration. Similarly, every EPI volume is sampled to the target space in a single resampling.

If I'm reading this correctly, you currently replace the boldref with the reverse-PE EPI, post-SDC correction, for both HMC and coregistration. That seems pretty doable. It would result in a bit of reordering some operations, but the overall structure of the workflow would be minimally affected.

[...]

PR #1934, opened 2020-01-15
ENH: Adds the `--bold2t1w-init` arg, which can be set to `'header'` when functional and anatomical volumes are collected in the same session/have good approximate starting registrations

noahbenson:

This PR adds the `--coreg-init-header` flag, which instructs the BOLD coregistration pipelines to pass the `init='header'` option to `BBRegister` and the `init_header=True` option to `MRICoreg`. This provides a solution to issue #1842 (we have verified that this PR fixes the processing of the data that spawned this issue).

It's likely that some of this code has not been written using the conventions of fmriprep; if someone can give me some guidance on this, I'd be happy to clean it up before any merge.

The doc-strings of all functions that accept/use/pass-along the `coreg_init_header` option have been updated to reflect the option. I am not certain if there are other places in the repo that should also have updated documentation, but I'm happy to make those changes if someone points me in the right direction.


fmriprep/workflows/base.py
```diff
@@ -70,6 +70,7 @@ def init_fmriprep_wf(
     use_bbr,
     use_syn,
     work_dir,
+    coreg_init_header=False,
```
effigies: We don't pass default values. This was an explicit decision made a long time ago. IIRC it was to avoid silent failures to pass values where appropriate.

```suggestion
    coreg_init_header,
```

fmriprep/cli/run.py
```diff
@@ -222,6 +222,9 @@ def get_parser():
         '--fs-license-file', metavar='PATH', type=Path,
         help='Path to FreeSurfer license key file. Get it (for free) by registering'
              ' at https://surfer.nmr.mgh.harvard.edu/registration.html')
+    g_fs.add_argument(
+        '--coreg-init-header', action='store_true', dest='coreg_init_header', default=False,
+        help='Use header information to initialize volume corergistration')
```
effigies: This should not be a FreeSurfer-specific option, as we also have an FSL implementation that should behave the same.

I would suggest something like `--bold2t1w-init` and take the options "register" and "header", defaulting to "register". This pairs it naturally with `--bold2t1w-dof`, IMO.

fmriprep/workflows/bold/registration.py
```diff
@@ -449,7 +456,7 @@ def init_bbreg_wf(use_bbr, bold2t1w_dof, omp_nthreads, name='bbreg_wf'):
 
     mri_coreg = pe.Node(
         MRICoregRPT(dof=bold2t1w_dof, sep=[4], ftol=0.0001, linmintol=0.01,
-                    generate_report=not use_bbr),
+                    generate_report=not use_bbr, no_cras0=init_header),
         name='mri_coreg', n_procs=omp_nthreads, mem_gb=5)
```
effigies: What purpose is `mri_coreg` serving now? Should we just remove the node if we're using the header?

fmriprep/workflows/bold/registration.py
```diff
@@ -489,10 +496,13 @@ def init_bbreg_wf(use_bbr, bold2t1w_dof, omp_nthreads, name='bbreg_wf'):
...
-        BBRegisterRPT(dof=bold2t1w_dof, contrast_type='t2', registered_file=True,
-                      out_lta_file=True, generate_report=True),
-        name='bbregister', mem_gb=12)
+    if init_header:
+        bbregister = BBRegisterRPT(dof=bold2t1w_dof, contrast_type='t2', registered_file=True,
+                                   out_lta_file=True, generate_report=True, init='header')
+    else:
+        bbregister = BBRegisterRPT(dof=bold2t1w_dof, contrast_type='t2', registered_file=True,
+                                   out_lta_file=True, generate_report=True)
+    bbregister = pe.Node(bbregister, name='bbregister', mem_gb=12)
```
effigies: This could be done instead by reverting and adding:

```Python
if init_header:
    bbregister.inputs.init = 'header'
```

That said, if we want to switch to `register` and `header`, it might end up being slightly different. Just want to draw your attention.

fmriprep/workflows/bold/registration.py
noahbenson: This is a good question--I don't have experience using `mri_coreg` myself, so I'm not sure how valuable it is in this situation. Regardless, skipping mri_coreg was also my intuition when I first started--i.e., that the mri_coreg node should not be linked up if `init_header` is true. However, this turned out to be substantially harder to get working--my edits removing mri_coreg resulted in a successful registration, but they disrupted downstream nodes that I'm not familiar with, so I opted to go for this (minimal) change to the existing pipeline.

fmriprep/workflows/base.py
noahbenson: I don't totally follow the policy here: you don't specify default values in [the API / any / some] function definitions? Where should the default values be coded up?

effigies: The policy is in response to this possible case: Imagine you set `coreg_init_header=False`, and then some time you call it, you forget to add `coreg_init_header=coreg_init_header`, then it gets disconnected from the option the user passes to the CLI, and tracking it down can be a headache.

The policy applies to the `init_*_wf()` functions, with the exception of things like `name`, which are permissible parameters but don't correspond with high-level knobs on functionality.

In this case the default value is in the CLI argument parser. If what needs to be passed through requires processing into some other value(s) then it might get a bit more complicated.

fmriprep/cli/run.py
noahbenson: So I'm not familiar with FSL or its way of doing things--do you know how an option like `--bold2t1w-init=header` gets passed to FLIRT? I have made a stab at adding it (using the `apply_xfm` input option, but this isn't something I use and is probably not correct.

noahbenson:
I just pushed several updates to address the comments so far. Note that I'm currently testing these updates, so possible I will push another in the next few hours if there appear to be bugs.

Note that I'm not an FSL user, so my attempt to patch the behavior into FSL is probably wrong--is there someone on the fmriprep side who can help with that?

effigies:
Yup, I can check out the FSL side of things.

JWinawer:
Hi all,

Just wondering what the status is on this pull request, and if it might be incorporated soon? 

Thanks
Jon Winawer

effigies:
Hi @JWinawer. Sorry, I'll try to get around to this, this week.

@noahbenson Any chance of resolving merge conflicts? This is our fault for letting this PR get stale, so we can do it if you prefer. Also, I don't think I'm going to have time to do the FSL implementation, so maybe the best approach is just to raise a `NotImplementedError` if people try to do `--bold2t1w-init header` with the FSL pipeline.

noahbenson:
@effigies The review comments and FSL errors are easy enough to incorporate! I actually already looked at merging the code a couple days ago, and my conclusion was that I could do it but that I probably won't get around to finishing it for a few days--there does appear to be some basic restructuring of how parameters get passed around in the new version that I need to make sure I understand before committing anything.

I'll make an attempt to finish this over the next week; ask here for some clarification.

noahbenson:
@effigies So I've pushed a bunch of changes that I think merge things, but I'm having some trouble running tests locally; I'm getting an error about `Could not find data for participant(s):...` that I've never seen before where the first character of my participant label gets truncated in the message. I've verified that the directories are getting mounted inside the container fine. Before I go on a bug-hunt inside the docker,  is there any chance that this is a bug introduced in something I pulled from master or is it definitely my bug?
(Also, just FYI, the automated `test_pytest` failure seems to be a time-out issue rather than a code issue.)

effigies:
@noahbenson Thanks for updating. This is on my todo list for tomorrow.

PR #1937, opened 2020-01-17
FIX: Include all functional runs in reports, establish consistent ordering

### LINKED ISSUE #1923: fmriprep html not including all functional reports
Did some runs of fmriprep 1.5.4 and noticed that if I have some functional images that include the `run` identifier, but others don't, only the reports for the functional images that include the `run` identifier are included in the final html report.

Looking in the working directory, it appears all the `svg` files are being produced, but they are not being copied over to the derivatives figures directory.

I modified `ds005` to have two different tasks (one with a `run` identifier and one without) and ran the following fmriprep command:

```
fmriprep-docker -i poldracklab/fmriprep:1.5.4 -w ${PWD}/work ${PWD}/ds005 ${PWD}/derivatives --sloppy --notrack --nthreads 3 --fs-no-reconall --fs-license-file ${PWD}/license.txt --output-spaces func -vv
```
[here is the data](https://osf.io/urjk4/download) including the modified ds005, the working directory and the derivatives output.

EDIT:
my first guess is the problem starts in niworkflows [where we are taking the combination of all available entities](https://github.com/poldracklab/niworkflows/blob/9a07072cc52351a0fe85648fa5e864f9b47192df/niworkflows/reports/core.py#L314) (not all inputs will have a value for all the entities)

jdkent:
ref #1923

N/A

Should this be merged into the maint/1.5.x branch or the master branch, i'm afraid I have not been following the branching strategies closely.

[also, I should probably not build fmriprep on my account](https://circleci.com/gh/jdkent/fmriprep/tree/pin_niworkflows)

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -24,7 +24,7 @@ install_requires =
     nibabel >=2.2.1
     nipype >=1.3.1
     nitime
-    niworkflows ~= 1.1.3
+    niworkflows ~= 1.0.4
     numpy
     pandas
     psutil >=5.4
--- a/setup.cfg
+++ b/setup.cfg
@@ -24,7 +24,7 @@ install_requires =
     nibabel >=2.2.1
     nipype >=1.3.1
     nitime
-    niworkflows ~= 1.0.3
+    niworkflows ~= 1.1.3
     numpy
     pandas
     psutil >=5.4
--- a/fmriprep/data/reports-spec.yml
+++ b/fmriprep/data/reports-spec.yml
@@ -27,7 +27,7 @@ sections:
       overlaid on the participant's T1w template.
     subtitle: Surface reconstruction
 - name: Functional
-  ordering: session,task,run
+  ordering: session,task,acquisition,ceagent,reconstruction,direction,run,echo
   reportlets:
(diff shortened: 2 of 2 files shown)


effigies:
New features go into master. Bug fixes and documentation improvements go into `maint/1.5.x`.

There are changes to CIFTI behavior in niworkflows 1.1.x that will make this inappropriate for `maint/1.5.x`. If your changes can be backported on top of niworkflows 1.0.3, we could make a 1.0.4 release of niworkflows and pin to that.

Otherwise, we can just update the pin on `master` and this fix will be in fMRIPrep 20.0.0.

jdkent:
I'll take a look and see if the changes can be applied on top of niworkflows 1.0.3.

setup.cfg
```diff
@@ -24,7 +24,7 @@ install_requires =
     nibabel >=2.2.1
     nipype >=1.3.1
     nitime
-    niworkflows ~= 1.0.3
+    niworkflows ~= 1.1.3
```
effigies:
```suggestion
    niworkflows ~= 1.0.4
```

Do not commit until https://pypi.org/project/niworkflows/1.0.4/ exists.

PR #1945, opened 2020-01-22
[WIP,ENH] Add Grand Mean Scaling

### LINKED ISSUE #1210: Grand mean scaling
It would be beneficial to add grand mean scaling to all outputs:

From http://mindhive.mit.edu/node/71
> grand mean scaling: An analysis step in which the voxel values in every image are divided by the average global mean intensity of the whole session. This effectively removes any mean global differences in intensity between sessions. This is different than global scaling! This step makes a good deal of sense in fMRI, because differences between sessions can be substantial. By performing it at the first (within-subject) level, as well, it means you don't have to do it at the second (between-subject) level, since the between-subject differences are already removed as well. This step is performed by default by all the major analysis software packages.

jdkent:
fixes #1210

Question: Should this be built against maint/1.5.x branch instead of master (it looks like sdcflows depends on niworkflows < 1.1.0, but fmriprep master requires niworkflows > 1.1.0).

### DIFF
--- a/fmriprep/interfaces/confounds.py
+++ b/fmriprep/interfaces/confounds.py
@@ -79,7 +79,6 @@ class GrandMeanScaling(SimpleInterface):
 
     def _run_interface(self, runtime):
         import nibabel as nib
-        from nipype.utils.filemanip import fname_presuffix
         func_img = nib.load(self.inputs.in_func)
         func_mask = nib.load(self.inputs.in_mask).get_fdata().astype(bool)
 
--- a/fmriprep/interfaces/confounds.py
+++ b/fmriprep/interfaces/confounds.py
@@ -8,7 +8,6 @@ Handling confounds
     >>> import os
     >>> import pandas as pd
     >>> import numpy as np
-    >>> import nibabel as nib
 
 """
 import os
@@ -16,6 +15,7 @@ import re
 import shutil
 import numpy as np
 import pandas as pd
+import nibabel as nib
 from nipype import logging
 from nipype.utils.filemanip import fname_presuffix
 from nipype.interfaces.base import (
@@ -32,7 +32,8 @@ class GrandMeanScalingInputSpec(BaseInterfaceInputSpec):
                    desc='input BOLD time-series (4D file)')
     in_mask = File(exists=True, mandatory=True,
                    desc='3D brain mask')
-    grand_mean = traits.Int(desc='grand mean scale value')
+    grand_mean = traits.Int(mandatory=True,
+                            desc='grand mean scale value')
 
 
 class GrandMeanScalingOutputSpec(TraitedSpec):
@@ -65,6 +66,8 @@ class GrandMeanScaling(SimpleInterface):
     >>> scale.inputs.in_mask = 'func_mask.nii.gz'
     >>> scale.inputs.grand_mean = 10000
     >>> res = scale.run()
+    >>> res.outputs.out_func == "func_scaled.nii.gz"
+    True
     >>> scaled_data = nib.load(res.outputs.out_func).get_fdata()
     >>> round(scaled_data[func_mask.astype(bool)].mean())
     10000.0
@@ -78,15 +81,25 @@ class GrandMeanScaling(SimpleInterface):
     output_spec = GrandMeanScalingOutputSpec
 
     def _run_interface(self, runtime):
-        import nibabel as nib
         func_img = nib.load(self.inputs.in_func)
         func_mask = nib.load(self.inputs.in_mask).get_fdata().astype(bool)
 
         func_data = func_img.get_fdata()
 
+        # ensure func_data and func_mask are the same shape
+        if func_data.shape != func_mask.shape:
+            msg = ("mask is not the same shape as the input_image.\n"
+                   "mask shape: {mshape} != image shape: {dshape}\n"
+                   "mask: {mask}\n"
+                   "image: {image}").format(mshape=func_mask.shape,
+                                            dshape=func_data.shape,
+                                            mask=self.inputs.in_mask,
+                                            image=self.inputs.in_func)
+            raise ValueError(msg)
+
         scaling_factor = self.inputs.grand_mean / func_data[func_mask].mean()
 
-        func_data[func_mask] = func_data[func_mask] * scaling_factor
+        func_data[func_mask] *= scaling_factor
 
         out = fname_presuffix(self.inputs.in_func, suffix='_scaled')
         func_img.__class__(func_data, func_img.affine, func_img.header).to_filename(out)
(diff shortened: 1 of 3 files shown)


fmriprep/interfaces/confounds.py
```diff
@@ -7,6 +7,8 @@
 
     >>> import os
     >>> import pandas as pd
+    >>> import numpy as np
+    >>> import nibabel as nib
```
jdkent: This import did not appear to work in my `GrandMeanScaling` doctest, thus its repetition below, any thoughts on how these imports are interpreted?

fmriprep/interfaces/confounds.py
```diff
@@ -25,6 +27,75 @@
...
+    .. testcleanup::
+
+    >>> tmpdir.cleanup()
+    """
+
+    input_spec = GrandMeanScalingInputSpec
+    output_spec = GrandMeanScalingOutputSpec
+
+    def _run_interface(self, runtime):
+        import nibabel as nib
```
mgxd: this should probably be a top level import

fmriprep/interfaces/confounds.py
```diff
@@ -25,6 +27,75 @@
...
+    >>> nib.Nifti1Image(func_data, affine).to_filename('func.nii.gz')
+    >>> func_mask = np.eye(10).astype(int)
+    >>> nib.Nifti1Image(func_mask, affine).to_filename('func_mask.nii.gz')
+
+    >>> scale = GrandMeanScaling()
+    >>> scale.inputs.in_func = 'func.nii.gz'
+    >>> scale.inputs.in_mask = 'func_mask.nii.gz'
+    >>> scale.inputs.grand_mean = 10000
+    >>> res = scale.run()
+    >>> scaled_data = nib.load(res.outputs.out_func).get_fdata()
```
mgxd: can you add a line ensuring `res.outputs.out_func == "func_scaled.nii.gz"`

fmriprep/interfaces/confounds.py
```diff
@@ -25,6 +27,75 @@
 LOGGER = logging.getLogger('nipype.interface')
 
 
+class GrandMeanScalingInputSpec(BaseInterfaceInputSpec):
+    in_func = File(exists=True, mandatory=True,
+                   desc='input BOLD time-series (4D file)')
+    in_mask = File(exists=True, mandatory=True,
+                   desc='3D brain mask')
+    grand_mean = traits.Int(desc='grand mean scale value')
```
mgxd: since this is required and no default is set, let's make this mandatory too

fmriprep/interfaces/confounds.py
```diff
@@ -25,6 +27,75 @@
...
+    output_spec = GrandMeanScalingOutputSpec
+
+    def _run_interface(self, runtime):
+        import nibabel as nib
+        func_img = nib.load(self.inputs.in_func)
+        func_mask = nib.load(self.inputs.in_mask).get_fdata().astype(bool)
+
+        func_data = func_img.get_fdata()
+
+        scaling_factor = self.inputs.grand_mean / func_data[func_mask].mean()
```
mgxd: we might want to add a quick check for equal shape to the functional image here

fmriprep/interfaces/confounds.py
```diff
@@ -25,6 +27,75 @@
...
+
+        func_data[func_mask] = func_data[func_mask] * scaling_factor
```
mgxd:
```suggestion
        func_data[func_mask] *= scaling_factor
```

fmriprep/interfaces/confounds.py
mgxd: I guess this one is not necessary if you are doing it again in GMS doctest

jdkent:
I think the largest conceptual problem I'm having connecting this to the larger workflow is whether to apply the grand mean scaling before pushing the outputs into their final space (e.g., `T1w`, `MNI`, etc.) or after. If we want to do registration in single-shot, then the only time I will have the appropriate data and mask is after the transform is applied, meaning I will have to apply grand mean scaling separately to each output space provided. 

That is the path forward I am seeing currently unless someone else has another solution I'm not seeing.

jdkent:
Okay, I think I'm at least partially caught up on `outputnode` and `poutputnode`. I see two different strategies moving forward.

1. Do Grand Mean Scaling Once
    - do grand mean scaling after head motion correction with the generated mask.
    - invert the hmc transformation
    - apply inverted hmc transformation to grand mean scaled image
    - pass the grand mean scaled image to the rest of the workflow

Pros:
  - only perform grand mean scaling once
  - the code should be modular

Cons:
  - potentially expensive computation to reverse/apply the hmc transform
  - does inverting the transform cause any interpolation issues (I don't think so since it's a linear transform, but I do not want to presume)
  - small errors may occur for the grand mean after transforming the image (e.g., the grand mean may be different in other output spaces)


2. Do Grand Mean Scaling on each volume output space
    - perform grand mean scaling on native and t1w spaces separately
    - reorganize workflow to have the `poutputnode` and `outputnode` of `bold_std_trans_wf` be grand mean scaled before being passed on to other outputs.

Pros:
  - grand mean scaling will be exactly correct for each output space
 
Cons:
  - have to apply grand mean scaling to each output space
  - have to reorganize workflow to apply grand mean scaling correctly
    - the code changes will most likely not be very modular


I am slightly leaning towards option 1, but I will hold on this until there is consensus on what should be done

sorry, I thought this was going to be relatively straightforward, but it has turned out to be more intensive than I've anticipated.

emdupre:
Just wanted to ping in that I'm a little nervous about having this on by default -- It will make it harder to integrate with some post-fMRIPrep workflows, particularly for multi-echo. Would it be possible to make this an optional adjustment ?

effigies:
Finally getting around to looking at this. I see @jdkent's points about the practical pros/cons of when to apply GMS, but I think they're somewhat beside the point. I would be inclined to reject option 2 for a simple reason: it can be done post-fMRIPrep. The arguments I see are:

1) Pre-resampling
   * Pro: All outputs have the same source values, differing only by resampling
   * Con: Some might not want it
2) Post-reampling
   * Pro: Easy to do post-fMRIPrep
   * Con: Variations in values relating to resampling exist prior to rescaling

If the effects of performing GMS post-resampling are statistically problematic (does anybody have a reference?) then I think we don't have much choice but to make it an option. But if not, my inclination is to leave it be.

jdkent:
I do not have a reference saying that performing GMS post-resampling is problematic, perhaps this would be a good candidate for a post-fmriprep plugin.

effigies:
Closing this. If it turns out that pre-resampling GMS is in demand, we can revisit making it an option. It should not be on-by-default.

PR #1951, opened 2020-01-27
FIX: SyN SDC logic failing in `--force-syn` cases

effigies:

Depends on https://github.com/poldracklab/sdcflows/pull/82.

Using a `docker/` branch to get a testable Docker image out pre-release.


effigies:
This fixes my reproduction of the issue. Should see if we can get a regression test in sdcflows itself.

effigies:
I think a bug-fix (patch) release of sdcflows makes sense. The inputs/outputs didn't change. We just fixed a case where one of the outputs was undefined.

PR #1983, opened 2020-02-14
FIX: Revert MNI152NLin2009cAsym as default BOLD output

### LINKED ISSUE #1979: No default output spaces in 20.0.0rc1
Up to now we've always had MNI152NLin2009cAsym as a default output space. Now I get 
```
UserWarning: fMRIPrep will not generate preprocessed derivatives because none of `--output-spaces`, `--use-aroma`, or `--cifti-output` were set.
```

I would suggest we restore the default behavior.

mgxd:
Closes #1979

### DIFF
--- a/fmriprep/cli/run.py
+++ b/fmriprep/cli/run.py
@@ -129,7 +129,7 @@ def get_parser():
              'option is not enabled, standard EPI-T1 coregistration is performed '
              'using the middle echo.')
     g_conf.add_argument(
-        '--output-spaces', nargs='+', action=OutputReferencesAction, default=SpatialReferences(),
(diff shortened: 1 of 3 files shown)


fmriprep/cli/run.py
```diff
@@ -718,7 +718,14 @@ def parse_spaces(opts):
 
     """
     spaces = opts.output_spaces
-    spaces.checkpoint()
+    if not spaces.references and not spaces.is_cached():
+        warnings.warn(
+            "No output spaces have been set - fMRIPrep will add `MNI152NLin2009cAsym` by default."
```
oesteban: I think we can do this without any warnings.

PR #1989, opened 2020-02-19
DOC: Integrate intersphinx, drop external module wrapping

oesteban:
Beef up the intersphinx game, replace old links to internal function and
class definitions by new ones from niworkflows, sdcflows, smriprep & co.

Remove the import stubs (anatomical and fieldmap workflows & interfaces)
as they are not required anymore.


oesteban:
The workflow seems to work locally. I'm going to merge so that we test more thoroughly with the external modules without wrapping.

Issue #1986, opened 2020-02-16
--cifti-output issues for data processed in fsLR-32k space in 20.0.1rc1


helloTC:
Hi, I installed the latest version of fmriprep (v20.0.1rc) in my servers by using pip command. I would like to get processed data in fsLR-den32k space. It looks fsLR no long need to be included in the --output-spaces in this lastest version because after running 
`fmriprep test_fmriprep test_fmriprep_output participant -w work --output-spaces fsLR --cifti-output`
I got an error as 
> ValueError: space identifier "fsLR" is invalid.
Valid identifiers are: MNI152Lin, MNI152NLin2009cAsym, MNI152NLin6Asym, MNI152NLin6Sym, MNIInfant, MNIPediatricAsym, NKI, OASIS30ANTs, PNC, fsaverage, T1w, T2w, anat, fsnative, func, run, sbref, session.

While I noticed in [documentation](https://github.com/poldracklab/fmriprep/blob/maint/20.0.x/docs/spaces.rst#standard-spaces) of the v20.0.1rc still implied that the users need to include fsLR in the --output-spaces, it seems a bug here.

Issue #1887 allows CIFTI output when using --cifti-output and will compliant with HCP grayordinate space (cortex in fsLR, subcortical in MNI152NLin6Asym). 
However, after running command only with --cifti-output: `fmriprep test_fmriprep test_fmriprep_output participant -w work -cifti-output`, I got one another error as 

> Process Process-2:
Traceback (most recent call last):
  File "/usr/local/neurosoft/miniconda3/lib/python3.6/multiprocessing/process.py", line 258, in _bootstrap
    self.run()
  File "/usr/local/neurosoft/miniconda3/lib/python3.6/multiprocessing/process.py", line 93, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/local/neurosoft/miniconda3/lib/python3.6/site-packages/fmriprep/cli/run.py", line 676, in build_workflow
    work_dir=str(work_dir),
  File "/usr/local/neurosoft/miniconda3/lib/python3.6/site-packages/fmriprep/workflows/base.py", line 267, in init_fmriprep_wf
    use_syn=use_syn,
  File "/usr/local/neurosoft/miniconda3/lib/python3.6/site-packages/fmriprep/workflows/base.py", line 623, in init_single_subject_wf
    use_syn=use_syn,
[...]

PR #1993, opened 2020-02-20
DOC: Clean up surface outputs

mgxd:
Mentioned in #1986

### DIFF
--- a/docs/outputs.rst
+++ b/docs/outputs.rst
@@ -143,15 +143,14 @@ mid-thickness surface mesh::
 
 Surface output spaces include ``fsnative`` (full density subject-specific mesh),
 ``fsaverage`` and the down-sampled meshes ``fsaverage6`` (41k vertices) and
-``fsaverage5`` (10k vertices, default). Additionally, the BOLD can be resampled to
-``fsLR`` surfaces, a hybrid version of the ``fsaverage`` atlas with both hemispheres
-in precise geographic correspondence, with varying mesh densities: 32k (default), 59k, and 164k.
+``fsaverage5`` (10k vertices, default).
 
 **Grayordinates files**.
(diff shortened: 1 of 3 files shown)


docs/outputs.rst
```diff
@@ -150,8 +150,8 @@ in precise geographic correspondence, with varying mesh densities: 32k (default)
 **Grayordinates files**.
 CIFTI is a container format that holds both volumetric (regularly sampled in a grid)
 and surface (sampled on a triangular mesh) samples.
-Sub-cortical time series are volumetric (supported spaces: ``MNI152NLin6Asym``), while cortical
-time series are sampled to surface (supported spaces: ``fsLR``).
+Sub-cortical time series are volumetric (space: ``MNI152NLin6Asym``), while cortical
```
oesteban: Since this is specified by HCP, I think adding the spaces names only clutters:
```suggestion
Sub-cortical time series are sampled on a regular grid derived from one MNI template, while cortical
```

docs/outputs.rst
```diff
@@ -150,8 +150,8 @@ in precise geographic correspondence, with varying mesh densities: 32k (default)
...
+time series are sampled to surface (space: ``fsLR``).
```
oesteban:
```suggestion
time-series are sampled on surfaces projected from the [Glasser2016]_ template.
```

With the Glasser reference being:
> Glasser, M., Coalson, T., Robinson, E. et al. A multi-modal parcellation of human cerebral cortex. Nature 536, 171–178 (2016). https://doi.org/10.1038/nature18933

Perhaps, it would be nice to also add a link to the cifti spec of HCP: https://www.nitrc.org/forum/attachment.php?attachid=333&group_id=454&forum_id=1955

PR #1996, opened 2020-02-25
ENH: Include low res CIFTI templates in Docker image

mgxd:
Closes #1995

Issue #1995, opened 2020-02-25
--cifti-output error singularity version 20.0.0


saigerutherford:
Cifti output spaces does not seem to be working in the singularity container version 20.0.0.

Here is the command I am running fMRIPrep with:
```module load singularity

FMRIPREP=/sw/examples/singularity/fmriprep-20.0.0.simg
SURF_LICENSE=${SCRIPTDIR}/data/fs_license.txt

BIDS_DIR=/tmp/workflow_usefs_${SUB}/BIDS
OUTPUT_DIR=/tmp/workflow_usefs_${SUB}/derivatives
WORK_DIR=/tmp/workflow_usefs_${SUB}/work

singularity run --cleanenv $FMRIPREP                        \
    $BIDS_DIR $OUTPUT_DIR participant                       \
    --nthreads ${NTHREADS}                                  \
    --mem-mb ${MEMMB}                                       \
    --fs-license-file=$SURF_LICENSE                         \
    --fs-subjects-dir ${OUTPUT_DIR}/freesurfer              \
    --participant-label=sub-${SUB}                          \
    -w $WORK_DIR                                            \
    --ignore slicetiming                                    \
    --cifti-output                                         \
    --use-aroma                                             \
    --aroma-melodic-dimensionality -100                     \
    --output-spaces anat fsLR MNI152NLin6Asym:res-2
```

And here is the error output:
```[saruther@gl-login1 2060]$ tail -f 06fmriprep.log 
        Summary:                Available Tasks:        Available Modalities: 
        25 Files, 3.55GB        task-gaze               T1w                   
        1 - Subject             task-jovi               T2w                   
        1 - Session             task-rest               bold                  
                                                        fieldmap              


	If you have any questions, please post on https://neurostars.org/tags/bids.

Making sure the input data is BIDS compliant (warnings can be ignored in most cases).
200225-12:51:44,324 nipype.workflow IMPORTANT:
	 
    Running fMRIPREP version 20.0.0:
      * BIDS dataset path: /tmp/workflow_usefs_2060/BIDS.
      * Participant list: ['2060'].
      * Run identifier: 20200225-125144_8a8b4fba-3af8-416a-ba3b-083451d3c2ee.

    Spatial References:
    - Reference(space='anat', spec={})
    - Reference(space='fsLR', spec={})
    - Reference(space='MNI152NLin6Asym', spec={'res': '2'})
    
200225-12:51:49,779 nipype.utils WARNING:
[...]

PR #1997, opened 2020-02-25
CI: Test that Docker image can run a common set of output spaces without network access

effigies:
##  Changes proposed in this pull request

Let's reduce the odds of #1995 coming up again. 

* Adds a `--network` option to `fmriprep-docker` that allows us to pass the `--network` option to Docker
* Set `--network=none` for ds005, which uses AROMA, SyN-SDC, and outputs to a number of spaces including HCP-style CIFTI.

Hopefully this will fail, and then will pass with #1996 in.


effigies:
First failed. Merged in #1996.

PR #1998, opened 2020-02-26
[NOMRG] UX: Display errors as summary/details elements

### LINKED ISSUE #1991: it should be much more obvious that errors can be clicked on
image

It is not at _all_ obvious that these are expandable by clicking on them. Instead of round bullets, these should be rotating triangles.

effigies:

Uses https://github.com/poldracklab/niworkflows/pull/464 to generate reports. Check the output of ds054 for an example error, when it's run.

Closes #1991.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -24,7 +24,7 @@ install_requires =
     nibabel >= 3.0.1
     nipype >= 1.3.1
     nitime
-    niworkflows ~= 1.1.7
+    niworkflows @ git+https://github.com/effigies/niworkflows.git@bbce974af87788218d16d8dac27802c2738cac8f
     numpy
     pandas
     psutil >= 5.4


effigies:
Here's an example report with an error: https://12090-53175327-gh.circle-artifacts.com/0/tmp/ds054/derivatives/fmriprep/sub-100185.html

@dmd Any comments?

effigies:
I've canceled the non-ds054 tests. We'll have a full run before any release on the off chance that this is somehow affecting other tests.

effigies:
No point merging, since we're going to have a couple bug fixes in niworkflows 1.1.8. Will keep open until we release niworkflows and update the PIN.

dmd:
> Here's an example report with an error: https://12090-53175327-gh.circle-artifacts.com/0/tmp/ds054/derivatives/fmriprep/sub-100185.html
> 
> @dmd Any comments?

Exactly what I envisioned, thanks.

effigies:
Included in 2590fed406a77874f6ab5dc498d246bc0f0db4bc.

Issue #2008, opened 2020-02-28
Installation instructions for standalone fmriprep are out of date, causing crashes.


bbfrederick:
We've recently updated our standalone fmriprep installation fmriprep 1.5.8 (and will now go to 20.0.x) from 1.4.1.  We have found that AROMA processing crashes with the following error:

`
Node Name: fmriprep_wf.single_subject_017_wf.func_preproc_task_session3_wf.ica_aroma_wf.ica_aroma
/data/latncase/MDD_STRESS/fmriprep/test_ERISONE2/derivatives/fmriprep/sub-017/log/20200227-134551_1da3a78f-5ea2-4ce5-90ee-b7f94ab0e167/crash-20200228-001011-mai14-ica_aroma-2e188a62-9b80-471c-8dbc-0dfcb69d4583.txt
Working Directory: /data/latncase/MDD_STRESS/fmriprep/work/fmriprep_wf/single_subject_017_wf/func_preproc_task_session3_wf/ica_aroma_wf/ica_aroma
Inputs:

TR: 2.0
args: -np
compress_report: auto
denoise_type: nonaggr
dim:
environ: {}
feat_dir:
fnirt_warp_file:
in_file:
mask:
mat_file:
melodic_dir:
motion_parameters:
out_dir: out
out_report: ica_aroma_reportlet.svg
report_mask:
Traceback (most recent call last):
  File "/PHShome/bbf2/software/anaconda3/envs/fmriprep158/lib/python3.7/site-packages/nipype/pipeline/plugins/legacymultiproc.py", line 67, in run_node
    result["result"] = node.run(updatehash=updatehash)
  File "/PHShome/bbf2/software/anaconda3/envs/fmriprep158/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 516, in run
    result = self._run_interface(execute=True)
  File "/PHShome/bbf2/software/anaconda3/envs/fmriprep158/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 635, in _run_interface
    return self._run_command(execute)
  File "/PHShome/bbf2/software/anaconda3/envs/fmriprep158/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 741, in _run_command
    result = self._interface.run(cwd=outdir)
  File "/PHShome/bbf2/software/anaconda3/envs/fmriprep158/lib/python3.7/site-packages/nipype/interfaces/base/core.py", line 397, in run
[...]

PR #2009, opened 2020-02-28
Updated standalone installation requirements

bbfrederick:
This PR updates the external dependencies listed in installation.rst for standalone installation to match the packages and versions found in the Dockerfile.
There is also a typo fixed in CONTRIBUTING.md

This should close issue #2008, which I raised today.

The need for an updated version of ICA-AROMA is pretty clear, as fmriprep now uses a flag that doesn't exist in the older version.  I used the version and download link used in the Dockerfile - if there's a more appropriate permalink that might be better.
The bids-validator version has been updated to match the Dockerfile.
I can't see any indication that c3d is installed in the Dockerfile, so I removed it as a dependency.

This is a documentation only fix, which modifies docs/installation.rst and CONTRIBUTING.md


docs/installation.rst
```diff
@@ -144,10 +144,9 @@ the ``fmriprep`` package:
 - FSL_ (version 5.0.9)
 - ANTs_ (version 2.2.0 - NeuroDocker build)
 - AFNI_ (version Debian-16.2.07)
-- `C3D <https://sourceforge.net/projects/c3d/>`_ (version 1.0.0)
 - FreeSurfer_ (version 6.0.1)
-- `ICA-AROMA <https://github.com/rhr-pruim/ICA-AROMA/>`_ (version 0.4.1-beta)
-- `bids-validator <https://github.com/bids-standard/bids-validator>`_ (version 1.1.0)
+- `ICA-AROMA <https://github.com/oesteban/ICA-AROMA/archive/v0.4.5.tar.gz>`_ (version 0.4.5)
```
effigies: Ugh. I didn't realize we have an unofficial release in here.

@oesteban I see @maartenmennes merged https://github.com/maartenmennes/ICA-AROMA/pull/43. Perhaps we should pin:

```suggestion
- `ICA-AROMA <https://github.com/maartenmennes/ICA-AROMA/archive/e8d7a58.tar.gz>`_ (commit e8d7a58, post v0.4.4-beta)
```

WDYT?

bbfrederick: Seems reasonable to me.

bbfrederick: Well, my reading of the Dockerfile was not in fact as careful as I thought it was.  C3D _is_ installed when other packages are pulled from Neurodebian (as convert3d).  So the line:

- `C3D <https://sourceforge.net/projects/c3d/>`_ (version 1.0.0)

Should _NOT_ be removed from the standalone dependencies.  Should I open a new PR, or can this one be fixed?

effigies: This one can be fixed. Just push to the same branch and it'll update.

PR #2015, opened 2020-03-03
ENH: Add t2w/flair as potential --ignore targets

### LINKED ISSUE #2011: Add `--ignore t2w` and `--ignore flair` options
This is probably most easily implemented by clearing subject data entries in `init_single_subject_wf`, just after:

https://github.com/poldracklab/fmriprep/blob/1641529a685c6c23ce2f98ccce06da56280aed98/fmriprep/workflows/base.py#L485-L493

To be consistent with `--ignore fieldmap`, we could pass `ignore` into sMRIPrep's workflows. It's unclear to me that this would be overall cleaner. We could also consider moving `--ignore fieldmap` into `init_single_subject_wf`, though that would be an additional PR.

This adds a CLI option, so I'm marking 20.1.0, but retains backwards compatibility, so I can see an argument for 20.0.2.

This is based on discussion in [Neurostars #2412](https://neurostars.org/t/fmriprep-t2-images/2412/11).

mgxd:
Closes #2011 
This adds new options to the `--ignore` flag, allowing users to avoid using T2w/FLAIR images during FreeSurfer processing.

### DIFF
--- a/fmriprep/workflows/base.py
+++ b/fmriprep/workflows/base.py
@@ -491,6 +491,10 @@ def init_single_subject_wf(
     else:
         subject_data = collect_data(layout, subject_id, task_id, echo_idx,
                                     bids_filters=bids_filters)[0]
+        if 'flair' in ignore and subject_data['flair']:
+            subject_data['flair'] = []
+        if 't2w' in ignore and subject_data['t2w']:
+            subject_data['t2w'] = []
 
     # Make sure we always go through these two checks
     if not anat_only and subject_data['bold'] == []:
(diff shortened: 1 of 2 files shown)


fmriprep/workflows/base.py
```diff
@@ -491,6 +491,10 @@ def init_single_subject_wf(
     else:
         subject_data = collect_data(layout, subject_id, task_id, echo_idx,
                                     bids_filters=bids_filters)[0]
+        if 'flair' in ignore and subject_data['flair']:
+            subject_data['flair'] = []
+        if 't2w' in ignore and subject_data['t2w']:
```
effigies: No harm in replacing empty lists.

```suggestion
        if 'flair' in ignore:
            subject_data['flair'] = []
        if 't2w' in ignore:
```

PR #2017, opened 2020-03-03
MAINT: Bump to 3.7+

mgxd:
also removes a python 2 test of the docker wrapper, and bumps licenses up a year

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -233,6 +233,19 @@ jobs:
               /src/fmriprep/fmriprep -svx --doctest-modules
 
       - run:
+          name: Test fmriprep-wrapper (Python 2)
+          command: |
+            export PY2=$(pyenv versions | grep '2\.' |
+                         sed -e 's/.* 2\./2./' -e 's/ .*//')
+            pyenv local $PY2
+            echo -n "Python version: "
+            python --version
+            pip install --upgrade pip setuptools
+            pip install --upgrade wrapper/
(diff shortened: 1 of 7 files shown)


.circleci/config.yml
```diff
@@ -232,19 +232,6 @@ jobs:
...
-            export PY2=$(pyenv versions | grep '2\.' |
-                         sed -e 's/.* 2\./2./' -e 's/ .*//')
-            pyenv local $PY2
-            echo -n "Python version: "
-            python --version
-            pip install --upgrade pip setuptools
-            pip install --upgrade wrapper/
-            which fmriprep-docker
-            fmriprep-docker -i poldracklab/fmriprep:latest --help
-            fmriprep-docker -i poldracklab/fmriprep:latest --version
```
effigies: This is intentionally left available in Python 2 because Python 2 is still the default Python on a lot of systems, and so it's the cheapest way to get people going in Docker.

effigies:
Please also adjust:

https://github.com/poldracklab/fmriprep/blob/2cf3e494b3921f9b9549e6caa686a737bc28d118/setup.cfg#L21

https://github.com/poldracklab/fmriprep/blob/2cf3e494b3921f9b9549e6caa686a737bc28d118/setup.cfg#L16-L18

https://github.com/poldracklab/fmriprep/blob/2cf3e494b3921f9b9549e6caa686a737bc28d118/.travis.yml#L9

PR #2020, opened 2020-03-04
FIX: Properly mount local templates

mgxd:
This PR re-enables use of custom templates after the output spaces refactoring.

The local directory where the custom template exists is now mounted to the TemplateFlow home within the container.

Related issue: https://neurostars.org/t/custom-template-fmriprep-20-0-1/6232

### DIFF
--- a/wrapper/fmriprep_docker.py
+++ b/wrapper/fmriprep_docker.py
@@ -428,6 +428,8 @@ def main():
         for space in opts.output_spaces:
             if space.split(':')[0] not in (TF_TEMPLATES + NONSTANDARD_REFERENCES):
                 tpl = os.path.basename(space)
+                if not tpl.startswith('tpl-'):
+                    raise RuntimeError("Custom template %s requires a `tpl-` prefix" % tpl)
                 target = '/home/fmriprep/.cache/templateflow/' + tpl
                 command.extend(['-v', ':'.join((os.path.abspath(space), target, 'ro'))])
                 spaces.append(tpl[4:])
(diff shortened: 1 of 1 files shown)


wrapper/fmriprep_docker.py
```diff
@@ -427,9 +427,10 @@ def main():
         spaces = []
         for space in opts.output_spaces:
             if space.split(':')[0] not in (TF_TEMPLATES + NONSTANDARD_REFERENCES):
-                target = '/imports/' + os.path.basename(space)
+                tpl = os.path.basename(space)
+                target = '/home/fmriprep/.cache/templateflow/' + tpl
                 command.extend(['-v', ':'.join((os.path.abspath(space), target, 'ro'))])
-                spaces.append(target)
+                spaces.append(tpl[4:])
```
effigies: It looks like we're baking in some assumptions that the directory will be called `tpl-XYZ`. We should either require that (error if `tpl[:4] != 'tpl-'`) or be a bit more forgiving (`if not tpl.startswith('tpl-'): tpl = 'tpl-' + tpl`).

mgxd: Sure, we can add a check to ensure the directory is compliant with templateflow standards.

effigies: This may be obvious, but in case you're about to do a significant amount of work, I would really keep this very simple. If the error is going to show up early and be clear inside fMRIPrep proper, then we really don't need to do anything here.

I'm mostly worried about if somebody types `--output-spaces /my/template` and gets strange errors about "tpl-late not found", or something like that.

mgxd: since this check is happening within the niworkflows SpatialReference, that would probably be the best place for that. Since the wrapper behavior is a little different (it has to actually mount the directory), we can include a check here too

PR #2018, opened 2020-03-03
ENH: Config module to replace long list of downstreamed arguments

oesteban:
~~Includes addressing #2019, although~~ I think that one could be addressed separately to simplify this already huge PR.

Requires: https://github.com/nipreps/niworkflows/pull/475

## Summary
This module implements the memory structures to keep a consistent, singleton config.

### Example
```Python
>>> from fmriprep import config
>>> config_file = config.execution.work_dir / '.fmriprep.toml'
>>> config.to_filename(config_file)
>>> # Call build_workflow(config_file, retval) in a subprocess
>>> with Manager() as mgr:
>>>     from .workflow import build_workflow
>>>     retval = mgr.dict()
>>>     p = Process(target=build_workflow, args=(str(config_file), retval))
>>>     p.start()
>>>     p.join()
>>> config.load(config_file)
>>> # Access configs from any code section as:
>>> value = config.section.setting
```
The module also has a :py:func:`to_filename` function to allow writting out
the settings to hard disk in *ToML* format, which looks like:

    [nipype]
    crashfile_format = "txt"
    get_linked_libs = false
    nprocs = 8
    omp_nthreads = 8
    plugin = "MultiProc"
    resource_monitor = false
    stop_on_first_crash = false
    version = "1.5.0-dev+gdffcb7815"

    [execution]
    bids_dir = "/data/openfmri/ds000005"
    boilerplate_only = false
    exec_env = "posix"
    fs_license_file = "/opt/freesurfer/license.txt"
    fs_subjects_dir = "/data/openfmri/ds000005/derivatives/freesurfer-6.0.1"
    log_dir = "/data/openfmri/ds000005/derivatives/fmriprep/logs"
    log_level = 15
    low_mem = false
    md_only_boilerplate = false
    notrack = true
    output_dir = "/data/openfmri/ds000005/derivatives"
    reports_only = false
    run_uuid = "20200302-174345_9ba9f304-82de-4538-8c3a-570c5f5d8f2f"
    participant_label = [ "01",]
    version = "20.0.1+13.g3ca42930.dirty"
    work_dir = "/home/oesteban/tmp/fmriprep-ds005/fprep-work2"
    write_graph = false

    [workflow]
    anat_only = false
    aroma_err_on_warn = false
    aroma_melodic_dim = -200
    bold2t1w_dof = 6
    cifti_output = false
    fmap_bspline = false
    force_syn = false
    hires = true
    ignore = []
    internal_spaces = "MNI152NLin2009cAsym"
[...]

PR #2021, opened 2020-03-04
MNT: Pin minor series of nipype, major of nibabel

effigies:
Based on a conversation in https://github.com/nipy/nipype/pull/3180, it seems prudent to pin to minor series of nipype. Nipype is on bug-fix-only patch releases. This fits well with our proposed rules for minor release series. Upgrading the nipype pin to a new minor release series would only be valid in minor releases of fMRIPrep.

I also pinned nibabel 3.0. Nibabel has very strict backwards-compatibility constraints within a major version series, so upgrading to a new minor release should not introduce any concerns even in bug-fix releases.

We might want to consider assessing other dependencies. With the prospect of long-term releases, unpinned dependencies become riskier.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -21,8 +21,8 @@ classifiers =
 python_requires = >=3.5
 install_requires =
     indexed_gzip >= 0.8.8
-    nibabel >= 3.0.1
-    nipype >= 1.3.1
+    nibabel ~= 3.0
+    nipype ~= 1.4.0
     nitime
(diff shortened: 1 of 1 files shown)


effigies:
@oesteban I don't think #2018 should stop us from merging to `maint/20.0.x`. I'm okay with delaying merging `maint/20.0.x` into `master` if it starts to present merge conflicts.

PR #2032, opened 2020-03-11
FIX: remedy GIFTI resampling template files

### LINKED ISSUE #2025: --cifti-output incorrect resampling from fsaverage_den-164k to fsLR_den-32k
Hi all,

I am using fMRIPrep 20.0.1 and hoping to get preprocessed data in fsLR_den-32k space. I ran the following on our cluster:

```
singularity run \
    --cleanenv -B /nfs:/nfs $FMRIPREP \
    $BIDS_DIR $OUTPUT_DIR participant \
    --n_cpus $SLURM_NTASKS_PER_NODE \
    --omp-nthreads 4 \
    --fs-license-file=$SURF_LICENSE \
    --participant-label=$SUBLABEL \
    --ignore slicetiming \
    --dummy-scans 12 \
    --cifti-output \
    -w $WORK_DIR
```
It finished with no errors, but the resulting cifti time-series clearly does not look right.

fmriprep_artifact

Not sure if this is producing the issue but it appears that the area inputs to Connectome Workbench's -metric-resample are incorrect. 'area_metrics' is set to True, which means that metric overlay files (e.g. fsaverage.L.midthickness_va_avg.164k_fsavg_L.shape.gii) should be passed. It appears that anatomical surface files (tpl-fsLR_space-fsaverage_hemi-L_den-164k_midthickness.surf.gii) are used instead. If I run the following command using metric area files the output appears sensible:

```
wb_command -metric-resample \
$WORK_DIR/fmriprep_wf/single_subject_212_wf/func_preproc_ses_baseline_task_dsp_run_01_wf/bold_surf_wf/_target_fsaverage/update_metadata/mapflow/_update_metadata1/rh.fsaverage.gii \
~/standard_mesh_atlases/resample_fsaverage/fsaverage_std_sphere.R.164k_fsavg_R.surf.gii \
~/standard_mesh_atlases/resample_fsaverage/fs_LR-deformed_to-fsaverage.R.sphere.32k_fs_LR.surf.gii \
ADAP_BARY_AREA \
~/space-fsLR_hemi-R_den-91k_bold.gii \
-area-metrics ~/standard_mesh_atlases/resample_fsaverage/fsaverage.R.midthickness_va_avg.164k_fsavg_R.shape.gii \
~/standard_mesh_atlases/resample_fsaverage/fs_LR.R.midthickness_va_avg.32k_fs_LR.shape.gii
```

fmriprep_noartifact

Best,
James

mgxd:
Fixes #2025 

This adds any previous release of the  `20.0.x` series to bad versions.


setup.cfg
```diff
@@ -33,7 +33,7 @@ install_requires =
     sdcflows ~= 1.2.0
     smriprep ~= 0.5.2
     tedana >= 0.0.5
-    templateflow >= 0.4.2
+    templateflow >= 0.5.0
```
effigies: What changes are introduced in this minor version release?

mgxd: A new template (rodent) and changes to `fsLR` / `fsaverage` templates.

effigies: Does that affect any non-CIFTI outputs?

mgxd: I'd be surprised if they did - 99.9% sure of "no"

mgxd:
surfaces look much better.

screenshot

.github/auto-comment.yml
```diff
@@ -1,71 +0,0 @@
-pullRequestOpened: |
```
oesteban: Let's remove this on a different PR

.versions.json
```diff
@@ -5,6 +5,12 @@
 		"1.5.0rc2": "Nipype unstable",
 		"1.5.0": "Nipype unstable",
 		"1.5.1rc1": "Nipype unstable",
-		"1.5.1rc2": "Nipype unstable"
+		"1.5.1rc2": "Nipype unstable",
+		"20.0.0rc1": "fsLR resampling error",
```
oesteban:
```suggestion
		"20.0.0rc1": "fsLR resampling error (problematic only when using the --cifti-output flag)",
```

setup.cfg
```diff
@@ -33,7 +33,7 @@ install_requires =
     sdcflows ~= 1.2.0
     smriprep ~= 0.5.2
     tedana >= 0.0.5
-    templateflow >= 0.4.2
+    templateflow ~= 0.5.1rc1
```
oesteban: esp. with RCs (`~=` will not match 0.5.1), I think we want this:
```suggestion
    templateflow >= 0.5.1rc1
```
Also, templateflow does not generally break it's API - so it boils down to the actual templates in the skeleton. I think it is safe to use a blanket `>=` here.

.versions.json
oesteban: And so on, so forth for the others

.github/auto-comment.yml
mgxd: we can, though at this point its more spam than good

.github/auto-comment.yml
```diff
@@ -0,0 +1,71 @@
+pullRequestOpened: |
```
oesteban: It would be best if rolling back this removal didn't show in the diff.

.github/auto-comment.yml
oesteban: for ourselves - if we move towards more community-driven, something like this would be useful. Since that should be happening soon, let's address this elsewhere (and probably defer the conversation to the community).

.github/auto-comment.yml
effigies: @oesteban TBC are you requesting a rebase so that it's not in the history at all? The change is not visible in the PR "files changed" tab unless you look at specific commits.

oesteban: Oh right, yes, then it's okay I don't think editing the history is worth the pain.

effigies:
Latest changes very unlikely to break the build and another will be necessary for any release, so I went ahead and merged prematurely.

PR #2036, opened 2020-03-14
ENH: Side-track *sMRIPrep* if a path of anatomical derivatives is given

### LINKED ISSUE #1175: Support independent jobs across sessions
It's certain that we will see more and more datasets like MyConnectome and Midnight Scan Club. To efficiently process such dataset we need to support submitting independent jobs across sessions. In practice this would mean:

- addition of `session` analysis level and `--session_label` flag for selecting sessions (more details at https://groups.google.com/forum/#!topic/bids-apps-dev/Vnx5HHudqq4)
- addition of `participant_anat` analysis level to be run before any `session` levels

oesteban:
Summary
-------
This PR utilizes poldracklab/smriprep#107 to allow a fast skip of the
anatomical workflow evaluation, and the need of maintaining the work
directory for it.

Closes #1175

  * [x] Added an ``--anat-derivatives`` flag to the command line.
  * [x] Integrated the new flag with the new config module.
  * [x] Estimate existing derivatives using *sMRIPrep*'s tooling.
  * [x] Pass in the existing derivatives and poldracklab/smriprep#107
    skips the anatomical workflow fulfilling the workflow API.

Potential problems
------------------
This PR does not contain any measures to preserve the reproducibility of
*fMRIPrep* if the fast-track is used.


oesteban:
I'm hitting this poldracklab/fmriprep#2041 because *sMRIPrep* stores the fsnative-to-T1w (and inverse) matrices in ITK format to the output folder, but *fMRIPrep* expects them to be LTA for LTA concatenation (with the bbregister's LTA).

I think this is the best reason to push nitransforms forward (and actually the reason why we started with it at all originally). It should be capable of:
- Replacing the LTA -> ITK conversions currently done with Freesurfer tools (sMRIPrep)
- Replacing the LTA -> FSL -> ITK conversions with FreeSurfer and C3D (fMRIPrep)
- Replacing all LTA concatenations and storing ITK concatenated (or lists, if we want to keep full provenance and full floating-point accuracy) transforms (both sMRIPrep and fMRIPrep).

Preparing nitransforms for this (generating a package, fixing poldracklab/nitransforms#64) will require a fair amount of work.

WDYT? - should we try to get over with this asap, or find a workaround (i.e., convert back to LTA with FS tools) and finalize later?

oesteban:
Unless I'm missing something, [this report](https://5220-53608443-gh.circle-artifacts.com/0/tmp/ds005/derivatives/fmriprep/sub-01.html) shows results using NiTransforms instead of ``mri_concatenate_lta``, LTA convert and c3d (for LTA -> FSL -> ITK). I'll check in more depth tomorrow, but they look good at first glance.

The final step is to alter the smoke tests to use the new flag (and the proposed feature). It's working locally on ds005, though.

Before that final step, it might be appropriate to have a first review here (cc/ @mgxd, @effigies).

oesteban:
@effigies, @mgxd this seems to be settling in. Let's see the latest CircleCI workflow. If that comes out green, I'll add the fasttrack to ds054 and ds210.

I think it's about time to finalize poldracklab/smriprep#107 and get that one merged - WDYT?

Another thought is that this particular feature will make it easier to quickly tests out things on other extensions (i.e., infants). If we are given a dataset with all the anatomical derivatives we need, we can work right away on the adaptations to the functional workflow (hopefully just the sampling of CIFTI2 files) without being held up by anatomical issues.

oesteban:
Okay, the ``--no-fs-reconall`` jobs are not properly working on Circle. I believe the ordering of the segments of the segmentation is not correct. Will look into this asap

fmriprep/cli/parser.py
```diff
@@ -99,6 +99,11 @@ def _bids_filter(value):
              "https://fmriprep.readthedocs.io/en/%s/faq.html#"
              "how-do-I-select-only-certain-files-to-be-input-to-fMRIPrep" % (
                  currentv.base_version if is_release else 'latest'))
+    g_bids.add_argument(
+        "--anat-derivatives", action='store', nargs="?", metavar="PATH", type=PathExists,
```
mgxd, line 103: why is this setting nargs? can you pass `--anat-derivatives` without any path?

oesteban: Yes, that would point it to the default folder (`<fmriprep_output>/fmriprep/`). But I will double-check just in case.

fmriprep/workflows/bold/resampling.py
```diff
@@ -113,7 +112,8 @@ def select_target(subject_id, space):
                          mem_gb=DEFAULT_MEMORY_MIN_GB)
     resampling_xfm = pe.Node(LTAConvert(in_lta='identity.nofile', out_lta=True),
                              name='resampling_xfm')
-    set_xfm_source = pe.Node(ConcatenateLTA(out_type='RAS2RAS'), name='set_xfm_source')
+    merge_xfm = pe.Node(niu.Merge(2), name="merge_xfm", run_without_submitting=True)
+    concat_xfm = pe.Node(ConcatenateXFMs(out_fmt="fs"), name="concat_xfm")
```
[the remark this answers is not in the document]
oesteban: I think it's left empty because the `filename =` metadata is not set. But there's no failsafe for that in nitransforms right now, so one could think of setting it and getting bitten by this.

fmriprep/cli/parser.py
oesteban: Okay, I have removed the `nargs` because there wasn't any reason to have it. Good catch!

docs/faq.rst
```diff
@@ -271,3 +266,56 @@ Therefore, if your *fMRIPrep* process crashes and you attempt to re-run it reusi
 as much as it could from the previous run, you can either make sure that
 the default ``$PWD/work/`` points to a reasonable, reusable path in your environment or
 configure a better location on your with ``-w <PATH>``.
+
+What is the *anatomical fast-track*?
```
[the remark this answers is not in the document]
oesteban: Separated to the usage section. Hopefully, that sounds more appropriate.

fmriprep/workflows/bold/registration.py
```diff
@@ -685,12 +679,27 @@ def init_fsl_bbr_wf(use_bbr, bold2t1w_dof, name='fsl_bbr_wf'):
...
 
+    if sloppy is True:
+        downsample = pe.Node(niu.Function(
+            function=_conditional_downsampling, output_names=["out_file", "out_mask"]),
+            name='downsample')
+        workflow.connect([
+            (inputnode, downsample, [("t1w_brain", "in_file")]),
+            (wm_mask, downsample, [("out", "in_mask")]),
+            (downsample, flt_bbr, [('out_file', 'reference'),
+                                   ('out_mask', 'wm_seg')]),
```
oesteban, line 694: With the utilization of cached, high-resolution anatomical derivatives, the time it took for flirt/bbr to run exploded. With this patch, I minimized flirt/bbr from ~5min to ~11sec (on my desktop, it must be a bit longer on CircleCI). Some other nodes have extended their duration a bit as when used pre-computed anatomical derivatives, but I think now the timing is reasonable.

I used this simple script: https://gist.github.com/oesteban/0c17459936deb18ac81377db4b4763c2 to measure times. We should consider running something like that in fmriprep to get some stats about execution etc.

setup.cfg
```diff
@@ -25,14 +26,15 @@ install_requires =
...
     nipype ~= 1.4
     nitime
-    niworkflows >= 1.2.0rc4, < 1.3
+    nitransforms >= 20.0.0rc3,<20.2
+    niworkflows >= 1.2.0rc4, < 1.4
     numpy
     pandas
     psutil >= 5.4
-    pybids >= 0.10.0
+    pybids >= 0.10.2
```
oesteban, line 34: This is necessary (uses build_path)

oesteban:
@effigies, @mgxd can you find some time window to take a second pass on this? Let's try to get it merged before Friday - WDYT?

effigies:
Yup. Will have a look this afternoon.

fmriprep/workflows/bold/registration.py
```diff
@@ -456,39 +464,25 @@ def init_bbreg_wf(use_bbr, bold2t1w_dof, omp_nthreads, name='bbreg_wf'):
...
                                 ('subject_id', 'subject_id'),
                                 ('in_file', 'source_file')]),
         # Output ITK transforms
-        (inputnode, lta_concat, [('fsnative2t1w_xfm', 'in_lta2')]),
-        (lta_concat, lta2fsl_fwd, [('out_file', 'in_lta')]),
-        (lta_concat, lta2fsl_inv, [('out_file', 'in_lta')]),
-        (inputnode, fsl2itk_fwd, [('t1w_brain', 'reference_file'),
-                                  ('in_file', 'source_file')]),
-        (inputnode, fsl2itk_inv, [('in_file', 'reference_file'),
-                                  ('t1w_brain', 'source_file')]),
```
effigies, line 480: These were performing the work of ensuring that the ITK-orientation (qform) was being used in the conversion. I guess we're ensuring that `qform == sform` and that's all encoded in the LTA?

oesteban: Sorry I don't know how to address the question, so I'll go step by step and then see if that was what you wanted to clarify.

1. LTA does include q/s-form information in the file, and it is necessary to interpret the transform only if it is VOX2VOX type.
2. NiTransforms converts all LTAs with VOX2VOX type into RAS2RAS type so that the s/q-form information is detached from the transform (i.e., it operates in physical coordinates).
3. NiTransforms writes an ITK transform. ITK transforms are completely(*) unaware of the notion of the images' orientation. (*) this is not 100% true for the case of displacements fields, but this is not the case here.
4. These connections enforced observing q/s-forms because FSL embeds that information within the transform.

Did that get at the bottom of your point?

effigies: Ah, right. It's FSL that does a VOX2VOX. I think we're good. Thanks.

docs/usage.rst
```diff
@@ -100,9 +99,62 @@ would be equivalent to the latest example: ::
     ...
 
 
+.. _prev_derivs:
+
+Reusing precomputed derivatives
+-------------------------------
+Reusing a previous, partial execution of *fMRIPrep*
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+*fMRIPrep* will pick up where it left off a previous execution, so long the work directory
```
mgxd, line 108:
```suggestion
*fMRIPrep* will pick up where it left off a previous execution, so long as the work directory
```

setup.cfg
```diff
@@ -25,14 +26,15 @@ install_requires =
     nibabel ~= 3.0
     nipype ~= 1.4
     nitime
+    nitransforms >= 20.0.0rc3,<20.2
```
mgxd, line 29: should we pin this to master as well?

wrapper/fmriprep_docker.py
```diff
@@ -164,8 +164,15 @@ def merge_help(wrapper_help, target_help):
...
-                            'fs-subjects-dir', 'use-plugin'])
+    expected_overlap = set([
+        'anat-derivatives',
+        'fs-license-file',
+        'fs-subjects-dir',
+        'h',
+        'use-plugin',
+        'version',
+        'w',
+    ])
```
mgxd, line 175: no need to convert between types
```suggestion
    expected_overlap = {
        'anat-derivatives',
        'fs-license-file',
        'fs-subjects-dir',
        'h',
        'use-plugin',
        'version',
        'w',
    }
```

setup.cfg
oesteban: we'll need to set a release pin to cut the RC, so I'd leave this. 20.2 should be tolerant enough to allow nitransforms progress without breaking anything used in here.

oesteban:
brace yourselves

PR #2037, opened 2020-03-15
ENH: Adapt fMRIPrep to upcoming sMRIPrep API changes

oesteban:
In particular, poldracklab/fmriprep#167 introduces a few changes that need to be considered before #1829 is merged:

  - The new `skull_strip_mode` argument for the anatomical workflow (poldracklab/smriprep#167)
  - The functional workflow does not require parameterized inputs
  - The anatomical workflow no longer produces a brain-masked conversion of the T1w.

### DIFF
--- a/fmriprep/workflows/base.py
+++ b/fmriprep/workflows/base.py
@@ -222,16 +222,15 @@ It is released under the [CC0]\
         freesurfer=config.workflow.run_reconall,
         hires=config.workflow.hires,
         longitudinal=config.workflow.longitudinal,
-        num_t1w=len(subject_data['t1w']),
         omp_nthreads=config.nipype.omp_nthreads,
         output_dir=str(config.execution.output_dir),
         reportlets_dir=reportlets_dir,
         skull_strip_fixed_seed=config.workflow.skull_strip_fixed_seed,
-        # skull_strip_mode='force',
+        skull_strip_mode='force',
         skull_strip_template=Reference.from_string(
             config.workflow.skull_strip_template)[0],
         spaces=spaces,
-        # t1w=subject_data['t1w'],
+        t1w=subject_data['t1w'],
     )
 
     workflow.connect([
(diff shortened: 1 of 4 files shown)


oesteban:
This build should be ready to pass. Comments are welcome :)

setup.cfg
```diff
@@ -21,16 +21,16 @@ python_requires = >=3.7
...
-    niworkflows @ git+https://github.com/nipreps/niworkflows.git@76f831f1f8fbf6c277fdf92535639bbf11d84527
+    niworkflows ~= 1.1.9
     numpy
     pandas
     psutil >= 5.4
     pybids >= 0.10.0
     pyyaml
     sdcflows ~= 1.2.0
-    smriprep ~= 0.5.2
+    smriprep >= 0.6.0rc1
```
mgxd:
```suggestion
    smriprep ~= 0.6.0rc1
```

We should keep this strict to avoid future API changes (i.e. 0.7.x)

oesteban:
> Looks like it'll only require adding a flag to the CLI and maybe modifying an output, right?

Yup. The option and a new line in the config module for this option.

fmriprep/workflows/base.py
```diff
@@ -222,16 +222,15 @@ def init_single_subject_wf(subject_id):
         freesurfer=config.workflow.run_reconall,
         hires=config.workflow.hires,
         longitudinal=config.workflow.longitudinal,
-        num_t1w=len(subject_data['t1w']),
         omp_nthreads=config.nipype.omp_nthreads,
         output_dir=str(config.execution.output_dir),
         reportlets_dir=reportlets_dir,
         skull_strip_fixed_seed=config.workflow.skull_strip_fixed_seed,
-        # skull_strip_mode='force',
+        skull_strip_mode='force',
```
mgxd: surely this will have a config option?

setup.cfg
oesteban: I might be wrong, but I think `~=` here will not match 0.6.0 (without the rc). What about:
```suggestion
    smriprep >= 0.6.0rc1,<0.7
```

fmriprep/workflows/base.py
oesteban: yes (we are talking about that right now in #1829 )

setup.cfg
mgxd: yup, that works.

PR #2039, opened 2020-03-17
ENH: Add a ``--skull-strip-t1w`` argument to skip brain extraction

oesteban:
Supersedes #1829.

@adelavega can you check locally this works for you?

### DIFF
--- a/docs/workflows.rst
+++ b/docs/workflows.rst
@@ -68,7 +68,8 @@ single reference template (see `Longitudinal processing`_).
     Likewise, if you know your images are not skull-stripped and the heuristic incorrectly
     determines that they are, you can force skull stripping with ``--skull-strip-t1w force``.
     The default behavior of detecting pre-extracted brains may be explicitly requested with
-    ``---skull-strip-t1w auto``.
+    ``---skull-strip-t1w auto``, which will use a heuristic to check if each image is
+    already masked.
 
 See also *sMRIPrep*'s
 :py:func:`~smriprep.workflows.anatomical.init_anat_preproc_wf`.
--- a/docs/workflows.rst
+++ b/docs/workflows.rst
@@ -55,6 +55,19 @@ single reference template (see `Longitudinal processing`_).
         t1w=['sub-01/anat/sub-01_T1w.nii.gz'],
     )
 
+.. important::
+
+    Occasionally, openly shared datasets may contain preprocessed anatomical images
+    as if they are unprocessed.
+    Particularly, brain-extracted images will crash the anatomical workflow of *fMRIPrep*.
+    It is possible to make *fMRIPrep* adapt to such a circumstance by setting the
+    command line argument ``---skull-strip-t1w auto``.
+    If you know all your T1w images are brain-extracted (i.e., they only contain the brain
+    and have been zeroed elsewhere), you can skip brain extraction for all the subjects
+    with ``---skull-strip-t1w skip``.
+    The regular functioning (i.e., attempt brain extraction always) can be achieved with
+    ``---skull-strip-t1w force``.
+
 See also *sMRIPrep*'s
 :py:func:`~smriprep.workflows.anatomical.init_anat_preproc_wf`.
 
--- a/docs/workflows.rst
+++ b/docs/workflows.rst
@@ -59,14 +59,16 @@ single reference template (see `Longitudinal processing`_).
 
     Occasionally, openly shared datasets may contain preprocessed anatomical images
     as if they are unprocessed.
-    Particularly, brain-extracted images will crash the anatomical workflow of *fMRIPrep*.
-    It is possible to make *fMRIPrep* adapt to such a circumstance by setting the
-    command line argument ``---skull-strip-t1w auto``.
-    If you know all your T1w images are brain-extracted (i.e., they only contain the brain
-    and have been zeroed elsewhere), you can skip brain extraction for all the subjects
-    with ``---skull-strip-t1w skip``.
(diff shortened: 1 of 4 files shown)


docs/workflows.rst
```diff
@@ -55,6 +55,19 @@ single reference template (see `Longitudinal processing`_).
         t1w=['sub-01/anat/sub-01_T1w.nii.gz'],
     )
 
+.. important::
+
+    Occasionally, openly shared datasets may contain preprocessed anatomical images
+    as if they are unprocessed.
+    Particularly, brain-extracted images will crash the anatomical workflow of *fMRIPrep*.
+    It is possible to make *fMRIPrep* adapt to such a circumstance by setting the
+    command line argument ``---skull-strip-t1w auto``.
```
mgxd: since this isn't the default, we should document what it does here.

```suggestion
    command line argument ``---skull-strip-t1w auto``, which will use a heuristic to check if each image is already masked
```

effigies:
Why is the default not "auto"?

adelavega:
I'd think `auto` should be the default as well, otherwise if a user is changing that option its likely they know they need to skip brain extraction. That said, I can understand if we don't think the algorithm is reliable enough (yet?)

fmriprep/config.py
```diff
@@ -439,6 +439,9 @@ class workflow(_Config):
     """Fix a seed for skull-stripping."""
     skull_strip_template = "OASIS30ANTs"
     """Change default brain extraction template."""
+    skull_strip_t1w = "force"
```
effigies:
```suggestion
    skull_strip_t1w = "auto"
```

fmriprep/cli/parser.py
```diff
@@ -209,6 +209,11 @@ def _bids_filter(value):
     g_ants.add_argument('--skull-strip-fixed-seed', action='store_true',
                         help='do not use a random seed for skull-stripping - will ensure '
                              'run-to-run replicability when used with --omp-nthreads 1')
+    g_ants.add_argument(
+        '--skull-strip-t1w', action='store', choices=('auto', 'skip', 'force'), default='force',
```
effigies:
```suggestion
        '--skull-strip-t1w', action='store', choices=('auto', 'skip', 'force'), default='auto',
```

docs/workflows.rst
```diff
@@ -55,6 +55,19 @@ single reference template (see `Longitudinal processing`_).
...
+    If you know all your T1w images are brain-extracted (i.e., they only contain the brain
+    and have been zeroed elsewhere), you can skip brain extraction for all the subjects
+    with ``---skull-strip-t1w skip``.
+    The regular functioning (i.e., attempt brain extraction always) can be achieved with
+    ``---skull-strip-t1w force``.
```
effigies:
```suggestion
    In the case of brain-extracted (skull-stripped) T1w images, attempting to perform
    brain extraction again will often have poor results and may cause *fMRIPrep* to crash.
    By default, *fMRIPrep* will attempt to detect these cases using a heuristic to check if the
    T1w image is already masked.
    If this heuristic fails, and you know your images are skull-stripped, you can skip brain
    extraction with ``--skull-strip-t1w skip``.
    Likewise, if you know your images are not skull-stripped and the heuristic incorrectly
    determines that they are, you can force skull stripping with ``--skull-strip-t1w force``.
    The default behavior of detecting pre-extracted brains may be explicitly requested with
    ``---skull-strip-t1w auto``.
```

oesteban:
I'd rather keep the `'force'` mode as default, since it is currently the "official" operation mode. This flag was suggested to cover a particular use case (when the user is not aware of having T1w images already masked), and crashing first seems like a conservative but effective approach.

Finally, this feature should be superseded by evolutions of poldracklab/smriprep#107 (and #2036) where one can feed fMRIPrep with subsets of precalculated anatomical derivatives.

Is that convincing?

effigies:
I don't see the point of having auto at all, then. The idea was to have something that didn't require using a CLI flag most of the time.

oesteban:
Removing `'auto'` would effectively be the same as #1829 (i.e., a boolean switch). However, when writing this I thought it was worth having `'auto'` for when your dataset has both full and brain-extracted T1w, and you don't know which participants while you trust the heuristic. I can add a warning saying "use with caution" to emphasize this point.

Without a benchmark that says the heuristic works with very high sensitivity and specificity, I feel resistant to making it the default.

oesteban:
For instance @nicholst et al., are processing ds001 and have found *fMRIPrep* gives them very bad results for sub-04.

Looking at the reports, they discovered that the T1w is skull-stripped.

I think having an implicit option would have been potentially dangerous to their type of analysis because it would have been really easy to miss that *fMRIPrep* actually run different workflows for that subject and the others (and this is assuming `'auto'` worked perfectly on all cases).

Setting the default to `'force'` (i.e., the original *fMRIPrep* behavior before this patch) makes everything explicit.

docs/workflows.rst
oesteban: (this suggestion has been added)

nicholst:
Setting `force` as default makes sense for continuity of usage.  If people get some experience with `auto` and find it does a good job catching extracted brains, then later it could be switched to `auto`.

adelavega:
Works for me. FWIW I just tested in on `ds001110` with `auto` and it worked well.

oesteban:
Thanks for the fast check and your patience, @adelavega. This took a lot longer than we would've liked (for the most part, because of me) - many thanks.

PR #2048, opened 2020-03-23
MAINT: Clean up console scripts

oesteban:
- [x] Removed the "sample-openfmri" scripts ([now kept here](https://gist.github.com/oesteban/51e0302292d9a1d3090a2807c100a698)).
- [x] Transferred the bold-mask script over to NiWorkflows (see nipreps/niworkflows#488)


oesteban:
> this looks fine to me - are we sure users aren't using them in console?

The brainmask executable will be still available via niworkflows. The other two, I honestly doubt anyone (including us) is using them.

PR #2065, opened 2020-04-08
REL: 20.1.0rc2

mgxd:
- fixes for troublesome dependencies
- smriprep's latest API changes

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -25,16 +25,16 @@ install_requires =
     nibabel ~= 3.0
     nipype ~= 1.4
     nitime
-    niworkflows >= 1.2.0rc1,<1.3
+    niworkflows >= 1.2.0rc2, < 1.3
     numpy
     pandas
     psutil >= 5.4
     pybids >= 0.10.0
     pyyaml
     sdcflows ~= 1.2.0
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -50,15 +50,17 @@ datalad = datalad
...
+    sphinx >= 1.5.3, < 3
     sphinx-argparse
     sphinx_rtd_theme
     sphinxcontrib-napoleon
 docs =
     %(doc)s
-duecredit = duecredit
+duecredit =
+    duecredit
+    citeproc-py != 0.5.0
```
oesteban, line 63: is this implemented?

effigies: I think this issue has been fixed?

mgxd: i wasn't sure after following https://github.com/brechtm/citeproc-py/issues/94

effigies: Right, I guess it will fail for citeproc-py 0.5.0, but do we generally add carve-outs like this for things up the dependency tree that have bugs?

mgxd: sure - we can probably just remove this hack now that 0.5.1 is out.

PR #500, opened 2017-05-02
[WIP] RF: Rename ants_nthreads -> omp_nthreads

effigies:
This PR moves `ants_nthreads` to `omp_nthreads` to indicate per-process thread limit, rather than ANTs-specific thread limits.

I will also move `recon-all` to respect `omp_nthreads`, rather than `nthreads`.

Closes #499

PR #2071, opened 2020-04-10
FIX: Restore omp-nthreads calculation

### LINKED ISSUE #2070: --omp-nthreads no longer maxing out at 8
Version: 20.1.0rc1

Without setting `--nthreads` or `--omp-nthreads`, I'm getting both set to nprocs.

The old formula for `--omp-nthreads` was `max(1, min(8, nthreads - 1))`.

This probably changed along with the massive configuration object.

effigies:

This restores the default value of` omp-nthreads` based on `n-cpus` established in #500 and removed in #2018.

Fixes #2070.

### DIFF
--- a/fmriprep/cli/parser.py
+++ b/fmriprep/cli/parser.py
@@ -316,7 +316,6 @@ def parse_args(args=None, namespace=None):
     opts = parser.parse_args(args, namespace)
     config.execution.log_level = int(max(25 - 5 * opts.verbose_count, logging.DEBUG))
     config.from_dict(vars(opts))
-    config.loggers.init()
 
     # Initialize --output-spaces if not defined
     if config.execution.output_spaces is None:
--- a/fmriprep/cli/parser.py
+++ b/fmriprep/cli/parser.py
@@ -102,7 +102,7 @@ def _build_parser():
 
     g_perfm = parser.add_argument_group('Options to handle performance')
     g_perfm.add_argument(
-        '--nprocs', '--nthreads', '--n_cpus', '-n-cpus', action='store', type=PositiveInt,
+        '--nprocs', '--nthreads', '--n_cpus', '--n-cpus', action='store', type=PositiveInt,
         help='maximum number of threads across all processes')
     g_perfm.add_argument('--omp-nthreads', action='store', type=PositiveInt,
                          help='maximum number of threads per-process')
--- a/fmriprep/config.py
+++ b/fmriprep/config.py
@@ -251,7 +251,7 @@ class nipype(_Config):
     """Estimation in GB of the RAM this workflow can allocate at any given time."""
     nprocs = os.cpu_count()
     """Number of processes (compute tasks) that can be run in parallel (multiprocessing only)."""
-    omp_nthreads = os.cpu_count()
+    omp_nthreads = None
     """Number of CPUs a single process can access for multithreaded execution."""
     plugin = 'MultiProc'
     """NiPype's execution plugin."""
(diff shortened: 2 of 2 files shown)


fmriprep/cli/parser.py
```diff
@@ -102,7 +102,7 @@ def _bids_filter(value):
 
     g_perfm = parser.add_argument_group('Options to handle performance')
     g_perfm.add_argument(
-        '--nprocs', '--nthreads', '--n_cpus', '-n-cpus', action='store', type=PositiveInt,
+        '--nprocs', '--nthreads', '--n_cpus', '--n-cpus', action='store', type=PositiveInt,
```
effigies, line 105: Sneaking this one in to avoid a separate issue/PR.

fmriprep/cli/parser.py
```diff
@@ -316,7 +316,6 @@ def parse_args(args=None, namespace=None):
     opts = parser.parse_args(args, namespace)
     config.execution.log_level = int(max(25 - 5 * opts.verbose_count, logging.DEBUG))
     config.from_dict(vars(opts))
-    config.loggers.init()
```
effigies, line 319: This is already called in `from_dict()`.

oesteban:
I'll look into this tomorrow

fmriprep/config.py
```diff
@@ -304,6 +304,9 @@ def init(cls):
             }
         })
 
+        if cls.omp_nthreads is None:
+            cls.omp_nthreads = max(1, min(cls.nprocs - 1 if cls.nprocs > 1 else os.cpu_count(), 8))
```
mgxd, line 308: will this ever be 1 (is the outer max necessary)?

effigies: Oh, maybe not. I think the if/else avoids it.

fmriprep/config.py
```diff
@@ -304,6 +304,9 @@ def init(cls):
             }
         })
 
+        if cls.omp_nthreads is None:
+            cls.omp_nthreads = min(cls.nprocs - 1 if cls.nprocs > 1 else os.cpu_count(), 8)
```
oesteban, line 308:
```suggestion
            cls.omp_nthreads = min(os.cpu_count(), 8)
```
I've been willing to remove that nprocs - 1 for a long while. I think we overstepped on the responsibilities of the OS' scheduling. There's really no gain in having a bunch (if any) of fast, lightweight processes running in parallel to antsRegistration in one less cpu.

If you want to have one process running next to antsRegistration with 8 CPUs, it is much better to have --omp-nthreads 8 and make --nprocs 9 than removing one from ants.

effigies: IMO the default behavior should be the best performance (on a typical dataset) without overcommitting processors, and the flags can be used to adapt. Have we tested whether leaving a core free to clear the small-job queue performs better or worse than `omp_nthreads = min(8, cpu_count())`?

If we haven't tested, let's make that a separate issue and just get this one in to get us back to the behavior users have come to expect.

oesteban: Sure, just to be clear - I'm not endorsing here to overcommit processors, I'm just saying that removing one available thread to processes does not make much sense.

And yet, we allow users to set that via the `--omp-settings` flag anyways.

That said, I'm fine to go ahead with this and do `omp_nthreads = min(8, cpu_count())` on a different PR. I haven't tested quantitatively, but in my experience, I'm sure it is much better. That `- 1` forces me to always set `--omp-nthreads`, which is a bit annoying.

effigies: Opened #2078. Thanks for your forbearance.

PR #498, opened 2017-05-01
[RTM] ENH: Use --template flag to specify normalization target

effigies:
This PR separates the normalization directive (`--output-space template`) from the specification of a normalization target via `--template <TEMPLATE>`.

Currently only `MNI152NLin2009cAsym` is supported, but this lays the groundwork for additional targets with minimally intrusive modifications.

See discussion in #487.

Depends on poldracklab/niworkflows#158.

PR #2073, opened 2020-04-10
FIX: Update anatomical connections with *sMRIPrep*

mgxd:
relies on https://github.com/nipreps/niworkflows/pull/498 and https://github.com/poldracklab/smriprep/pull/174

### DIFF
--- a/fmriprep/workflows/base.py
+++ b/fmriprep/workflows/base.py
@@ -276,7 +276,7 @@ tasks and sessions), the following preprocessing was performed.
 
         workflow.connect([
             (anat_preproc_wf, func_preproc_wf,
-             [(('outputnode.t1w_preproc', _pop), 'inputnode.t1w_preproc'),
+             [('outputnode.t1w_preproc', 'inputnode.t1w_preproc'),
               ('outputnode.t1w_mask', 'inputnode.t1w_mask'),
               ('outputnode.t1w_dseg', 'inputnode.t1w_dseg'),
               ('outputnode.t1w_aseg', 'inputnode.t1w_aseg'),
--- a/setup.cfg
+++ b/setup.cfg
@@ -25,14 +25,14 @@ install_requires =
     nibabel ~= 3.0
     nipype ~= 1.4
     nitime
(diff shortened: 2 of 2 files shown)


fmriprep/workflows/base.py
```diff
@@ -275,7 +275,7 @@ def init_single_subject_wf(subject_id):
 
         workflow.connect([
             (anat_preproc_wf, func_preproc_wf,
-             [(('outputnode.t1w_preproc', _pop), 'inputnode.t1w_preproc'),
+             [('outputnode.t1w_preproc', 'inputnode.t1w_preproc'),
```
oesteban, line 278: You only modified the N4-only workflow -- are you positive this will work with the default brain extraction workflow?

oesteban: Okay, I just saw #498.

mgxd: local testing with https://github.com/poldracklab/smriprep/pull/174 seems to work with both branches of `--skull-strip-t1w` - `brain_extraction_wf` and `n4_only_wf`

mgxd:
once we cut a new version of smriprep / niworkflows, this can go in

oesteban:
Let's cut a new RC, can you take on that?

oesteban:
BTW, are you positive a new release of niworkflows is necessary? Is smriprep pinning a particular commit of niworkflows right now?

mgxd:
i'm fairly sure, since https://github.com/nipreps/niworkflows/commit/50fec8b8873cd76cbdbd7c39305e676c9990ce99 isn't in rc3

oesteban:
Then this line https://github.com/poldracklab/smriprep/blob/master/setup.cfg#L28 should be something like

```
smriprep @ git+https://github.com/nipreps/niworkflows.git@master
```

isn't it?

oesteban:
What I mean, forcing ourselves to have linked dependencies would preempt messing up with releases. If I look at smriprep now, I would think it does not require a niworkflows release. Does it make sense? Should we update our contribution procedures to make this more salient?

PR #2080, opened 2020-04-14
MAINT: reduce number of warnings

mgxd:
follows the effort of https://github.com/poldracklab/mriqc/pull/838 in hopes of reducing warnings out of the user's control.

- Avoid niworkflow imports at top level
- Port over warnings filter from mriqc

### DIFF
--- a/fmriprep/_warnings.py
+++ b/fmriprep/_warnings.py
@@ -1,0 +1,36 @@
+"""Manipulate Python warnings."""
+import warnings
+import logging
+
+_wlog = logging.getLogger("py.warnings")
+_wlog.addHandler(logging.NullHandler())
+
+
+def _warn(message, category=None, stacklevel=1, source=None):
+    """Redefine the warning function."""
+    if category is not None:
+        category = type(category).__name__
+        category = category.replace("type", "WARNING")
+
+    logging.getLogger("py.warnings").warning(
+        f"{category or 'WARNING'}: {message}"
+    )
+
+
+def _showwarning(message, category, filename, lineno, file=None, line=None):
+    _warn(message, category=category)
+
+
+warnings.warn = _warn
+warnings.showwarning = _showwarning
+
+# warnings.filterwarnings("ignore", category=FutureWarning)
+# warnings.filterwarnings("ignore", category=DeprecationWarning)
+# warnings.filterwarnings("ignore", category=ResourceWarning)
+# # cmp is not used by mriqc, so ignore nipype-generated warnings
+# warnings.filterwarnings("ignore", "cmp not installed")
+# warnings.filterwarnings(
+#     "ignore", "This has not been fully tested. Please report any failures."
+# )
+# warnings.filterwarnings("ignore", "sklearn.externals.joblib is deprecated in 0.21")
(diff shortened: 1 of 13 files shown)


fmriprep/_warnings.py
```diff
@@ -0,0 +1,36 @@
...
+# warnings.filterwarnings("ignore", category=FutureWarning)
+# warnings.filterwarnings("ignore", category=DeprecationWarning)
+# warnings.filterwarnings("ignore", category=ResourceWarning)
+# # cmp is not used by mriqc, so ignore nipype-generated warnings
+# warnings.filterwarnings("ignore", "cmp not installed")
+# warnings.filterwarnings(
+#     "ignore", "This has not been fully tested. Please report any failures."
+# )
+# warnings.filterwarnings("ignore", "sklearn.externals.joblib is deprecated in 0.21")
+# warnings.filterwarnings("ignore", "can't resolve package from __spec__ or __package__")
```
effigies, line 36: If we don't need these, let's just remove them.

fmriprep/config.py
```diff
@@ -89,24 +80,23 @@
...
-warnings.showwarning = redirect_warnings
+if not hasattr(sys, "_is_pytest_session"):
+    sys._is_pytest_session = False  # Trick to avoid sklearn's FutureWarnings
+# Disable all warnings in main and children processes only on production versions
+if not any((
+    "+" in __version__,
+    __version__.endswith(".dirty"),
+    os.getenv("FMRIPREP_DEV", "0").lower() in ("1", "on", "true", "y", "yes")
+)):
+    os.environ["PYTHONWARNINGS"] = "ignore"
```
effigies, line 99: Circle is not showing the annoying warnings, but it looks like this should be permitting them.

mgxd: from @oesteban mriqc PR:
> If development, warnings.showwarning and warnings.warn are monkeypatched to redirect to the 'py.warnings' logger. The logger writes to /dev/null, but that can we change for debugging purposes.

it might be a good idea to keep these visible in dev versions though, so we can change this.

mgxd: i disabled this monkeypatch on development versions by default, but added a check for `FMRIPREP_WARNINGS` that will enable it.

fmriprep/config.py
```diff
@@ -89,24 +79,30 @@
...
+    "+" in __version__,
+    __version__.endswith(".dirty"),
+    os.getenv("FMRIPREP_DEV", "0").lower() in ("1", "on", "true", "y", "yes")
+)):
+    from ._warnings import logging
+    os.environ["PYTHONWARNINGS"] = "ignore"
+elif os.getenv("FMRIPREP_WARNINGS", "0").lower() in ("1", "on", "true", "y", "yes"):
+    # allow disabling warnings on development versions
+    # https://github.com/poldracklab/fmriprep/pull/2080#discussion_r409118765
+    from .warnings import logging
```
oesteban, line 103:
```suggestion
    from ._warnings import logging
```

?

mgxd: yes, silly me.

PR #2081, opened 2020-04-15
REL: 1.5.10

effigies:
Proposal for release of 1.5.10, which fixes a bug that has been present in *all* versions of fMRIPrep, although it would only affect runs with phase-difference fieldmaps in some orientation besides RAS+.


CHANGES.rst
```diff
@@ -1,3 +1,14 @@
+1.5.10 (April 15, 2020)
+=======================
+Bug-fix release in the 1.5.x series.
+
+This release fixes a bug for phase-difference fieldmaps that are not in RAS+ orientation.
```
mgxd, line 5: to emphasize
```suggestion
This release fixes a bug for **phase-difference fieldmaps that are not in RAS+ orientation**.
```

effigies:
Yeah, assuming we're all on board with this as both a correct diagnosis and the right fix, then I'll handle the sdcflows/fmriprep releases needed for 20.0.x as well.

.versions.json
```diff
@@ -1,10 +1,82 @@
 {
```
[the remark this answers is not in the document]
effigies: We should push it once we have a 20.0.x fix. Otherwise it'll warn people with no useful action to take.

CHANGES.rst
oesteban:
```suggestion
This release fixes a bug for **phase-difference fieldmaps that are not in RAS+ orientation**.
When the non-canonical orientation (any other than RAS+) included a reordering of axis
(e.g. ARS+), *fMRIPrep* crashed when the number of voxels of the swapped dimensions did not
match.
Otherwise, the bug introduced a poor masking - oftentimes unnoticeable if the original orientation
was LAS+)
```

CHANGES.rst
```diff
@@ -1,3 +1,14 @@
+1.5.10 (April 15, 2020)
+=======================
+Bug-fix release in the 1.5.x series.
+
+This release fixes a bug for **phase-difference fieldmaps that are not in RAS+ orientation**.
```
oesteban, line 5:
```suggestion
This release fixes a bug for **phase-difference fieldmaps that are not in RAS+ orientation**.
When the non-canonical orientation (any other than RAS+) included a reordering of axis
(e.g. ARS+), *fMRIPrep* crashed when the number of voxels of the swapped dimensions did not
match.
Otherwise, the bug introduced a poor masking - oftentimes unnoticeable if the original orientation
was LAS+)
```

effigies:
Thanks for resubmitting.

effigies:
It's gotten late, so I'll aim to do this tomorrow morning. I'm going to reword the release text slightly and squash the commits.

effigies:
Rebased and tagged in b529274bc37c2d29ab5a859f69379fa6ab4a4eeb.

PR #2086, opened 2020-04-16
ENH: Keep track and allow reuse of random seeds with the config module

mgxd:
~~This PR sets and tracks `ANTS_RANDOM_SEED` environmental for increased run-to-run reproducibility. By default, this value is set to a string representation of the current date in `YYYYMMDD` format. Additionally, a parser option `--ants-fixed-seed` is added to allow manually setting this option.~~

This PR adds a new section to the config, *seeds*, that tracks and sets random seeds. Firstly, a master seed is generated, or can be passed in via the new `--random-seed` argument. This seed initiates the PRNG and reproducibly generates program specific seeds. ANTs is the only seed currently tracked.

NOTE: `ANTS_RANDOM_SEED` is only used for some ANTs interfaces (https://github.com/ANTsX/ANTsR/issues/210#issuecomment-538812296)

### DIFF
--- a/fmriprep/cli/parser.py
+++ b/fmriprep/cli/parser.py
@@ -8,7 +8,6 @@ from .. import config
 
 def _build_parser():
     """Build parser object."""
-    from datetime import datetime
     from functools import partial
     from pathlib import Path
     from argparse import (
@@ -46,8 +45,6 @@ def _build_parser():
         from json import loads
         if value and Path(value).exists():
             return loads(Path(value).read_text())
-
-    _date_to_num = datetime.now().strftime('%Y%M%d')
 
     verstr = f'fMRIPrep v{config.environment.version}'
     currentv = Version(config.environment.version)
@@ -177,6 +174,10 @@ https://fmriprep.readthedocs.io/en/%s/spaces.html""" % (currentv.base_version
     g_conf.add_argument(
         '--dummy-scans', required=False, action='store', default=None, type=int,
         help='Number of non steady state volumes.')
+    g_conf.add_argument(
+        '--random-seed', action='store', type=int, default=None,
+        help='Initialize the random seed for the workflow'
+    )
 
     # ICA_AROMA options
     g_aroma = parser.add_argument_group('Specific options for running ICA_AROMA')
@@ -214,9 +215,7 @@ https://fmriprep.readthedocs.io/en/%s/spaces.html""" % (currentv.base_version
     g_ants.add_argument('--skull-strip-fixed-seed', action='store_true',
                         help='do not use a random seed for skull-stripping - will ensure '
                              'run-to-run replicability when used with --omp-nthreads 1 and '
-                             '--ants-fixed-seed <int>')
-    g_ants.add_argument('--ants-fixed-seed', action='store', default=_date_to_num,
-                        help='Seed to use for antsRegistration')
+                             'matching --master-random-seed <int>')
     g_ants.add_argument(
         '--skull-strip-t1w', action='store', choices=('auto', 'skip', 'force'), default='force',
         help="determiner for T1-weighted skull stripping ('force' ensures skull "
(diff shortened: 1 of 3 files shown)


fmriprep/cli/parser.py
```diff
@@ -46,6 +47,8 @@ def _bids_filter(value):
         if value and Path(value).exists():
             return loads(Path(value).read_text())
 
+    _date_to_num = datetime.now().strftime('%Y%M%d')
```
mgxd, line 50: if this isn't random enough, we could always elect for a more "random" seed, aka
```python
random.choice(range(10000))
```

effigies: I'm fine with the current date, as all we're looking for is reproducibility. If we do want to poll the current PRNG for something with more bits of entropy, I would suggest using `numpy.random.randint`. A bit more direct than `choice(range(n))`.

effigies:
What if we just have one master seed, and generate the derivative seeds as the first n seeds to pop out of the PRNG?

oesteban:
> What if we just have one master seed, and generate the derivative seeds as the first n seeds to pop out of the PRNG?

I think this is the right approach. I would keep those first n seeds as named fields in the config, so that the order of seeds is not a problem when inserting a new sub-workflow requiring seeds between two existing ones (in the nipype workflow build sense)

effigies:
We should either put a master seed in the config or all seeds in the config, but not both. Consider the case where I specify the master seed and the ANTs seed; what's the process for reliably generating any other seeds? Obviously they'll be recorded, but the logic gets complicated to ensure that the PRNG state is the same at the start of run regardless of the exact set of provided/generated seeds.

mgxd:
Okay, this is ready for another batch of reviews. I've adjusted the implementation to create a *seeds* section within the config to monitor each seed we generate / use.

One question I've been pondering - instead of exposing the random seed directly (via `--random-seed`), should the command line option just be a high level `--reuse-config` switch that reuses certain options (including the master seed) if a previous config file exists. ATM, I'm leaning towards the latter.

fmriprep/cli/parser.py
```diff
@@ -174,6 +174,10 @@ def _bids_filter(value):
     g_conf.add_argument(
         '--dummy-scans', required=False, action='store', default=None, type=int,
         help='Number of non steady state volumes.')
+    g_conf.add_argument(
+        '--random-seed', action='store', type=int, default=None,
+        help='Initialize the random seed for the workflow'
+    )
```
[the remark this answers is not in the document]
mgxd: yep, let's punt reuse to a separate PR, since we'll be pulling more than random seeds.

if we want to preserve the current output structure, my workflow for reuse was essentially:
- check if `fmriprep/sub-N/log` exists
- use the latest mtime of potential uuid folders
- load up toml

I still think we should be explicit when reusing a previous config file, whether its in the workdir or outputs.

mgxd:
> the config file is reused if the work directory is kept

is this currently implemented?

oesteban:
I believe so.

mgxd:
it looks like its being assigned and written to disk before loading?

https://github.com/poldracklab/fmriprep/blob/2a90550e8b810ffb7e4508aacd8acc9e5416c101/fmriprep/cli/run.py#L7-L16

https://github.com/poldracklab/fmriprep/blob/2a90550e8b810ffb7e4508aacd8acc9e5416c101/fmriprep/cli/parser.py#L311-L318

https://github.com/poldracklab/fmriprep/blob/1d93d9e5599d08347ba2b85a4d4590a4c9273814/fmriprep/cli/run.py#L27-L28

anyways, we can add in the behavior in a future PR - perhaps with `--reuse-config-file` using the following priority
1) manually passed in config
2) config in working directory
3) latest config in output directory

WDYT?

oesteban:
I would go with 2 and 3. Perhaps a new flag, but only to preempt the reuse
of the config.

effigies:
> We should either put a master seed in the config or all seeds in the config, but not both.

So it looks like we did both, but only the master seed has any effect. I guess this is useful for reporting, but it feels potentially confusing

oesteban:
We are not exposing the config file to the bulk of users, and I'd like to push hard to keep that if we can. I guess that allows this particular feature to be confusing.

On the other hand, having the seeds that have been actually set will help us trace back related errors.

Happy to reconsider my all-in position w.r.t. not exposing the config file to the users.

PR #2092, opened 2020-04-18
DOC: Pin nipype with new sphinx extension syntax

effigies:
Testing nipy/nipype#3206.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -54,7 +54,7 @@ doc =
     packaging
     pydot >= 1.2.3
     pydotplus
-    sphinx >= 1.5.3
+    sphinx >= 1.8
     sphinx-argparse
     sphinx_rtd_theme
     sphinxcontrib-napoleon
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -52,7 +52,7 @@ doc =
     packaging
     pydot >= 1.2.3
     pydotplus
-    sphinx >= 1.5.3, < 3
+    sphinx >= 1.5.3
```
oesteban, line 55: when the new API of sphinx started? as early as 1.5.3?

effigies: 1.8

oesteban: Then
```suggestion
    sphinx >= 1.8
```
I guess?

effigies: Yup. I've got a few commits I haven't pushed yet. I was just using this to test for the moment. I would say let's make this a real PR when we depend on 1.5.0.

PR #2099, opened 2020-04-22
ENH: add options to patch in other dependencies

mgxd:
While testing the wrapper, I realize we don't make it easy to patch all our "in-house" dependencies.

This PR removes package specific `--patch-x` calls and instead adds a new argument, `--patch`, which supports one or more key value pairs of package, package path.

For example:
```
fmriprep-docker ... --patch fmriprep=/code/fmriprep/fmriprep smriprep=/code/smriprep/smriprep
```

### DIFF
--- a/wrapper/fmriprep_docker.py
+++ b/wrapper/fmriprep_docker.py
@@ -218,6 +218,15 @@ def is_in_directory(filepath, directory):
 def get_parser():
     """Defines the command line interface of the wrapper"""
     import argparse
+
+    class ToDict(argparse.Action):
+        def __call__(self, parser, namespace, values, option_string=None):
+            d = {}
+            for kv in values:
+                k, v = kv.split("=")
+                d[k] = os.path.abspath(v)
+            setattr(namespace, self.dest, d)
+
     parser = argparse.ArgumentParser(
         description=__doc__,
         formatter_class=argparse.ArgumentDefaultsHelpFormatter,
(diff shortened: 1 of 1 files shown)


effigies:
Was thinking about this recently. It's getting to be a lot of options where it started with three. What if we switch it to `--patch pkg=path`?

mgxd:
@effigies I agree - and given the fact those options are developer specific and we're gearing up for new minor version, I see this as a good time to purge the current options altogether.

wrapper/fmriprep_docker.py
```diff
@@ -371,11 +373,12 @@ def main():
...
 
     # Patch working repositories into installed package directories
-    for pkg in ('fmriprep', 'niworkflows', 'nipype'):
-        repo_path = getattr(opts, 'patch_' + pkg)
-        if repo_path is not None:
-            command.extend(['-v',
-                            '{}:{}/{}:ro'.format(repo_path, PKG_PATH, pkg)])
+    if opts.patch:
+        for pkg, repo_path in opts.patch.items():
+            if repo_path is not None:
```
effigies, line 378: Is it possible for `repo_path` to be `None`?

wrapper/fmriprep_docker.py
```diff
@@ -279,15 +288,8 @@ def get_parser():
...
-    g_dev.add_argument('-f', '--patch-fmriprep', metavar='PATH',
-                       type=os.path.abspath,
-                       help='working fmriprep repository')
-    g_dev.add_argument('-n', '--patch-niworkflows', metavar='PATH',
-                       type=os.path.abspath,
-                       help='working niworkflows repository')
-    g_dev.add_argument('-p', '--patch-nipype', metavar='PATH',
-                       type=os.path.abspath,
-                       help='working nipype repository')
+    g_dev.add_argument('--patch', nargs="+", metavar="PACKAGE=PATH", action=ToDict,
```
effigies, line 291:
```suggestion
    g_dev.add_argument('-p', '--patch', nargs="+", metavar="PACKAGE=PATH", action=ToDict,
```

mgxd: i'd prefer to leave this explicit tbh, especially since `-p` was used for a specific patch before. though if you feel strongly, we can add it in.

wrapper/fmriprep_docker.py
mgxd: i don't think so, but wanted to cover any bases i missed. let's remove it and see what happens

wrapper/fmriprep_docker.py
effigies: I don't feel strongly.

PR #2101, opened 2020-04-24
FIX: Rename use_syn config option to use_syn_sdc to match CLI

### LINKED ISSUE #1941: `--use-syn-sdc` no longer respected when `--ignore fieldmaps` is passed
Prior to 1.5.3, `--use-syn-sdc --ignore fieldmaps` would ignore fieldmaps, and use SyN-SDC.

As of 9e2320365467f7c3d35d43de4b1cc8dad3636975 (#1886), access to [any SDC algorithm](https://github.com/poldracklab/fmriprep/commit/9e2320365467f7c3d35d43de4b1cc8dad3636975#diff-e705526667f071ca0c3b04c2bd96272bR351-R354) is restricted by `--ignore fieldmaps`.

This leaves users who would prefer to use SyN over fieldmaps stranded.

Ref: https://neurostars.org/t/syn-sdc-not-working-with-more-recent-fmriprep-version/5824

effigies:

Proposing in case this is the correct fix.

* FIX: Rename use_syn config option to use_syn_sdc to match CLI

Closes #1941 (again).

### DIFF
--- a/fmriprep/config.py
+++ b/fmriprep/config.py
@@ -456,7 +456,7 @@ class workflow(_Config):
     """Run ICA-:abbr:`AROMA (automatic removal of motion artifacts)`."""
     use_bbr = None
     """Run boundary-based registration for BOLD-to-T1w registration."""
-    use_syn = None
+    use_syn_sdc = None
     """Run *fieldmap-less* susceptibility-derived distortions estimation
     in the absence of any alternatives."""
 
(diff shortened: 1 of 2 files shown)


mgxd:
the cleanest way would probably be to add `dest=use_syn` to the parser argument

effigies:
I can do that, but why is that cleaner? SyN isn't SDC-specific, so `use_syn` is less clear to me.

Let me know if you still want me to switch.

mgxd:
i meant in terms of lines changed, but IDRC

effigies:
I'd prefer keeping the variable name consistent with the CLI just to keep things simple.

PR #2107, opened 2020-04-28
ENH: ability to reuse existing config file

mgxd:
Allows loading configuration files from previous fmriprep runs.

By default, all fields besides `run_uuid` are copied. Most options can be overwritten by CLI options.

### DIFF
--- a/fmriprep/cli/parser.py
+++ b/fmriprep/cli/parser.py
@@ -289,6 +289,7 @@ https://fmriprep.readthedocs.io/en/%s/spaces.html"""
     )
     g_conf.add_argument(
         "--random-seed",
+        dest="_random_seed",
         action="store",
         type=int,
         default=None,
--- a/fmriprep/cli/parser.py
+++ b/fmriprep/cli/parser.py
@@ -543,10 +543,15 @@ def parse_args(args=None, namespace=None):
     parser = _build_parser()
     opts = parser.parse_args(args, namespace)
 
-    new_config = os.getenv('FMRIPREP_NO_REUSE_CONFIG') or opts.reports_only
+    new_config = os.getenv('FMRIPREP_NO_REUSE_CONFIG')
     if not new_config:
         # attempt to reuse previous config
-        config.load_previous_config(opts.output_dir, opts.work_dir, opts.participant_label)
+        config.load_previous_config(
+            opts.output_dir,
+            opts.work_dir,
+            participant=opts.participant_label,
+            new_uuid=not opts.reports_only,
+        )
 
     config.execution.log_level = int(max(25 - 5 * opts.verbose_count, logging.DEBUG))
     config.from_dict(vars(opts))
--- a/fmriprep/cli/parser.py
+++ b/fmriprep/cli/parser.py
@@ -476,13 +476,6 @@ https://fmriprep.readthedocs.io/en/%s/spaces.html"""
         "aggregation, not reportlet generation for specific nodes.",
     )
     g_other.add_argument(
-        "--run-uuid",
-        action="store",
-        default=None,
-        help="Specify UUID of previous run, to include error logs in report. "
-        "No effect without --reports-only.",
-    )
-    g_other.add_argument(
         "--write-graph",
         action="store_true",
         default=False,
@@ -541,12 +534,19 @@ discourage its usage."""
 
 def parse_args(args=None, namespace=None):
     """Parse args and run further checks on the command line."""
+    import os
     import logging
     from niworkflows.utils.spaces import Reference, SpatialReferences
     from niworkflows.utils.misc import check_valid_fs_license
 
     parser = _build_parser()
     opts = parser.parse_args(args, namespace)
+
+    new_config = os.getenv('FMRIPREP_NO_REUSE_CONFIG') or opts.reports_only
+    if not new_config:
+        # attempt to reuse previous config
+        config.load_previous_config(opts.output_dir, opts.work_dir, opts.participant_label)
+
     config.execution.log_level = int(max(25 - 5 * opts.verbose_count, logging.DEBUG))
     config.from_dict(vars(opts))
 
--- a/fmriprep/config.py
+++ b/fmriprep/config.py
@@ -630,7 +630,7 @@ def init_spaces(checkpoint=True):
     workflow.spaces = spaces
 
 
-def load_previous_config(output_dir, work_dir, participant=None):
+def load_previous_config(output_dir, work_dir, participant=None, new_uuid=True):
     """
     Search for existing config file, and carry over previous options.
     If a participant label is specified, the output directory is searched for
@@ -642,10 +642,6 @@ def load_previous_config(output_dir, work_dir, participant=None):
         Existing config found and loaded
 
     """
-    # settings to avoid reusing
-    skip = {
-        'execution': ('run_uuid',)
-    }
     conf = None
     if participant:
         # output directory
@@ -663,6 +659,8 @@ def load_previous_config(output_dir, work_dir, participant=None):
 
     # load existing config
     if conf is not None:
+        # settings to avoid reusing
+        skip = {} if new_uuid else {'execution': ('run_uuid',)}
         print(f"Reusing config file {conf}")
         load(conf, skip=skip)
         return True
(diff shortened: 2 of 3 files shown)


fmriprep/config.py
```diff
@@ -627,3 +627,39 @@ def init_spaces(checkpoint=True):
...
+    if conf is None:
+        # work directory
+        conf = Path(work_dir) / f'.{CONFIG_FILENAME}'
+        if not conf.exists():
+            conf = None
+
+    # load existing config
+    if conf is not None:
+        print(f"Reusing config file {conf}")
+        load(conf)
```
oesteban, line 662: this should load in the old uuid. Why don't you just override it here? (instead of within the parser)

mgxd: we could, but then we'd have the logic in two separate places (here + attribute definition).

in the parser, it is hit once regardless of the presence of an existing config

mgxd:
@oesteban sounds good, I've taken your comments into consideration for the latest commit.

fmriprep/cli/parser.py
```diff
@@ -541,12 +534,18 @@ def _bids_filter(value):
...
     """Parse args and run further checks on the command line."""
+    import os
     import logging
     from niworkflows.utils.spaces import Reference, SpatialReferences
     from niworkflows.utils.misc import check_valid_fs_license
 
     parser = _build_parser()
     opts = parser.parse_args(args, namespace)
+
+    if os.getenv('FMRIPREP_NO_REUSE_CONFIG') is not None:
```
oesteban, line 545: I'm not 100% positive this would be the right thing to do:
```suggestion
    # IMPORTANT: For the --reports-only operation, the last UUID must be retrieved.
    if os.getenv('FMRIPREP_NO_REUSE_CONFIG') is not None or opts.reports_only is not None:
```
but I'm sure we want to do something like this with `--reports-only`. Probably this feature does not work anymore, but that's out of the scope of the PR. Let's try to minimize the impact of removing ``--run-uuid`` somehow.

fmriprep/config.py
```diff
@@ -347,7 +353,7 @@ class execution(_Config):
     the command line) as spatial references for outputs."""
     reports_only = False
     """Only build the reports, based on the reportlets found in a cached working directory."""
-    run_uuid = '%s_%s' % (strftime('%Y%m%d-%H%M%S'), uuid4())
+    run_uuid = _gen_run_uuid()
```
oesteban, line 356:
```suggestion
    run_uuid = f"{strftime('%Y%m%d-%H%M%S')}_{uuid4()}"
```

fmriprep/config.py
```diff
@@ -81,9 +83,7 @@
     import sys
     import random
 
-    from uuid import uuid4
```
oesteban, line 84:
```suggestion
    from uuid import uuid4
```

fmriprep/config.py
```diff
@@ -81,9 +83,7 @@
...
     from pathlib import Path
-    from time import strftime
```
oesteban, line 86:
```suggestion
    from time import strftime
```

fmriprep/config.py
```diff
@@ -157,6 +157,12 @@
     pass
 
 
+def _gen_run_uuid():
+    from uuid import uuid4
+    from time import strftime
+    return f"{strftime('%Y%m%d-%H%M%S')}_{uuid4()}"
+
+
```
oesteban, line 165: Let's not pollute with more short-lived utility functions, whenever convenient.
```suggestion
```

fmriprep/cli/parser.py
```diff
@@ -541,12 +534,18 @@ def _bids_filter(value):
...
+        # reuse config if available
+        config.load_previous_config(opts.output_dir, opts.work_dir, opts.participant_label)
+
     config.execution.log_level = int(max(25 - 5 * opts.verbose_count, logging.DEBUG))
     config.from_dict(vars(opts))
```
oesteban, line 550: This will rewrite all the options from the command line. Now that we removed opts.run_uuid, this should be safe - can you have a deep look into what should/can/may be overwritten here?

fmriprep/config.py
mgxd: unless we plan on add functionality to `config.load` to avoid certain fields, the easiest way to do this is to set run_uuid again after loading. it seems sloppy to manage the specific syntax in two places, whereas having a utility function will make maintenance easier in the long run

fmriprep/config.py
```diff
@@ -624,3 +630,40 @@ def init_spaces(checkpoint=True):
...
+        # work directory
+        conf = Path(work_dir) / f'.{CONFIG_FILENAME}'
+        if not conf.exists():
+            conf = None
+
+    # load existing config
+    if conf is not None:
+        print(f"Reusing config file {conf}")
+        load(conf)
+        execution.run_uuid = _gen_run_uuid()
```
oesteban, line 666:
```suggestion
        load(conf, skip=("execution.run_uuid",))
```

Maybe it would be more generalizable to inform the load function of fields that should be skipped? (this suggestion requires the implementation of the skip arg)

That said, skipping ``execution.run_uuid`` will be necessary/desired in all instances?

fmriprep/config.py
oesteban: I think adding a `skip` argument to `config.load` could be very useful here (and in the future) - and in this particular case workarounds your legitimate concern.

fmriprep/config.py
```diff
@@ -167,15 +169,16 @@ def __init__(self):
...
 
     @classmethod
-    def load(cls, settings, init=True):
+    def load(cls, settings, init=True, ignore=None):
         """Store settings from a dictionary."""
         for k, v in settings.items():
+            if ignore and k in ignore:
+                continue
             if v is None:
                 continue
```
oesteban, line 178:
```suggestion
        ignore =  ignore or {}
        for k, v in settings.items():
            if k in ignore or v is None:
                continue
```

fmriprep/config.py
```diff
@@ -167,15 +169,16 @@ def __init__(self):
...
             if k in cls._paths:
                 setattr(cls, k, Path(v).absolute())
-                continue
-            if hasattr(cls, k):
+            elif hasattr(cls, k):
```
oesteban, line 181: what's the use case for this change (if -> elif)

fmriprep/config.py
```diff
@@ -510,18 +511,19 @@ def init(cls):
...
-    """Master seed used to generate all other tracked seeds"""
+    """Master random seed to initialize the Pseudorandom Number Generator (PRNG)"""
     ants = None
     """Seed used for antsRegistration, antsAI, antsMotionCorr"""
 
     @classmethod
     def init(cls):
-        cls.master = workflow.random_seed
+        if cls._random_seed is not None:
+            cls.master = cls._random_seed
```
oesteban, line 523: Can you elaborate a bit more about this change? It looks more compact than before, but I'm not sure I fully understand what are you trying to improve here.

fmriprep/config.py
```diff
@@ -624,3 +628,41 @@ def init_spaces(checkpoint=True):
...
+        # work directory
+        conf = Path(work_dir) / f'.{CONFIG_FILENAME}'
+        if not conf.exists():
+            conf = None
+
+    # load existing config
+    if conf is not None:
+        # settings to avoid reusing
+        skip = {} if new_uuid else {'execution': ('run_uuid',)}
+        print(f"Reusing config file {conf}")
```
oesteban, line 664: should we send this to stderr or use the loggers config (if the latter, we probably want to log a message after the fact - "Loaded previous config file {conf}", and do it after line 665)?

fmriprep/config.py
mgxd: it was just to remove the continue

fmriprep/config.py
mgxd: sure, i originally put `random_seed` within the `execution` config class, but this PR shifts it to a private variable within the `seeds` class. Additionally, this check needs to be added when reusing configs to ensure seeds are being reused, while also allowing command line options to override.

oesteban:
I think we can push this to 20.2 for our peace of mind - wdyt?

mgxd:
i'm fine with that 👍

oesteban:
@mgxd, may I ask you to submit another PR just running `black fmriprep/config.py` first, so that this, #2177, and future changes are easier to maintain?

mgxd:
sure, this needs to be given another look anyways

PR #1800, opened 2019-09-30
[WIP, FIX] remove bold_mask generation from t2s

emdupre:
## Changes proposed in this pull request

Addresses #1786  

Removes `bold_mask` generation from `bold_t2s_wf` in the case of multi-echo data. 

## Documentation that should be reviewed

N/A.

PR #2108, opened 2020-04-28
PIN: nipreps/niworkflows#485

effigies:
Merged nipreps/master into nipreps/niworkflows#485 and pushed to my own remote. Opening PR to check that we get sensible brain masks on CI.

cc @oesteban If this passes, I'll merge the niworkflows PR.

Issue #1010, opened 2018-02-27
Multi-echo: replace CompCor with ME-ICA


oesteban:
Comes from #891

emdupre:
This will likely involve working with [ME-ICA/tedana](https://github.com/me-ica/tedana/) !

emdupre:
tedana is gearing up for its first release 🎉  
I could open a PR incorporating it here, since it will be available on pypi, but I have a few questions:

- [MDP](http://mdp-toolkit.sourceforge.net/) is still a dependency for tedana, although we're planning to [remove this in a later release](https://github.com/ME-ICA/tedana/issues/14). Would you prefer to wait until this is updated, to avoid introducing another dependency ?
- I'm planning to just sub it out for the AROMA connections, and pass along the T2* map we're already generating. I don't think it makes sense to allow for both AROMA and tedana, but if you have thoughts otherwise, I'd love to hear them !

tsalo:
Regarding the use of AROMA _and_ ME-ICA, what if the ICA step was separated from the AROMA step? Tedana can perform its classification on existing decompositions, as long as they're in the right format, which _should_ be BEP-002 compatible. I know that there are plans to replace the current AROMA dependency (#1784), so as long as the new implementation can accept BIDS-compatible decomposition files, then component classification (whether tedana or AROMA) becomes modular.

Tedana outputs its classifications in addition to the denoised data, so if fMRIPrep retained the classifications and the decomposition file(s), but _not_ the denoised data, then the tedana-selected noise components could be retained in the confounds file just like AROMA noise components, and users could denoise the preprocessed data as they see fit.

effigies:
@tsalo We have a user that wants to use fMRIPrep + Tedana denoising. Right now, they're extracting files from the working directory, denoising, and resampling. From your comment, it sounds like an alternative approach (short of doing this correctly inside fMRIPrep) would be for them to create confounds and write them out in TSV files. Is that correct, and is it as simple as performing MELODIC on the optimal combination that comes out of the T2SMap node?

tsalo:
For tedana denoising, you need the STC+MC echo-wise files in native space in order to calculate the metrics necessary for the classification step. While tedana does support external ICAs like MELODIC, you still need access to the partially preprocessed echo-wise data.

[...]

PR #2109, opened 2020-04-29
ENH: Remove t2s-coreg

### LINKED ISSUE #1786: RFC: Updating t2s_coreg for MEEPI
Recently, @shotgunosine showed on [NeuroStars](https://neurostars.org/t/registrations-and-brainmasks/4876) that there are some reasons to be concerned about MEEPI processing with `t2s_coreg`. 

Specifically, because T2* maps are calculated voxelwise, voxels with invalid T2* values are filled with zeroes, and the images therefore appear "skullstripped" or "masked" in some way (I'll call them "clipped," here). This is causing problems with last brain masking and dropping regions with lower signal, such as the OFC. 

There's discussion internally to `tedana` about alternative T2* fitting methods (see https://github.com/ME-ICA/tedana/issues/315) beyond a simple log-linear fit, which might show less "clipping." This is a longer term discussion at this point, though.

At this point, I see three paths forward.

1. Replace those voxels with invalid T2* values with values from some other echo. I have implemented a draft of this [here](https://github.com/ME-ICA/tedana/pull/393), although this feels more like a band-aid than a real solution.

2. Update how fMRIPrep handles its bold_ref images, such that this "clipping" isn't a concern. This one feels the least likely to me, but maybe it's achievable.

3. Remove the `t2s_coreg` option altogether. @handwerkerd expressed that he personally has not seen strong benefits to using the generated T2* image in coregistration, despite it being promoted in the previous ME-ICA pipeline. In this case, we could just continue to use the first echo.

I'd love feedback on which, if any, of these paths make sense to you !

tsalo:
Resolves #1786. This builds off of #1800, but removes the T2*-based coregistration option completely, based on the most recent comments in #1786 and discussion with @emdupre.

- Remove `--t2s-coreg` option from CLI
- Remove `t2s_coreg`-related steps from workflows.
- Remove mention of `--t2s-coreg` option from docs/workflows.rst.

- In `docs/workflows.rst`, I remove a sentence about `--t2s-coreg`.


fmriprep/workflows/bold/t2s.py
```diff
@@ -74,26 +70,23 @@ def init_bold_t2s_wf(echo_times, mem_gb, omp_nthreads,
...
     workflow.connect([
         (inputnode, t2smap_node, [('bold_file', 'in_files')]),
-        (t2smap_node, outputnode, [('optimal_comb', 'bold')]),
-        (t2smap_node, skullstrip_t2smap_wf, [('t2star_map', 'inputnode.in_file')]),
-        (skullstrip_t2smap_wf, outputnode, [
-            ('outputnode.mask_file', 'bold_mask'),
-            ('outputnode.skull_stripped_file', 'bold_ref_brain')]),
+        (t2smap_node, outputnode, [
+            ('optimal_comb', 'bold'),
+            ('t2star_map', 'bold_ref_brain')]),
```
effigies, line 89: This workflow is now `inputnode -> t2smap_node -> outputnode`, so the workflow itself can be replaced in the calling function by `t2smap_node`. Unless it's useful to keep this space open to expand the workflow in the future?

tsalo: I don't know if implementing it would require that `bold_t2s_wf` remain a workflow, but in https://github.com/poldracklab/fmriprep/issues/1010#issuecomment-619209589 I'm proposing a way to incorporate tedana's component classification into fMRIPrep. It might be worth it to keep the workflow around until #1010 is resolved at least.

I can remove `bold_ref_brain` now.

effigies: Sure. Let's keep this change minimal and we can reconsider architecture later. I realized there's also the workflow description that would need to be moved out of this separate module.

tsalo:
@mgxd thank you for catching that. It should be good now.

mgxd:
LGTM, thanks!

oesteban:
@effigies @mgxd merge? I'd like to see how #2108 is affected by this.

effigies:
LGTM. Merge at will, @oesteban.

Shotgunosine:
Sorry for the naive question, when would I expect to see this change in a release?

effigies:
20.1.0, which is coming Real Soon Now.

Seriously, we're trying to wrap up as quickly as possible, but everything's taking longer than we intended.

Out of curiosity, is this a feature that you needed?

Shotgunosine:
If I've followed all of the references right, I think this PR started with an issue I had with masking multi-echo rest sequences:
https://neurostars.org/t/registrations-and-brainmasks/4876/16

I had forgotten about it and ran fmriprepv20.0.5 without selecting a single echo (which was my work around) and found that a lot of the rest sequences have brain excluded from the brain mask around the base of the brain, which was the original issue:

Screen Shot 2020-05-08 at 11 05 52 AM
I'm hoping that the new masking implemented here resolves this so that i can use the optimally combined rest instead of just a single echo.

Shotgunosine:
@tsalo, I know I'm lat to the party here, but did you by any chance check to see if the masking here works on this dataset: https://openneuro.org/datasets/ds002156/versions/1.0.0

tsalo:
@Shotgunosine I haven't checked the masking at all, sorry. I build this PR off of @emdupre's initial work, which I believe dealt with the masking more than what I added, which was just removing the t2s_coreg option from the workflow (and thus didn't touch non-t2s-coreg masking).

_However_, there was a glaring masking bug in the t2smap workflow in tedana <= 0.0.8 (see https://github.com/ME-ICA/tedana/issues/544). I think that that bug could be responsible for at least some of the problems, and we should probably check how the masking works once 0.0.9 is released and I've pinned that version in fMRIPrep (see #2117).

Shotgunosine:
Ok, sounds good. Looking at #1800 it seems like she verified that those changes worked with my data, so I'm optimistic that 20.1 will help when it gets here.

emdupre:
In https://github.com/poldracklab/fmriprep/pull/2109/commits/04cf152402647e0c45eb9f0a61e17daf322fe85a I moved all of the `bold_mask` generation steps outside of the T2* interface, so it should be using the same `bold_mask` that's generated from the SDC workflow. Which, in the case of multi-echo data, will be from the first echo rather than from the optimal combination ! So I don't _think_ that the masking that you're mentioning @tsalo will really impact, here (but happy to discuss more out-of-band !).

Regardless, yes, I ran on yours locally and it looked right @Shotgunosine ! But fMRIPrep has had a lot of updates since then, so I'd be curious to re-run with a patched docker image for the current master.

effigies:
To test with `master`:

```
fmriprep-docker -i poldracklab/fmriprep:unstable ...
```

Shotgunosine:
I tested with unstable yesterday and it seems like it got much better masks on that test subject.
Screen Shot 2020-05-20 at 3 03 48 PM

PR #2117, opened 2020-05-05
[MAINT] Pin tedana version

tsalo:
This PR pins the tedana version to 0.0.9a1, given several breaking changes since 0.0.8.

## Changes proposed in this pull request:
- Closes ME-ICA/tedana#540.
- Pin tedana version to 0.0.9a1. There are a couple of breaking changes in the newest tedana release, including the following:
    - The t2smap workflow outputs nii.gz files instead of nii files.
    - The output files have been renamed as well, to better reflect BIDS conventions.
- Additionally, tedana's added a slightly slower, more accurate method for T2*/S0 estimation using nonlinear estimation. I've changed the default to use this new "curvefit" method instead of the log-linear approximation that was originally used.

- The `init_bold_t2s_wf` workflow `__desc__` has been changed to reflect the improved T2*/S0 estimation method.
- I also changed the title of the workflow description in the docs from `T2* Driven Coregistration` to `T2*-driven echo combination`. This fixes an oversight on my part from #2109.


fmriprep/interfaces/multiecho.py
```diff
@@ -15,8 +15,7 @@
 >>> os.chdir(datadir)
 
 """
-import os
-from nibabel.filename_parser import splitext_addext
+import os.path
```
mgxd, line 18: no need to change this?
```suggestion
import os
```

fmriprep/interfaces/multiecho.py
```diff
@@ -39,13 +38,22 @@ class T2SMapInputSpec(CommandLineInputSpec):
                              mandatory=True,
                              minlen=3,
                              desc='echo times')
+    fittype = traits.Enum('curvefit', 'loglin',
+                          argstr='--fittype %s',
+                          position=3,
+                          mandatory=False,
```
mgxd, line 44: should be the default
```suggestion
```

setup.cfg
```diff
@@ -35,7 +35,7 @@ install_requires =
     pyyaml
     sdcflows @ git+https://github.com/nipreps/sdcflows.git@master
     smriprep @ git+https://github.com/poldracklab/smriprep.git@master
-    tedana >= 0.0.5
+    tedana ~= 0.0.9a1
```
mgxd, line 38: will this pick up 0.0.9? When pinning RCs in the past, we've had to do this manually, i.e.
`>= 0.0.9a1, < 0.0.10`

tsalo: It was able to grab 0.0.9a1 in the tests, but I wasn't thinking about 0.0.9. Should I switch to `>= 0.0.9a1, < 0.0.10` so it grabs 0.0.9 once it's released, or is it okay to pin to 0.0.9a1?

mgxd: i would lean towards allowing grabbing a more stable version, but i'm not familiar with tedana's release cycle so whatever you think would be best

tsalo: That should be fine. I'm happy to change it.

mgxd: we can always increase/drop the maximum version, thanks!

PR #2123, opened 2020-05-08
ENH: Enable filtering for ANY or NONE in ``--bids-filter-file``

bpinsard:

When loading from `bids-filter-file` from json the pybids query values NONE and ANY cannot be used (cannot be serialized).

Here is a quick fix for that problem, will need to be pushed to smriprep as well, and tested.
json `null` is converted to bids.layout.Query.NONE
json string `'__any__'` is converted to bids.layout.Query.ANY (could change the naming if we prefer `'*'` but it usually means anything including empty, maybe `'+'` then ?)

The FAQ section on how to select images when multiple one available.

### DIFF
--- a/fmriprep/cli/parser.py
+++ b/fmriprep/cli/parser.py
@@ -40,10 +40,19 @@ def _build_parser():
         value = str(value)
         return value.lstrip('sub-')
 
+    def _filter_pybids_none_any(dct):
+        import bids
+        for k,v in dct.items():
+            if v is None:
+                dct[k] = bids.layout.Query.NONE
+            elif v == '__any__':
+                dct[k] = bids.layout.Query.ANY
+        return dct
+
     def _bids_filter(value):
         from json import loads
         if value and Path(value).exists():
-            return loads(Path(value).read_text())
+            return loads(Path(value).read_text(), object_hook=_filter_pybids_none_any)
 
     verstr = f'fMRIPrep v{config.environment.version}'
     currentv = Version(config.environment.version)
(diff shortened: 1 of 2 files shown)


fmriprep/cli/parser.py
```diff
@@ -40,10 +40,19 @@ def _drop_sub(value):
         value = str(value)
         return value.lstrip('sub-')
 
+    def _filter_pybids_none_any(dct):
+        import bids
+        for k, v in dct.items():
+            if v is None:
+                dct[k] = bids.layout.Query.NONE
```
effigies, line 47: PyBIDS should already treat `None` like `Query.None`.

fmriprep/cli/parser.py
```diff
@@ -40,10 +40,19 @@ def _drop_sub(value):
...
+            elif v == '__any__':
```
effigies, line 48: What do you think about using a wildcard?

```suggestion
            elif v == '*':
```

fmriprep/cli/parser.py
bpinsard: Weird, I tried and it did not, and now it does. 
Maybe my sleep deprived brain is just making a lot of typos. :/
I will remove it.

fmriprep/cli/parser.py
```diff
@@ -40,10 +40,17 @@ def _drop_sub(value):
         value = str(value)
         return value.lstrip('sub-')
 
+    def _filter_pybids_none_any(dct):
+        import bids
+        for k, v in dct.items():
+            if v == '__any__':
+                dct[k] = bids.layout.Query.ANY
+        return dct
```
effigies, line 48: The switch to `"*"` seems to have gotten lost in the merge.

```suggestion
        return {k: bids.layout.Query.ANY if v == "*" else v
                for k, v in dct.items()}
```

Issue #2049, opened 2020-03-23
Reporting of pre-run FS in fs-subjects-dir in reports


utooley:
Version: 20.0.2
See Neurostars [post](https://neurostars.org/t/detection-of-fs-subjects-dir-in-20-0-2/6342) about this.

I have pre-run Freesurfer and pointed fMRIprep to the directory using `--fs-subjects-dir`, but the report still says “FreeSurfer reconstruction: Run by fMRIPrep”.  I'm almost positive that fMRIPrep is detecting the pre-run surfaces, given the contents of the working dir and the log, but the reporting is wrong, perhaps due to the lines [here](https://github.com/poldracklab/fmriprep/blob/0d0f62e0132d648867a8ecbb72c7670a9e77cf33/fmriprep/interfaces/reports.py#L102-L113)? I'm not sure I understand what's going on there with deciding based on `echo`, though.

`_inputs.pklz` in `working_dir/fsdir_run_20200323_143435_c23f0b8d_4707_4593_9b85_dc891f5fefe5` has the correct Freesurfer subjects directory:

```
Singularity fmriprep-20-0-2.simg:/mnt> nipypecli show _inputs.pklz 
{'derivatives': '/mnt/derivatives/fmriprep_t2',
 'freesurfer_home': '/opt/freesurfer',
 'overwrite_fsaverage': False,
 'spaces': [],
 'subjects_dir': '/mnt/derivatives/freesurfer_t2'}
```

And the logs show something like this (too fast for Freesurfer to run!)
```
[Node] Finished "fmriprep_wf.single_subject_CBPD0000_wf.func_preproc_ses_02_task_rest_run_02_wf.bold_bold_trans_wf.bold_reference_wf.enhance_and_skullstrip_bold_wf.pre_mask_dilate".
200323-16:31:13,460 nipype.workflow INFO:
	 [Node] Setting-up "fmriprep_wf.single_subject_CBPD0000_wf.anat_preproc_wf.surface_recon_wf.autorecon1" in "/tmp/fmriprep_wd/fmriprep_wf/single_subject_CBPD0000_wf/anat_preproc_wf/surface_recon_wf/autorecon1".
200323-16:31:13,651 nipype.interface INFO:
	 recon-all complete : Not running
200323-16:31:13,652 nipype.workflow INFO:
	 [Node] Running "autorecon1" ("smriprep.interfaces.freesurfer.ReconAll"), a CommandLine Interface with command:
echo recon-all: nothing to do
[...]

PR #2124, opened 2020-05-11
FIX: correct previous recon-all reporting

mgxd:
Closes https://github.com/poldracklab/fmriprep/issues/2049

Turns out we weren't passing in the `sub-` identifier, which in turn made ReconAll (rightfully) assume the subject wasn't run.

This also imports the modified `ReconAll` interface used in smriprep.

### DIFF
--- a/fmriprep/interfaces/reports.py
+++ b/fmriprep/interfaces/reports.py
@@ -16,7 +16,7 @@ from nipype.interfaces.base import (
     traits, TraitedSpec, BaseInterfaceInputSpec,
     File, Directory, InputMultiObject, Str, isdefined,
     SimpleInterface)
-from nipype.interfaces import freesurfer as fs
+from smriprep.interfaces.freesurfer import ReconAll
 from niworkflows.utils.bids import BIDS_NAME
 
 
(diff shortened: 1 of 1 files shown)


effigies:
Issue introduced in https://github.com/nipreps/niworkflows/pull/299, released in niworkflows 0.7.0, and thus affecting fMRIPrep 1.3.0+.

Can you target `maint/20.0.x`, and then we can just merge this forward?

mgxd:
untangling some more webs:

```python
In [1]: from smriprep.interfaces import freesurfer as fs                                                                                                                          
In [2]: fs.ReconAll                                                                                                                                                               
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-2-53ff371892c6> in <module>
----> 1 fs.ReconAll

AttributeError: module 'niworkflows.interfaces.freesurfer' has no attribute 'ReconAll'

In [3]: from smriprep.interfaces.freesurfer import ReconAll
In [4]: exit
```

this seems hazardous given `smriprep.interfaces.freesurfer` exists..

effigies:
Ugh. That's a nasty one.

fmriprep/interfaces/reports.py
```diff
@@ -16,7 +16,7 @@
     traits, TraitedSpec, BaseInterfaceInputSpec,
     File, Directory, InputMultiObject, Str, isdefined,
     SimpleInterface)
-from nipype.interfaces import freesurfer as fs
+from smriprep.interfaces.freesurfer import ReconAll
```
effigies, line 19: I'm not actually sure we need that. All that patch allows us to do is to specify a collection of `steps` instead of a `directive`. We are testing the directive `all` below.

mgxd: i added that import to keep the `recon-all` calls consistent - lmk if you feel strongly to change that

effigies: I don't feel strongly.

mgxd:
before:
https://13616-53175327-gh.circle-artifacts.com/0/tmp/ds005/derivatives/fmriprep/sub-01.html

after:
https://13789-53175327-gh.circle-artifacts.com/0/tmp/ds005/derivatives/fmriprep/sub-01.html#Summary

Issue #1697, opened 2019-07-09
No confounds file due to IndexError for tcompcor 


zenkavi:
The following error resulted in not having a confounds file for that run of the subject. Everything else looked fine but I might be missing something.

```
Node Name: fmriprep_wf.single_subject_100068_wf.func_preproc_task_machinegame_run_001_wf.bold_confounds_wf.tcompcor
/oak/stanford/groups/russpold/data/ds000054/0.0.4/derivatives/fmriprep_1.4.0/fmriprep/sub-100068/log/20190704-092934_a41c660f-f278-453c-b0f6-56dc37edc4bb/crash-20190704-200938-zenkavi-tcompcor-34a37fb7-b906-4389-8d3f-3dee188531cc.txt
Working Directory: /tmp/work/fmriprep_wf/single_subject_100068_wf/func_preproc_task_machinegame_run_001_wf/bold_confounds_wf/tcompcor
Inputs: 
components_file: tcompcor.tsv
failure_mode: NaN
header_prefix: t_comp_cor_
high_pass_cutoff: 128.0
ignore_initial_volumes: 1
mask_files: ['/tmp/work/fmriprep_wf/single_subject_100068_wf/func_preproc_task_machinegame_run_001_wf/bold_confounds_wf/tcc_tfm/sub-100068_T1w_corrected_xform_rbrainmask_eroded_trans_boldmsk.nii.gz']
mask_index:
mask_names:
merge_method:
num_components:
percentile_threshold: 0.05
pre_filter: cosine
realigned_file: /tmp/work/fmriprep_wf/single_subject_100068_wf/func_preproc_task_machinegame_run_001_wf/bold_bold_trans_wf/merge/vol0000_xform-00000_merged.nii.gz
regress_poly_degree: 1
repetition_time: 1.0
save_metadata: True
save_pre_filter: True
use_regress_poly:
variance_threshold: 0.5
Traceback (most recent call last):
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/plugins/multiproc.py", line 69, in run_node
    result['result'] = node.run(updatehash=updatehash)
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 472, in run
    result = self._run_interface(execute=True)
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 563, in _run_interface
    return self._run_command(execute)
[...]

PR #2128, opened 2020-05-19
FIX: Ensure correct WM and CSF masks are picked in confounds workflow

oesteban:
Following poldracklab/smriprep#189.


oesteban:
Hey @mgxd, can you double-check this locally before merging (same you did for the sMRIPrep counterpart).

I have run this on the subject that was failing in #1697 (ds054/sub-100068) and it seems okay. The run condition in my settings is ``--fs-no-reconall`` without``--anat-derivatives``.

mgxd:
looks fine to me, thanks!

Issue #2052, opened 2020-03-25
Improve run-to-run reliability of confounds


oesteban:
This is a meta-request for inclusion as part of [the 20.1.0 LTS release](https://github.com/poldracklab/fmriprep/milestone/38), to reference issues that will help with the run-to-run reliability of confounds on surrogate projects (sMRIPrep, SDCFlows, NiWorkflows):

- BOLD reference/brainmask ([niworkflows#485](https://github.com/nipreps/niworkflows/pull/485)), which will require some additional testing (facilitated with [niworkflows#488](https://github.com/nipreps/niworkflows/pull/488))
- Replace FSL FAST with FS' aseg whenever possible ([smriprep#170](https://github.com/poldracklab/smriprep/issues/170))

oesteban:
Both ideas are now implemented. Closing.

PR #2130, opened 2020-05-20
FIX: Revise the reproducibility of CompCor masks

oesteban:
This PR revises the implementation of CompCor masks in an attempt to make it closer to the original proposal and at the same time address some recurring problems of *fMRIPrep*'s tCompCor implementation.

Finally, with the more careful resampling of prior knowledge from the anatomical scan, this refactor should also make the aCompCor components more run-to-run repeatable.

## aCompCor

The massaging of CompCor masks is now done in anatomical space where it is more precise, and a careful resampling to BOLD space follows.

The implementation deviates from Behzadi et al. Their original implementation thresholded the CSF and the WM partial-volume masks at 0.99 (i.e., 99% of the voxel volume is filled with a particular tissue), and then binary eroded that 2 voxels:

> Anatomical data were segmented into gray matter, white matter, and CSF partial volume maps using the FAST algorithm available in the FSL software package (Smith et al., 2004). Tissue partial volume maps were linearly interpolated to the resolution of the functional data series using AFNI (Cox, 1996). In order to form white matter ROIs, the white matter partial volume maps were thresholded at a partial volume fraction of 0.99 and then eroded by two voxels in each direction to further minimize partial voluming with gray matter. CSF voxels were determined by first thresholding the CSF partial volume maps at 0.99 and then applying a threedimensional nearest neighbor criteria to minimize multiple tissue partial voluming. Since CSF regions are typically small compared to white matter regions mask, erosion was not applied.

This particular procedure is not generalizable to BOLD data with different voxel zooms as the mathematical morphology operations will be scaled by those. Also, from reading the excerpt above and the tCompCor description, I (@oesteban)
believe that they always operated slice-wise given the large slice-thickness of their functional data.

Instead, *fMRIPrep*'s implementation deviates from Behzadi's implementation on two aspects:

  * the masks are prepared in high-resolution, anatomical space and then projected into BOLD space; and,
  * instead of using binary erosion, a dilated GM map is generated -- thresholding the corresponding PV map at 0.05 (i.e., pixels containing at least 5% of GM tissue) and then subtracting that map from the CSF, WM and CSF+WM (combined) masks. This should be equivalent to eroding the masks, except that the erosion only happens at direct interfaces with GM.

When the probseg maps provene from FreeSurfer's ``recon-all`` (i.e., they are discrete), binary maps are *transformed* into some sort of partial volume maps by means of a Gaussian smoothing filter with sigma adjusted by the size of the
BOLD data.

## tCompCor

In the case of *tCompCor*, this commit removes the heavy erosion of the brain mask because 1) that wasn't part of the original proposal by Behzadi et al., and 2) the erosion was the potential source of errors from numpy complaining that it can't take from an empty axis of an array.

> Based on these results, we chose a 2% threshold (∼20–30 voxels per slice) as a reasonable empirical threshold that effectively identified voxels with the highest fractional variance of physiological noise.

Although they do the calculation slice-wise, this commit rolls tCompCor back to calculate the 2% threshold on the whole-brain mask.

I would appreciate a deep review from @rciric & @jdkent to ensure my interpretation of Behzadi's paper is correct. I'm also tagging @effigies for a more top-level check of the changes. Further feedback from others is of course very much appreciated.

Resolves: #2129.
References: #2052.

### DIFF
--- a/fmriprep/utils/confounds.py
+++ b/fmriprep/utils/confounds.py
@@ -1,0 +1,138 @@
+"""Utilities for confounds manipulation."""
+
+
+def mask2vf(in_file, zooms=None, out_file=None):
+    """
+    Convert a binary mask on a volume fraction map.
+
+    The algorithm simply applies a Gaussian filter with the kernel size scaled
+    by the zooms given as argument.
+
+    """
+    import numpy as np
+    import nibabel as nb
+    from scipy.ndimage import gaussian_filter
+
+    img = nb.load(in_file)
+    imgzooms = np.array(img.header.get_zooms()[:3], dtype=float)
+    if zooms is None:
+        zooms = imgzooms
+
+    zooms = np.array(zooms, dtype=float)
+    sigma = 0.5 * (zooms / imgzooms)
+    if len(sigma) > 1:
+        sigma = tuple(sigma)
+
+    data = gaussian_filter(img.get_fdata(dtype=np.float32), sigma=sigma)
+
+    max_data = np.percentile(data[data > 0], 99)
+    data = np.clip(data / max_data, a_min=0, a_max=1)
+
+    if out_file is None:
+        return data
+
+    hdr = img.header.copy()
+    hdr.set_data_dtype(np.float32)
+    nb.Nifti1Image(data.astype(np.float32), img.affine, hdr).to_filename(out_file)
+    return out_file
+
+
+def acompcor_masks(in_files, is_aseg=False, zooms=None):
+    """
+    Generate aCompCor masks.
+
+    This function selects the CSF partial volume map from the input,
+    and generates the WM and combined CSF+WM masks for aCompCor.
+
+    The implementation deviates from Behzadi et al.
+    Their original implementation thresholded the CSF and the WM partial-volume
+    masks at 0.99 (i.e., 99% of the voxel volume is filled with a particular tissue),
+    and then binary eroded that 2 voxels:
+
+    > Anatomical data were segmented into gray matter, white matter,
+    > and CSF partial volume maps using the FAST algorithm available
+    > in the FSL software package (Smith et al., 2004). Tissue partial
+    > volume maps were linearly interpolated to the resolution of the
+    > functional data series using AFNI (Cox, 1996). In order to form
+    > white matter ROIs, the white matter partial volume maps were
+    > thresholded at a partial volume fraction of 0.99 and then eroded by
+    > two voxels in each direction to further minimize partial voluming
+    > with gray matter. CSF voxels were determined by first thresholding
+    > the CSF partial volume maps at 0.99 and then applying a threedimensional
+    > nearest neighbor criteria to minimize multiple tissue
+    > partial voluming. Since CSF regions are typically small compared
+    > to white matter regions mask, erosion was not applied.
+
+    This particular procedure is not generalizable to BOLD data with different voxel zooms
+    as the mathematical morphology operations will be scaled by those.
+    Also, from reading the excerpt above and the tCompCor description, I (@oesteban)
+    believe that they always operated slice-wise given the large slice-thickness of
+    their functional data.
+
+    Instead, *fMRIPrep*'s implementation deviates from Behzadi's implementation on two
+    aspects:
+
+      * the masks are prepared in high-resolution, anatomical space and then
+        projected into BOLD space; and,
+      * instead of using binary erosion, a dilated GM map is generated -- thresholding
+        the corresponding PV map at 0.05 (i.e., pixels containing at least 5% of GM tissue)
+        and then subtracting that map from the CSF, WM and CSF+WM (combined) masks.
+        This should be equivalent to eroding the masks, except that the erosion
+        only happens at direct interfaces with GM.
+
+    When the probseg maps provene from FreeSurfer's ``recon-all`` (i.e., they are
+    discrete), binary maps are *transformed* into some sort of partial volume maps
+    by means of a Gaussian smoothing filter with sigma adjusted by the size of the
+    BOLD data.
+
+    """
+    from pathlib import Path
+    import numpy as np
+    import nibabel as nb
+    from scipy.ndimage import binary_dilation
+    from skimage.morphology import ball
+
+    csf_file = in_files[2]  # BIDS labeling (CSF=2; last of list)
+    # Load PV maps (fast) or segments (recon-all)
+    gm_vf = nb.load(in_files[0])
+    wm_vf = nb.load(in_files[1])
+    csf_vf = nb.load(csf_file)
+
+    # Prepare target zooms
+    imgzooms = np.array(gm_vf.header.get_zooms()[:3], dtype=float)
+    if zooms is None:
+        zooms = imgzooms
+    zooms = np.array(zooms, dtype=float)
+
+    if not is_aseg:
+        gm_data = gm_vf.get_fdata() > 0.05
+        wm_data = wm_vf.get_fdata()
+        csf_data = csf_vf.get_fdata()
+    else:
+        csf_file = mask2vf(
+            csf_file,
+            zooms=zooms,
+            out_file=str(Path("acompcor_csf.nii.gz").absolute()),
+        )
+        csf_data = nb.load(csf_file).get_fdata()
+        wm_data = mask2vf(in_files[1], zooms=zooms)
+
+        # We do not have partial volume maps (recon-all route)
+        gm_data = np.asanyarray(gm_vf.dataobj, np.uint8) > 0
+
+    # Dilate the GM mask
+    gm_data = binary_dilation(gm_data, structure=ball(3))
+
+    # Output filenames
+    wm_file = str(Path("acompcor_wm.nii.gz").absolute())
+    combined_file = str(Path("acompcor_wmcsf.nii.gz").absolute())
+
+    # Prepare WM mask
+    wm_data[gm_data] = 0  # Make sure voxel does not contain GM
+    nb.Nifti1Image(wm_data, gm_vf.affine, gm_vf.header).to_filename(wm_file)
+
+    # Prepare combined CSF+WM mask
+    comb_data = csf_data + wm_data
+    comb_data[gm_data] = 0  # Make sure voxel does not contain GM
+    nb.Nifti1Image(comb_data, gm_vf.affine, gm_vf.header).to_filename(combined_file)
+    return [csf_file, wm_file, combined_file]
--- a/fmriprep/utils/confounds.py
+++ b/fmriprep/utils/confounds.py
@@ -20,8 +20,6 @@ def mask2vf(in_file, zooms=None, out_file=None):
 
     zooms = np.array(zooms, dtype=float)
     sigma = 0.5 * (zooms / imgzooms)
-    if len(sigma) > 1:
-        sigma = tuple(sigma)
 
     data = gaussian_filter(img.get_fdata(dtype=np.float32), sigma=sigma)
 
--- a/fmriprep/workflows/bold/confounds.py
+++ b/fmriprep/workflows/bold/confounds.py
@@ -141,8 +141,8 @@ def init_bold_confs_wf(
     from ...interfaces.confounds import aCompCorMasks
 
     gm_desc = (
-        'a probabilistic segmentation from the *aseg* segmentation' if freesurfer
-        else 'the corresponding partial volume map'
+        "dilating a GM mask extracted from the FreeSurfer's *aseg* segmentation" if freesurfer
+        else "thresholding the corresponding partial volume map at 0.05"
     )
 
     workflow = Workflow(name=name)
@@ -168,11 +168,10 @@ voxels within the brain mask.
 For aCompCor, three probabilistic masks (CSF, WM and combined CSF+WM)
 are generated in anatomical space.
 The implementation differs from that of Behzadi et al. in that instead
-of eroding the masks by 2 pixels on BOLD space, a liberal mask of the
-GM is obtained by thresholding {gm_desc} at 0.05.
-Then this mask is substracted from the WM and the combined CSF+WM masks
-to ensure components are not extracted from voxels containing a minimal
-fraction of GM.
+of eroding the masks by 2 pixels on BOLD space, the aCompCor masks are
+subtracted a mask of pixels that likely contain a volume fraction of GM.
+This mask is obtained by {gm_desc}, and it ensures components are not extracted
+from voxels containing a minimal fraction of GM.
 Finally, these masks are resampled into BOLD space and binarized by
 thresholding at 0.99 (as in the original implementation).
 Components are also calculated separately within the WM and CSF masks.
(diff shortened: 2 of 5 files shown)


oesteban:
Ready for you @rciric

oesteban:
The horrible contours on our subsampled dataset will be addressed by nipreps/niworkflows#544.

fmriprep/utils/confounds.py
```diff
@@ -0,0 +1,138 @@
...
+
+    img = nb.load(in_file)
+    imgzooms = np.array(img.header.get_zooms()[:3], dtype=float)
+    if zooms is None:
+        zooms = imgzooms
+
+    zooms = np.array(zooms, dtype=float)
+    sigma = 0.5 * (zooms / imgzooms)
+    if len(sigma) > 1:
+        sigma = tuple(sigma)
```
mgxd, line 24: is this check necessary? (will `sigma` ever be `<= 1`?)

rciric:
Hello,

Apologies for the late response. I agree that morphological erosion using an isotropic fixed-voxel-size structuring element would generalize poorly to new resolutions. Reading the Behzadi paper again, I’m somewhat confused about the decision to warp the data into functional space before applying the erosion, and the motivation for doing so is unclear to me. I completely agree with the decision to perform the erosion in anatomical space, and the adaptive kernel size is a great idea.

Eroding using a low-threshold GM mask intuitively sounds like an improvement over conventional methods, but I do think that we should ideally examine the correlations of WM and CSF time series with whole-brain global signal when using the new approach to determine whether there are any major changes from the conventional approach. In particular, we would expect that limiting partial volume effects keeps the tissue-specific correlation with global signal reasonably low. This is of particular concern for those researchers who believe strongly that global mean signal should not be discarded (although I do not find myself in that camp).

My recommendation for tCompCor is similar: the proposed approach sounds very reasonable, although I think it wouldn't hurt to more clearly characterise the noise ROI and the component time series correlations at some point. tCompCor appears to be used with somewhat less frequency than aCompCor.

To verify, are the new aCompCor masks also going to define the regions over which the (global) mean WM and CSF signals are extracted?

The literature is not entirely uniform in use of anatomical CompCor approaches. In addition to the original method from Behzadi and colleagues, a similar [aCompCor methodology proposed by Muschelli and colleagues](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4043948/) is used with considerable frequency. In particular, the aCompCor evaluated by our group (and by Parkes et al. to my knowledge) more closely matched the Muschelli approach. By computing separate decompositions for each compartment mask and for the union mask, fMRIPrep already supports the most significant difference between the approaches. It's not, however, immediately clear to me whether Muschelli and colleagues performed their mask erosions in native anatomical space or in a standard space, since the maps they used were provided by a combined segmentation-normalisation procedure from SPM:

> Subject-specific tissue probability maps generated during SPM’s unified segmentation-normalization process were restricted using a 99% probability threshold. To further reduce the risk of capturing signals of interest from adjacent grey matter voxels, the resulting WM mask was eroded using MATLAB’s imerode function, and the CSF mask was constrained to areas within the ALVIN mask of the ventricles (Kempton et al., 2011).

As a potential aid for evaluating the masks we obtain, I'm including here for reference the subject-wise voxel inclusion frequency map for WM and CSF masks in the Muschelli study.
image
> (from Muschelli et al., 2014)

Also potentially of note, [Power and colleagues](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3849338/) performed their mask erosion using FreeSurfer. I'm not sure whether this erosion was performed in volumetric or surface space, and I only have a passing familiarity with FreeSurfer. In any case, there is substantial precedent for eroding masks in subject anatomical space. This approach was most similar to what our group did in the benchmarking paper, although we performed our erosions morphologically in volumetric space.

> Each subject’s MP-RAGE underwent segmentation by FreeSurfer (version 5.0). Segments corresponding to the lateral ventricles, the white matter, the gray matter, and all within-brain voxels were extracted. 5 versions of the white matter and ventricular masks were formed: uneroded, and erosions 1–4 times. All masks were resliced to 3 mm isotropic dimensions to match the BOLD data. Maximal erosion (4×) masks were preferred; if maximal erosion yielded empty masks, lesser erosions were progressively considered until a mask with qualifying voxels was found. This relaxation occurred infrequently for white matter masks. Erosions of 1 were often used for CSF masks. These masks were used to extract nuisance signals, to calculate DV and SD within the masks, and to define the voxel timeseries displayed in Part I.

The description in [a later work](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5296400/) suggests to me that the erosion was likely also in volumetric space, although I again cannot say with confidence. This study also includes an example of eroded masks (corresponding to the deepest WM colour below, although I seem to recall that many of the example masks I saw from the WashU group were restricted to deeper voxels still.)

image
> (from Power et al., 2017)

Overall, I agree with the premises of this revision, although I believe that further practical characterisation of fMRIPrep's CompCor procedure, potentially in the form of empirical correlations and voxel inclusion maps, would be welcome at some future stage.

effigies:
I don't think the FreeSurfer method described could apply to the surface, as there is no representation of white matter or CSF on the surface. It sounds like they used the aseg segmentation.

oesteban:
Hi @rciric, happy to do this:

> we should ideally examine the correlations of WM and CSF time series with whole-brain global signal when using the new approach to determine whether there are any major changes from the conventional approach. In particular, we would expect that limiting partial volume effects keeps the tissue-specific correlation with global signal reasonably low.

If I understand correctly and to say that this PR improves the situation, we should calculate the correlation of global signal with the time-series extracted from the different CompCor masks. Then, if these correlations are not significantly larger than those before this PR, we can say these new masks at least don't make worse masks for compcor. Is that right?

oesteban:
> It sounds like they used the aseg segmentation.

Yes, I think this is correct.AAAW

fmriprep/workflows/bold/confounds.py
```diff
@@ -249,8 +241,11 @@ def init_bold_confs_wf(
         acompcor.inputs.repetition_time = metadata['RepetitionTime']
 
     # Global and segment regressors
+    gs_select = pe.Node(niu.Select(index=[0, 1]), name="gs_select",
+                        run_without_submitting=True)
     signals_class_labels = ["csf", "white_matter", "global_signal"]
-    mrg_lbl = pe.Node(niu.Merge(3), name='merge_rois', run_without_submitting=True)
+    mrg_lbl = pe.Node(niu.Merge(2, ravel_inputs=True), name='merge_rois',
+                      run_without_submitting=True)
```
effigies, line 248:
```suggestion
    merge_rois = pe.Node(niu.Merge(2, ravel_inputs=True), name='merge_rois',
                         run_without_submitting=True)
```

fmriprep/workflows/bold/confounds.py
```diff
@@ -327,66 +323,37 @@ def init_bold_confs_wf(
...
         # aCompCor
-        (inputnode, acompcor, [('bold', 'realigned_file')]),
-        (inputnode, acompcor, [('skip_vols', 'ignore_initial_volumes')]),
-        (acc_tfm, acc_msk, [('output_image', 'roi_file')]),
-        (acc_msk, mrg_lbl_cc, [('out', 'in1')]),
-        (csf_msk, mrg_lbl_cc, [('out', 'in2')]),
-        (wm_msk, mrg_lbl_cc, [('out', 'in3')]),
-        (mrg_lbl_cc, acompcor, [('out', 'mask_files')]),
+        (inputnode, acompcor, [("bold", "realigned_file")]),
+        (inputnode, acompcor, [("skip_vols", "ignore_initial_volumes")]),
```
effigies, line 337:
```suggestion
        (inputnode, acompcor, [("bold", "realigned_file"),
                               ("skip_vols", "ignore_initial_volumes")]),
```

fmriprep/workflows/bold/confounds.py
```diff
@@ -220,21 +204,29 @@ def init_bold_confs_wf(
...
+    acc_msk_tfm = pe.MapNode(ApplyTransforms(
+        interpolation='Gaussian', float=False), iterfield=["input_image"],
+        name='acc_msk_tfm', mem_gb=0.1)
+    acc_msk_brain = pe.MapNode(ApplyMask(), name="acc_msk_brain",
+                               iterfield=["in_file"])
+    acc_msk_bin = pe.MapNode(Binarize(thresh_low=0.99), name='acc_msk_bin',
+                             iterfield=["in_file"])
     acompcor = pe.Node(
         ACompCor(components_file='acompcor.tsv', header_prefix='a_comp_cor_', pre_filter='cosine',
                  save_pre_filter=True, save_metadata=True, mask_names=['combined', 'CSF', 'WM'],
```
effigies, line 220: `aCompCorMasks` returns CSF, WM, Combined. Do we need to reorder like:

```suggestion
                 save_pre_filter=True, save_metadata=True, mask_names=['CSF', 'WM', 'combined'],
```

fmriprep/workflows/bold/confounds.py
```diff
@@ -327,66 +323,37 @@ def init_bold_confs_wf(
...
         (inputnode, signals, [('bold', 'in_file')]),
-        (csf_tfm, csf_msk, [('output_image', 'roi_file')]),
-        (csf_msk, mrg_lbl, [('out', 'in1')]),
-        (wm_tfm, wm_msk, [('output_image', 'roi_file')]),
-        (wm_msk, mrg_lbl, [('out', 'in2')]),
-        (inputnode, mrg_lbl, [('bold_mask', 'in3')]),
+        (acc_msk_bin, gs_select, [('out_file', 'inlist')]),
+        (gs_select, mrg_lbl, [('out', 'in1')]),
+        (inputnode, mrg_lbl, [('bold_mask', 'in2')]),
         (mrg_lbl, signals, [('out', 'label_files')]),
```
effigies, line 357:
```suggestion
        (gs_select, merge_rois, [('out', 'in1')]),
        (inputnode, merge_rois, [('bold_mask', 'in2')]),
        (merge_rois, signals, [('out', 'label_files')]),
```

fmriprep/utils/confounds.py
```diff
@@ -0,0 +1,138 @@
...
+    if out_file is None:
+        return data
+
+    hdr = img.header.copy()
+    hdr.set_data_dtype(np.float32)
+    nb.Nifti1Image(data.astype(np.float32), img.affine, hdr).to_filename(out_file)
+    return out_file
+
+
+def acompcor_masks(in_files, is_aseg=False, zooms=None):
```
effigies, line 40: I don't really like sticking major pieces of machinery into a `utils` module and having the interfaces be thin wrappers on a hidden function. I think at one point we were doing something like:

```Python
class AlgorithmInterface:
        self._results = _algorithm(**self.inputs.get())
        return runtime

def _algorithm(arg1, arg2):
    ...
    return {"out1": ...}
```

And the justification was that it would be easier to test a function than an interface. We're not testing this, and now it's being hidden away into an unassuming corner of the code.

I don't want to hold up the PR, but I would suggest we think through our organization and apply it consistently.

oesteban: > I don't want to hold up the PR, but I would suggest we think through our organization and apply it consistently.

Yup, this would be good. Let's make sure we find a time after the LTS. At first my only comment would be that sphinx does not take private functions by default (that's why this is public). That said, if when we meet we decide to have everything together, fine by me - this was kind of an arbitrary decision.

fmriprep/workflows/bold/confounds.py
oesteban: Good catch 👍

oesteban (commented):
Thanks for the review :)

effigies:
The masks in reports look pretty bad.

effigies:
Pushed your previous commit to `upstream/compcormasks` to generate reports. The good news is: my changes didn't cause the bad-looking reports.

oesteban:
> The masks in reports look pretty bad.

You mean for ds054 and ds210 (i.e., `--fs-no-reconall`), right?

I have mostly tested with recon-all - that's why I haven't lent much attention to ds054 and ds210. That said, I'm surprised that the combined masks look so bad.

To recapitulate:

Before this PR we were probably representing the combined mask (as stated in the caption of the reportlet), and it never ever contained CSF. I think our CSF mask was so exclusive that very few voxels were in it (and hence, the empty take errors from numpy). - I say this just in case you think ds005's reports look awful too.

Now, the masks with recon-all look pretty good. I've checked all of them after running ds114, and @shotgunosine has also generated a good bunch (which looked good to me and @rciric will soon assess after his quals).

Then, we have the `--fs-no-reconall` path. I know I have run ds054 without reconall at full resolution and that gave me sound results. This might be a result of the very big pixels of the WM mask with the heavily subsampled dataset. Let me run a bunch of dataset. Perhaps, I should start with ds114 (and `--fs-no-reconall`), and see how different the masks are w.r.t. the default.

effigies:
You're right, I didn't look at ds005 because it takes so much longer to complete.

Do the voxels near the edge of the brain mask occur in higher resolution images? Are they problematic? I assume they're classified as CSF, but I would worry you get GM signal bleed-over (possibly resulting in the high variance for tCompCor).

oesteban:
> Do the voxels near the edge of the brain mask occur in higher resolution images? Are they problematic? I assume they're classified as CSF, but I would worry you get GM signal bleed-over (possibly resulting in the high variance for tCompCor).

Are you referring to pink (aCompCor) or blue (tCompCor)? I guess they are not pink because the new implementation is quite careful - if they are selected, the certainty of them being CSF is very high.

Regarding tCompCor, this is expected (which does not mean good). I particularly asked @rciric about this, and he didn't seem too concerned. Please note that the original Behzadi paper DID NOT erode the brain mask (as we were doing), and also, their top percentile was calculated slice-wise (with veeery thick slices). My impression is that the original implementation of tCompCor wasn't great because of this.

We may want to come back to the previous implementation, but I definitely see the new tCompCor masks picking up voxels in the brain stem and external parts of the commissures - which is a good thing, I believe.

effigies:
> Are you referring to pink (aCompCor) or blue (tCompCor)?

They show up in both.

Fair enough.

oesteban:
Okay, running ds054 and ds210 at their original resolution with `--fs-no-reconall`. Will report back when done.

oesteban:
> Okay, running ds054 and ds210 at their original resolution with `--fs-no-reconall`. Will report back when done.

Masks look okay (better with FS, but that's the drill). I'm building a new singularity image (which required cutting an RC of sMRIPrep) to test these masks including the fix from Chris to make the reportlet show the combined mask instead of the wm mask.

As soon as I have both datasets run with the latest image I'll ask @rciric for his trained eye.

Meanwhile, please have a look into #2245 - I think it would be worth pushing into 20.1.x to be able to do the correlation w/global-signal test.

effigies:
@oesteban @rciric What's the status of this? I think we're very close to a release, so let me know if I can help on anything.

oesteban:
My plan is to quickly make a PR to add the CompCor time series to the bar
plot of correlations with the GS (and remove the components from there).
That if @rciric thinks it's a good idea.

Worst case scenario, the plots will become more readable.

effigies:
Is this PR ready to merge, then?

oesteban:
Yes, I think that's safe

Issue #2131, opened 2020-05-21
Use upstreamed nitransforms from niworkflows


oesteban:
After merging nipreps/niworkflows#525, only one codebase should be maintained.

PR #2132, opened 2020-05-22
ENH: Finish the upstreaming of NiTransforms interfaces to NiWorkflows

oesteban:
Resolves: #2131


oesteban:
Actually, I missed out on one import.

Issue #2137, opened 2020-05-23
Reuse of same work directory between subjects with 20.1.0


ariekahn:
I'm testing fmriprep 20.1.0 via singularity (`20.1.0rc2+177.ge3d3bc51`).

I was running my subjects in parallel jobs on an SGE cluster, and noticed that they started failing quickly with missing outputs, and appeared to be searching for the wrong subject IDs.

For instance, once instance that was started with the following argments:
```
singularity run -B $SBIA_TMPDIR:/tmp --cleanenv $SIMG \
    $PROJECT_DIR/data/ \
    $PROJECT_DIR/derived/ \
    participant \
    --nprocs 9 \
    --mem_mb 65536 \
    --fs-license-file $PROJECT_DIR/extra/license.txt \
    --fs-subjects-dir $PROJECT_DIR/derived/freesurfer \
    --participant-label GLS006 \
    -w $PROJECT_DIR/work \
    --skull-strip-fixed-seed \
    --bold2t1w-dof 9 \
    --dummy-scans 10 \
    --output-spaces MNI152NLin2009cAsym:res-native MNI152NLin6Asym:res-2
```

produced the following run:
```
    Running fMRIPREP version 20.1.0rc2+177.ge3d3bc51:
      * BIDS dataset path: /cbica/projects/GraphLearning/project/data.
      * Participant list: ['GLS020'].
      * Run identifier: 20200522-234318_57e47270-f002-4929-8b35-d51e400b07b6.
      * Output spaces: MNI152NLin2009cAsym:res-native MNI152NLin6Asym:res-2.
```

which in turn led to the multiple runs clobbering each other.

It looks like this was due to how the current build writes `.fmriprep.toml` to the root of the work directory, which gets overwritten if I have multiple instances running in parallel.

 With previous versions, fmriprep would create subdirectories for `fsdir_run_<runid>` and `single_subject_<subjectid>` for each instance, but it looks like the root of the work directory is being used now for both `bids.db` and `.fmriprep.toml`.

I'm not sure what correct behavior should be here, which is why I'm opening this issue. I can't find any information in the documentation, and the one parallel example I see for SLURM appears to re-use the same work directory. So I'm not sure if this is incorrect usage on my part, or the documentation needs to be updated.

mgxd:
@ariekahn this certainly sounds like a bug likely introduced during the migration to the config file, I'll take a look - thanks for piloting before the release!

oesteban:
[...]

PR #2138, opened 2020-05-24
FIX: Temporary config file in work directory gets clobbered in parallel jobs

mgxd:
A test fix for #2137. Also ran `black` on `fmriprep/cli/run.py`.

 @ariekahn if you have the time and would like to pilot this, please do 😄

### DIFF
--- a/fmriprep/cli/run.py
+++ b/fmriprep/cli/run.py
@@ -26,7 +26,15 @@ def main():
     # CRITICAL Save the config to a file. This is necessary because the execution graph
     # is built as a separate process to keep the memory footprint low. The most
     # straightforward way to communicate with the child process is via the filesystem.
-    config_file = config.execution.work_dir / ".fmriprep.toml"
+    config_file = config.execution.work_dir / ".".join(
+        (
+            ".fmriprep",
+            config.execution.participant_label[0]
+            if config.execution.participant_label
+            else "",
+            "toml",
+        )
+    )
     config.to_filename(config_file)
 
     # CRITICAL Call build_workflow(config_file, retval) in a subprocess.
@@ -120,9 +128,14 @@ def main():
         # Bother users with the boilerplate only iff the workflow went okay.
         boiler_file = config.execution.output_dir / "fmriprep" / "logs" / "CITATION.md"
         if boiler_file.exists():
-            if config.environment.exec_env in ("singularity", "docker", "fmriprep-docker"):
+            if config.environment.exec_env in (
+                "singularity",
+                "docker",
+                "fmriprep-docker",
+            ):
                 boiler_file = Path("<OUTPUT_PATH>") / boiler_file.relative_to(
-                    config.execution.output_dir)
+                    config.execution.output_dir
+                )
             config.loggers.workflow.log(
                 25,
                 "Works derived from this fMRIPrep execution should include the "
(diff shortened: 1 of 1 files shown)


ariekahn:
Definitely, giving it a test right now.

effigies:
Why not just tag with the run UUID?

fmriprep/cli/run.py
```diff
@@ -26,7 +26,15 @@ def main():
...
-    config_file = config.execution.work_dir / ".fmriprep.toml"
+    config_file = config.execution.work_dir / ".".join(
+        (
+            ".fmriprep",
+            config.execution.participant_label[0]
+            if config.execution.participant_label
+            else "",
+            "toml",
+        )
+    )
```
oesteban, line 37: I agree with @effigies this is prone to leave use-cases uncovered. E.g., running several runs on the same participant but different ``--bids-filters``. A solid way of making sure of this is using a temporary directory.

```suggestion
    config_file = Path(mkdtemp()) / ".fmriprep.toml"
```
(adding `from tempfile import mkdtemp` somewhere around lines 10-13).
I think at some point in the end, we copy the config file to a folder with the uuid under derivatives (like crashfiles). We should copy this temporal file and delete the folder then.

oesteban: I can't seem to find it (where we copy the config to the derivatives folder).

Anyways, it could be nice to store it in `config.execution.output_dir / "fmriprep" / "logs" / f"config-{config.execution.run_uuid}.toml"` or similar.

You could set this path at the very beginning, although there could be some implications related to #2107 and the reutilization of uuids. Other than that, since this is not I/O bound, writing to a temp folder will not buy us much and directly using a filename on the output folder (or work directory, either way) tagged with the uuid might be more effective.

mgxd: Sure, we can generate a run specific config file. I designed this with #2107 in mind, but I'm fine with relying on just the config in the output directory.

As for the output config path, I think I prefer the current implementation better (putting all the config, crashes, etc. in the same directory). We may however want to put the most recent config at:

`out / "fmriprep" / subject / "log" / "latest-fmriprep.toml"`
for ease of viewing

ariekahn:
@mgxd That eliminated the problem I was running into, just fyi

fmriprep/cli/run.py
oesteban: > Sure, we can generate a run specific config file

I think this is a requirement, not a possibility here. Then, #2107 should work off of this bugfix.



> As for the output config path, I think I prefer the current implementation better (putting all the config, crashes, etc. in the same directory)

I'm fine with this. I just checked and we do write out a file `out / "fmriprep" / subject / "log" / uuid / "fmriprep.toml"`. So maybe it's just a matter of working only with this file.



> We may however want to put the most recent config at:
> 
> `out / "fmriprep" / subject / "log" / "latest-fmriprep.toml"`
> for ease of viewing

uuids have the date and time in the beginning. It's just a matter of picking the latest. From the user's perspective, this might be easier. However, we are printing out the config in the log, and that's normally what gets reported.

oesteban: Okay, we are indeed writing out the config to the derivatives folder:

https://github.com/poldracklab/fmriprep/blob/e3d3bc51dbf03215e3e4d2746d8aaacdd9afb84d/fmriprep/workflows/base.py#L79-L82

But now I'm realizing this is not a great place. This folder is generated and written at the subject level (regardless ``--bids-filters``, btw), and the problematic file producing the clobbering works at the run level. At the run level, one may want to process several participants, and thus how you decide the participant under which you will write this config?. I'm fine having written the config file underneath the processing of each subject under the subject's log folder (as we are currently doing), but a long term solution for this involves either writing to a temporary folder or tagging the name with the uuid. Then #2107 would need to adapt accordingly (I guess this could be done extracting the most recent config from the subjects involved in the processing).

Issue #2145, opened 2020-05-27
LTA concatenation with NiTransforms producing wrong transforms


oesteban:
A problem using `ConcatenateXFMs` from the new `niworkflows.interfaces.nitransforms` module resulted in surface-sampled results looking like:

![Screen Shot 2020-05-26 at 4 10 41 PM](https://user-images.githubusercontent.com/598470/83048765-f6d6c000-9ffe-11ea-8edf-edd9111f0763.png)

because of the sampler receives the wrong transformation.

The problem is easily noticeable on the carpetplot:
![image (1)](https://user-images.githubusercontent.com/598470/83048934-3c938880-9fff-11ea-9893-41845a9e209a.png)

Thanks @mgxd for catching this and making all the necessary diagnosis.

Issue #2121, opened 2017-11-04
Use lta_convert for simplicity


oesteban:
Comes from https://github.com/poldracklab/fmriprep/issues/745#issuecomment-338356352

EDIT:
This issue is a reminder that this code:

https://github.com/oesteban/fmriprep/blob/eb3c838f1bed031c9f02d54b3db3bdcae2dad8dd/fmriprep/workflows/bold/registration.py#L455-L463

might be hacky and hard to maintain. A solution using FS' lta_convert or nitransforms would be more sustainable

oesteban:
3903beed9aecfe805eea5aa31cdac102e8ffbb97

Issue #2118, opened 2020-05-06
CIFTI space output RPI orientation


saigerutherford:
The HCP task analysis scripts require the subcortical volumetric data to have RPI orientation, but the fMRIPrep (version 20.0.4) CIFTI output has LPI orientation for the subcortical volumetric data which creates an error when merging the task contrast maps back into CIFTI space (the script splits the CIFTI file into surface & volume, runs task analysis, then merges back together). If anyone would like to run HCP task analysis on this data they need to flip the subcortical data to be RPI orientation in order for the HCP task scripts to work.

mgxd:
@saigerutherford the CIFTI subcortical orientation has been changed to be compatible with HCP analysis in the upcoming 20.1.0 LTS release. The latest release candidate `20.1.0rc2` includes this change, if you would like to pilot one of your participants.

mgxd:
@saigerutherford now that 20.1.0 has been released, could you try running one participant and see if the orientation is fixed?

oesteban:
@saigerutherford could you reopen this issue if you see we really didn't fix it? Also confirming this worked for you would be great, but I'll close this issue for now.

Issue #2041, opened 2020-03-18
Use nitransforms for LTA-ITK conversions


oesteban:
Replacing this: https://github.com/poldracklab/fmriprep/blob/1d93d9e5599d08347ba2b85a4d4590a4c9273814/fmriprep/workflows/bold/registration.py#L456-L463

(will require poldracklab/nitransforms#64)

oesteban:
3903beed9aecfe805eea5aa31cdac102e8ffbb97

PR #2146, opened 2020-05-27
FIX: Generate proper LTA transform prior BOLD sampling on surfaces

oesteban:
This is a temporary patch before we go all the way in with NiTransforms
in the sampling of BOLD on surfaces.

The anatomical _fast-track_ required to expose the fsnative-to-T1w
transform in the derivatives folder (which we were already doing in ITK
format).

When fMRIPrep ran without the fast-track, then the LTA transform would
be directly passed in without conversions. The fast-track PR forced the
implementation to use the ITK version.

This, in conjunction with the little trick to stick the BOLD shape and
zooms into the LTA (i.e., using ``lta_concatenate`` with an identity
transform with those features, shape and zooms, as moving) resulted in
an overly complex workflow that I partially implemented with
NiTransforms.

This PR gets rid of the concatenation with identity trick, using
NiTransforms to generate a transform equivalent to the concatenated LTA
we used to generate before the fast-track was introduced.

Resolves: #2145
Related: #2118, #2041, #2121.

### DIFF
--- a/fmriprep/workflows/bold/resampling.py
+++ b/fmriprep/workflows/bold/resampling.py
@@ -89,7 +89,6 @@ The BOLD time-series were resampled onto the following surfaces
 
     get_fsnative = pe.Node(FreeSurferSource(), name='get_fsnative',
                            run_without_submitting=True)
-    tonii = pe.Node(fs.MRIConvert(out_type='niigz'), name='tonii')
 
     def select_target(subject_id, space):
         """Get the target subject ID, given a source subject ID and a target space."""
@@ -129,8 +128,7 @@ The BOLD time-series were resampled onto the following surfaces
         (inputnode, rename_src, [('source_file', 'in_file')]),
         (inputnode, itk2lta, [('source_file', 'src_file'),
                               ('t1w2fsnative_xfm', 'in_file')]),
-        (get_fsnative, tonii, [('T1', 'in_file')]),
-        (tonii, itk2lta, [('out_file', 'dst_file')]),
+        (get_fsnative, itk2lta, [('T1', 'dst_file')]),
         (inputnode, sampler, [('subjects_dir', 'subjects_dir'),
                               ('subject_id', 'subject_id')]),
         (itersource, targets, [('target', 'space')]),
--- a/fmriprep/workflows/bold/resampling.py
+++ b/fmriprep/workflows/bold/resampling.py
@@ -69,13 +69,8 @@ def init_bold_surf_wf(
         BOLD series, resampled to FreeSurfer surfaces
 
     """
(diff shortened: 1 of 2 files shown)


fmriprep/workflows/bold/resampling.py
```diff
@@ -86,12 +81,16 @@ def init_bold_surf_wf(
...
-        niu.IdentityInterface(fields=['source_file', 't1w_preproc', 'subject_id', 'subjects_dir',
+        niu.IdentityInterface(fields=['source_file', 'subject_id', 'subjects_dir',
                                       't1w2fsnative_xfm']),
         name='inputnode')
     itersource = pe.Node(niu.IdentityInterface(fields=['target']), name='itersource')
     itersource.iterables = [('target', surface_spaces)]
 
+    get_fsnative = pe.Node(FreeSurferSource(), name='get_fsnative',
+                           run_without_submitting=True)
+    tonii = pe.Node(fs.MRIConvert(out_type='niigz'), name='tonii')
```
effigies, line 92: If all you need from the T1.mgz is the shape/affine, then you don't need to convert to NIfTI. Nibabel supports MGH files just fine.

fmriprep/workflows/bold/resampling.py
```diff
@@ -127,20 +123,19 @@ def select_target(subject_id, space):
...
         (inputnode, rename_src, [('source_file', 'in_file')]),
-        (inputnode, resampling_xfm, [('source_file', 'source_file'),
-                                     ('t1w_preproc', 'target_file')]),
-        (inputnode, concat_xfm, [('source_file', 'moving'),
-                                 ('t1w_preproc', 'reference')]),
-        (inputnode, merge_xfm, [('t1w2fsnative_xfm', 'in2')]),
+        (inputnode, itk2lta, [('source_file', 'src_file'),
+                              ('t1w2fsnative_xfm', 'in_file')]),
+        (get_fsnative, tonii, [('T1', 'in_file')]),
+        (tonii, itk2lta, [('out_file', 'dst_file')]),
```
effigies, line 133: So:

```suggestion
        (get_fsnative, itk2lta, [('T1', 'dst_file')]),
```

mgxd:
testing off 1d417a9, surfaces look muuuuch better!
Screen Shot 2020-05-27 at 1 51 09 PM

fmriprep/workflows/bold/resampling.py
oesteban: I wasn't convinced about nitransforms' utilization of nibabel... I'll give your suggestion a try

oesteban: okay, nicely works - pretty slick.

oesteban:
Good to go?

Issue #2142, opened 2020-05-25
Inconsistent header/data information in aparcaseg


ariekahn:
version: 20.1.0 master (as of patch for #2138)

I'm getting odd behavior loading in any of the functional space `aparcaseg` files:

```python
from nilearn.image import load_img, new_img_like
img = load_img('GLS005_ses-2_task-graphlocalizer_space-MNI152NLin2009cAsym_desc-aparcaseg_dseg.nii.gz')
print('From fmriprep:')
print(f'data type: {img.get_data_dtype()}\n')

print('Unique values:')
with np.printoptions(precision=3, suppress=True):
    print(np.unique(img.dataobj))
    
img2 = new_img_like(x, np.round(data).astype('<i2'))
img2.to_filename('img2.nii.gz')
img2 = load_img('img2.nii.gz')
print('After Resaving:')
print(f'data type: {img2.get_data_dtype()}\n')
print('Unique values:')
with np.printoptions(precision=3, suppress=True):
    print(np.unique(img2.dataobj))
```

```
From fmriprep:
data type: int16

Unique values:
[   0.       1.987    4.006    4.999    6.987    8.011    9.999   10.992
   11.986   13.011   14.005   14.998   15.992   16.986   18.01    24.003
   25.991   28.009   30.99    40.989   43.007   44.001   45.988   47.013
   49.      49.994   50.988   52.012   53.006   54.      58.005   59.993
   62.011   63.005   77.009   84.99   250.994  251.988  253.013  254.006
  255.    1000.002 1000.996 1001.989 1003.014 1005.002 1005.995 1006.989
 1008.014 1009.007 1010.001 1010.995 1011.988 1013.013 1014.007 1015.
 1015.994 1016.988 1018.012 1019.006 1020.    1020.993 1021.987 1023.012
 1024.005 1024.999 1025.993 1026.986 1028.011 1029.005 1029.998 1030.992
 1031.986 1033.011 1034.004 1034.998 2000.004 2000.998 2001.992 2002.985
 2005.004 2005.997 2006.991 2007.985 2009.009 2010.003 2010.997 2011.99
 2013.015 2014.009 2015.002 2015.996 2016.99  2018.014 2019.008 2020.002
 2020.995 2021.989 2023.014 2024.008 2025.001 2025.995 2026.989 2028.013
 2029.007 2030.001 2030.994 2031.988 2033.013 2034.006 2035.   ]

After Resaving:
data type: int16

Unique values:
[...]

PR #2147, opened 2020-05-27
FIX: ``MultiLabel`` interpolations should not use ``float=True``

oesteban:
This is the root cause of #2142 (although our posterior handling of
the data types in the DerivativesDatasink wasn't great and that
justified nipreps/niworkflows#527).

I found many instances where ``ApplyTransforms`` was used in combination
with ``float=True``. This is why the data array contained floating point
values.

After using this patch, most of the ``_dseg.nii.gz`` files in the output
dropped from ~400KB down to ~25KB (confirming the size is changing on
hard disk). Visually checking one of these files seemed to give the
same labels.

Resolves: #2142

### DIFF
--- a/fmriprep/workflows/bold/confounds.py
+++ b/fmriprep/workflows/bold/confounds.py
@@ -492,11 +492,11 @@ def init_carpetplot_wf(mem_gb, metadata, cifti_output, name="bold_carpet_wf"):
 
     # Warp segmentation into EPI space
     resample_parc = pe.Node(ApplyTransforms(
-        float=True,
+        dimension=3,
         input_image=str(get_template(
             'MNI152NLin2009cAsym', resolution=1, desc='carpet',
             suffix='dseg', extension=['.nii', '.nii.gz'])),
-        dimension=3, default_value=0, interpolation='MultiLabel'),
+        interpolation='MultiLabel'),
         name='resample_parc')
 
     # Carpetplot and confounds plot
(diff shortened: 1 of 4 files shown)


fmriprep/workflows/bold/confounds.py
```diff
@@ -492,11 +492,11 @@ def init_carpetplot_wf(mem_gb, metadata, cifti_output, name="bold_carpet_wf"):
 
     # Warp segmentation into EPI space
     resample_parc = pe.Node(ApplyTransforms(
-        float=True,
+        default_value=0,
         input_image=str(get_template(
             'MNI152NLin2009cAsym', resolution=1, desc='carpet',
             suffix='dseg', extension=['.nii', '.nii.gz'])),
-        dimension=3, default_value=0, interpolation='MultiLabel'),
```
mgxd, line 499: will this ever not be `dimension=3`?

oesteban: nope, but `dimension=3` is the default, so no need to be set (same for `default_value`)

mgxd: https://github.com/nipy/nipype/blob/4c48a185e4d190071f78a6766ca34c71ff61e10f/nipype/interfaces/ants/resampling.py#L307-L317

from that help string it seems it is inferred if not provided, I think leaving it would be best

Issue #2150, opened 2020-05-29
Slightly confusing reports with SDC and multiecho


Shotgunosine:
Version: 20.1.0
What I was doing:
I've rerun fMRIPrep 20.1.0 on my longitudinal sample with single echo task based fMRI and multi-echo resting state. 

What I expected:
A single section of the report for each task based and resting state scan.

What actually happened:
In reviewing the reports, I was initially confused because there are two entries for each rest scan:

> Reports for: session 04, task rest, run 1.
> Reports for: session 04, task rest, run 1, echo 1.

It looks like the SDC report is getting split out and listed separately as echo 1, while the rest of the steps are shown under "Reports for: session 04, task rest, run 1." To make it more confusing, the steps under  "Reports for: session 04, task rest, run 1." list SDC:

> Summary
> 
>     Repetition time (TR): 2.5s
>     Phase-encoding (PE) direction: Anterior-Posterior
>     Slice timing correction: Applied
>     Susceptibility distortion correction: PEB/PEPOLAR (phase-encoding based / PE-POLARity)
>     Registration: FreeSurfer bbregister (boundary-based registration, BBR) - 6 dof
>     Non-steady-state volumes: 4 (Warning: 3 automatically detected)

I've checked four or five subject now and this has occurred for each of them, so I'm pretty sure this will replicate for any multi-echo scan with SDC, see https://openneuro.org/datasets/ds002156/versions/1.0.0 for one such example.

effigies:
Thanks. This is likely to be a pretty simple fix, so we can target 20.1.x.

oesteban:
I've noticed this is indeed happening with our smoke test (ds210). Agree with @effigies this should be easy (just a `dismiss_entities=("echo",)` in the corresponding DerivativesDataSink).

PR #2160, opened 2020-06-01
FIX: exclude "echo" entity from sdc reports

mgxd:
A test fix for #2150


mgxd:
ok, things look right:
https://14291-53175327-gh.circle-artifacts.com/0/tmp/ds210/derivatives/fmriprep/sub-02.html

oesteban:
Shouldn't this PR target `maint/20.1.x` ?

mgxd:
We could (and probably should), but I don't think we've introduced any `20.1.x` breaking changes yet?

oesteban:
Yup, a simple merge should do. I just realized that we need to start operating that way ASAP.

PR #2168, opened 2020-06-03
ENH: Ensure nipype telemetry is just pinged once

### LINKED ISSUE #2140: Ping Nipype's eTelemetry just once on ``config`` load - then disable.
fMRIPrep should ping Nipype's eTelemetry just once. We should do it (silently) when the config module is loaded first and then set the env variable to disable it going forward:

1. Read the environment and determine whether the user has not explicitly disabled the eTelemetry ping.
2. Before setting up the loggers, ping eTelemetry if the user has not disabled it, making sure we do not write anything to screen.
3. Set enviornment variable to disable eTelemetry pings.

mgxd:
Closes #2140 

adds `disable_telemetry` property to `config.nipype` to control whether the initial latest version is done. afterwards, both `NIPYPE_NO_ET` and the config value `check_version` are set to disable future checks

### DIFF
--- a/fmriprep/config.py
+++ b/fmriprep/config.py
@@ -244,6 +244,8 @@ class nipype(_Config):
 
     crashfile_format = 'txt'
     """The file format for crashfiles, either text or pickle."""
+    disable_telemetry = bool(os.getenv("NIPYPE_NO_ET") is None)
+    """Disable interface telemetry"""
     get_linked_libs = False
     """Run NiPype's tool to enlist linked libraries for every interface."""
     memory_gb = None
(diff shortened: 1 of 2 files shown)


fmriprep/config.py
```diff
@@ -244,6 +244,8 @@ class nipype(_Config):
 
     crashfile_format = 'txt'
     """The file format for crashfiles, either text or pickle."""
+    disable_telemetry = bool(os.getenv("NIPYPE_NO_ET") is None)
```
oesteban, line 247: having a config for this is a bad idea - this is one of those options we don't want to reuse when reusing the config file. Please check mgxd/fmriprep#3 for an alternative.

oesteban:
Circle was getting a lot of debug traces for the nipype ping, so this should be easy to check 👍 . We should also check that etelemetry is recording the initial ping.

PR #2180, opened 2020-06-06
FIX: Coerce license path to pathlike

mgxd:
#2165 dropped the conversion from str -> path-like for the wrapper's FreeSurfer license binding

### DIFF
--- a/wrapper/fmriprep_docker.py
+++ b/wrapper/fmriprep_docker.py
@@ -237,6 +237,7 @@ def get_parser():
 
     def _is_file(path, parser):
         """Ensure a given path exists and it is a file."""
+        path = os.path.abspath(path)
         if not os.path.isfile(path):
             raise parser.error(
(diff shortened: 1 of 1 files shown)


wrapper/fmriprep_docker.py
```diff
@@ -237,6 +237,7 @@ def __call__(self, parser, namespace, values, option_string=None):
 
     def _is_file(path, parser):
         """Ensure a given path exists and it is a file."""
+        path = os.path.abspath(path)
```
effigies, line 240: I would suggest that we switch to `return os.path.abspath(path)`, so that the exception matches the user input.

Feel free to disregard and merge, if you disagree. This is a good fix.

mgxd: I'll merge it in as is, I feel the full path to the invalid license would be helpful when debugging

Issue #2174, opened 2020-06-05
Report number of echoes (or "single-echo") in BOLD summary reports


oesteban:
That will help to diagnose problems like #2153

PR #2184, opened 2020-06-09
ENH: Report number of echoes in BOLD summary.

oesteban:
Builds on the new echo list extraction of #1803 to determine whether the
dataset is single-echo, multi-echo (and how many echoes are processed),
or multi-echo but processing only one of the echoes (e.g., using
``--bids-filter``).

Resolves: #2174.

### DIFF
--- a/fmriprep/interfaces/reports.py
+++ b/fmriprep/interfaces/reports.py
@@ -32,6 +32,7 @@ FUNCTIONAL_TEMPLATE = """\
 \t\t<ul class="elem-desc">
 \t\t\t<li>Repetition time (TR): {tr:.03g}s</li>
 \t\t\t<li>Phase-encoding (PE) direction: {pedir}</li>
+\t\t\t<li>{multiecho}</li>
 \t\t\t<li>Slice timing correction: {stc}</li>
 \t\t\t<li>Susceptibility distortion correction: {sdc}</li>
 \t\t\t<li>Registration: {registration}</li>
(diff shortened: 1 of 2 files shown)


oesteban:
This seems to be functioning, thumbs up?

fmriprep/interfaces/reports.py
```diff
@@ -218,10 +220,20 @@ def _generate_segment(self):
...
 
+        multiecho = "Single-echo EPI sequence."
+        n_echos = len(self.inputs.echo_idx)
+        if n_echos == 1:
+            multiecho = (
+                f"Multi-echo EPI sequence: only echo {self.inputs.echo_idx[0]} processed "
+                "in single-echo mode."
+            )
+        if n_echos > 2:
+            multiecho = (f"Multi-echo EPI sequence: {n_echos} echoes.")
```
tsalo, line 231: What about `n_echos == 2`? Perhaps that should be labeled `"Dual-echo EPI sequence"`.

oesteban: That will require a bit more of logic: at this point we should never see ``length == 2`` because dual echoes are split very early before this

tsalo: Ah, I see. Would it be better to not be explicit in the default then? I.e., not "single-echo EPI sequence", since it's not necessarily going to be single-echo? On the other hand, dual-echo is going to be pretty rare...

oesteban: In reality, this is to report how fMRIPrep worked - so I don't think it is inaccurate to say that the dataset was single-echo. Let's go with this and then refine is some user requests to have the possibility to see in the reports dual-echo as such.

PR #2185, opened 2020-06-09
ENH: load .bids_ignore file and feed it to pybids.BIDSLayout

bpinsard:

Ignore files in `.bidsignore` (as it is not used by pybids for now).
Can be useful in cases where we use BIDS extension with entities which are not yet supported by pybids and thus cannot be used to filter the fmriprep input data.

Currently, it might not be able to deal with the full .gitignore syntax as it only uses `fnmatch`. For instance the `!` might not work.

Also, I can't remember why the `BIDSLayout` is created with `validate=False`? For lesion masks maybe?

TODO.
Not sure in which part of the doc this should be documented.

### DIFF
--- a/fmriprep/config.py
+++ b/fmriprep/config.py
@@ -403,12 +403,13 @@ class execution(_Config):
             from bids.layout import BIDSLayout
             work_dir = cls.work_dir / 'bids.db'
             work_dir.mkdir(exist_ok=True, parents=True)
+            bids_ignore = load_bidsignore(cls.bids_dir)
             cls._layout = BIDSLayout(
                 str(cls.bids_dir),
                 validate=False,
                 # database_path=str(work_dir),
                 ignore=("code", "stimuli", "sourcedata", "models",
-                        "derivatives", re.compile(r'^\.')))
+                        "derivatives", re.compile(r'^\.')) + bids_ignore)
         cls.layout = cls._layout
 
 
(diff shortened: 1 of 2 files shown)


oesteban:
@effigies, shouldn't pyBIDS automatically pick up the `.bidsignore`?

oesteban:
> Also, I can't remember why the `BIDSLayout` is created with `validate=False`? For lesion masks maybe?

I think so. Also because we rely on the user not setting `--skip-bids-validation`.

oesteban:
> Not sure in which part of the doc this should be documented.

I think we may want to open a subsection under `docs/usage.rst` for the BIDS filters option.

effigies:
Sorry, I missed this. No `.bidsignore` is a BIDS validator concept, and not a PyBIDS one. It is not inherent to BIDS. There have been discussions about merging it, but the problem is that there are two reasonable interpretations:

1. Do not index the ignored data; it's just here for completeness, and tools should not attempt to interpret unspecified data.
2. Ignore warnings on the ignored data; it's here to be used by a tool that knows to look for it.

mgxd (commented):
Looking through this now, perhaps it makes more sense to leave non-standard layout initialization up to the user (using an already generated index of the layout with the newly added `--bids-database-dir` flag). Then, when https://github.com/bids-standard/pybids/pull/650 is merged, it will be much easier for users to generate custom layouts, such as adding `ignore` values.

effigies:
Agreed. I don't think this is a productive way forward as it will make fMRIPrep behave very idiosyncratically relative to other BIDS Apps. Let's close this, and I will review the PyBIDS CLI PR...

PR #2201, opened 2020-06-23
FIX: skip t1w check if anat-derivatives provided.

### LINKED ISSUE #2197: Is T1w necessary when precomputed anat is given?
When using precomputed anat, I am getting this message:

```
Exception: No T1w images found for participant 01. All workflows require T1w images.
```

It is because in our project, we have separate sub-datasets, one for anatomical and multiple functional.

Is the raw T1w used in the functional part of the pipeline, or this check could be skipped if smriprep output is given? 

Thanks!

bpinsard:

Fixes #2197 .
In the case where a precomputed anatomical derivatives are provided with `--anat-derivatives`, skip the check for an existing T1w.

For now the side-tracking of smriprep will produce a warning when the anat-derivatives does not match the necessary input. Should it produce an error and die instead?. I think that this option will be mainly used by skilled users, and they would rather know if fast-track didn't work.

### DIFF
--- a/fmriprep/workflows/base.py
+++ b/fmriprep/workflows/base.py
@@ -138,6 +138,7 @@ def init_single_subject_wf(subject_id):
         subject_data['t2w'] = []
 
     anat_only = config.workflow.anat_only
+    anat_precomp = config.execution.anat_derivatives is not None
     # Make sure we always go through these two checks
     if not anat_only and not subject_data['bold']:
         task_id = config.execution.task_id
--- a/fmriprep/workflows/base.py
+++ b/fmriprep/workflows/base.py
@@ -276,7 +276,7 @@ It is released under the [CC0]\
         ])
     else:
         workflow.connect([
-            (anat_preproc_wf, bids_info, [('outputnode.t1w_preproc', 'in_file')]),
+            (bidssrc, bids_info, [(('bold', fix_multi_T1w_source_name), 'in_file')]),
             (anat_preproc_wf, summary, [('outputnode.t1w_preproc', 't1w')]),
             (anat_preproc_wf, ds_report_summary, [('outputnode.t1w_preproc', 'source_file')]),
             (anat_preproc_wf, ds_report_about, [('outputnode.t1w_preproc', 'source_file')]),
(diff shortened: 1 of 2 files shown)


fmriprep/workflows/base.py
```diff
@@ -147,7 +148,7 @@ def init_single_subject_wf(subject_id):
                 subject_id, task_id if task_id else '<all>')
         )
 
-    if not subject_data['t1w']:
+    if not anat_derivatives and not subject_data['t1w']:
```
oesteban, line 151: This check should go after the branch in line 220 because you can imagine the expectations are not met and the T1w is not provided either.

bpinsard:
There is also a check for an existing T1w in `niworkflows.interfaces.bids.BIDSDataGrabber`.
Before I move forward with a PR to niworkflows, do you agree with the changes?

bpinsard:
With the current changes, I still have two problems with the following nodes that requires path to raw T1w file:
`fmriprep_wf.single_subject_01_wf.ds_report_about`
`fmriprep_wf.single_subject_01_wf.bids_info`


What would be the best approach with these nodes when using anat fast-track?
Skip them? (might not be possible for BidsInfo which apparently convert the grabbed T1w file to BIDS entities).
Replace the raw filename with the preprocessed filenames for T1w,T2w?

bpinsard:
Sorry, left this behind for a while.
Tests are failing because it requires the changes in niworkflows#545.

effigies:
@bpinsard Can you install your niworkflows branch by updating this line?

https://github.com/poldracklab/fmriprep/blob/658c0ad603aafaa247b75819deffdec35c7a2ca3/setup.cfg#L30

setup.cfg
```diff
@@ -27,7 +27,7 @@ install_requires =
     nipype >= 1.5
     nitime
     nitransforms >= 20.0.0rc3,<20.2
-    niworkflows @ git+https://github.com/nipreps/niworkflows.git@master
+    niworkflows @ git+https://github.com/nipreps/niworkflows.git@refs/pull/545/head
```
effigies, line 30:
```suggestion
    niworkflows @ git+https://github.com/nipreps/niworkflows.git@master
```

effigies:
Cool. Tests passing, so merged in niworkflows. Reverting to the dependency on `master`, and I think this can go in.

mgxd:
Cool, thanks @bpinsard!

bpinsard:
Thanks for your help and feedback!

PR #2202, opened 2020-06-25
Add --no-tty option to fmriprep-docker.py

kjamison:
Adding an option to the wrapper to omit the "-it" argument to docker, which avoids crashes in non-interactive environments (See https://github.com/nipreps/smriprep/issues/209 and https://github.com/nipreps/smriprep/pull/216#issuecomment-649134073)

### DIFF
--- a/wrapper/fmriprep_docker.py
+++ b/wrapper/fmriprep_docker.py
@@ -292,7 +292,9 @@ def get_parser():
                        help='Set custom environment variable within container')
     g_dev.add_argument('-u', '--user', action='store',
                        help='Run container as a given user/uid')
-
+    g_dev.add_argument('--no-tty', action='store_true',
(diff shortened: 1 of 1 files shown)


mgxd:
hey @kjamison, thanks for the PR! The reason the tests are failing is that your branch is quite outdated, could you update your commit to the HEAD? You should be able to do this by:

```
git remote add upstream git@github.com:poldracklab/fmriprep.git
git pull --rebase upstream master
<update changes in wrapper/fmriprep_docker.py>
git push -f origin add_no_tty_option
```

kjamison:
Hmm that's weird. I cloned right before making that change.

Anyway I just redid it.

-Keith

mgxd:
great - following up in #2204

PR #2203, opened 2020-06-26
ENH: Leverage BIDSLayout's database_path

### LINKED ISSUE #2019: Use ``BIDSLayout.database_path``
In general, it's the case that we run several subjects (sometimes a lot) all stored within the same BIDS dataset. Using the new pybids feature to store the index to hard-disk seems like an effective way of minimizing indexing times.

Problems that will need to be addressed:
- Parallelism: race conditions in generating the first indexed database (although I'm not sure this would be a real problem).
- Reproducibility: does pybids have any ways of checking whether the index should be updated? (cc/ @effigies)
- Where this index should be stored?:
  - The BIDS dataset root.
  - The output folder (I'm not sure of this one, as one would probably like to index the derivatives downstream).
  - The work directory (in many scenarios, e.g., running on a local scratch, this would defeat the purpose of this utility).

mgxd:
Closes #2019 

### Additions
- Adds support for `BIDSLayout.database_path`
  - If no database path is specified, one will be created in `<working-directory>/<run-uuid>`
  - If it is specified, will just use that path

### Changes
- A `<run-uuid>` directory is created at the top level of the `<working-directory>`, and the previously generated `config-<run-uuid>.toml` files are now located within each directory and renamed to `config.toml`

### DIFF
--- a/fmriprep/cli/parser.py
+++ b/fmriprep/cli/parser.py
@@ -156,7 +156,7 @@ def _build_parser():
         "with an alternative processing tool (NOT RECOMMENDED).",
     )
     g_bids.add_argument(
-        "--database-path",
+        "--bids-database-dir",
         metavar="PATH",
         type=PathExists,
         help="Path to an existing PyBIDS database folder, for faster indexing "
--- a/fmriprep/cli/parser.py
+++ b/fmriprep/cli/parser.py
@@ -155,6 +155,12 @@ def _build_parser():
         help="Reuse the anatomical derivatives from another fMRIPrep run or calculated "
         "with an alternative processing tool (NOT RECOMMENDED).",
     )
+    g_bids.add_argument(
+        "--database-path",
+        metavar="PATH",
+        type=PathExists,
+        help="Path to SQLite database indicies for BIDS dataset"
+    )
 
     g_perfm = parser.add_argument_group("Options to handle performance")
     g_perfm.add_argument(
--- a/fmriprep/cli/parser.py
+++ b/fmriprep/cli/parser.py
@@ -159,7 +159,8 @@ def _build_parser():
         "--database-path",
         metavar="PATH",
         type=PathExists,
-        help="Path to SQLite database indicies for BIDS dataset"
+        help="Path to an existing PyBIDS database folder, for faster indexing "
+             "(especially useful for large datasets)."
     )
 
     g_perfm = parser.add_argument_group("Options to handle performance")
--- a/fmriprep/cli/run.py
+++ b/fmriprep/cli/run.py
@@ -26,7 +26,8 @@ def main():
     # CRITICAL Save the config to a file. This is necessary because the execution graph
     # is built as a separate process to keep the memory footprint low. The most
     # straightforward way to communicate with the child process is via the filesystem.
-    config_file = config.execution.work_dir / f"config-{config.execution.run_uuid}.toml"
+    config_file = config.execution.work_dir / config.execution.run_uuid / "config.toml"
+    config_file.parent.mkdir(exist_ok=True, parents=True)
     config.to_filename(config_file)
 
     # CRITICAL Call build_workflow(config_file, retval) in a subprocess.
--- a/fmriprep/config.py
+++ b/fmriprep/config.py
@@ -332,14 +332,14 @@ class execution(_Config):
     """A path where anatomical derivatives are found to fast-track *sMRIPrep*."""
     bids_dir = None
     """An existing path to the dataset, which must be BIDS-compliant."""
+    bids_database_dir = None
+    """Path to the directory containing SQLite database indices for the input BIDS dataset."""
     bids_description_hash = None
     """Checksum (SHA256) of the ``dataset_description.json`` of the BIDS dataset."""
     bids_filters = None
     """A dictionary of BIDS selection filters."""
     boilerplate_only = False
     """Only generate a boilerplate."""
-    database_path = None
-    """Path to the directory containing SQLite database indices for the input BIDS dataset."""
     debug = False
     """Run in sloppy mode (meaning, suboptimal parameters that minimize run-time)."""
     echo_idx = None
@@ -385,7 +385,7 @@ class execution(_Config):
     _paths = (
         'anat_derivatives',
         'bids_dir',
-        'database_path',
+        'bids_database_dir',
         'fs_license_file',
         'fs_subjects_dir',
         'layout',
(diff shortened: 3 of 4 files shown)


fmriprep/config.py
```diff
@@ -401,12 +404,14 @@ def init(cls):
         if cls._layout is None:
             import re
             from bids.layout import BIDSLayout
-            work_dir = cls.work_dir / 'bids.db'
-            work_dir.mkdir(exist_ok=True, parents=True)
+            _db_path = cls.database_path or (cls.work_dir / cls.run_uuid / 'bids.db')
```
oesteban, line 407: Is `bids.db` meant to be a folder? (because bidslayout will create a folder)

fmriprep/cli/parser.py
```diff
@@ -151,6 +151,12 @@ def _bids_filter(value):
         help="Reuse the anatomical derivatives from another fMRIPrep run or calculated "
         "with an alternative processing tool (NOT RECOMMENDED).",
     )
+    g_bids.add_argument(
+        "--database-path",
+        metavar="PATH",
+        type=PathExists,
+        help="Path to SQLite database indicies for BIDS dataset"
```
oesteban, line 158:
```suggestion
        help="Path to an existing PyBIDS database folder, for faster indexing "
             "(especially useful for large datasets)."
```

fmriprep/cli/run.py
```diff
@@ -26,7 +26,8 @@ def main():
     # CRITICAL Save the config to a file. This is necessary because the execution graph
     # is built as a separate process to keep the memory footprint low. The most
     # straightforward way to communicate with the child process is via the filesystem.
-    config_file = config.execution.work_dir / f"config-{config.execution.run_uuid}.toml"
+    config_file = config.execution.work_dir / config.execution.run_uuid / "config.toml"
+    config_file.parent.mkdir(exist_ok=True, parents=True)
```
oesteban, line 30: This is a bit unrelated. Are you 100% there are no side-effects?

fmriprep/config.py
```diff
@@ -338,6 +338,8 @@ class execution(_Config):
     """A dictionary of BIDS selection filters."""
     boilerplate_only = False
     """Only generate a boilerplate."""
+    database_path = None
+    """Path to directory containing SQLite database indicies for BIDS dataset"""
```
oesteban, line 342:
```suggestion
    """Path to the directory containing SQLite database indices for the input BIDS dataset."""
```

fmriprep/config.py
```diff
@@ -401,12 +404,14 @@ def init(cls):
         if cls._layout is None:
             import re
             from bids.layout import BIDSLayout
-            work_dir = cls.work_dir / 'bids.db'
-            work_dir.mkdir(exist_ok=True, parents=True)
+            _db_path = cls.database_path or (cls.work_dir / cls.run_uuid / 'bids.db')
+            _db_path.mkdir(exist_ok=True, parents=True)
+            _reset_db = bool(cls.database_path is not None)
```
oesteban, line 409: This will reindex the dataset if `--database_path` is provided, correct? What is the need for this?

fmriprep/config.py
oesteban: That said, keeping an index under the `run_uuid` folder will make the caching useless in 99% of the cases, as normally a new uuid will be generated. I would rather keep the general database under work_dir.

mgxd: > Is `bids.db` meant to be a folder? (because bidslayout will create a folder)

yes, i kept the name you used in your original implementation.

> That said, keeping an index under the `run_uuid` folder will make the caching useless in 99% of the cases, as normally a new uuid will be generated. I would rather keep the general database under work_dir.

Well, there's a trade off there too. If we generate at the top-level `work_dir`, we have to handle potential race conditions of database index initializing in cases where multiple participant labels are run concurrently.

fmriprep/cli/run.py
mgxd: this change cascaded from the addition of the <run-uuid> subdirectory within the work dir. I think it cleans up the structure a bit, organizing the `config-<run-uuid>.toml` top-level files.

fmriprep/config.py
mgxd: I think this is a mistake, the indexing should only be reset if database_path was not provided.

fmriprep/config.py
```diff
@@ -401,14 +404,16 @@ def init(cls):
...
+            _db_path.mkdir(exist_ok=True, parents=True)
             cls._layout = BIDSLayout(
                 str(cls.bids_dir),
                 validate=False,
-                # database_path=str(work_dir),
+                database_path=_db_path,
+                reset_database=cls.database_path is None,
                 ignore=("code", "stimuli", "sourcedata", "models",
                         "derivatives", re.compile(r'^\.')))
+            cls.database_path = _db_path
```
effigies, line 416: I think this implies reusing the cache if you're reusing the config?

effigies:
Overall I think this looks good. I agree that we should make it run-specific if it isn't passed as an explicit argument, because the alternative is complicated filesystem-dependent behavior. Left a comment to clarify about what happens when config reuse becomes possible. I don't think we do want to reuse database paths in that case, as there's no requirement that the indexed dataset be the same.

fmriprep/config.py
mgxd: I added this since the config is saved and reloaded during workflow generation
https://github.com/poldracklab/fmriprep/blob/3b1dabdb3b5205ed48771a9fa0d9638f61c80d36/fmriprep/cli/run.py#L29-L49
and I wanted to ensure it wouldn't be re-indexed on reload.

Good point, we will have to differentiate between intra/inter-run config reuse - though I think that is something to worry about in the context of the reuse PR.

fmriprep/config.py
oesteban: > Well, there's a trade off there too. If we generate at the top-level `work_dir`, we have to handle potential race conditions of database index initializing in cases where multiple participant labels are run concurrently.

You mean in NFS systems? I believe directory creation is the only atomic operation in NFS systems - meaning it is either present or not present.

What about the following:

- if the database dir exists in the work_dir, it passed in to BIDSLayout as is and not forced to reindex.
- if the database dir does not exist, can we initialize it to the run_uuid folder and then move to the work_dir (except some other process did already and it exists)?

effigies: I think asking people to pre-create a database cache if they want to use it without re-indexing (and thus from multiple concurrent runs) is fine. The alternative seems to be asking for subtle bugs.

mgxd: after talking about this with @oesteban, we agreed it would be nice to provide a CLI tool to generate a database index of a given BIDS dataset. @effigies do you think that would be something we could/should include in pybids, or rather in another spinoff package?

effigies: That could easily go in pybids, but it would need to be flexible. An index depends on initialization parameters, so if two apps use different initialization parameters, a user would need to know enough to initialize two separate indices.

PR #2208, opened 2020-07-06
FIX: Drop leading "." on BIDSLayout queries

mgxd:
This updates any BIDSLayout queries to ensure compatibility after PyBIDS 0.11.0.


effigies:
The goal was to move toward using the initial dot, not away. Is something going wrong?

mgxd:
yeah, the following query is now failing (showing up in the docs build)

```python
import templateflow.api as tf
tf.get('MNI152NLin2009cAsym', resolution=1, desc='carpet', suffix='dseg', extension=['.nii', '.nii.gz'])
```

effigies:
Oh, I see, this is due to the fix you added. We were previously helping people by adding/dropping `.`s even when they provided their own config. That became unwieldy due to the shift.

Does it make more sense to update templateflow to include the `.` in the extension?

PR #2240, opened 2020-08-18
ENH: Reuse config

### LINKED ISSUE #2187: Reuse of config file
Allow loading configuration files from previous *fMRIPrep* runs.

effigies:
Restart of #2107. Last comments were about restyling with `black`, so I restyled, and rebased all of #2107 on top of this, restyling in the process.

Closes #2187.

### DIFF
--- a/fmriprep/cli/parser.py
+++ b/fmriprep/cli/parser.py
@@ -318,6 +318,7 @@ https://fmriprep.readthedocs.io/en/%s/spaces.html"""
     )
     g_conf.add_argument(
         "--random-seed",
+        dest="_random_seed",
         action="store",
         type=int,
         default=None,
@@ -505,13 +506,6 @@ https://fmriprep.readthedocs.io/en/%s/spaces.html"""
         "aggregation, not reportlet generation for specific nodes.",
     )
     g_other.add_argument(
-        "--run-uuid",
-        action="store",
-        default=None,
-        help="Specify UUID of previous run, to include error logs in report. "
-        "No effect without --reports-only.",
-    )
-    g_other.add_argument(
         "--write-graph",
         action="store_true",
         default=False,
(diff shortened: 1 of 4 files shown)


fmriprep/cli/parser.py
```diff
@@ -570,11 +564,23 @@ def _bids_filter(value):
...
+
+    new_config = os.getenv('FMRIPREP_NO_REUSE_CONFIG')
+    if not new_config:
+        # attempt to reuse previous config
+        config.load_previous_config(
+            opts.output_dir,
+            opts.work_dir,
+            participant=opts.participant_label,
+            new_uuid=not opts.reports_only,
+        )
```
effigies, line 582: This should not be a default behavior.

fmriprep/config.py
```diff
@@ -573,18 +573,19 @@ def init(cls):
...
     def init(cls):
-        cls.master = workflow.random_seed
+        if cls._random_seed is not None:
+            cls.master = cls._random_seed
         if cls.master is None:
             cls.master = random.randint(1, 65536)
         random.seed(cls.master)  # initialize the PRNG
-
         # functions to set program specific seeds
         cls.ants = _set_ants_seed()
```
effigies, line 590: This should depend on the master seed, so that it can be reproduced by setting the master seed.

fmriprep/config.py
```diff
@@ -683,3 +686,41 @@ def init_spaces(checkpoint=True):
...
+    if conf is None:
+        # work directory
+        conf = Path(work_dir) / f".{CONFIG_FILENAME}"
+        if not conf.exists():
+            conf = None
+
+    # load existing config
+    if conf is not None:
+        # settings to avoid reusing
+        skip = {} if new_uuid else {"execution": ("run_uuid",)}
```
effigies, line 721: I think this is backwards. My reading is `opts.reports_only -> new_uuid=False -> skip = {"execution": ("run_uuid",)} -> new run_uuid generated`.

fmriprep/config.py
mgxd: isn't that already the case? the seed is set above, and then the method is called.

fmriprep/cli/parser.py
mgxd: I tend to agree here, but I think @oesteban felt otherwise?

fmriprep/config.py
mgxd: I think the idea behind the logic here is:

- if `conf` exists, we have found a previous config file with all options (including `run_uuid`)
- when loading this config, skip reusing the previous `run_uuid` and use the newly generated one (so future runs are still treated as unique)

mgxd: The idea here is:

- if `conf` exists, we have found a previous config file with all options (including `run_uuid`)
- when loading this config, skip reusing the previous `run_uuid` and use the newly generated one (so future runs are still treated as unique)

fmriprep/config.py
effigies: Oh, right. My eyes skipped over that.

effigies:
Rebased into a small set of commits. This is ready for review.

PR #2248, opened 2020-09-01
ENH: Output compcor masks if `--debug compcor` is passed

### LINKED ISSUE #2245: Write out "sanity" derivatives / introspection mode
In trying to extract the signal for each of the CompCor ROIs to then check their correlation with the global signal (as @rciric suggested in order to finalize #2130) I have realized that I will need to keep the working directory somewhere and then manually select the ROI masks.

I propose to add a new ``--introspect-mode`` flag that runs the workflow in some sort of "debug mode" and writes out to the derivatives folder additional information such as the compcor masks.

In addition, and independently of the ``--introspect-mode``, we could also internally keep some "sanity features", such as the timeseries from CompCor ROIs and use those to make decisions (i.e., very much like the fallback for bold-t1 registration). When in ``--introspect-mode``, these signals/features would be written out to the derivatives folder.

WDYT? @effigies @mgxd 

Note: this is also related in a way to @satra's longstanding request for more quantitative quality metrics, he referred to freesurfers when proposed, but I think it goes to the same point.

effigies:
Closes #2245. Builds on #2130 to avoid writing something that will need to be refactored...

Should be reviewed per-commit.


effigies:
Rebased for a cleaner merge, but the force push [diff](https://github.com/poldracklab/fmriprep/compare/a7ad411bdae3ede743526f18b1e4ce13690ac101..791b1a6efdfe9e4f4fc0e813a47d23b3b38389c3) is empty. This can be merged without waiting on CI, if we're happy with it.

@oesteban Are you okay with this approach?

.circleci/ds054_fasttrack_outputs.txt
```diff
@@ -9,12 +9,16 @@ fmriprep/logs/CITATION.tex
 fmriprep/sub-100185
 fmriprep/sub-100185/func
 fmriprep/sub-100185/func/sub-100185_task-machinegame_run-1_boldref.nii.gz
+fmriprep/sub-100185/func/sub-100185_task-machinegame_run-1_desc-acompcor_mask.nii.gz
 fmriprep/sub-100185/func/sub-100185_task-machinegame_run-1_desc-brain_mask.json
 fmriprep/sub-100185/func/sub-100185_task-machinegame_run-1_desc-brain_mask.nii.gz
 fmriprep/sub-100185/func/sub-100185_task-machinegame_run-1_desc-confounds_timeseries.json
 fmriprep/sub-100185/func/sub-100185_task-machinegame_run-1_desc-confounds_timeseries.tsv
+fmriprep/sub-100185/func/sub-100185_task-machinegame_run-1_desc-CSF_mask.nii.gz
```
oesteban, line 17:
```suggestion
fmriprep/sub-100185/func/sub-100185_task-machinegame_run-1_label-CSF_desc-acompcor_mask.nii.gz
```

I can't remember where we left the `label-` conversation with masks. But this is something we want to address.

As it is now, this can be read as _nice, and here is my CSF mask_, which is not true. This is an eroded CSF mask, so either we add the label (if BIDS compatible) or we cram it into the `desc` like `desc-acompcorCSF`.

.circleci/ds054_fasttrack_outputs.txt
```diff
@@ -9,12 +9,16 @@ fmriprep/logs/CITATION.tex
...
 fmriprep/sub-100185/func/sub-100185_task-machinegame_run-1_desc-preproc_bold.json
 fmriprep/sub-100185/func/sub-100185_task-machinegame_run-1_desc-preproc_bold.nii.gz
+fmriprep/sub-100185/func/sub-100185_task-machinegame_run-1_desc-tcompcor_mask.nii.gz
```
oesteban, line 20:
```suggestion
fmriprep/sub-100185/func/sub-100185_task-machinegame_run-1_desc-CompCorTemp_mask.nii.gz
```

a/tCompCor can be determined from the ROIs used. If we use `desc-CompCorXXX` then all the compcor-related masks will go together when files are listed ordered by name, which is convenient.

fmriprep/workflows/bold/outputs.py
```diff
@@ -338,6 +339,23 @@ def init_func_derivatives_wf(
                                         (('cifti_metadata', _read_json), 'meta_dict')])
         ])
 
+    if "compcor" in config.execution.debug:
+        ds_acompcor_masks = pe.Node(
+            DerivativesDataSink(
+                base_directory=output_dir, suffix="mask", desc=["CSF", "WM", "acompcor"],
+                compress=True),
+            name="ds_acompcor_masks", run_without_submitting=True)
```
oesteban, line 347: If my suggestion above is taken, I think these two data sinks could be reduced to just one.

oesteban (commented):
I like the approach, thanks for this. Left a few comments

effigies:
I don't have a strong feeling about the names, as these are debug outputs. I agree we probably shouldn't confuse them for non-eroded masks... What about `CompCor[ACWT]`?

oesteban:
> I don't have a strong feeling about the names, as these are debug outputs. I agree we probably shouldn't confuse them for non-eroded masks... What about `CompCor[ACWT]`?

Perfect!

fmriprep/workflows/bold/outputs.py
effigies: We'd also need a Merge node, so it's very little difference in verbosity, and I think explicit will be clearer.

oesteban (approved):

fmriprep/cli/parser.py
```diff
@@ -533,9 +533,16 @@ def _bids_filter(value):
...
+        "--debug",
+        action="store",
+        nargs="+",
+        choices=config.DEBUG_MODES + ("all",),
+        help="Debug mode(s) to enable. 'all' is alias for all available modes.",
+    )
+
     g_other.add_argument(
         "--sloppy",
-        dest="debug",
```
mgxd, line 538: dropping this will require renaming of current `debug` variables in the workflow:

https://github.com/poldracklab/fmriprep/blob/d7fc498b4d58a60ec174d0ec088770bdfed5483f/fmriprep/workflows/base.py#L240
https://github.com/poldracklab/fmriprep/blob/3ca515c4a63966480c0a93c0b36192e19d3888dd/fmriprep/workflows/bold/base.py#L336

mgxd (commented):
one concern:

oesteban:
Can't we just replace those with config.execution.sloppy?

mgxd:
yes, but I can't easily suggest the change 😄

fmriprep/cli/parser.py
effigies: Ah, good catch.

fmriprep/workflows/base.py
```diff
@@ -237,7 +237,7 @@ def init_single_subject_wf(subject_id):
     # Preprocessing of T1w (includes registration to MNI)
     anat_preproc_wf = init_anat_preproc_wf(
         bids_root=str(config.execution.bids_dir),
-        debug=config.execution.debug is True,
+        debug=config.execution.sloppy is True,
```
mgxd, line 240: might as well clean this up while we're touching it.
```suggestion
        debug=config.execution.sloppy,
```

mgxd (approved):

Issue #1746, opened 2019-08-26
Correlation plots are basically unreadable


effigies:
The labels are too large to match the rows/columns they are attached to. One approach to consider is breaking it into chunks using black lines, and identifying the clusters, such as a/tCompCor, cosine basis, motion parameters. Then there could be enough room to help identify the order of the handful of gross signals (GS, CSF, WM) and summary confounds (FD, DVARS).

effigies:
Oh, meant to link to a full-size report: http://datasets.datalad.org/labs/poldrack/ds003_fmriprep/sub-01.html

I'm guessing we were mostly looking at the dummy reports produced on Circle with a small TR count?

oesteban:
Closed via #2252

PR #2252, opened 2020-09-03
ENH: Revise confounds in confounds-correlation plots

oesteban:
Fixes: #1746

### DIFF
--- a/fmriprep/workflows/bold/confounds.py
+++ b/fmriprep/workflows/bold/confounds.py
@@ -339,7 +339,7 @@ were annotated as motion outliers.
         import pandas as pd
         return [
             col for col in pd.read_table(table, nrows=2).columns
-            if not col.startswith(("a_comp_cor_", "t_comp_cor_"))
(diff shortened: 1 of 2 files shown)


fmriprep/workflows/bold/confounds.py
```diff
@@ -387,6 +396,8 @@ def _pick_wm(files):
         (wm_tfm, wm_msk, [('output_image', 'roi_file')]),
         (wm_msk, mrg_lbl, [('out', 'in2')]),
         (inputnode, mrg_lbl, [('bold_mask', 'in3')]),
+        (acc_msk, mrg_lbl, [('out', 'in4')]),
+        (tcc_msk, mrg_lbl, [('out', 'in5')]),
```
effigies, line 400: Don't we want to inspect the `high_variance_masks` output of `tcompcor`, rather than just the input mask?

oesteban: probably yes, I'll check

oesteban: okay yes. the correlation of the mask currently connected will be always near 1.0. Just added a commit.

PR #2255, opened 2020-09-04
CI: Use 2020 resolver for pip to ensure correct resolution

effigies:
In PyPI checks, we see:

```
ERROR: After October 2020 you may experience errors when installing or updating packages. This is because pip will change the way that it resolves dependency conflicts.

We recommend you use --use-feature=2020-resolver to test your packages with the new resolver before it becomes the default.
```

This PR adds the suggested flag to ensure we won't get caught short in October.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -910,7 +910,7 @@ jobs:
             virtualenv --python=python sdist
             source sdist/bin/activate
             pip install --upgrade "pip>=19.1" numpy
-            pip install dist/fmriprep*.tar.gz
+            pip install --use-feature=2020-resolver dist/fmriprep*.tar.gz
             which fmriprep | grep sdist\\/bin
             INSTALLED_VERSION=$(fmriprep --version)
             INSTALLED_VERSION=${INSTALLED_VERSION%$'\r'}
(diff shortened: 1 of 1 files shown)


effigies:
@oesteban I'm not sure I understand the current

* `sdcflows<1.5,>=1.3.2`
* `nitransforms<20.2,>=20.0.0rc3`

Neither sdcflows 1.4 nor nitransforms 20.1 exist, so I'm not sure why we're preemptively supporting them.

effigies:
Since this only affects the pypi deploy tests, I canceled the rest. This can be merged, or we can throw in any changes we might want to make to the dependencies while we're here.

oesteban:
> Neither sdcflows 1.4 nor nitransforms 20.1 exist, so I'm not sure why we're preemptively supporting them.

For nitransforms it is just flexibility. We don't know yet what release plan (if there is one) there is for nitransforms, and its API should not really change.

For SDCFlows I just checked https://www.nipreps.org/devs/versions/ and it seems we did not contemplate a minor version bump with 20.2.x so this is my mistake. We should probably set `sdcflows ~= 1.3.2` for the next release.

PR #2256, opened 2020-09-04
DOC: Add FAQ entry for using pre-indexed layouts

mgxd:


docs/faq.rst
```diff
@@ -292,3 +292,17 @@ Some examples follow:
...
+How to decrease *fMRIPrep* runtime when working with large datasets?
+--------------------------------------------------------------------
+*fMRIPrep* leverages PyBIDS to produce a layout, which indexes the input BIDS dataset and facilitates file queries.
+Depending on the amount of files and metadata within the BIDS dataset, this process can be time-intensive.
+As of the 20.2.0 release, *fMRIPrep* supports the ``--bids-database-dir`` option, which can be used to pass in an already indexed BIDS layout.
+
+The default *fMRIPrep* layout can be generated by running the following shell command (requires PyBIDS 0.12.1 or greater)::
+
+  pybids layout <path-to-bids-dir> <path-to-save-index> --no-validate \
+  --ignore code --ignore stimuli --ignore sourcedata --ignore models --ignore derivatives --ignore m/^\./
```
effigies, line 306: These should already be ignored by default...

```suggestion
  pybids layout <path-to-bids-dir> <path-to-save-index> --no-validate
```

mgxd: Almost, but `derivatives` is missing from the [pybids defaults](https://github.com/bids-standard/pybids/blob/2fc912f7987b3e56472b6fb15286abee9e4cbe89/bids/layout/validation.py#L33-L34) but present in ours https://github.com/poldracklab/fmriprep/blob/d82bb4d7d652517e32029bc3840e39edc9907f93/fmriprep/config.py#L443-L450

Unless `BIDSLayout` filters those out separately?

effigies: Derivatives is a special case, and is always ignored. When  you use `--derivatives`, it adds derivatives in a separate stage, after the normal indexing.

mgxd: Cool, I'll update the BIDSLayout initialization for clarity

effigies: Yeah, I think it was like that because one version of PyBIDS somehow lost the automatic ignores, and we just put them in to avoid having to add a new minimum PyBIDS on all of the dependencies.

docs/faq.rst
```diff
@@ -292,3 +292,16 @@ Some examples follow:
     Please check the `link
     <https://neurostars.org/t/fmriprep-how-to-reuse-longitudinal-and-pre-run-freesurfer/4585/15>`__
     for further information on this approach.
+
+
+How to decrease *fMRIPrep* runtime when working with large datasets?
+--------------------------------------------------------------------
+*fMRIPrep* leverages PyBIDS to produce a layout, which indexes the input BIDS dataset and facilitates file queries.
+Depending on the amount of files and metadata within the BIDS dataset, this process can be time-intensive.
+As of the 20.2.0 release, *fMRIPrep* supports the ``--bids-database-dir`` option, which can be used to pass in an already indexed BIDS layout.
```
oesteban, line 301:
```suggestion
As of the 20.2.0 release, *fMRIPrep* supports the ``--bids-database-dir <database_dir>`` option,
which can be used to pass in an already indexed BIDS layout.
```

docs/faq.rst
```diff
@@ -292,3 +292,16 @@ Some examples follow:
...
+
+The default *fMRIPrep* layout can be generated by running the following shell command (requires PyBIDS 0.12.1 or greater)::
+
+  pybids layout <path-to-bids-dir> <path-to-save-index> --no-validate
```
oesteban, line 305:
```suggestion
  pybids layout <bids_root> <database_dir> --no-validate
```

docs/faq.rst
```diff
@@ -292,3 +292,16 @@ Some examples follow:
...
+
```
oesteban, line 306:
```suggestion

where ``<bids_root>`` indicates the root path of the BIDS dataset, and ``<database_dir>``
is the path to the pre-indexed layout - which is then passed into *fMRIPrep*.

```

docs/faq.rst
```diff
@@ -292,3 +292,16 @@ Some examples follow:
...
+Alternatively, a custom layout can be generated to provide finer-grain control of included and/or excluded files.
```
oesteban, line 307: I think I don't understand this. What do you mean here?

mgxd: I meant that alternative / additional `pybids layout --ignore` and `--force-index` arguments can be used to include / exclude files indexed. Given that this is a pretty rare use case, it might be best to just remove this bit

effigies:
```suggestion
By using the ``--force-index`` and ``--ignore`` options,
finer control can be achieved of what files are visible to fMRIPrep.

Note that any discrepancies between the pre-indexed database and
the BIDS dataset complicate the provenance of fMRIPrep derivatives.
If `--bids-database-dir` is used, the referenced directory should be
preserved for the sake of reporting and reproducibility.
```

bpinsard:
Could there be any implication if the database is pre-indexed without the same ignores as in fmriprep, as I think it is the case now with pybids cli?

https://github.com/poldracklab/fmriprep/blob/c1cd30514a38e0de950b1e644795945d000c099a/fmriprep/config.py#L443-L449

mgxd:
haha this is exactly what we discussed here: https://github.com/poldracklab/fmriprep/pull/2256#discussion_r483894778

bpinsard:
Ah yes, sorry I had not seen that comment, that answer my question.

Issue #2266, opened 2020-09-13
Missing template, although it should be prefetched by TemplateFlow


oesteban:
This is immediately related to #2265.

Although explicitly added when building the image (a7be31fc6742326d56ba113ace80fcd9330b451e), it seems that templates are re-downloaded again under some undetermined circumstances (https://app.circleci.com/pipelines/github/poldracklab/fmriprep/495/workflows/6f50b83f-77ec-4e64-bf81-61f14fe53474/jobs/15827).

PR #2268, opened 2020-09-15
FIX: More targetted TemplateFlow queries to work with all later releases

oesteban:
When putting together the 20.1.3 release, some version conflicts in
addition to the extension (dot-start) changes made the image be shipped
without important blobs.
The new queries should work well for any version of TemplateFlow, while
ensuring we don't get empty queries.

Resolves: #2266


mgxd (commented):
For the sake of cleaning up the Dockerfile, I think we should move away from the command-line python injection and use a dedicated helper script for this. I'll open up a draft shortly

Issue #2265, opened 2020-09-13
[CircleCI] Temporary failure in name resolution when fetching templates


oesteban:
https://app.circleci.com/pipelines/github/poldracklab/fmriprep/495/workflows/6f50b83f-77ec-4e64-bf81-61f14fe53474/jobs/15827

```
Downloading https://templateflow.s3.amazonaws.com/tpl-MNI152NLin2009cAsym/tpl-MNI152NLin2009cAsym_res-02_desc-fMRIPrep_boldref.nii.gz
Process Process-2:
Traceback (most recent call last):
  File "/usr/local/miniconda/lib/python3.7/site-packages/urllib3/connection.py", line 171, in _new_conn
    (self._dns_host, self.port), self.timeout, **extra_kw)
  File "/usr/local/miniconda/lib/python3.7/site-packages/urllib3/util/connection.py", line 56, in create_connection
    for res in socket.getaddrinfo(host, port, family, socket.SOCK_STREAM):
  File "/usr/local/miniconda/lib/python3.7/socket.py", line 748, in getaddrinfo
    for res in _socket.getaddrinfo(host, port, family, type, proto, flags):
socket.gaierror: [Errno -3] Temporary failure in name resolution

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/miniconda/lib/python3.7/site-packages/urllib3/connectionpool.py", line 600, in urlopen
    chunked=chunked)
  File "/usr/local/miniconda/lib/python3.7/site-packages/urllib3/connectionpool.py", line 343, in _make_request
    self._validate_conn(conn)
  File "/usr/local/miniconda/lib/python3.7/site-packages/urllib3/connectionpool.py", line 849, in _validate_conn
    conn.connect()
  File "/usr/local/miniconda/lib/python3.7/site-packages/urllib3/connection.py", line 314, in connect
    conn = self._new_conn()
  File "/usr/local/miniconda/lib/python3.7/site-packages/urllib3/connection.py", line 180, in _new_conn
    self, "Failed to establish a new connection: %s" % e)
urllib3.exceptions.NewConnectionError: <urllib3.connection.VerifiedHTTPSConnection object at 0x7f13b27ab860>: Failed to establish a new connection: [Errno -3] Temporary failure in name resolution

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/miniconda/lib/python3.7/site-packages/requests/adapters.py", line 445, in send
[...]

PR #2269, opened 2020-09-15
FIX: Dependency conflict between niworkflows and templateflow

oesteban:
Addresses:

```
ERROR: niworkflows 1.2.9 has requirement pybids~=0.10.2, but you'll have pybids 0.12.1 which is incompatible.
ERROR: niworkflows 1.2.9 has requirement templateflow<0.6.3,>=0.4.2, but you'll have templateflow 0.6.3 which is incompatible.
```

References: #2264
References: #2266

IMPORTANT: this commit should not be merged into master.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -31,12 +31,12 @@ install_requires =
     numpy
     pandas
     psutil >= 5.4
-    pybids >= 0.10.2
+    pybids ~= 0.10.2
     pyyaml
     sdcflows ~= 1.3.1
     smriprep ~= 0.6.1
     tedana >= 0.0.9a1, < 0.0.10
-    templateflow ~= 0.6
+    templateflow >=0.4.2, < 0.6.3
     toml
 test_requires =
     coverage


effigies (approved):
Looks reasonable. Because you're running Circle on your own account instead of poldracklab, I can't restart the failing test, but it looks like a one-off failure. Can you restart?

setup.cfg
```diff
@@ -31,12 +31,12 @@ install_requires =
...
     pandas
     psutil >= 5.4
-    pybids >= 0.10.2
+    pybids ~= 0.10.2
     pyyaml
     sdcflows ~= 1.3.1
     smriprep ~= 0.6.1
     tedana >= 0.0.9a1, < 0.0.10
-    templateflow ~= 0.6
+    templateflow >=0.4.2, <0.6.3
```
mgxd, line 39:
```suggestion
    templateflow >=0.4.2, < 0.6.3
```

mgxd (approved):

oesteban:
> it looks like a one-off failure. Can you restart?

That's #2265 (in addition to #2266).

Merging this after #2268 should do.

PR #2270, opened 2020-09-15
RF: Dockerfile

mgxd:
- Clean up extraneous bits
- Create dedicated script for template caching

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1020,7 +1020,7 @@ jobs:
       - run:
           name: Build fMRIPrep
           command: |
-            pip install --user twine  # For use in checking distributions
+            pip install --user --upgrade twine importlib-metadata  # For use in checking distributions
             THISVERSION=$( python get_version.py )
             THISVERSION=${THISVERSION%.dirty*}
             THISVERSION=${CIRCLE_TAG:-$THISVERSION}
@@ -1112,7 +1112,7 @@ jobs:
       - run:
           name: Upload packages to PyPI
           command: |
-            pip install --user twine
+            pip install --user --upgrade twine importlib-metadata
             twine upload dist/fmriprep* wrapper/dist/fmriprep*
 
   deployable:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1013,14 +1013,14 @@ jobs:
 
   test_deploy_pypi:
     docker:
-      - image: circleci/python:3.7.4
+      - image: circleci/python:3.8.5
     working_directory: /tmp/src/fmriprep
     steps:
       - checkout
       - run:
           name: Build fMRIPrep
           command: |
-            pip install --user --upgrade twine importlib-metadata  # For use in checking distributions
+            python -m pip install --user twine  # For use in checking distributions
             THISVERSION=$( python get_version.py )
             THISVERSION=${THISVERSION%.dirty*}
             THISVERSION=${CIRCLE_TAG:-$THISVERSION}
@@ -1029,7 +1029,7 @@ jobs:
             pip install --upgrade "pip>=19.1" numpy
             echo "${CIRCLE_TAG:-$THISVERSION}" > fmriprep/VERSION
             python setup.py sdist
-            pip wheel --no-deps -w dist/ .
+            python -m pip wheel --no-deps -w dist/ .
       - store_artifacts:
           path: /tmp/src/fmriprep/dist
       - run:
@@ -1038,11 +1038,11 @@ jobs:
             THISVERSION=$( python get_version.py )
             THISVERSION=${THISVERSION%.dirty*}
             THISVERSION=${CIRCLE_TAG:-$THISVERSION}
-            twine check dist/fmriprep*.tar.gz
+            python -m twine check dist/fmriprep*.tar.gz
             virtualenv --python=python sdist
             source sdist/bin/activate
-            pip install --upgrade "pip>=19.1" numpy
-            pip install dist/fmriprep*.tar.gz
+            python -m pip install --upgrade "pip>=19.1" numpy
+            python -m pip install dist/fmriprep*.tar.gz
             which fmriprep | grep sdist\\/bin
             INSTALLED_VERSION=$(fmriprep --version)
             INSTALLED_VERSION=${INSTALLED_VERSION%$'\r'}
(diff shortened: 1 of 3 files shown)


oesteban (approved):
I think this is a good idea in preparation of multistaged builds. also nice to clean up a bit.

Please rebase with master.

scripts/fetch_templates.py
```diff
@@ -0,0 +1,114 @@
...
+import templateflow.api as tf
+
+
+def fetch_MNI2009(validate=True):
+    # T1w
+    tf.get(
+        'MNI152NLin2009cAsym',
+        atlas=None,
+        resolution=(1, 2),
+        extension='.nii.gz'
```
oesteban, line 13: I think we can get rid of the extension selector, if you look at the just merged #2268 changes.

oesteban (approved):

oesteban:
Let's keep this for the 20.3.x series.

oesteban:
Please rebase :)

oesteban (approved):

Dockerfile
```diff
@@ -79,37 +79,34 @@ RUN apt-get update && \
...
-
-# Installing SVGO
-RUN curl -sL https://deb.nodesource.com/setup_10.x | bash -
-RUN apt-get install -y nodejs
-RUN npm install -g svgo
 
-# Installing bids-validator
-RUN npm install -g bids-validator@1.4.0
+# Installing SVGO and bids-validator
+RUN npm install -g svgo bids-validator@1.4.0 \
```
oesteban, line 104:
```suggestion
RUN npm install -g svgo bids-validator@1.5.4 \
```

Dockerfile
```diff
@@ -142,10 +139,8 @@ RUN conda install -y python=3.7.1 \
                      graphviz=2.40.1 \
                      traits=4.6.0 \
                      zlib; sync && \
-    chmod -R a+rX /usr/local/miniconda; sync && \
-    chmod +x /usr/local/miniconda/bin/*; sync && \
```
oesteban, line 146: Are we sure this glitch has been fixed?

.circleci/config.yml
```diff
@@ -40,14 +40,14 @@ jobs:
...
+                docker tag localhost:5000/ubuntu ubuntu:xenial-20200706
                 docker pull localhost:5000/fmriprep
                 docker tag localhost:5000/fmriprep nipreps/fmriprep:latest
                 docker tag localhost:5000/fmriprep nipreps/fmriprep
             else
                 echo "Pulling from Docker Hub"
-                docker pull ubuntu:xenial-20200114
-                docker tag ubuntu:xenial-20200114 localhost:5000/ubuntu
+                docker pull ubuntu:xenial-20200706
+                docker tag ubuntu:xenial-xenial-20200706 localhost:5000/ubuntu
```
oesteban, line 50:
```suggestion
                docker tag ubuntu:xenial-20200706 localhost:5000/ubuntu
```

Dockerfile
mgxd: our test datasets are passing on Circle - I don't think this is still necessary.

oesteban: yes, but this used to be the case and people would experience linking issues with singularity... not sure it is worth removing.

oesteban: Actually, it would be best to remove in a separate PR, so that it is clear within one atomic change.

oesteban: I think this is the only outstanding issue before merging this PR, correct?

Dockerfile
```diff
@@ -79,37 +79,34 @@ RUN apt-get update && \
...
-
-# Installing SVGO
-RUN curl -sL https://deb.nodesource.com/setup_10.x | bash -
-RUN apt-get install -y nodejs
-RUN npm install -g svgo
 
-# Installing bids-validator
-RUN npm install -g bids-validator@1.4.0
+# Installing SVGO and bids-validator
+RUN npm install -g svgo bids-validator@1.5.4 \
```
effigies, line 104:
```suggestion
RUN npm install -g svgo bids-validator@1.5.7 \
```

Dockerfile
```diff
@@ -189,18 +176,14 @@ RUN echo "${VERSION}" > /src/fmriprep/fmriprep/VERSION && \
     echo "include fmriprep/VERSION" >> /src/fmriprep/MANIFEST.in && \
     pip install --no-cache-dir "/src/fmriprep[all]"
 
-RUN install -m 0755 \
-    /src/fmriprep/scripts/generate_reference_mask.py \
-    /usr/local/bin/generate_reference_mask
```
effigies, line 194: Are we making this available in the container in another way? This has in the past been useful for helping debug the reference mask generation when it goes badly.

mgxd: wasn't this ported over to niworkflows already? https://github.com/nipreps/niworkflows/blob/3120884336dd9c3b6211eeb538ed5f2d1085ac78/setup.cfg#L52-L53

effigies: Oh, maybe. Have we tested that? It looks like it might not work, since there's no data sink.

https://github.com/nipreps/niworkflows/blob/master/niworkflows/cli/boldref.py

mgxd: the files should be available (albeit hidden) since we are setting the workflow base to the working directory. I opened up https://github.com/nipreps/niworkflows/issues/579 to address this concern.

In the meantime, I feel pretty comfortable removing this redundancy given that the tools are already in place should someone really need to use this under-the-radar helper script.

effigies: Sure. I think the issue is going to be with the default of `remove_unnecessary_outputs=True` in the nipype config. But I agree that we can move this discussion over to niworkflows.

effigies:
Rebase on `master`?

mgxd:
Hmmm, documenting the circle error here:

> pkg_resources.ContextualVersionConflict: (importlib-metadata 0.23 (/usr/local/lib/python3.7/site-packages), Requirement.parse('importlib-metadata>=1; python_version < "3.8"'), {'keyring'})

Not sure why this is now popping up/not handled by `pip`, but adding `importlib-metadata` here might do the trick
https://github.com/nipreps/fmriprep/blob/a42d9290d24298e7bce8bd8bef81ba8a147d04cc/.circleci/config.yml#L1029

effigies:
Yeah, I resolved that somewhere recently by just using 3.8. If we're not planning on doing this for LTS, maybe we should just bump to 3.8?

mgxd:
Wouldn't that also break the conflict?
> Requirement.parse('importlib-metadata>=1; python_version < "3.8"'), {'keyring'})

Looking at the [source code](https://github.com/python/importlib_metadata#importlib_metadata), it states it's intended to be ported to 3.8

effigies:
That's a constraint. It means "if python_version < 3.8, install importlib-metadata>=1". `importlib-metadata` is a backport of Python 3.8's stdlib `importlib.metadata`. (It may have started the other way, with intent to include in 3.8, but now it functions as a backport.)

effigies:
Here's the previous fix, btw: https://github.com/nipreps/smriprep/pull/234

effigies (approved):

mgxd:
let's get this in before it gets stale again

Issue #2289, opened 2020-09-25
Migrate to nipreps org


oesteban:
After the release of the LTS.

oesteban:
Some webhooks need to be revised.

oesteban:
We can cut a bug-fix or post-release with nipreps/niworkflows#573 to assess deployment

PR #2290, opened 2020-09-29
MAINT: Migrate from poldracklab to nipreps orgs

oesteban:
Resolves: #2289.

### DIFF
--- a/docs/docker.rst
+++ b/docs/docker.rst
@@ -14,7 +14,7 @@ In order to run *fMRIPrep* Docker images, the Docker Engine must be installed
 If you used *fMRIPrep* via Docker in the past, you might need to pull down a
 more recent version of the image: ::
 
-    $ docker pull poldracklab/fmriprep:<latest-version>
+    $ docker pull nipreps/fmriprep:<latest-version>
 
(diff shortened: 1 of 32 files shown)


docs/docker.rst
```diff
@@ -31,7 +31,7 @@ needed, e.g.::
 
     $ fmriprep-docker /path/to/data/dir /path/to/output/dir participant
     RUNNING: docker run --rm -it -v /path/to/data/dir:/data:ro \
-        -v /path/to_output/dir:/out poldracklab/fmriprep:1.0.0 \
+        -v /path/to_output/dir:/out nipreps/fmriprep:1.0.0 \
```
effigies, line 34:
```suggestion
        -v /path/to_output/dir:/out nipreps/fmriprep:20.2.0 \
```

effigies (commented):
Looks reasonable. I think we should probably push all 20.x.y tags to `nipreps/fmriprep` on DockerHub, to avoid having to say "well, for this version, use poldracklab/fmriprep".

oesteban:
> I think we should probably push all 20.x.y tags to `nipreps/fmriprep` on DockerHub, to avoid having to say "well, for this version, use poldracklab/fmriprep".

Currently doing this.

PR #2302, opened 2020-10-09
FIX: Clarify phase encoding direction, rather than axis

mgxd:
BOLD runs' `FunctionalSummary` nodes were returning PE-Axis instead of PE-Direction.

### DIFF
--- a/fmriprep/interfaces/reports.py
+++ b/fmriprep/interfaces/reports.py
@@ -194,10 +194,12 @@ class FunctionalSummary(SummaryInterface):
                 '(boundary-based registration, BBR) - %d dof' % dof,
                 'FreeSurfer <code>mri_coreg</code> - %d dof' % dof],
         }[self.inputs.registration][self.inputs.fallback]
-        if self.inputs.pe_direction is None:
-            pedir = 'MISSING - Assuming Anterior-Posterior'
-        else:
-            pedir = {'i': 'Left-Right', 'j': 'Anterior-Posterior'}[self.inputs.pe_direction[0]]
+        pedir = {
+            'i-': 'Left-Right',
+            'i': 'Right-Left',
+            'j-': 'Anterior-Posterior',
+            'j': 'Posterior-Anterior'
+        }.get(self.inputs.pe_direction, 'MISSING - Assuming Anterior-Posterior')
 
         if isdefined(self.inputs.confounds_file):
             with open(self.inputs.confounds_file) as cfh:
--- a/fmriprep/interfaces/reports.py
+++ b/fmriprep/interfaces/reports.py
@@ -33,6 +33,7 @@ FUNCTIONAL_TEMPLATE = """\
 \t\t<details open>
 \t\t<summary>Summary</summary>
 \t\t<ul class="elem-desc">
+\t\t\t<li>Original orientation: {ornt}</li>
 \t\t\t<li>Repetition time (TR): {tr:.03g}s</li>
 \t\t\t<li>Phase-encoding (PE) direction: {pedir}</li>
 \t\t\t<li>{multiecho}</li>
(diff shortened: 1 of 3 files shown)


fmriprep/interfaces/reports.py
```diff
@@ -194,10 +194,12 @@ def _generate_segment(self):
...
-        if self.inputs.pe_direction is None:
-            pedir = 'MISSING - Assuming Anterior-Posterior'
-        else:
-            pedir = {'i': 'Left-Right', 'j': 'Anterior-Posterior'}[self.inputs.pe_direction[0]]
+        pedir = {
+            'i-': 'Left-Right',
+            'i': 'Right-Left',
+            'j-': 'Anterior-Posterior',
+            'j': 'Posterior-Anterior'
+        }.get(self.inputs.pe_direction, 'MISSING - Assuming Anterior-Posterior')
```
effigies, line 202: This is only true for LAS images. We need the orientation of the original image, to be certain.

mgxd: Good point - I've added a commit to consider LAS/RAS. How extensive do you think our support should be for different orientations?

effigies:
We can get images in any orientation. Need something like:

```Python
def get_world_pedir(img, pe_direction):
    axes = (("Right", "Left"),
            ("Anterior", "Posterior"),
            ("Superior", "Inferior"))
    ax_idcs = {"i": 0, "j": 1, "k": 2}

    ornt = ''.join(nb.aff2axcodes(img.affine))
    axcode = ornt[ax_idcs[pe_direction[0]]]
    inv = pe_direction[1:] == "-"

    for ax in axes:
        for flip in (ax, ax[::-1]):
            if flip[not inv].startswith(axcode):
                return "-".join(flip)
    raise ValueError(f"Cannot determine world direction of phase encoding. Orientation: {ornt}; PE dir: {pe_direction}")
```

effigies:
Incidentally, for testing, I've recently written this, which allows you to generate an image of any orientation:

```Python
def generate_image(params):
    dataobj = np.zeros(params["shape"], params["dtype"]) 
    affine = np.eye(4) 
    rot = ornt.inv_ornt_aff(ornt.axcodes2ornt(params["orientation"]), (1, 1, 1)) 
    affine[:3, :3] = np.diag(params["zooms"][:3]) @ np.linalg.inv(rot[:3, :3]) 
    return nb.Nifti1Image(dataobj, affine)
```

(It's intended for generating datasets on the fly, which is why it's so parameterized.)

fmriprep/interfaces/tests/test_reports.py
```diff
@@ -0,0 +1,35 @@
...
+    ('LAS', 'j', 'Posterior-Anterior'),
+    ('LAS', 'i-', 'Left-Right'),
+    ('LAS', 'k-', 'Superior-Inferior'),
+    ('LPI', 'j', 'Anterior-Posterior'),
+    ('LPI', 'i-', 'Left-Right'),
+    ('LPI', 'k-', 'Inferior-Superior'),
+])
+def test_get_world_pedir(tmpdir, orientation, pe_dir, expected):
+    img = generate_img(orientation)
+    assert orientation == ''.join(nb.aff2axcodes(img.affine))
```
effigies, line 33: I guess we don't really need `generate_img` if we're using the orientation and not the image in `get_world_pedir`.

effigies (approved):

fmriprep/interfaces/reports.py
```diff
@@ -194,10 +199,8 @@ def _generate_segment(self):
                 '(boundary-based registration, BBR) - %d dof' % dof,
                 'FreeSurfer <code>mri_coreg</code> - %d dof' % dof],
         }[self.inputs.registration][self.inputs.fallback]
-        if self.inputs.pe_direction is None:
-            pedir = 'MISSING - Assuming Anterior-Posterior'
-        else:
-            pedir = {'i': 'Left-Right', 'j': 'Anterior-Posterior'}[self.inputs.pe_direction[0]]
+
+        pedir = get_world_pedir(self.inputs.pe_direction, self.inputs.orientation)
```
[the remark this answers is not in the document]
effigies: Seems to have gotten flipped again.

```suggestion
        pedir = get_world_pedir(self.inputs.orientation, self.inputs.pe_direction)
```

PR #2303, opened 2020-10-09
ENH: Add --output-layout CLI option to enable BIDS (YODA) mode

effigies:

This PR introduces a `--output-layout` argument that can take the values "legacy" or "yoda".

Legacy layout:

```
output_dir/
    fmriprep/   # BIDS Derivative
    freesurfer/ # FreeSurfer
```

YODA layout:

```
output_dir/         # BIDS Derivative
    sourcedata/
        freesurfer/ # FreeSurfer
```

The following command produces the following directory structure:

```
fmriprep sourcedata/rawdata . participant --output-layout yoda
```

```
$PWD/
    sourcedata/
        rawdata/     # Input
        freesurfer/  # FreeSurfer outputs (or pre-run)
    sub-01/
    ...
    dataset_description.json
```

The advantage of this approach is that one can construct the datalad dataset:

```
datalad create -c text2git ds00WXYZ-fmriprep
cd ds00WXYZ-fmriprep
mkdir sourcedata/ code/
datalad install -d . -s ///containers code/containers
datalad install -d . -s ///openneuro/ds00WXYZ sourcedata/rawdata
datalad install -d . -s /path/to/prerun/ds00WXYZ-freesurfer sourcedata/freesurfer
```

And turn it into a valid BIDS Derivatives dataset by running

```
datalad containers-run -n code/containers/bids-fmriprep \
    sourcedata/rawdata . participant --output-layout yoda
```

### Discussion

To do this, I've created two new config variables: `output-layout` and `fmriprep_dir`, where the latter is determined by the `layout`. Places where we've been using `output_dir / "fmriprep"` are now `fmriprep_dir`. This breaks some assumptions of `DerivativesDataSink`, but this can be overcome by simply setting the `out_path_base` to `""`.

Should we call it something other than "yoda"? That's not terribly informative to people not steeped in Datalad lore. In the long run I'd like to make it the default.


effigies:
cc @yarikoptic @franklin-feingold @jbwexler for comments on the proposed structure.

yarikoptic:
Note: text2git might not be good enough - some outputs iirc are huge text xml files and need a dedicated setting to get them to annex. 
I like yoda, but it also could be "bids" ;-)

effigies:
Maybe nested? Would rather not imply that the standard layout is not BIDS.

effigies:
@mgxd I want to add the option in 20.2.1, but we can't change the default. I'm okay with switching the default in 20.3.

mgxd:
Ah - i saw this was targeting master so I assumed 20.3.0. If we go the backporting route, I'm strongly in favor of an envvar switch rather than a new CLI option so we aren't violating our bug-fix release principles.

effigies:
I really don't like using environment variables for such drastic changes. It makes the provenance messy. I thought a flag that did not change anything by default would be safe by our rules, but I admit I did not check.

My reasoning here is that this would be valuable to LTS users who want to use datalad to manage dataset level provenance. Large projects are likely to settle on the LTS and benefit from datalad.

So I'm engaging in some motivated reasoning here, but I think it might be worth bending that rule, or adjusting it to permit cases like this in the future.

yarikoptic:
> Maybe nested? Would rather not imply that the standard layout is not BIDS.

but is it? ;)  
I think there is nothing wrong with acknowledging the truth. If anything, the current/original layout is something like "nested-bids" (bids nested in a directory + non-bids freesurfer), so IMHO "original" (or "nested-bids" or "nested-bids+freesurfer") and "bids" make sense to me and would be more factually correct ;)

effigies:
@bpinsard @mgxd @oesteban We should decide on whether a new CLI option can be put into the LTS.

effigies:
From meeting, fine to add CLI option in LTS. Will rebase on `maint/20.2.x` with fixes for CI-identified issues.

Following suggestions from @yarikoptic, will switch to "bids" and "legacy" modes, unless anybody has better ideas or strong opinions.

bpinsard:
(Sorry, I missed the notification.) Adding that data management option to the LTS while keeping default is good. Thanks!

.circleci/config.yml
```diff
@@ -439,6 +440,40 @@ jobs:
...
+                --nthreads 2 -vv
+      - run:
+          name: Checking outputs of partial fMRIPrep run
+          command: |
+            mkdir -p /tmp/${DATASET}/test
+            CHECK_OUTPUTS_FILE="${DATASET}_yoda_fasttrack_outputs.txt"
+            if [ -f /tmp/.nofasttrack ]; then
+                CHECK_OUTPUTS_FILE="${DATASET}_yoda_outputs.txt"
+            fi
+            find /tmp/${DATASET}/yoda \( -path */figures -o -path */log -path */sourcedata \) -prune -o -name "*" -print | sed s+/tmp/${DATASET}/++ | sort > /tmp/${DATASET}/test/outputs.out
```
effigies, line 474: @oesteban Since you wrote the original `find` command, perhaps you know why this isn't ignoring `sourcedata`?

effigies:
Love too have 2 hour feedback loops on the CI.

.circleci/config.yml
```diff
@@ -439,6 +440,40 @@ jobs:
...
+                --nthreads 2 -vv
+      - run:
+          name: Checking outputs of partial fMRIPrep run
+          command: |
+            mkdir -p /tmp/${DATASET}/test
+            CHECK_OUTPUTS_FILE="${DATASET}_yoda_fasttrack_outputs.txt"
+            if [ -f /tmp/.nofasttrack ]; then
+                CHECK_OUTPUTS_FILE="${DATASET}_yoda_outputs.txt"
+            fi
+            find /tmp/${DATASET}/yoda \( -path */figures -o -path */log -o -path */sourcedata \) -prune -o -name "*" -print | sed s+/tmp/${DATASET}/++ | sort > /tmp/${DATASET}/test/outputs.out
```
effigies, line 474: I reran with SSH to verify that this should work, and it does in bash. I don't understand why we're seeing failures. Any help would be appreciated.

effigies:
Good to merge?

fmriprep/cli/parser.py
```diff
@@ -685,10 +704,9 @@ def parse_args(args=None, namespace=None):
         )
 
     # Setup directories
-    config.execution.log_dir = output_dir / "fmriprep" / "logs"
+    config.execution.log_dir = config.execution.fmriprep_dir / "logs"
     # Check and create output and working directories
     config.execution.log_dir.mkdir(exist_ok=True, parents=True)
-    output_dir.mkdir(exist_ok=True, parents=True)
```
oesteban, line 691:
```suggestion
    config.execution.fmriprep_dir.mkdir(exist_ok=True, parents=True)
```
?

fmriprep/workflows/base.py
```diff
@@ -285,7 +285,7 @@ def init_single_subject_wf(subject_id):
     # Overwrite ``out_path_base`` of smriprep's DataSinks
     for node in workflow.list_node_names():
         if node.split('.')[-1].startswith('ds_'):
-            workflow.get_node(node).interface.out_path_base = 'fmriprep'
+            workflow.get_node(node).interface.out_path_base = ""
```
oesteban, line 288: I believe the whole loop can go away.

fmriprep/workflows/bold/base.py
```diff
@@ -573,12 +573,12 @@ def init_func_preproc_wf(bold_file):
         # And ensure echo is dropped from report
         for node in fmap_unwarp_report_wf.list_node_names():
             if node.split('.')[-1].startswith('ds_'):
-                fmap_unwarp_report_wf.get_node(node).interface.out_path_base = 'fmriprep'
+                fmap_unwarp_report_wf.get_node(node).interface.out_path_base = ""
                 fmap_unwarp_report_wf.get_node(node).inputs.dismiss_entities = ("echo",)
```
oesteban, line 577: I believe these loops can go away (setting `out_path_base` bc of the class modification above, and the `dismiss_entities` because of your PR to drop repeated entities).

oesteban (approved):
Minor nitpicks. LGTM overall

.circleci/config.yml
oesteban: I think you were missing an `-o`:

```suggestion
            find /tmp/${DATASET}/yoda \( -path */figures -o -path */log -o -path */sourcedata \) -prune -o -name "*" -print | sed s+/tmp/${DATASET}/++ | sort > /tmp/${DATASET}/test/outputs.out
```

.circleci/config.yml
oesteban: I see you caught the missing `-o`, sorry I couldn't have a look in time.

fmriprep/reports/core.py
```diff
@@ -0,0 +1,110 @@
...
+        from yaml import safe_load as load
+
+        settings = load(config.read_text())
+        self.packagename = self.packagename or settings.get("package", None)
+
+        # Removed from here: Appending self.packagename to self.root and self.out_dir
+        # In this version, pass reportlets_dir and out_dir with fmriprep in the path.
+
+        if self.subject_id is not None:
+            self.root = self.root / "sub-{}".format(self.subject_id)
```
mgxd, line 20:
```suggestion
            self.root = self.root / f"sub-{self.subject_id}"
```

setup.cfg
```diff
@@ -104,7 +105,7 @@ parentdir_prefix =
 
 [flake8]
 max-line-length = 99
-doctests = True
+doctests = False
```
mgxd, line 108: why has this changed?

mgxd (approved):
LGTM

setup.cfg
effigies: Because flake8 barfs on doctests that use variables that were provided by the `doctest_namespace` fixture, and I didn't want to rewrite the doctests that I copied from niworkflows.

fmriprep/workflows/base.py
effigies: I don't think so. These nodes will have another default `out_path_base`.

fmriprep/workflows/bold/base.py
effigies: These are coming from `sdcflows`, not fMRIPrep.

.circleci/config.yml
effigies: Well, it still didn't work. :-)

fmriprep/cli/parser.py
effigies: `log_dir` is a subdirectory of `fmriprep_dir`.

fmriprep/cli/parser.py
```diff
@@ -642,9 +652,18 @@ def parse_args(args=None, namespace=None):
     output_dir = config.execution.output_dir
     work_dir = config.execution.work_dir
     version = config.environment.version
+    output_layout = config.execution.output_layout
 
     if config.execution.fs_subjects_dir is None:
-        config.execution.fs_subjects_dir = output_dir / "freesurfer"
+        if output_layout == "bids":
```
yarikoptic, line 658: good -- you went for `"bids"` (not `"yoda"`), but then why all the examples articulate "YODA" and "yoda/"?   might confuse someone?

effigies:
Okay, fully de-yodafied. Everybody happy with a merge on green?

effigies:
Let's do it.

PR #2316, opened 2020-10-26
ENH: Add Docker authentication to increase pull rate limit

### LINKED ISSUE #2311: [Action Required] How to avoid Docker Hub rate limits for your CircleCI builds
Just forwarding an email from CircleCI:

> Hi there,
> 
> On November 1st, Docker Hub will begin limiting anonymous image pulls. We want to make sure you know how you might be impacted and what you can do to avoid interruptions to your workflow.
> 
> Adding Docker authentication to your pipeline config is the easiest way to avoid any service disruptions. If you use the Docker executor or pull Docker images when using the machine executor on CircleCI, we encourage you to authenticate. Because the anonymous API rate limits are based on IP addresses, they will impact CircleCI cloud customers. Authenticated users get higher per-user rate limits, regardless of IP.
> 
> We are currently working on a partnership with Docker to minimize the impact of this change for our users and will share more details as we get them.
> 
> For more information or to leave a question for us, please head over to Discuss or contact support.
> 
> Thanks and happy building,
> 
> \- The CircleCI team

We can either start using GH Packages and pull from their registry (probably we'll enjoy more bandwidth even now) or add the login.

mgxd:
Partially? closes #2311 

https://circleci.com/docs/2.0/private-images/

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,6 +25,7 @@ jobs:
             - build-v4-
           paths:
             - /tmp/docker
+            - /tmp/images
       - run:
           name: Docker authentication
           command: |
@@ -34,6 +35,14 @@ jobs:
       - run:
           name: Set up Docker registry
           command: |
+            if [[ -f /tmp/images/registry.tar.gz ]]; then
+              echo "Loading saved registry image"
+              docker load < /tmp/images/registry.tar.gz
+            else
+              echo "Pulling registry image from DockerHub"
+              docker pull registry:2
+              docker save registry:2 | gzip > /tmp/images/registry.tar.gz
+            fi
             docker run -d -p 5000:5000 --restart=always --name=registry \
                 -v /tmp/docker:/var/lib/registry registry:2
       - run:
@@ -104,6 +113,7 @@ jobs:
          key: build-v4-{{ .Branch }}-{{ epoch }}
          paths:
             - /tmp/docker
+            - /tmp/images
 
   get_data:
     machine:
@@ -251,6 +261,7 @@ jobs:
             - build-v4-
           paths:
             - /tmp/docker
+            - /tmp/images
       - restore_cache:
           keys:
             - data-v8-{{ .Revision }}
@@ -263,6 +274,13 @@ jobs:
       - run:
           name: Set-up a Docker registry
           command: |
+            if [[ -f /tmp/images/registry.tar.gz ]]; then
+              echo "Loading saved registry image"
+              docker load < /tmp/images/registry.tar.gz
+            else
+              echo "Pulling registry image from DockerHub"
+              docker pull registry:2
+            fi
             docker run -d -p 5000:5000 --restart=always --name=registry \
                 -v /tmp/docker:/var/lib/registry registry:2
       - run:
@@ -344,6 +362,7 @@ jobs:
             - build-v4-
           paths:
             - /tmp/docker
+            - /tmp/images
       - restore_cache:
           keys:
             - data-v8-{{ .Branch }}-{{ .Revision }}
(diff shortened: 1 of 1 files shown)


effigies:
Looks like this is going to be a problem with PRs, since we would be exposing credentials to any fork.

mgxd:
Good point - this has been [brought up](https://discuss.circleci.com/t/updated-authenticate-with-docker-to-avoid-impact-of-nov-1st-rate-limits/37567/49) in the circle discussion forum, but no workaround for the `machine` executor has been proposed yet.

effigies:
We can still wrap everything into a `if [[ -n "$DOCKER_PASS" ]]` block. That will boost speeds on the main repository's builds, and then PRs can use the caches, and only rarely hit Docker Hub.

mgxd:
Yup, though my worry is PR builds could hit the rate limit (I believe 100 pulls/day). I don't expect us to normally hit that cap, but something to monitor.

effigies:
Is that a serious concern, though? Most of our pulls are from the local registry, which is loaded from a cache. As long as `master` is building those caches regularly, there's little danger of even 1 pull per day from PRs.

mgxd:
I was under the impression every "Set-up Docker registry" step in our build is pulling the `registry` image - i.e. https://app.circleci.com/pipelines/github/nipreps/fmriprep/33/workflows/71306b88-b53e-4a3e-bef8-070d71ad5700/jobs/222/parallel-runs/0/steps/0-107

effigies:
Oh, good call. Maybe we can tarball that image.

.circleci/config.yml
```diff
@@ -26,9 +25,25 @@ jobs:
             - build-v4-
           paths:
             - /tmp/docker
+            - /tmp/images
```
effigies, line 28: Is it possible we'll want to clear the `registry:2` cache independent of the build cache?

effigies (approved):
LGTM. I think we could separate out the `registry:2` cache, but I don't think it's critical to do so.

.circleci/config.yml
mgxd: I think we'll update that image so infrequently that it's not really worth it to make a new cache 🤷‍♂️

effigies:
Let's do it. Can always fix, and it'd be good to confirm that PRs are avoiding hitting docker hub.

effigies:
... can't merge from my phone. Feel free to merge.

oesteban:
I have merged because I think this is good in the long term. However it seems CircleCI has partnered with Docker Hub to avoid exactly this problem - https://discuss.circleci.com/t/updated-authenticate-with-docker-to-avoid-impact-of-nov-1st-rate-limits/37567?mkt_tok=eyJpIjoiTVRsaU1HWmtNRE16TkRZdyIsInQiOiJWcVBBMFJZdlR3TFhSVmxYejJtbDJKa2dqY3ZoY1k0TVl1eUs4d3Vpc2dqOG1ZSTNibHVGd1pTMXFJeGhlTkdzcnJQRWtrcWIwZjc0N1BEVUlPSFNNXC9qdVFxOFpcLzZhUGFHS0x1RHNLVXNGak1id25WaUVcL3FiYktIWlpqcTgrOHNGdDRJaldKdlUrQWNIVmVJWVUxdUE9PSJ9

I just got the email, literally.

mgxd:
@oesteban I think the limit still applies to `machine` executors, which is what we're using (hopefully to take advantage of docker layer caching down the line).

https://discuss.circleci.com/t/updated-authenticate-with-docker-to-avoid-impact-of-nov-1st-rate-limits/37567/52
https://discuss.circleci.com/t/updated-authenticate-with-docker-to-avoid-impact-of-nov-1st-rate-limits/37567/58?u=mgxd

PR #2331, opened 2020-11-18
FIX: non-existing path or JSON syntax error for `--bids-filter-file` should raise on error.

bpinsard:

Check if provided bids-filter-file path exists if not raise an error.
This should avoid cases when the user provides file filters and the pipeline silently runs on other set of file. 

none

### DIFF
--- a/fmriprep/cli/parser.py
+++ b/fmriprep/cli/parser.py
@@ -82,6 +82,7 @@ def _build_parser():
     PathExists = partial(_path_exists, parser=parser)
     IsFile = partial(_is_file, parser=parser)
     PositiveInt = partial(_min_one, parser=parser)
+    BIDSFilter = partial(_bids_filter, parser=_bids_filter)
 
     # Arguments as specified by BIDS-Apps
     # required, positional arguments
@@ -145,7 +146,7 @@ def _build_parser():
         "--bids-filter-file",
         dest="bids_filters",
         action="store",
-        type=_bids_filter,
+        type=BIDSFilter,
         metavar="FILE",
         help="a JSON file describing custom BIDS input filters using PyBIDS. "
         "For further details, please check out "
--- a/fmriprep/cli/parser.py
+++ b/fmriprep/cli/parser.py
@@ -82,7 +82,7 @@ def _build_parser():
     PathExists = partial(_path_exists, parser=parser)
     IsFile = partial(_is_file, parser=parser)
     PositiveInt = partial(_min_one, parser=parser)
-    BIDSFilter = partial(_bids_filter, parser=_bids_filter)
+    BIDSFilter = partial(_bids_filter, parser=parser)
 
     # Arguments as specified by BIDS-Apps
     # required, positional arguments
--- a/fmriprep/cli/parser.py
+++ b/fmriprep/cli/parser.py
@@ -55,11 +55,17 @@ def _build_parser():
             for k, v in dct.items()
         }
 
-    def _bids_filter(value):
-        from json import loads
+    def _bids_filter(value, parser):
+        from json import loads, JSONDecodeError
 
-        if value and Path(value).exists():
-            return loads(Path(value).read_text(), object_hook=_filter_pybids_none_any)
+        if value:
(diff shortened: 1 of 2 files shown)


fmriprep/cli/parser.py
```diff
@@ -58,8 +58,11 @@ def _filter_pybids_none_any(dct):
     def _bids_filter(value):
         from json import loads
 
-        if value and Path(value).exists():
-            return loads(Path(value).read_text(), object_hook=_filter_pybids_none_any)
+        if value:
+            if Path(value).exists():
+                return loads(Path(value).read_text(), object_hook=_filter_pybids_none_any)
+            else:
+                raise parser.error(f"Path does not exist: <{value}>.")
```
mgxd, line 65: maybe a `FileNotFoundError` would be more helpful here?
```suggestion
            raise FileNotFoundError(f"Path does not exist: <{value}>.")
```

mgxd (approved):
LGTM - this should probably be backported to the LTS as well. It should be a relatively clean process if you change the target branch to `maint/20.2.x`

fmriprep/cli/parser.py
bpinsard: I followed what is done elsewhere in the parser: 
https://github.com/nipreps/fmriprep/blob/962f0b2e26d522d87ed786f12766001aa65e9dfd/fmriprep/cli/parser.py#L20-L24
but I have no problem changing it.

mgxd: sure that works too, though if you are following that example the function should be converted into a partial:

https://github.com/nipreps/fmriprep/blob/c707938e1462e77b1c3a1216a855ceb0aba70845/fmriprep/cli/parser.py#L76-L78

bpinsard: oh right, my mistake. Would there be any advantage of making it a partial for that specific option?

mgxd: i think its necessary to pass in the defined parser

effigies:
Should this target `maint/20.2.x`?

fmriprep/cli/parser.py
```diff
@@ -55,11 +55,14 @@ def _filter_pybids_none_any(dct):
...
+    def _bids_filter(value, parser):
         from json import loads
 
-        if value and Path(value).exists():
-            return loads(Path(value).read_text(), object_hook=_filter_pybids_none_any)
+        if value:
+            if Path(value).exists():
+                return loads(Path(value).read_text(), object_hook=_filter_pybids_none_any)
+            else:
+                raise parser.error(f"Path does not exist: <{value}>.")
```
effigies, line 65: Let's follow the lead of `_is_file` and delegate the basic check while adding a more informative error for bad JSON:

```suggestion
        path = _is_file(value, parser)
        try:
            return loads(path.read_text(), object_hook=_filter_pybids_none_any)
        except JSONDecodeError:
            parser.error(f"Invalid JSON file: <{value}>")
```

fmriprep/cli/parser.py
```diff
@@ -55,11 +55,14 @@ def _filter_pybids_none_any(dct):
             for k, v in dct.items()
         }
 
-    def _bids_filter(value):
+    def _bids_filter(value, parser):
         from json import loads
```
effigies, line 59:
```suggestion
        from json import loads, JSONDecodeError
```

mgxd:
@bpinsard I think the CI errors have been fixed, you'll need to rebase off the `maint/20.2.x` branch

effigies:
Rebase again?

fmriprep/cli/parser.py
effigies: Have we resolved this discussion? Would it be worth adding a test to make sure that it works, or are there tests in the CI that are covering the necessary cases?

mgxd: I think the partial call got lost in the latest rebase

effigies:
@bpinsard I added a test and fixed a bug. This LGTM at this point, feel free to merge when tests pass.

PR #2349, opened 2020-12-17
ENH: Mask multi-echo data with reference mask instead of echo-specific masks

### LINKED ISSUE #2348: Skull-strip first echo and apply mask to other echoes
For multi-echo EPI data, fMRIPrep currently skull-strips all echoes independently and then feeds those skull-stripped time series into the t2smap workflow for optimal combination. However, I believe that it would be preferable to skull-strip just the first echo, retain the brain mask from that echo, and then apply the mask to all echoes instead.

This stems from @emdupre's idea in [this Mattermost post](https://mattermost.brainhack.org/brainhack/pl/1ifqy75c3pfcmd778c681ftgtc).

tsalo:
Closes #2348.

- Instead of skull-stripping each echo-specific output of `bold_bold_trans_wf` before calling `bold_t2s_wf`, use the reference image-based mask from `bold_bold_trans_wf `.
- Add `bold_mask` input to `init_bold_t2s_wf`.
- Add `mask_file` input to `T2SMap` interface.

None


fmriprep/workflows/bold/t2s.py
```diff
@@ -79,7 +81,7 @@ def init_bold_t2s_wf(echo_times, mem_gb, omp_nthreads,
     t2smap_node = pe.Node(T2SMap(echo_times=list(echo_times)), name='t2smap_node')
 
     workflow.connect([
-        (inputnode, t2smap_node, [('bold_file', 'in_files')]),
+        (inputnode, t2smap_node, [('bold_file', 'in_files'), ('bold_mask', 'mask_file')]),
```
oesteban, line 84:
```suggestion
        (inputnode, t2smap_node, [('bold_file', 'in_files'),
                                  ('bold_mask', 'mask_file')]),
```

fmriprep/workflows/bold/base.py
```diff
@@ -539,12 +536,11 @@ def init_func_preproc_wf(bold_file):
                 ('outputnode.bold', 'bold_files')]),
             (join_echos, final_boldref_wf, [
                 ('bold_files', 'inputnode.bold_file')]),
-            (bold_bold_trans_wf, skullstrip_bold_wf, [
-                ('outputnode.bold', 'inputnode.in_file')]),
-            (skullstrip_bold_wf, join_echos, [
-                ('outputnode.skull_stripped_file', 'skullstripped_bold_files')]),
+            # use reference image mask used by bold_bold_trans_wf
+            (bold_bold_trans_wf, bold_t2s_wf, [
```
oesteban, line 540: I'd be careful to ensure that `bold_bold_trans_wf` does not depend on `bold_t2s_wf`. If this assumption is acceptable, then this seems the most reasonable solution to this problem.

tsalo: Is that because it would introduce a loop in the workflow? I agree that T2* mapping/optimal combination should happen after the native-space transforms are applied, mostly so that those transforms can be done all at once. There are solid reasons to do optimal combination after distortion correction, but I don't think that will have a very large effect on the data and I doubt that it's worth the extra effort since it would involve splitting up the transforms performed by `bold_bold_trans_wf` into two sets.

oesteban: > Is that because it would introduce a loop in the workflow?

Yes, Nipype will not handle loops.

tsalo:
The failures _seem_ like they're unrelated to the changes in this PR:

```python
_______________ ERROR collecting fmriprep/interfaces/__init__.py _______________
/src/fmriprep/fmriprep/interfaces/__init__.py:5: in <module>
    from niworkflows.interfaces import (
/usr/local/miniconda/lib/python3.7/site-packages/niworkflows/interfaces/freesurfer.py:10: in <module>
    from skimage import morphology as sim
/usr/local/miniconda/lib/python3.7/site-packages/skimage/__init__.py:121: in <module>
    from ._shared import geometry
skimage/_shared/geometry.pyx:1: in init skimage._shared.geometry
    ???
E   ValueError: numpy.ufunc size changed, may indicate binary incompatibility. Expected 216 from C header, got 192 from PyObject
```

oesteban:
@tsalo could you rebase to see if the errors have been fixed?

tsalo:
Done!

effigies:
@tsalo Quick bump to re-rebase.

tsalo:
@effigies done!

effigies (approved):
This looks reasonable to me, and the ds210 outputs look sensible.

effigies:
@oesteban @mgxd Any objection to sneaking this into 21.0?

mgxd (approved):

PR #2370, opened 2021-02-09
FIX: Ignore SBREF files if --ignore sbref is passed

### LINKED ISSUE #2369: --ignore sbref doesn't seem to work as expected
### What version of fMRIPrep are you using?

fMRIPrep version: 20.2.1 

### What kind of installation are you using? Containers (Singularity, Docker), or "bare-metal"?

singularity on HPC

### What is the exact command-line you used?

 ``` /usr/local/miniconda/bin/fmriprep --fs-license-file /work/freesurfer/license.txt /work/PhantomPain /work/PhantomPain/derivatives participant --participant-label testpilote3 -w /work/temp_data_PhantomPain --mem-mb 50000 --omp-nthreads 10 --nthreads 12 --ignore slicetiming sbref --fd-spike-threshold 0.5 --output-spaces MNI152NLin2009cAsym MNI152NLin6Asym --return-all-components```

### Have you checked that your inputs are BIDS valid?

yes, it does pass the bids-validator

### Did fMRIPrep generate the visual report for this particular subject? If yes, could you share it?

visual report here: https://www.dropbox.com/s/f8awapf73qxflue/export_fmriprep_report_ignore_sbref.zip?dl=0

### Can you find some traces of the error reported in the visual report (at the bottom) or in *crashfiles*?

nothing to declare

### Are you reusing previously computed results (e.g., FreeSurfer, Anatomical derivatives, work directory of previous run)?

I have been re-using FreeSurfer results but took care of removing the temporary files before launching a new execution.

Hi FMRIPREP experts,

I would like to report an observation on the behavior of fmriprep when passing ```--ignore sbref ``` to the command line:
- I was expecting this argument to actually ignore the sbref image when calculating the reference image which is used as a target for HMC. I don't think it is actually the case. Indeed, I am tracking down a misbehavior of mcflirt (see #2360 ) when using SBref as a target for realignment, which I think is maybe related to the difference of contrast between SBref images and the functional run. 
To check that, I relaunched fmriprep while having removed the sbref images from my dataset. Indeed the realignment results were quite different.
Later, I noticed that it is possible to use the ```--ignore sbref``` argument, which looked more convenient that having to remove the SBref images. However, launching fmriprep this way, I obtained the same realignment parameters than with the sbref case, even if the reference image was indeed calculated from the non-steady volumes in the end.

effigies:
We updated our message but not our actions.

Fixes #2369.

### DIFF
--- a/fmriprep/workflows/bold/base.py
+++ b/fmriprep/workflows/bold/base.py
@@ -200,6 +200,7 @@ def init_func_preproc_wf(bold_file):
     sbref_msg = f"No single-band-reference found for {os.path.basename(ref_file)}."
     if sbref_files and 'sbref' in config.workflow.ignore:
         sbref_msg = "Single-band reference file(s) found and ignored."
+        sbref_files = []
     elif sbref_files:
         sbref_msg = "Using single-band reference file(s) {}.".format(
             ','.join([os.path.basename(sbf) for sbf in sbref_files]))


julfou81:
This issue is solved as of fmriprep version 20.2.2. I could run run the option ```--ignore bref``` on my data set successfully using the singularity image.

PR #2380, opened 2021-03-04
PIN: NiPreps dependencies to development branches

mgxd:
Spawned after discussion with @oesteban in https://github.com/nipreps/fmriprep/pull/2372#issuecomment-790015350

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1146,9 +1146,9 @@ workflows:
       - test_deploy_pypi:
           filters:
             branches:
-              ignore:
-                - /docs\/.*/
-                - /docker\/.*/
+              only:
+                - /rel\/.*/
+                - /maint\/.*/
             tags:
               only: /.*/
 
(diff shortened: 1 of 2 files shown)


mgxd:
@oesteban was niworkflows 1.4.0rc5 released from a yet-to-be pushed branch?
https://github.com/nipreps/niworkflows/commit/2428186ac8cb927b96f69d73fb59d3c72e96c6a9

oesteban:
Oh yes, I need to merge `rel/1.4.0rc5` into master - sorry about that.

mgxd:
right...forgot about the dreaded sdcflows 2.0 migration...

oesteban:
Should we just close this one and use link dependencies only during testing PRs?

mgxd:
I still think there's some merit to pinning the closely related nipreps packages to master.

oesteban (approved):
Okay, let's try this and see how it goes. I think the most important change is that of the `test_deploy_pypi` on circle, and it seems like a good idea at this moment.

oesteban:
I don't think we can use dependency links for niworkflows and sdcflows before #2392 is merged. Why don't we just merge the changes to CircleCI's config here and update the dependencies there?

mgxd:
Totally agree - opened #2472 in favor of this

PR #2400, opened 2021-04-21
ENH: Set and track NumPy's random seed

mgxd:
Continuation of #2086

This PR explicitly sets the numpy random seed.

### DIFF
--- a/fmriprep/tests/test_config.py
+++ b/fmriprep/tests/test_config.py
@@ -95,6 +95,5 @@ def test_prng_seed(master_seed, ants_seed, numpy_seed):
         assert os.getenv("ANTS_RANDOM_SEED") == str(ants_seed)
 
     _reset_config()
-    assert os.getenv("ANTS_RANDOM_SEED") is None
     for seed in ('_random_seed', 'master', 'ants', 'numpy'):
         assert getattr(config.seeds, seed) is None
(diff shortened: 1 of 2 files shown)


oesteban:
Can you split the circleci maintenance into a quickfix to master?

oesteban (approved):

effigies:
The test should probably be mocked so that the config and environment are reset.

mgxd:
is 008853d what you had in mind @effigies ?

oesteban:
You'll probably need to customize something from:
https://github.com/nipreps/fmriprep/blob/674124ee80b3e2a8affddf005e910e4ca1c97cc0/fmriprep/workflows/tests.py#L10-L35

effigies:
No. You can use mock.patch to mock the environment, and probably the config object, too.

fmriprep/tests/test_config.py
```diff
@@ -51,3 +78,23 @@ def test_config_spaces():
...
+    seeds = config.seeds
+    with patch.dict(os.environ, {}):
+        seeds.load({'_random_seed': master_seed}, init=True)
+        assert getattr(seeds, 'master') == master_seed
+        assert seeds.ants == ants_seed
+        assert seeds.numpy == numpy_seed
+        assert os.getenv("ANTS_RANDOM_SEED") == str(ants_seed)
+
+    _reset_config()
+    assert os.getenv("ANTS_RANDOM_SEED") is None
```
effigies, line 98: Will fail if user has `ANTS_RANDOM_SEED` set.

```suggestion
```

effigies (approved):

PR #2403, opened 2021-04-23
PIN: tedana 0.0.9a1 for LTS branch

effigies:
tedana 0.0.9 requires scipy 1.3.3+ and scikit-learn 0.22+

Upgrading to 0.0.9+ would break our commitment to avoid non-bug-fix changes to underlying libraries during the LTS window.


effigies:
Going to go ahead and merge for now, since there isn't a clear decision that 0.0.10 will not break LTS constraints.

PR #2406, opened 2021-04-28
[ENH] Improve `_get_series_len` performance

### LINKED ISSUE #2405: Performance of `_get_series_len`
I am having some performance issues with creating fMRIPrep workflows for large datasets. As such, I did some profiling, and found that a lot of time is spent during creation of the workflow in `init_func_preproc_wf` on doing outlier detection. 

As far is I understand, this is to make sure that slice timing correction is only included if a sufficient number of volumes remains after removing non-steady state volumes in `init_bold_reference_wf`/`EstimateReferenceImage`. However, if `dummy_scans` is specified in the configuration, then this check may not be necessary, and workflows could be created twice as fast. 

Is my reasoning correct? Would you be interested in me contributing code to make this faster?

<https://github.com/nipreps/fmriprep/blob/674124ee80b3e2a8affddf005e910e4ca1c97cc0/fmriprep/workflows/bold/base.py#L221-L225>

https://github.com/nipreps/fmriprep/blob/674124ee80b3e2a8affddf005e910e4ca1c97cc0/fmriprep/workflows/bold/base.py#L879-L887

https://github.com/nipreps/niworkflows/blob/11ac7a18a403af717c0c2a29a33b7cadb75eabd7/niworkflows/interfaces/registration.py#L552-L557

HippocampusGirl:
Fixes #2405

When `dummy_scans` is provided in the configuration, we do not need to estimate the number of non-steady-state volumes, thus increasing performance of workflow creation.

This effectively halves the time needed for `init_func_preproc_wf` in my test dataset.

Not applicable

### DIFF
--- a/fmriprep/workflows/bold/base.py
+++ b/fmriprep/workflows/bold/base.py
@@ -220,7 +220,7 @@ def init_func_preproc_wf(bold_file):
     run_stc = (
         bool(metadata.get("SliceTiming"))
         and 'slicetiming' not in config.workflow.ignore
-        and (_get_series_len(ref_file) > 4 or "TooShort")
+        and (_get_series_len(ref_file, config.workflow.dummy_scans) > 4 or "TooShort")
     )
 
     # Build workflow
@@ -874,15 +874,13 @@ Non-gridded (surface) resamplings were performed using `mri_vol2surf`
     return workflow
 
 
-def _get_series_len(bold_fname):
+def _get_series_len(bold_fname, skip_vols):
     from niworkflows.interfaces.registration import _get_vols_to_discard
     img = nb.load(bold_fname)
     if len(img.shape) < 4:
         return 1
 
-    if isinstance(config.workflow.dummy_scans, int):
-        skip_vols = config.workflow.dummy_scans
-    else:
+    if skip_vols is None:
         skip_vols = _get_vols_to_discard(img)
 
     return img.shape[3] - skip_vols
--- a/fmriprep/workflows/bold/base.py
+++ b/fmriprep/workflows/bold/base.py
@@ -880,7 +880,10 @@ def _get_series_len(bold_fname):
     if len(img.shape) < 4:
         return 1
 
-    skip_vols = _get_vols_to_discard(img)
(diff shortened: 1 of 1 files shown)


fmriprep/workflows/bold/base.py
```diff
@@ -880,7 +880,10 @@ def _get_series_len(bold_fname):
     if len(img.shape) < 4:
         return 1
 
-    skip_vols = _get_vols_to_discard(img)
+    if isinstance(config.workflow.dummy_scans, int):
+        skip_vols = config.workflow.dummy_scans
+    else:
+        skip_vols = _get_vols_to_discard(img)
```
effigies, line 886: The `config` object is a hulking beast, and I'd prefer to keep its tentacles confined to `init_*_wf()` functions, and pass arguments to standalone functions, to make them easier to reason about.

```Python
def _get_series_len(bold_fname, skip_vols):
    from niworkflows.interfaces.registration import _get_vols_to_discard
    img = nb.load(bold_fname)
    if len(img.shape) < 4:
        return 1

    if skip_vols is None:
        skip_vols = _get_vols_to_discard(img)

    return img.shape[3] - skip_vols
```

And then, above, pass `config.workflow.dummy_scans` to `_get_series_len()`.

effigies (approved):
This is great! I appreciate the attention you pay to details like this that improve performance.

HippocampusGirl:
Thank you :-) I actually only noticed this by accident while profiling something else.

fmriprep/workflows/bold/base.py
oesteban: We probably want to make an overhaul of the documentation, on the code guidelines section) and:
- mention this (please do not use config deep in the workflow)
- recommend black for everything except the `workflow.connect` function

oesteban (approved):

fmriprep/workflows/bold/base.py
HippocampusGirl: I haven't actually looked at the code guidelines in a while, so I read them again just now. 

I realized that I probably also should have considered the branch naming/pull request naming, and added `[ENH]`.

PR #2421, opened 2021-06-02
PIN: nipreps/niworkflows bugfix branch

effigies:
Testing nipreps/niworkflows#641.


setup.cfg
```diff
@@ -27,7 +27,7 @@ install_requires =
     nipype >= 1.5.1
     nitime
     nitransforms >= 20.0.0rc3, < 20.2
-    niworkflows ~= 1.3.2
+    niworkflows @ git+https://github.com/nipreps/niworkflows@refs/pull/641/merge
```
effigies, line 30:
```suggestion
    niworkflows ~= 1.3.4
```

oesteban:
What should we do with this one? Should we patch `maint/20.2.x` and propagate?

effigies:
This should be closeable. The latest release should already have the updated niworkflows, though we should verify.

oesteban:
Follow up on #2458.

PR #2429, opened 2021-06-24
ENH: Relax requirement for PyBIDS databases to exist

effigies:

It can be useful to store the BIDS database separately from the working directory. For example, if you want to preserve the working directory on an HPC run, it will need to be put on (or synced to) a network store. There is no advantage to doing that with an SQLite database as well, so we should permit the location to be specified without pre-generating.

### DIFF
--- a/fmriprep/config.py
+++ b/fmriprep/config.py
@@ -439,7 +439,7 @@ class execution(_Config):
             if cls.bids_database_dir:
                 _db_path = cls.bids_database_dir
                 if not _db_path.exists():
-                    logging.getLogger("cli").warn(
+                    logging.getLogger("cli").warning(
                         f"Creating PyBIDS database directory: {_db_path}"
                     )
             else:
--- a/fmriprep/config.py
+++ b/fmriprep/config.py
@@ -436,9 +436,15 @@ class execution(_Config):
             import re
             from bids.layout import BIDSLayout
 
-            _db_path = cls.bids_database_dir or (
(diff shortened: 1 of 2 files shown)


mgxd (approved):
Seems reasonable - should we add a quick log warning users when it doesn't exist? and in the case of the docker wrapper, we should create it rather than docker to avoid permission issues

effigies:
Hmm. Seemed straightforward, but our execution config init has to come before the logging config init. I can just kind of log in a one-off way, but it won't be saved to the log file.

For the wrapper, Docker will do that automatically. Try `fmriprep-docker --shell --bids-database-dir /tmp/bidsdb` and observe that `/tmp/bidsdb` is created on the host.

mgxd:
I have a vague memory of file ownership problems when Docker is left to create volumes, but perhaps that's been updated - in any case running the wrapper with `-u $(id -u)` should probably suffice.

fmriprep/config.py
```diff
@@ -436,9 +436,15 @@ def init(cls):
             import re
             from bids.layout import BIDSLayout
 
-            _db_path = cls.bids_database_dir or (
-                cls.work_dir / cls.run_uuid / "bids_db"
-            )
+            if cls.bids_database_dir:
+                _db_path = cls.bids_database_dir
+                if not _db_path.exists():
+                    logging.getLogger("cli").warn(
```
effigies, line 442:
```suggestion
                    logging.getLogger("cli").warning(
```

PR #2434, opened 2021-06-30
FIX: Raise informative error when duplicate subworkflows are added

effigies:
Inspired by https://github.com/nipreps/nibabies/issues/82.

`workflow.connect()` is tab-shifted but otherwise unchanged.

### DIFF
--- a/fmriprep/workflows/base.py
+++ b/fmriprep/workflows/base.py
@@ -303,23 +303,35 @@ tasks and sessions), the following preprocessing was performed.
     for bold_file in subject_data['bold']:
         func_preproc_wf = init_func_preproc_wf(bold_file)
 
-        workflow.connect([
-            (anat_preproc_wf, func_preproc_wf,
-             [('outputnode.t1w_preproc', 'inputnode.t1w_preproc'),
-              ('outputnode.t1w_mask', 'inputnode.t1w_mask'),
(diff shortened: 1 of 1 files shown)


mgxd (commented):
Should we generalize this to the whole workflow building process? i.e.
https://github.com/nipreps/fmriprep/blob/237523c0926fd84b60d6249088c13167455e5a16/fmriprep/cli/workflow.py#L82

I know we are building within a subprocess, which might make this error reporting too complicated, but it would be nice to handle this in the case of every `connect` statement

effigies:
The nice thing about where I put it is that we have the specific file name as well as the exception, so we can be pretty precise. Wrapping the line you suggest would lose that.

Besides, what are the other cases where there might be a collision? The subject ID is in the subject workflow name, and all the entities are in the functional workflow name, so AFAIK the `.nii`/`.nii.gz` duplication is the only possibility. If we see others, it would be good to get them, but a quick survey of old issues showed one pybids bug, and a bunch of this.

mgxd:
Sounds good - I agree it does give a better message there. If the duplication error ever comes up again we can revisit this - but I'm fine with the changes as is

effigies:
Cool. Guess I'll wait on the nibabies issue, to see if there is some other way that we're hitting this.

mgxd:
The nibabies error was due to an internal bug, merging.

PR #2436, opened 2021-07-02
FIX: Add some clarity to BrokenProcessPool failures

mgxd:
To remove some ambiguity from that error.

Additionally, fMRIPrep will now raise a `RuntimeError` in addition to any `Exception` hit when running the workflow.


effigies:
Looks sensible.

oesteban (approved):

PR #2409, opened 2021-05-13
ENH: Bump Dockerfile

mgxd:
This PR bumps up the Docker image base to Ubuntu Focal LTS (20.04).

As a result, a number of packages were updated, including:
- AFNI
- FSL
- Connectome Workbench
- Graphviz
- Pandoc

Furthermore, Python was bumped to 3.8.x, and many of its packages were "modernized" accordingly.

This builds the foundation for #2392 , as it depends on updated scipy. Also closes #2382

PR #2443, opened 2021-07-14
MNT: Update BIDS validator to 1.8.0

effigies:
Follow-up to #2409, to test whether the HED refactor solved our issues.


effigies:
Failure: https://github.com/bids-standard/bids-validator/issues/1307

oesteban:
Can we update to =~1.6.0 for now?

mgxd:
> Can we update to =~1.6.0 for now?

I believe `1.5.7` was the last version that did not require network access

effigies:
Yeah, the current pin is the last working version. I was seeing if our issue was resolved. It may have been, but we bundled a significant change in a patch release.

effigies:
Passed the validation stage. This should be good to go on green.

PR #2446, opened 2021-07-15
MAINT: Some dependencies cannot be left unpinned

oesteban:
This PR limits the latest acceptable version of networkx, nilearn and tedana, as they have started to show errors when building the docker image.


oesteban:
Okay - no more errors. I'm going to merge this in.

effigies:
Was confused by this, since we'd dealt with it in `maint/20.2.x` and `master`. Just added a tag to make it a bit clearer.

PR #2449, opened 2021-07-15
MAINT: Keep pip and setuptools with support for Python 3.5

oesteban:
Follows #2446.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -289,7 +289,7 @@ jobs:
             pyenv local $PY3
             echo -n "Python version: "
             python --version
-            pip install --upgrade pip setuptools
+            pip install --upgrade "pip<21" "setuptools<51"
             pip install --upgrade wrapper/
             which fmriprep-docker
             fmriprep-docker -i poldracklab/fmriprep:latest --help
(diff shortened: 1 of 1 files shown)


mgxd (approved):
LGTM

mgxd:
What's the expected lifecycle of older, non-LTS minor releases?

oesteban:
Good question. But in the case of the spatial normalization bug it is
something we want to fix even if support of 20.1.x were ended.

effigies:
Wouldn't it just make sense to recommend upgrading to 20.2, since this affects all 20.1?

oesteban:
> Wouldn't it just make sense to recommend upgrading to 20.2, since this affects all 20.1?

Yes, I think we are already doing that anyways. Some people may have started some long study with 20.1 and will want to stick with it - although the best advice would be upgrading indeed.

I'd be happy to call 20.1.x and end support after this bug is fixed. Only if something as big as this one surfaces, we might want to provide a fixed release. If the patch wouldn't be as easy, then we may want to ban 20.1.x completely - but let's give this series a chance, right?

oesteban:
> What's the expected lifecycle of older, non-LTS minor releases?

Should we open a pinned issue, start conversations and maybe raise this question in a roundup with some thought-through proposal?

Issue #2455, opened 2021-07-18
Back port docker auth to maint/20.2.x in Circleci


oesteban:
### What is the exact command-line you used?
```
<place your command line here>
```

### Can you find some traces of the error reported in the visual report (at the bottom) or in *crashfiles*?
```
<please copy&paste all the information you can gather about errors>
```

### fMRIPrep log
If you have access to the output logged by fMRIPrep, please make sure to attach it as a text file to this issue.

PR #2456, opened 2021-07-19
MAINT: Back port CircleCI configuration from dev branch

oesteban:
- Addresses issues when authenticating docker hub
- Updates CircleCI's machine image, and related changes.

Resolves: #2455.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,12 +1,16 @@
-version: 2
-jobs:
+version: 2.1
+orbs:
+  docker: circleci/docker@1.6.0
 
+jobs:
   build:
     environment:
       TZ: "/usr/share/zoneinfo/America/Los_Angeles"
       SCRATCH: "/scratch"
     machine:
-      image: ubuntu-1604:202007-01
+      # https://discuss.circleci.com/t/linux-machine-executor-images-2021-april-q2-update/39928
+      # upgrade Docker version
+      image: ubuntu-2004:202104-01
     working_directory: /tmp/src/fmriprep
     steps:
       - checkout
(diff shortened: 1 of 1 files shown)


oesteban:
Any obvious caveat or issue that I'm not seeing?

.circleci/config.yml
```diff
@@ -117,15 +123,14 @@ jobs:
...
     working_directory: /home/circleci/data
     steps:
       - restore_cache:
           keys:
-            - data-v8-{{ .Branch }}-{{ .Revision }}
-            - data-v8-{{ .Branch }}
-            - data-v8-
+            - data-v9-{{ .Branch }}-{{ .Revision }}
+            - data-v9-{{ .Branch }}
+            - data-v9-
```
effigies, line 133: Why are we dropping this cache? Abundance of caution?

.circleci/config.yml
```diff
@@ -1236,7 +1223,7 @@ workflows:
             - deployable
           filters:
             branches:
-              only: master
+              only: maint/20.2.x
             tags:
```
effigies, line 1227: I don't think we want to push `:unstable` from the `maint/20.2.x` branch.

effigies (approved):
Sent some small comments, but overall this looks fine.

.circleci/config.yml
oesteban: Actually, now that you made me think about it, we want to keep v8 and, if ever changed, do not update to v9 (which is used by master), but by something like `data-20.2-v1` etc.

It's been long since the last time we made changes to the data, and with `nipreps-data` this cache should become more stable and manageable with datalad (rather than just circle).

Issue #2454, opened 2021-07-16
CircleCI's cache tags are ambiguous


oesteban:
e.g., this tag:
https://github.com/nipreps/fmriprep/blob/7cb3117a165e75feba79233e0a27308e34c17d24/.circleci/config.yml#L26

fails with git tags because ``{{ .Branch }}`` is empty.

PR #2457, opened 2021-07-19
MAINT: Revise CircleCI's cache tags to avoid mixups between builds

oesteban:
As a result of the ambiguity of these tags, a tag 20.2.2 was deployed to docker hub containing the 20.1.4 build last Friday.

The misuse of ``{{ epoch }}`` as well as too liberal template matching when loading cached docker images resulted in this situation.

Resolves: #2454.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,7 +23,8 @@ jobs:
             fi
       - restore_cache:
           keys:
-            - build-v4-{{ .Branch }}-{{ epoch }}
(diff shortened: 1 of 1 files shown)


oesteban:
Any obvious problem / caveat?

oesteban:
Seems to be working properly. Pushing it in.

effigies:
Do we need to do a similar thing in `maint/20.2.x`?

oesteban:
Not sure it is worth the pain, but if we again are releasing two tags at once this could possibly happen again.

PR #2459, opened 2021-07-19
MAINT: Remove traces of Xenial base

mgxd:
I don't think we're ever hitting the `else` block, but fallback to focal base now that the Dockerfile has been updated.


oesteban:
Can we do this after #2456 and #2457 are merged?

mgxd:
This is pretty low priority since the registry is usually used, but just to note this is not to be backported

Issue #2411, opened 2021-05-15
sentry-sdk is imported unconditionally by the CLI, but not part of install_requires


effigies:
See: https://neurostars.org/t/fmriprep-error-using-ubuntu-20-04-2/19294

The issue is that `sentry-sdk` is imported unconditionally, but not part of `install_requires`.

1) If it's intended to be optional, we should guard against `ImportError`s.
2) If it's required, we should add to `install_requires`.
3) If we want people to be able to install fMRIPrep as a library and not purely as a runtime CLI, then we should update the [installation instructions](https://fmriprep.org/en/stable/installation.html#manually-prepared-environment-python-3-7) to install `fmriprep[sentry]`. Possibly move `sentry` to `[cli]` or something to make it clearer that the reason for adding it is to enable the CLI.

### What version of fMRIPrep are you using?

20.2.1

### What kind of installation are you using? Containers (Singularity, Docker), or "bare-metal"?

bare-metal

### What is the exact command-line you used?
```bash
bids_root_dir=$HOME/openneuro_tutorial/ds000102
preprocess_output_dir=$HOME/openneuro_tutorial/ds000102/fmriprep
subj=01
 
export TEMPLATEFLOW_HOME=$HOME/openneuro_tutorial/templateflow
export FS_LICENSE=$HOME/openneuro_tutorial/license.txt

fmriprep   $bids_root_dir $preprocess_output_dir participant \
   --participant-label $subj \
   --skip_bids_validation \
   --md-only-boilerplate \
   --fs-license-file $HOME/openneuro_tutorial/license.txt \
   --fs-no-reconall \
   --output-spaces MNI152NLin2009cAsym:res-1 \
   --nthreads $nthreads \
   --stop-on-first-crash \
   -w $HOME
```

### Have you checked that your inputs are BIDS valid?

Unknown

### Did fMRIPrep generate the visual report for this particular subject? If yes, could you share it?

No

### Can you find some traces of the error reported in the visual report (at the bottom) or in *crashfiles*?

No

### Are you reusing previously computed results (e.g., FreeSurfer, Anatomical derivatives, work directory of previous run)?

Unknown

### fMRIPrep log
If you have access to the output logged by fMRIPrep, please make sure to attach it as a text file to this issue.

```
    File "/home/taliw/.local/bin/fmriprep", line 11, in <module>
    load_entry_point('fmriprep==20.2.1', 'console_scripts', 'fmriprep')()
[...]

PR #2460, opened 2021-07-19
FIX: Unprotected import of ``sentry_sdk``, which is not a dependency

oesteban:
Resolves: #2411.


effigies (approved):

oesteban:
Tested locally on an environment without sentry. Current `maint/20.2.x` crashed as expected and this patch worked as expected.

PR #2461, opened 2021-07-19
ENH: Detect 3D BOLD series, warn and return empty workflow

### LINKED ISSUE #1931: Handle single-volume BOLD series
Currently, single-volume BOLD series give inscrutable errors (#1924). We should either accept them or add an early check to reject/skip them.

I would suggest when creating the bold workflow, we simply read the header, and if it's single-volume raise a warning and return an empty workflow. There are likely too many places where we assume multiple volumes, and the cost/benefit of adding support seems high at the moment.

effigies:
Closes #1931.

### DIFF
--- a/fmriprep/workflows/bold/base.py
+++ b/fmriprep/workflows/bold/base.py
@@ -146,9 +146,9 @@ def init_func_preproc_wf(bold_file):
 
     workflow = Workflow(name=wf_name)
 
-    if nb.load(bold_file).shape[3:] <= (1,):
+    if nb.load(bold_file).shape[3:] <= (5,):
         config.loggers.workflow.warning(
-            f"3D BOLD series <{ref_file}> detected. Skipping."
+            f"Too short BOLD series  (<= 5 timepoints). Skipping processing of <{ref_file}>."
         )
         return workflow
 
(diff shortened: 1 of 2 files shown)


fmriprep/workflows/bold/base.py
```diff
@@ -144,6 +144,15 @@ def init_func_preproc_wf(bold_file):
     from niworkflows.interfaces.utils import DictMerge
     from sdcflows.workflows.base import init_sdc_estimate_wf, fieldmap_wrangler
 
+    workflow = Workflow(name=wf_name)
+
+    if nb.load(bold_file).shape[3:] <= (1,):
```
oesteban, line 149: We ensure N > 5 later if STC is to be run. I'd say we can safely do that directly here.
```suggestion
    if nb.load(bold_file).shape[3:] <= (5,):
```

fmriprep/workflows/bold/base.py
```diff
@@ -144,6 +144,15 @@ def init_func_preproc_wf(bold_file):
...
+        config.loggers.workflow.warning(
+            f"3D BOLD series <{ref_file}> detected. Skipping."
```
oesteban, line 151:
```suggestion
            f"Too short BOLD series  (<= 5 timepoints). Skipping processing of <{ref_file}>."
```

oesteban (approved):
Suggesting to make this more strict and demand >5 timepoints.

fmriprep/workflows/bold/base.py
```diff
@@ -144,6 +144,15 @@ def init_func_preproc_wf(bold_file):
...
+        )
+        return workflow
+
+
```
effigies, line 155:
```suggestion

```

effigies:
Sure.

Issue #1847, opened 2019-11-01
Documentation says res- field of ``--output-spaces`` is in mm units


oesteban:
@bbfrederick

bbfrederick:
On this page (https://fmriprep.readthedocs.io/en/1.5.0/spaces.html), the documentation does not say explicitly, but very strongly implies, that :res-XX means XX in mm:

> Therefore, fMRIPrep will run nonlinear registration processes against the template T1w image corresponding to all the standard spaces supplied with the argument --output-spaces. By default, fMRIPrep will resample the preprocessed data on those spaces (labeling the corresponding outputs with the space-<template-identifier> BIDS entity) but keeping the original resolution of the BOLD data to produce smaller files, more consistent with the original data gridding. However, many users will be interested in utilizing a coarse gridding (typically 2mm isotropic) of the target template. Such a behavior can be achieved applying modifiers to the template identifier, separated by a : character. For instance, --output-spaces MNI152NLin6Asym:res-2 MNI152NLin2009cAsym will generate preprocessed BOLD 4D files on two standard spaces (MNI152NLin6Asym, and MNI152NLin2009cAsym) with the template’s 2mm isotropic resolution for the data on MNI152NLin6Asym space and the original BOLD resolution (say, e.g., 2x2x2.5 [mm]) for the case of MNI152NLin2009cAsym.

emdupre:
Could we update the example provided here to mirror [this one](https://github.com/bids-standard/bids-specification/pull/301/files#diff-ee997314dc6a77ebab0d882ed7945467R93) from the BIDS specification proposal ?

Personally, I think specifying alphabetic rather than numeric resolution values (e.g., `res-hi` or `res-lo`) makes it clearer that the JSON sidecar should be consulted for the exact resolution.

oesteban:
That would mean we need to refactor TemplateFlow... which I think makes sense but I'm not sure I have the bandwidth

PR #2466, opened 2021-07-20
DOC: Clarify that ``res-2`` entity does not mean 2mm

oesteban:
@bbfrederick brought up the problem that, although the documentation does not state the above explicitly, it is the most plausible reading.

This PR adds a danger admonition to clarify the ambiguity.

Follow-up: templateflow/templateflow#92.
Resolves: #1847.


oesteban:
@bbfrederick I took the liberty of setting you as a reviewer for this PR, please let me know if you would like I keep doing so sparingly.

That said, given the small size of the changes, we might go ahead with it soon if you can't find a moment to check (or are on vacation, which is also highly likely at this point of the year).

bbfrederick (approved):
I think this is quite clear.

PR #2468, opened 2021-07-20
FIX: Address the problems of a sloppy merge

oesteban:
I precipitated the merge of #2461 without observing three issues:

- The bold workflow is provided with a list (not a str) with ME-EPI
- Test data (ds054) for CircleCI has only 5 timepoints, which triggered workflow skipping.
- Inexisting variable name required in string interpolation when logging the skip warning.

Both problems are addressed in this PR.

Continues: #2461.


oesteban:
Tested locally with ds054 (the subsampled version) and with ds210.

PR #2471, opened 2021-07-20
MAINT: Add missing OASIS30 WM/BS probsegs

mgxd:
These templates are usually required during anatomical processing.

https://github.com/nipreps/niworkflows/blob/32a33d61ab64e262a4ebcf8afcf232d70c2f92b3/niworkflows/anat/ants.py#L391-L403


oesteban:
Should this be merged into `maint/20.2.x` first and then upstream?

mgxd:
No, since `maint/20.2.x` is not using this helper script to fetch templates

effigies (approved):

PR #2474, opened 2021-07-22
DOC: Refactor README and citing information

oesteban:
This PR moves citation information and text about the boilerplate over to www.nipreps.org. The citing section has been eliminated, which was long overdue because we haven't kept it up-to-date over time.

We could consider to grab the boilerplate our CircleCI tests generate and insert them automatically in the documentation at build step.

For the moment, the README is revised and the citation and acknowledgement information updated.


docs/usage.rst
```diff
@@ -48,6 +48,30 @@ The command-line interface of the docker wrapper
...
+
+
+Limitations and reasons not to use *fMRIPrep*
+---------------------------------------------
+
+1. Very narrow :abbr:`FoV (field-of-view)` images oftentimes do not contain
+   enough information for standard image registration methods to work correctly.
+   Also, problems may arise when extracting the brain from these data.
+   Supporting these particular images is already a future line of the development
+   road-map.
```
effigies, line 60: We have made some progress here...

```suggestion
1. Very narrow :abbr:`FoV (field-of-view)` images oftentimes do not contain
   enough information for standard image registration methods to work correctly.
   Also, problems may arise when extracting the brain from these data.
   fMRIPrep supports pre-aligned BOLD series, and accepting pre-computed
   derivatives such as brain masks is a target of future effort.
```

effigies (approved):
Otherwise :+1:.

PR #2488, opened 2021-08-03
MAINT: Codename convention proposal - female neuroscientists

oesteban:
This PR proposes that each fMRIPrep's new series (YY.MINOR.x) receives the name of a female neuroscientist for codename.

To get this convention started, I propose 21.0.x will carry the codename of *Kreek*, after [Mary Jeanne Kreek](https://en.wikipedia.org/wiki/Mary_Jeanne_Kreek), who passed away last March.

Although codenames do not have any effect on functionality, I hope this initiative is received warmly by @nipreps/fmriprep-contributors and I will be pestering you all for some time to make sure there are no objections. Please express your opinions in this PR before it gets merged.

I hope we have a clearer path to adopt these and other changes to the tool when the governance document is pushed forward.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,8 @@
-21.0.0 (TBD)
-============
-First major release in the 21.0.x series.
+21.0.0 *Kreek* (TBD)
+====================
+First major release in the 21.0.x series, codename *Kreek* (after
+`Mary Jeanne Kreek (1937-2021) <https://en.wikipedia.org/wiki/Mary_Jeanne_Kreek>`__,
+neurobiologist specializing in the study and treatment of addiction).
 Upstreams bugfix #2444 (feed *NiTransforms* with LTAs of type RAS2RAS)
 
 .. admonition:: New license - Apache 2.0
@@ -33,6 +35,7 @@ A full list of changes can be found below.
 * DOC: Point documentation at *SDCFlows* and remove SDC section (#2470)
 * DOC: Adds ``--mem`` metavar (#2378)
 * DOC: Skull stripping is forced by default (#2345)
+* MAINT: Codename convention proposal - female neuroscientists (#2488)
 * MAINT: Relicense +20.3.x - BSD-3-Clause -> Apache License 2.0 (#2325)
 * MAINT: Add missing OASIS30 WM/BS probsegs (#2471)
 * MAINT: Update BIDS validator to 1.8.0 (#2443)
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -1,6 +1,30 @@
 .. include:: links.rst
 
-----------
+-----------------------------------
+*fMRIPrep*'s versioning and changes
+-----------------------------------
+*fMRIPrep* follows the `NiPreps conventions <https://www.nipreps.org/devs/releases/>`__
+for the versioning scheme.
+In short, the basic release form is ``YY.MINOR.PATCH``, so the first release of
+2020 was 20.0.0.
+The most relevant implication to the user is that compatibility of intermediate and
+final results is guaranteed through a given series (in other words, ``YY.MINOR``).
+For instance, version 20.1.4 must be compatible with processing done with 20.1.0.
+
+**Long-term support (LTS) releases**.
+Some releases marked as LTS are special and have a `longer support window
+<https://www.nipreps.org/devs/releases/#long-term-support-series>`__.
+
+**Release codenames**.
+Release series (``YY.MINOR``) share a common *codename*.
+The codename will be one name drawn from `Wikipedia's list of women neuroscientists
+<https://en.wikipedia.org/wiki/List_of_women_neuroscientists>`__.
+This convention was initiated with the 21.0.x «*Kreek*» series, which are named after
+`Mary Jeanne Kreek (1937-2021) <https://en.wikipedia.org/wiki/Mary_Jeanne_Kreek>`__.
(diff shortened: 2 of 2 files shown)


effigies:
I'm generally not a fan of code names; I think they're supposed to make software friendlier and more mnemonic, but honestly I lose track of the Debian, Ubuntu and OSX names and wish people would just use numbers. Not sure if I'm the odd one out, here, though.

This also doesn't feel like it will necessarily be received well by everyone. The intent is obviously to honor scientists, and the choice of female scientists is presumably to help correct the erasure of women. It's not at all clear to me that this is how people would want to be honored, and it could even backfire if fMRIPrep starts out-ranking the actual work of those scientists when people search their names on search engines.

If I'm out-voted here are my suggestions:

1) Avoid living scientists. They don't need fMRIPrep (or discussions of it) in their search results.
2) Consider how to get ahead of any problems with search engines putting fMRIPrep (or works citing fMRIPrep, or threads discussing fMRIPrep) ahead of the work or bios of the scientists whose names we use. As an idea, writing up a bio of each person and linking to their work might at least help make sure pagerank pulls them up with us. And if that fails, at least having information about the person in our docs would be informative to people trying to look them up.
3) Take it seriously. This adds a component of arbitrating the relative significance of women in neuroscience, and I'd rather we didn't offend people by skipping over a person well-known in a subfield we're less familiar with. Especially if it's one where we go to a kind of adjacent field for someone in our intellectual tradition but who wouldn't have considered herself a neuroscientist.
4) Use an alphabetical scheme, like Ubuntu does. This will make it easy to compare versions.

oesteban:
Thanks for the feedback, it seems to me everyone deserved a bit more context which I failed to provide. That said, this is not a hill I'm willing to die in, so I'll just try to clarify some aspects of the proposal.

- Code names *are more pain than gain* (as Yarik summarized). I guess that depends on how much you will rely upon code names. I'll edit the proposal so that it's clear that *the chosen names must carry no semantics in terms of versions or new features*. In other words, our code names would not be more than an excuse to have a new link to a female neuroscientist with each new series. Thus far we are writing the release notes by hand and we don't release new series so often so I don't see how this can interfere with our automated deployment. Similarly, these code names are not meant to be used as an alternative to the version numbers.
- The intent of the proposal is to make our release notices and changelog more engaging - besides the cold list of changes we propose the reader a neuroscientist. Proposing other sources of inspiration seemed less engaging for being off-topic. The choice of limiting to women is just a gesture to show commitment within our environment. Also, because it may help make fMRIPrep more engaging to women (our contributors' list is mostly males).

   > It's not at all clear to me that this is how people would want to be honored, and it could even backfire if fMRIPrep starts out-ranking the actual work of those scientists when people search their names on search engines.

  I think saying this is to honor people is giving too much credit to fMRIPrep. The proposal just wants to disseminate and nurture curiosity about female neuroscientists - that's why I didn't think much of "reasons" or mechanisms to select these codenames. Because this should avoid any relevance or importance criteria, I'm proposing selecting at random below. I'm not sure how that out-ranking actual work may ever happen, does this comment refer to the pagerank below?

  > 1. Avoid living scientists. They don't need fMRIPrep (or discussions of it) in their search results.

  This sounds very reasonable to me.

  > 2. Consider how to get ahead of any problems with search engines putting fMRIPrep (or works citing fMRIPrep, or threads discussing fMRIPrep) ahead of the work or bios of the scientists whose names we use. As an idea, writing up a bio of each person and linking to their work might at least help make sure pagerank pulls them up with us. And if that fails, at least having information about the person in our docs would be informative to people trying to look them up.

  I think our currency is citations, not pagerank. As long as we do not take any actions that may bias the normal flow of citations (and use this to direct more to fmriprep in ways I can't imagine right now), we should be okay IMHO. That said, I'd be happy to try to pull their pagerank with us. Maybe the way of paying back is more about helping improve their pages at Wikipedia.

  > 3. Take it seriously. This adds a component of arbitrating the relative significance of women in neuroscience, and I'd rather we didn't offend people by skipping over a person well-known in a subfield we're less familiar with. Especially if it's one where we go to a kind of adjacent field for someone in our intellectual tradition but who wouldn't have considered herself a neuroscientist.

  Let's stipulate that the names will be selected at random within the shortlist of deceased people in the list of Wikipedia. The shortlist would be updated before issuing a new code name, the seed will be something fixed (e.g., the date of the first LTS release), and the script stored under `.maint/` in the repo. Making the choice at random eliminates the risk of offending anyone.

  > 4. Use an alphabetical scheme, like Ubuntu does. This will make it easy to compare versions.

  If we prefer that version numbers are used, why facilitate codename comparison? I'd prefer not to have any semantics, as said above.

> So it could be a log line/message at the beginning (or the end, since that is what people might typically see) honoring whomever/however you decide to honor. With such message it could be even more informative than just a name but provide related factoid.

I believe this is actually more involved - we need to write some code that does this and is executed every time. Then, if the run failed we might need to pull this out (if we did not print it in the beginning). Otherwise, many people will copy their logs verbatim on forums and here - we may start seeing problems of our support attempts clobbering the actual hits for a given scientist.

However, I find it interesting to add that factoid alongside the name with the `--version` argument of fMRIPrep. That should not be tracked by search engines and feeds that curiosity we want to nurture even in off-line environments as it can be hard-coded.

effigies:
This has been [pocket vetoed](https://en.wikipedia.org/wiki/Pocket_veto). If this is something you think we should reconsider, let's discuss at a nipreps call, take a vote, and commit to doing this work if the vote passes.

PR #2489, opened 2021-08-03
FIX: Simplify STC logic for too short BOLD series

oesteban:
After #2461 and #2468, the ``TooShort`` value for running STC could not be assigned.
This PR ignores manually set dummy timepoints, although the case when this could be problematic is largely a very edge case (say you have only 7 timepoints and manually set 3 as dummy).

On the flip side, this PR drops the use of a function discontinued in the latest niworkflows release, fixing a bug ahead of time (currently popping out in #2392).

### DIFF
--- a/fmriprep/workflows/bold/base.py
+++ b/fmriprep/workflows/bold/base.py
@@ -246,11 +246,10 @@ def init_func_preproc_wf(bold_file):
         # If fieldmaps are not enabled, activate SyN-SDC in unforced (False) mode
         fmaps = {'syn': False}
 
-    # Short circuits: (True and True and (False or 'TooShort')) == 'TooShort'
+    # Check whether STC must/can be run
     run_stc = (
         bool(metadata.get("SliceTiming"))
         and 'slicetiming' not in config.workflow.ignore
-        and (_get_series_len(ref_file, config.workflow.dummy_scans) > 4 or "TooShort")
     )
 
     # Build workflow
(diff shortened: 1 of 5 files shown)


fmriprep/workflows/bold/base.py
```diff
@@ -905,18 +904,6 @@ def init_func_preproc_wf(bold_file):
...
 
 
-def _get_series_len(bold_fname, skip_vols):
-    from niworkflows.interfaces.registration import _get_vols_to_discard
-    img = nb.load(bold_fname)
-    if len(img.shape) < 4:
-        return 1
-
-    if skip_vols is None:
-        skip_vols = _get_vols_to_discard(img)
```
effigies, line 915: We lose the case where the series has >5  volumes but <5 taking into account dummy scans.

oesteban: Yes, this is what I meant (and definitely failed to convey) above.

oesteban:
After some thinking, I believe the problem we were trying to avert with the check removed by this PR should actually end up in an error. However, the specific error that `3dTShift` issues (which fMRIPrep bubbles up) is really hard to understand.

My last commit 18f5e72 patches the interface so that the error is anticipated by the interface itself and a more understandable message is delivered to the user. CircleCI's build of ds210 should now crash with the new error (I'll fix that in an upcoming commit once this is confirmed to work properly).

I'm aware of the cost of not calculating steady states at workflow building time is delaying the crash  - but it is preferable that the user suffers an error down the line rather than just skipping STC and cross fingers that they will read the reports and the boilerplate, IMHO.

oesteban:
Okay, the new edit seems to work well. Now ds210 is crashing at the STC node with:

```
RuntimeError: Insufficient length of BOLD data (5 time points) after discarding 1 nonsteady-state (or 'dummy') time points.
```

I'm going to switch this test to have `--disable slicetiming` and add one entry to the FAQ.

oesteban:
New FAQ - https://fmriprep--2489.org.readthedocs.build/en/2489/faq.html#error-in-slice-timing-correction-insufficient-length-of-bold-data-after-discarding-nonsteady-states

effigies (approved):

PR #2444, opened 2021-07-15
FIX: Feed *NiTransforms* with LTAs of type RAS2RAS

oesteban:
As reported by @eburkevt back on October 18, 2020, resampled BOLD time-series in standard spaces do not look good anymore.
This PR uses FreeSurfer to cater RAS2RAS transforms to *NiTransforms*.
This may resolve the problem for *fMRIPrep* regardless of the resolution of poldracklab/nitransforms#125

Addresses: #2307.

PR #2508, opened 2021-08-19
FIX: Avoid unnecessary connections based on branching logic

mgxd:
After #2444, `init_bbreg_wf` would crash when `bold2t1w_init == 'register'` and `use_bbr is False`.

Originally implemented in https://github.com/nipreps/nibabies/pull/95


effigies (approved):

oesteban:
@mgxd I just upstreamed this to master, just to be sure - Was that expected?

mgxd:
yes, as long as #2444 was upstreamed as well

oesteban:
Yes, that was upstreamed already.

Issue #2512, opened 2021-08-23
ERROR: traits.trait_errors.TraitError: The 'tr' trait of a FunctionalSummaryInputSpec instance must be a float, but a value of None <class 'NoneType'> was specified.


BlackHoleExelion:
Ran with Fmriprep_20.2.3 in singularity. Also tested with Fmriprep_20.2.0 and got the same error.

Here is the command and error message: 

>(base) [xxx@singularity01 ~]$ singularity run --writable-tmpfs --cleanenv \
> -B /project/oathes_group:/mnt \
> /project/oathes_group/singularity_images/fmriprep_20.2.3.simg \
> -w /mnt/fmriprep/temp_fmriprep \
> --fs-license-file /mnt/fs_licenses/freesurfer_license.txt \
> --skip_bids_validation \
> --skull-strip-template OASIS30ANTs \
> --output-spaces MNI152NLin2009cAsym fsaverage5 T1w fsnative \
> --cifti-output 91k \
> --bold2t1w-dof 6 \
> --aroma-melodic-dimensionality -200 \
> --dvars-spike-threshold 1.5 \
> --fd-spike-threshold 0.5 \
> /mnt/BIDS/ZAPR01 /mnt/fmriprep/output participant --participant-label C262

> 210823-16:40:58,101 nipype.workflow IMPORTANT:
> 
>     Running fMRIPREP version 20.2.3:
>       * BIDS dataset path: /mnt/BIDS/ZAPR01.
>       * Participant list: ['C262'].
>       * Run identifier: 20210823-163855_b93f51a8-c988-493b-9fb8-2b09026ddc87.
>       * Output spaces: MNI152NLin2009cAsym:res-native fsaverage:den-10k T1w fsnative.
>       * Pre-run FreeSurfer's SUBJECTS_DIR: /mnt/fmriprep/output/freesurfer.
> 210823-16:41:02,9 nipype.workflow INFO:
>          No single-band-reference found for sub-C262_ses-Baseline_task-nback02_run-01_bold.nii.gz.
> Process Process-2:
> Traceback (most recent call last):
>   File "/usr/local/miniconda/lib/python3.7/multiprocessing/process.py", line 297, in _bootstrap
>     self.run()
>   File "/usr/local/miniconda/lib/python3.7/multiprocessing/process.py", line 99, in run
>     self._target(*self._args, **self._kwargs)
[...]

PR #2513, opened 2021-08-23
FIX: Permit missing TR to show PyBIDS error at workflow construction time

effigies:
Should make #2512 have a clearer error message.


mgxd (approved):

effigies:
Note that this relies on https://github.com/bids-standard/pybids/pull/593, included in version 0.11.0. We require 0.11.1+, so this should be safe.

Issue #2477, opened 2021-07-25
Slice-timing-correction induced time shift holds potential for error in subsequent fMRI analysis


theoschaefer:
Hi everyone,

with this post, we would like to point to a potentially serious issue introduced by fMRIPrep’s slice-timing correction (STC) which affects further processing of fMRI data and might therefore be relevant for a large number of users.

STC in fMRIPrep is referenced to the middle slice which leads to a time-shift in the onsets by 0.5 TR. Importantly, users might not be explicitly aware of this time shift and might therefore not consider it for subsequent analyses, especially since defaults of other software packages are sometimes misaligned (e.g. nilearn’s function “nilearn.glm.first_level.FirstLevelModel'' has a default microtime onset of 0). This may inadvertently lead to model misspecification for a potentially large number of users. Currently, it is not clear how many users are affected by this issue but an informal survey of fMRIPrep users at our institute makes us think this may be quite prevalent. The impact is currently also not very clear but correcting for the effect (i.e. considering the time shift in subsequent analyses) strongly improved the results of all users involved at our institute, so we think the impact may be high. The upside of this effect is that accurately taking the time shift into account should only improve model specification and therefore lead to stronger results.

While the reference to the middle slice is currently documented at two locations in fMRIPrep ([here](https://fmriprep.org/en/stable/workflows.html#slice-time-correction) and [here](https://raw.githubusercontent.com/oesteban/fmriprep/38a63e9504ab67812b63813c5fe9af882109408e/docs/_static/fmriprep-workflow-all.png)), it does not appear in the usage notes or in the output documentation.

For this reason, it might first be necessary to explicitly inform users in the documentation about the time shift at many different locations and - importantly - explain the consequence for statistical analyses after fMRIPrep is completed. This could include an explicit warning box along the lines of “Caution! Slice timing correction shifts the data by half a TR. This means that you need to adjust your onset files, the microtime onset or the parameters in your first-level model function”.  The information could be included in “Processing Pipeline Details”, “Outputs of fMRIprep”, “Citing fMRIPrep” and in the methods template.

Here are some additional ideas on how to deal with the time-shift issue:

1. More information in the outputs, e.g. by emphasizing the new reference time in the output or by providing new time frames in the output (e.g. [1., 3., 5., ...] instead of [0., 2., 4., …] for TR = 2)

2. Adjust the STC reference to the first slice to make fMRIPrep more modular. This has to be accompanied by additional documentation to avoid similar errors in the other direction. 

[...]

PR #2520, opened 2021-08-26
ENH: Add flag for STC reference time and set in all cases

### LINKED ISSUE #2516: Slice timing correction options

effigies:
This PR partially address #2477 by providing a flag `--slice-time-ref` that can take a float from 0-1 or special values `"first"` and `"middle"`.

This also always sets the `-tzero` flag for 3dTshift (in cases where we run STC) to `min + frac * (max - min)`, which will generally be `frac * max`. This *should* be the same as 3dTshift's default of `mean(SliceTiming)`, as well as `frac * max`, but I update the text to be explicit as well.

Millisecond precision is used in reporting.

Closes #2516.

### DIFF
--- a/fmriprep/workflows/bold/stc.py
+++ b/fmriprep/workflows/bold/stc.py
@@ -107,7 +107,7 @@ BOLD runs were slice-time corrected to {tzero:0.3g}s ({frac:g} of slice acquisit
     inputnode = pe.Node(niu.IdentityInterface(fields=['bold_file', 'skip_vols']), name='inputnode')
     outputnode = pe.Node(niu.IdentityInterface(fields=['stc_file']), name='outputnode')
 
-    LOGGER.log(25, f'BOLD series will be slice-timing corrected to an offset of {tzero:.3g}.')
+    LOGGER.log(25, f'BOLD series will be slice-timing corrected to an offset of {tzero:.3g}s.')
 
     # It would be good to fingerprint memory use of afni.TShift
     slice_timing_correction = pe.Node(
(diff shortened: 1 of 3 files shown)


fmriprep/workflows/bold/stc.py
```diff
@@ -93,22 +93,29 @@ def init_bold_stc_wf(metadata, name='bold_stc_wf'):
     from niworkflows.engine.workflows import LiterateWorkflow as Workflow
     from niworkflows.interfaces.header import CopyXForm
 
+    slice_times = metadata["SliceTiming"]
+    first, last = min(slice_times), max(slice_times)
+    frac = config.workflow.slice_time_ref
+    tzero = first + frac * (last - first)
```
effigies, line 99: Does it make sense to explicitly round to millisecond resolution?

```suggestion
    tzero = np.round(first + frac * (last - first), 3)
```

effigies (commented):
Overall I think this is ready to go. Please let me know if I haven't addressed your concerns, @martinhebart.

fmriprep/workflows/bold/stc.py
effigies: Oh, @mgxd forgot to ask about this.

mgxd: it looks like afni's tshift assigns this as a [float datatype](https://github.com/afni/afni/blob/25e77d564f2c67ff480fa99a7b8e48ec2d9a89fc/src/3dTshift.c#L201), which supports up to 6 decimal digits of precision. though that may be excessive

effigies: True. I'm just imagining that any extra significant digits are rounding errors from performing this calculation. And we report it to `{:.03g}`, so this would ensure that what is sent to AFNI and reported to the user are the same.

mgxd: I have no strong opinion, fine with rounding to millisecond

nitschalex:
Thanks for adding this @effigies!

PR #2529, opened 2021-09-08
CI: Run pytest through GitHub actions

effigies:

I want to run pre-release tests on fMRIPrep to make sure that upcoming releases like numpy, nilearn or pybids aren't going to break things. GitHub actions seems the lightest-touch way of doing this.

~~Starting with stable tests. Will add pre-release if this works.~~

Started with stable tests, added pre-release tests. Added a test to ensure that we exercise pybids.

### DIFF
--- a/.github/workflows/pre-release.yml
+++ b/.github/workflows/pre-release.yml
@@ -1,0 +1,63 @@
+name: Pre-release checks
+
+on:
+  push:
+    branches:
+      - master
+      - maint/*
+  pull_request:
+    branches:
+      - master
+      - maint/*
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  pre-release:
+    # Check pre-releases of dependencies on stable Python
+    runs-on: ${{ matrix.os }}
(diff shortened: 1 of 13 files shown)


oesteban (approved):
This looks solid, thanks!

The only question is whether, going forward, we should try to use conda whenever possible instead of virtual environments and other solutions. As more and more of the tools we use are distributed with conda (and I'm thinking of ANTs as the most prominent example), we probably want CI and our Dockerfiles to leverage it more.

effigies:
If we're running tests that require these as dependencies, then sure. We don't currently have any pytest-runnable tests that depend on ANTs, do we?

oesteban:
> If we're running tests that require these as dependencies, then sure. We don't currently have any pytest-runnable tests that depend on ANTs, do we?

No, that's why I think this PR can and should be merged. But for the future, I would lean towards standardizing the conda setup, even if we have to pay some increased installation price. Having a homogeneous setup across projects will help maintainability through time.

Issue #2497, opened 2021-08-06
Warning should be ERROR - *Incorrect phase-encoding direction, assuming PA (posterior-to-anterior)*


oesteban:
When running "fieldmapless" we get this message *Incorrect phase-encoding direction, assuming PA (posterior-to-anterior)* when the phase encoding direction is not specified.

Although fieldmapless runs without knowledge of the polarity, applying estimation on the wrong axis (when assuming AP/PA) could be extremely problematic.

This warning, IMHO, should be an error.

This affects the 20.2.x series and will likely become irrelevant with the integration of SDCFlows 2.0 (the new API makes the phase encoding direction a required parameter).

effigies:
This has been done 21.0. I would vote for keeping the behavior as-is for 20.2.

oesteban:
Sounds reasonable.

PR #2530, opened 2021-09-09
FIX: Restore SyN-SDC

effigies:


fmriprep/workflows/base.py
```diff
@@ -312,27 +312,34 @@ def init_single_subject_wf(subject_id):
...
         )
 
         config.loggers.workflow.debug(
             f"{len(fmap_estimators)} fieldmap estimators found: "
             f"{[e.method for e in fmap_estimators]}"
         )
 
+        if "fieldmaps" in config.workflow.ignore:
+            fmap_estimators = [f for f in fmap_estimators
+                               if f.method == fm.EstimatorType.ANAT]
```
mgxd, line 337: i don't think `fm` is defined yet here

fmriprep/workflows/base.py
```diff
@@ -424,13 +431,44 @@ def init_single_subject_wf(subject_id):
...
+
+        elif estimator.method == fm.EstimatorType.ANAT:
+            from sdcflows.workflows.fit.syn import init_syn_preprocessing_wf
+            sources = [str(s.path) for s in estimator.sources if s.suffix == "bold"]
+            layout = config.execution.layout
+            syn_preprocessing_wf = init_syn_preprocessing_wf(
+                omp_nthreads=config.nipype.omp_nthreads,
+                debug=config.execution.sloppy,
+                auto_bold_nss=True,
+                t1w_inversion=True,
```
mgxd, line 448: an aside: i don't really understand this option - why are we inverting the t1w? cc @oesteban

mgxd (commented):
Had a quick look, but looks like ds005 doesn't undergo any fieldmap processing, despite the inclusion of the `--use-syn-sdc` flag. Not sure why though..

effigies:
Yeah, I'm trying to debug the workflow creation to figure out what's failing.

fmriprep/workflows/base.py
effigies: Because T2* contrast is generally inverted from T1w, we invert to give the registration algorithm a more similar target.

We haven't really tested keeping it static and using, e.g., a mutual information metric.

effigies:
So I think the issue is that we now require `PhaseEncodingDirection` to run. Adding it to ds000005, I start accessing these paths.

effigies:
@oesteban It seems that we're no longer able to run SyN without the following defined for a BOLD file:

* `PhaseEncodingDirection`
*  One of:
    * `TotalReadoutTime`
    * `EffectiveEchoSpacing`
    * `EchoSpacing` and `ParallelReductionFactorInPlane`
    * `WaterFatShift` and `EPIFactor`

Presumably we should be able to have some kind of fallback, since we were able to do it before. What would be the best way to do this in the new SDCFlows?

effigies:
Okay, patched in the metadata. It's a little hacky, but I couldn't readily see a cleaner way of doing it.

effigies:
Well, it runs, but it's a substantial regression:

Screenshot from 2021-09-10 16-57-41
Screenshot from 2021-09-10 16-57-48
Screenshot from 2021-09-10 16-58-02
Screenshot from 2021-09-10 16-58-17

oesteban:
Sorry for the slow response.

> @oesteban It seems that we're no longer able to run SyN without the following defined for a BOLD file:
> 
>     * `PhaseEncodingDirection`
> 
>     * One of:
>       
>       * `TotalReadoutTime`
>       * `EffectiveEchoSpacing`
>       * `EchoSpacing` and `ParallelReductionFactorInPlane`
>       * `WaterFatShift` and `EPIFactor`
> 
> 
> Presumably we should be able to have some kind of fallback, since we were able to do it before. What would be the best way to do this in the new SDCFlows?

This is intended. You could convince me to allow setting `TotalReadoutTime` to 1.0 if it is not found in the metadata, but we should stop inferring AP/PA when phase encoding direction is not present. SDCFlows (or any other tool) should not by any means attempt correction with the assumption that distortion only happens along one axis if that axis is unknown.

The second part of the problem (the readout time) could be worked-around - but big fat warnings should be set in place. Here's the problem: now SDCFlows calculates a fieldmap and the actual value of the readout time is necessary to write out the fieldmap in Hz. Since it is a scaling factor, if all runs of the dataset are acquired with exactly the same readout time, it can safely be set to 1.0 in estimation of the fieldmap and reconstruction of the distortion. However, these fieldmaps should not be used for anything else (e.g., averaging fieldmaps to calculate a new "standardized" average across studies) because the units (or more precisely, the scale) are unknown.

>  a substantial regression:

This is expected, there are some substantial changes made to the workflow that have made it very different from the original implementation. The most relevant are:
- **New reference workflow**. With the new epi-reference workflow, the contrast of the reference has a little different distribution and I believe the registration parameters are now more sensible (which in the long term is good, but in the short term justifies having bad results).
- **Rotation to canonical**. For oblique datasets our fieldmap less registration scheme was theoretically flawed because ITK/ANTs work on physical coordinates and the PE axis is therefore not aligned anymore with the data array axis. This may have introduced a strong modification on the workflow (and is prone to bugs).
- I made the tunning with DWI's b-zeros.

In addition to this, I don't remember exactly whether the T1w inversion is enabled by default. I think in the long term it could be disabled when using mutual information, but I don't know what the tool is doing at the moment by default.

Overall, I'll have a deep look into this later this week. I apologize for the disappointment it must be to see this large regression up front.

oesteban:
Ha, I forgot I had already expressed my concern about the missing PE - #2497.

oesteban:
I'm picking up on where you left this. The first outstanding problem is that the syn registration was not correctly initialized.

I reorganized the syn workflow in two parts:

- [preprocessing](https://github.com/nipreps/sdcflows/blob/4b464fac3ca63af7a8596265ce630c29917849d5/sdcflows/workflows/fit/syn.py#L342-L898)
- [actually fitting](https://github.com/nipreps/sdcflows/blob/4b464fac3ca63af7a8596265ce630c29917849d5/sdcflows/workflows/fit/syn.py#L94-L339)

The idea is that the preprocessing massages the T1w and brings it into rough alignment with the EPI, through an Affine transform. However, for ds005 this registration step was far from acceptable. I am playing now with better parameters that bring the T1w into a seemingly good alignment into the EPI (for subject 02 of ds005). I'll keep playing with this.

oesteban:
I'm still testing on ds005/sub-01 locally, but this is getting closer to the end line. Results from SDCFlows are starting to seem sufficiently okay to go forward with this PR and separate the problem it its two elements: actual integration (to be addressed within fMRIPrep) and internals and improvement (to be addressed in SDCFlows proper).

I will update as soon as I get new local results.

oesteban:
@mgxd @effigies this is ready for merge, I believe. There's one issue though (see  #2606) that has held me back to apply the fieldmap estimation to ME data. In other words, in the results for ds210 on Circle you will find that the SyN estimation is executed and produces the expected outputs, but the fieldmaps are not applied.

effigies:
Mask is much improved after clearing the working directory.

PR #2540, opened 2021-09-15
ENH: Add SliceTimingCorrected and StartTime fields to BOLD metadata

### LINKED ISSUE #2539: Metadata indicating whether or not slice-timing correction has been performed
We've been discussing setting metadata to indicate the temporal offset that slice-timing correction induces in a BOLD series, but there would be no distinction in that metadata between not performing STC and aligning to the first slice (offset 0).

It may still be useful to indicate whether STC has been applied. The use case that was presented to me is that there are workflows that average over the acquisition window, rather than sampling the HRF in the center, which would be invalid when STC has realigned all signal to a single time point.

The final metadata would end up in BIDS functional derivatives, but it seems worthwhile to get a proof-of-concept in fMRIPrep ASAP.

effigies:

Closes #2539.
Addresses #2477.
Implements provisional https://github.com/bids-standard/bids-specification/issues/836

This does not update the `toffset` field in NIfTI at present.

One thing to address: when we're calculating the acquisition time (TA), we use `st[-1] + (st[1] - st[0])`, which treats each slice as having a duration. When we're calculating tzero, we use `(st[-1] - st[0]) * frac + st[0]`, which treats each slice as an instantaneous acquisition. So tzero is a half "slice duration" less than half TA (`tzero = (t[0] + TA) / 2 - (t[1] - t[0]) / 2`).

Our tzero calculation is consistent with AFNI's default, but while we're here, maybe we want to adjust it to be `tzero = (t[0] + TA) / 2`. Or is the calculation of TA incorrect?

cc @martinhebert @theoschaefer @nitschalex @oliver-contier @kellyhennigan

### DIFF
--- a/fmriprep/workflows/bold/outputs.py
+++ b/fmriprep/workflows/bold/outputs.py
@@ -42,19 +42,45 @@ def prepare_timing_parameters(metadata):
     Examples
     --------
 
+    .. testsetup::
+
+        >>> from unittest import mock
+
+    If SliceTiming metadata is absent, then the only change is to note that
+    STC has not been applied:
+
     >>> prepare_timing_parameters(dict(RepetitionTime=2))
-    {'RepetitionTime': 2}
+    {'RepetitionTime': 2, 'SliceTimingCorrected': False}
     >>> prepare_timing_parameters(dict(RepetitionTime=2, DelayTime=0.5))
-    {'RepetitionTime': 2, 'DelayTime': 0.5}
-    >>> prepare_timing_parameters(dict(RepetitionTime=2, SliceTiming=[0.0, 0.2, 0.4, 0.6]))
-    {'RepetitionTime': 2, 'DelayTime': 1.2}
+    {'RepetitionTime': 2, 'DelayTime': 0.5, 'SliceTimingCorrected': False}
     >>> prepare_timing_parameters(dict(VolumeTiming=[0.0, 1.0, 2.0, 5.0, 6.0, 7.0],
     ...                                AcquisitionDuration=1.0))
-    {'VolumeTiming': [0.0, 1.0, 2.0, 5.0, 6.0, 7.0], 'AcquisitionDuration': 1.0}
-    >>> prepare_timing_parameters(dict(VolumeTiming=[0.0, 1.0, 2.0, 5.0, 6.0, 7.0],
-    ...                                SliceTiming=[0.0, 0.2, 0.4, 0.6, 0.8]))
-    {'VolumeTiming': [0.0, 1.0, 2.0, 5.0, 6.0, 7.0], 'AcquisitionDuration': 1.0}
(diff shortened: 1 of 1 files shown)


theoschaefer:
Hi Chris,

I just remember that SPM defines TA as `TA = TR - (TR / number_of_slices)` (e.g. [see here](http://andysbrainblog.blogspot.com/2012/11/slice-timing-correction-in-spm.html)). That would result in a shorter TA (minus one slice duration) compared to your calculation. I do not know if treating slices as durations or instances is better (although intuitively durations makes more sense to me), but I would say consistency is most important. Thus, I would recommend to change either the tzero calculation to `tzero = (st[0] + TA) / 2` as you suggested or treat slices as instantaneous acquisitions and keep the old tzero calculation.

fmriprep/workflows/bold/outputs.py
```diff
@@ -42,26 +42,55 @@ def prepare_timing_parameters(metadata):
...
+    and the ``StartTime`` indicates a series offset.
+
+    >>> with mock.patch("fmriprep.config.workflow.ignore", []):
+    ...     prepare_timing_parameters(dict(RepetitionTime=2, SliceTiming=[0.0, 0.2, 0.4, 0.6]))
+    {'RepetitionTime': 2, 'SliceTimingCorrected': True, 'DelayTime': 1.2, 'StartTime': 0.3}
+    >>> with mock.patch("fmriprep.config.workflow.ignore", []):
+    ...     prepare_timing_parameters(dict(VolumeTiming=[0.0, 1.0, 2.0, 5.0, 6.0, 7.0],
+    ...                                    SliceTiming=[0.0, 0.2, 0.4, 0.6, 0.8]))
+    {'VolumeTiming': [0.0, 1.0, 2.0, 5.0, 6.0, 7.0], 'SliceTimingCorrected': True,
+     'AcquisitionDuration': 1.0, 'StartTime': 0.4}
```
oesteban, line 71:
```suggestion
    {'VolumeTiming': [0.0, 1.0, 2.0, 5.0, 6.0, 7.0], 'SliceTiming'=[0.4, 0.4, 0.4, 0.4, 0.4],
     'AcquisitionDuration': 1.0}
```
What are your thoughts about making it explicit that the `SliceTiming` metadata is a list of uniform values? This would be readily supported by BIDS and leaves no room for confusion - even if you try to run STC over again on these data, the algorithm should notice you are sampling on-grid all the time.

oesteban (commented):
Just a reminder of the option we talked about in a meeting some days ago. Once that idea is set back, this LGTM.

fmriprep/workflows/bold/outputs.py
effigies: I opened an issue on BIDS: https://github.com/bids-standard/bids-specification/issues/879

oesteban (approved):

Issue #2210, opened 2020-07-08
Susceptibility distortion correction (SDC) using Spin Echo Fieldmaps 


sameera2004:
Hi,

I pre-processed resting state data using fMRIprep 20.1.1. This data set contain 20 subjects, 2 sessions for each subject, and 4 runs for each session. I used SpinEcho field maps to do the susceptibility distortion correction (SDC). The SDC with SpinEcho’s worked for majority of the data. However, it worsen the SDC’s in certain subjects. Below is the command that I used to run fMRIprep;

singularity run --cleanenv --bind /mnt/fMRIprep /mnt/tools/fMRIprep_sing/fmriprep-20.1.1.simg /mnt/fMRIprep/scR21 /mnt/fMRIprep/scR21_outputs -w /mnt/fMRIprep/scR21_outputs/derivatives/scratch --fs-license-file $PWD/license.txt --write-graph --cifti-output 91k participant --participant-label 50015 50016 50020 50024 50027 50033 50042 50137 50144 --ignore slicetiming --output-spaces T1w MNI152NLin2009cAsym:res-2 fsaverage fsLR
I am wondering if anybody else has experienced similar issues. Please let me know.

Thanks in advance
Best Regards
Sameera

<img width="920" alt="Screen Shot 2020-07-07 at 5 20 14 PM" src="https://user-images.githubusercontent.com/5701652/86956918-e5b5be80-c127-11ea-975d-e79d86ff836c.png">
<img width="943" alt="Screen Shot 2020-07-07 at 5 20 25 PM" src="https://user-images.githubusercontent.com/5701652/86956925-e8181880-c127-11ea-8fdf-ac603c8df1ee.png">

oesteban:
Hi @sameera2004, I assume that the above is before SDC and the below shows the image after SDC? If that is the case, you might have a wrong `PhaseEncodingDirection` parameter in your sidecar JSON. Seems the correction was executed in the opposed direction.

jxvansne:
Hi @oesteban, Sameera works in my lab. I believe the two images she posted were an AP and a PA phase encode direction run (we alternate directions from one run to the next, HCP style) after SDC for a single subject.

[...]

PR #2547, opened 2021-09-16
ENH: Refactor workflow to adopt *SDCFlows*' new correction workflow

oesteban:
This PR addresses the problem reported in #2210.

Along with the specific changes in SDCFlows, this PR:

- Resamples BOLD data in its native space using new SDCFlows' utilities.
- Introduces a new ``IdentityInterface`` node called ``bold_final`` that acts as buffer to reduce LoC and branches when dealing with data with/without fieldmaps (and helpful too for ME-EPI, although this is not fully implemented).
- ROIs plot now uses the output of unwarping through a `RobustAverage` node, therefore fixing one issue reported in #2210.
- preproc datasink of BOLD in native space now connected to the ``bold_final`` node, which contains HMC'ed + SDC'ed (if fieldmaps found) BOLD in native space. Now, it is matched with the ROIs plot, reducing a little the opportunities for divergence.

**This PR DOES NOT solve (or even leaves without implementation)**:
- Resampling into T1w and standard spaces when fieldmap is present - the same error for the bold data reported in #2210 is expected to happen, as the displacements field generated by *SDCFlows* does not look right all times (nipreps/sdcflows#236)
- Mapping of surfaces into EPI
- ME-EPI + fieldmaps (still trying to resolve problems with single-echo, will look into this at a later time).
-

### DIFF
--- a/fmriprep/workflows/bold/base.py
+++ b/fmriprep/workflows/bold/base.py
@@ -620,7 +620,6 @@ Non-gridded (surface) resamplings were performed using `mri_vol2surf`
             (inputnode, func_derivatives_wf, [
                 (("bold_file", combine_meepi_source), "inputnode.source_file"),
             ]),
-            (join_echos, bold_final, [("bold_files", "bold")]),
             (join_echos, bold_t2s_wf, [("bold_files", "inputnode.bold_file")]),
             (bold_t2s_wf, split_opt_comb, [("outputnode.bold", "in_file")]),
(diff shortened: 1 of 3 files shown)


oesteban:
@mgxd I've rebased on master after running black on ``workflows/bold/base.py``, so the diff should be much cleaner now. Some little stylistic changes remained, but now they are just residual changes I don't think will disturb the review process.

oesteban:
This is holding me up a little, so I'll merge.

PR #2551, opened 2021-09-20
FIX: Shrink Docker image size

mgxd:
Addresses #2545

Exclude and remove many unneeded binaries / files.

To cut down the size a bit further, we may want to `strace` and compile a list of the freesurfer / fsl binaries we are using, as there are still many included that are not being used.

Building off this PR:

```
$ docker images

REPOSITORY                           TAG                 IMAGE ID       CREATED          SIZE
nipreps/fmriprep                     size                e0634fde686b   21 minutes ago   13.9GB
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -129,26 +129,68 @@ RUN apt-get update -qq \
     && mkdir -p /opt/fsl-5.0.11 \
     && curl -fsSL --retry 5 https://fsl.fmrib.ox.ac.uk/fsldownloads/fsl-5.0.11-centos6_64.tar.gz \
     | tar -xz -C /opt/fsl-5.0.11 --strip-components 1 \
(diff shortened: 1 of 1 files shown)


mgxd:
Okay, this is passing and reduces the image size substantially. Only fear is some path not being tested by our 3 datasets and the dependency is missing, but I think it would be relatively easy to add it back in.

Ideally, we would want to bundle these steps to avoid download + filtering, but I don't know if there'd be any licensing issues to consider.

effigies:
In principle it seems fine, especially if we're going to do another RC. Speaking of which, do we have things to get in before an RC?

effigies:
We've got a few fixes where the workflow wasn't running or the docker image itself was difficult to use, so I think let's cut one today.

PR #2567, opened 2021-10-01
REL: 20.2.4

### LINKED ISSUE #2292: DOC: Repeat that internal spaces not listed within --output-spaces will not be written out to derivatives
> 2) there is a special call out pointing out that data in spaces needed for intermediate calculations (like MNI152NLin6 for aroma) which are not specified in the output-spaces list are deleted after the results are resampled. That should be repeated in the last line of “Preprocessing blocks depending on standard templates”.

### LINKED ISSUE #2507: BUG: scl_slope/scl_inter in mask and dseg headers are set to nan, should be 1/0
Dear fmriprep experts,

I have just finished the preprocessing and everything seems alright since no errors were reported.
However, I have noticed that I cannot open the “sub-XX_desc-brain_mask.nii” file.

I use MRIcron to open these images and when I click on it it says “Invalid floating point operation. Press OK to ignore and risk data corruption. Press Abort to kill the program”.

Basically, I cannot even explore the mask since it contains NaN. Any idea why that is the case? 
I have notices WARNINGS during the anatomical and functional processing in which data types were converted from float32 to uint8 for the brain masks. These are the only files I cannot open. I wonder whether the two things are related

I am looking forward to hearing from your opinions
best wishes
Loris

effigies:
Release prep for 20.2.4.

Closes #2292.
Closes #2507.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,20 @@
+20.2.4 (October 01, 2021)
+=========================
+Bug-fix release in the 20.2.x LTS series.
+
+This release includes some fixes to minor bugs, and includes enhancements
+to allow the user to specify a different slice-timing-correction (STC) target.
+
+Each BOLD series will contain ``SliceTimingCorrected`` metadata to indicate
+whether STC was performed. If performed, the ``StartTime`` metadata field will
+be set to indicate the onset time of the BOLD series.
+
+  * FIX: Avoid unnecessary connections based on branching logic (#2508)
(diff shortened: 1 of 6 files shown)


CHANGES.rst
```diff
@@ -1,3 +1,20 @@
+20.2.4 (October 01, 2021)
```
effigies, line 1: Took longer to get to this point than I thought. Should probably wait until Monday to release.

```suggestion
20.2.4 (October 04, 2021)
```

PR #2570, opened 2021-10-04
CI: Add style checks (missing since Travis got throttled)

effigies:
Had to kill and re-tag 20.2.4 because I found a bug with `flake8`, which made me realize we aren't doing style checks anymore.

Adding in a very simple way to `maint/20.2.x`. Once merged into `master`, I will move this inside the `.maint/ci` checks.


effigies:
@mgxd @oesteban Any objections?

mgxd (approved):

PR #2576, opened 2021-10-06
ENH: Better integration of *SDCFlows*' unwarping

oesteban:
This PR builds on top of #2547, unifying the final BOLDref computation.


fmriprep/workflows/bold/base.py
```diff
@@ -623,6 +641,10 @@ def init_func_preproc_wf(bold_file, has_fieldmap=False):
             (join_echos, bold_t2s_wf, [("bold_files", "inputnode.bold_file")]),
             (bold_t2s_wf, split_opt_comb, [("outputnode.bold", "in_file")]),
             (split_opt_comb, bold_t1_trans_wf, [("out_files", "inputnode.bold_split")]),
+            (bold_t2s_wf, bold_final, [
+                ("outputnode.bold", "bold"),
+                ("outputnode.t2star_map", "boldref"),
```
effigies, line 646: Note that we should not be using `t2star_map` as the reference BOLD image. See #2574.

PR #2605, opened 2021-10-18
FIX: Optimize build job on CircleCI

oesteban:
Pulling miniconda is a waste of time for most Circle builds, since it will only be necessary when that particular step of the docker build needs to be re-executed.


effigies:
Looks like it's getting pulled anyway.

```
 ---> d664ee2c3874
Step 23/44 : COPY --from=nipreps/miniconda:py38_1.3.2 /opt/conda /opt/conda
py38_1.3.2: Pulling from nipreps/miniconda

afab3874: Pulling fs layer 
6f40b3be: Pulling fs layer 
10c60a31: Pulling fs layer 
8d7a2e34: Pulling fs layer 
330ada3c: Pulling fs layer 
a77f238c: Pulling fs layer 
Digest: sha256:4d0dc0fabb794e9fe22ee468ae5f86c2c8c2b4cd9d7b7fdf0c134d9e13838729
Status: Downloaded newer image for nipreps/miniconda:py38_1.3.2
 ---> Using cache
```

I wonder if using the hash in the `--from=` bit will let it recognize that it doesn't need to fetch? e.g.

```Dockerfile
COPY --from=nipreps/miniconda@sha256:4d0dc0fabb794e9fe22ee468ae5f86c2c8c2b4cd9d7b7fdf0c134d9e13838729 /opt/conda /opt/conda
```

oesteban:
We cross-posted - let me try the digest idea.

oesteban:
> Hmm. No difference.

I would have expected that. I need to send an empty commit and force a rebuild to be sure it's not going to pick the cached layer.

Issue #2591, opened 2021-10-12
Failed registrations between BOLD and T1w


HippocampusGirl:
I am experiencing a significant number of failed registrations between BOLD and T1w for multiple datasets. We noticed that the images appear to be rotated basically in the wrong direction after registration. 

We also re-ran some of the registrations with FSL BBR and ANTs, which worked without any problem.

Sadly we are not able to share any of the underlying data, but I am happy to help diagnose the issue.

<img width="738" alt="screenshot" src="https://user-images.githubusercontent.com/789054/136358195-446a0393-2de0-4b36-ba33-b4db7c1673b5.png">

### What version of fMRIPrep are you using?

20.2.3

### What kind of installation are you using? Containers (Singularity, Docker), or "bare-metal"?

Singularity, via [HALFpipe](https://github.com/HALFpipe/HALFpipe/)

### What is the exact command-line you used?

singularity run --containall --bind /:/ext halfpipe-halfpipe-1.2.0.sif

### Are you reusing previously computed results (e.g., FreeSurfer, Anatomical derivatives, work directory of previous run)?

No, we were running from scratch

### Have you checked that your inputs are BIDS valid?

They seem to be missing some metadata, but are valid.

![remmina_mindandbrain server_localhost_20211012-6165](https://user-images.githubusercontent.com/789054/136901918-6e6b9a48-034f-43d6-932f-12ecdb060691.png)

### Do you know if this used to work in the past? 

We were also having these issues with version 20.2.1, but were hoping that #2307 would fix it.

### Are you running it with the --fs-no-reconall flag? 

Yes

### And are you running with susceptibility distortion correction?

No

### Did fMRIPrep generate the visual report for this particular subject? If yes, could you share it?

See screenshot above. I was asked not to share the full image, please let me know if the provided information is sufficient for diagnosing the issue.

### Can you find some traces of the error reported in the visual report (at the bottom) or in *crashfiles*?

There are no crashfiles.

### fMRIPrep log
If you have access to the output logged by fMRIPrep, please make sure to attach it as a text file to this issue.

The log file is quite large for the entire dataset. I will re-run a single subject to generate a smaller one and post that.

oesteban:
> Sadly we are not able to share any of the underlying data, but I am happy to help diagnose the issue.

But, you should be able to share the visual report, right? :D

[...]

PR #2607, opened 2021-10-19
FIX: registration issue between T1w and BOLD in specific datasets

HippocampusGirl:

- Switch to using `mri_coreg` in `fsl_bbr_wf` instead of `flirt` as suggested by @effigies and @oesteban in #2591
- Add `-basescale 1` parameter to the `flirt` boundary-based registration call to avoid the scaling behavior there too. As far as I can tell, this does not have side effects, but it may make sense to wait for a response from the `flirt` developers before merging this one.

No changes to documentation

### DIFF
--- a/fmriprep/workflows/bold/registration.py
+++ b/fmriprep/workflows/bold/registration.py
@@ -140,7 +140,8 @@ def init_bold_reg_wf(
                                bold2t1w_init=bold2t1w_init, omp_nthreads=omp_nthreads)
     else:
         bbr_wf = init_fsl_bbr_wf(use_bbr=use_bbr, bold2t1w_dof=bold2t1w_dof,
-                                 bold2t1w_init=bold2t1w_init, sloppy=sloppy)
+                                 bold2t1w_init=bold2t1w_init, sloppy=sloppy,
+                                 omp_nthreads=omp_nthreads)
 
     workflow.connect([
         (inputnode, bbr_wf, [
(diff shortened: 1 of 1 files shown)


effigies:
Hmm. If we're adding `-basescale 1` to the BBR step, then it feels like adding `-basescale 1` to the initial FLIRT step is less intrusive than switching to `mri_coreg`.

Looking at [the source](https://github.com/InstitutoDOr/FriendENGINE/blob/cb29c1171c91a5c5e26d94b63bca9505ce171fb7/fsl/flirt/flirt.cc#L144-L162):

```C++
void set_basescale(const string& filenameA, const string& filenameB)
{
  if (!(gOptions->force_basescale)) { 
    // only try to automatically determine if it was not requested by the user
    volume<float> volA, volB;
    read_volume_hdr_only(volA,filenameA);
    read_volume_hdr_only(volB,filenameB);
    float maxdimA = Max(Max(volA.xdim(),volA.ydim()),volA.zdim());
    float maxdimB = Max(Max(volB.xdim(),volB.ydim()),volB.zdim());
    if (Max(maxdimA,maxdimB)>12) {  // over 150% of largest (8mm) scale
      gOptions->basescale = Max(maxdimA,maxdimB)/8;
    }
    float mindimA = Min(Min(volA.xdim(),volA.ydim()),volA.zdim());
    float mindimB = Min(Min(volB.xdim(),volB.ydim()),volB.zdim());
    if (Min(mindimA,mindimB)<0.75) { // between 0.5 and 1 mm 
      gOptions->basescale = Min(mindimA,mindimB);
    }
  }
}
```

`-basescale 1` only has an effect if a voxel dimension is > 12mm or <0.75mm, so I think we can be reasonably confident that the number of datasets affected will be quite small.

I'm okay moving to `mri_coreg` if everyone wants to keep on that way, but I think I have a slight preference for just setting `-basescale 1` on both steps of the FSL pipeline for now, and then moving to a common pipeline in future versions.

Sorry for jumping one way and then the other.

mgxd:
can this be closed now that #2624 and #2625 are in?

PR #2624, opened 2021-11-11
FIX: Add `-basescale 1` parameter to avoid `flirt` scaling

effigies:
- Disables `flirt` scaling of input transformation matrices, which may
  lead to issues with high-resolution T1w data (see #2591)

This is #2607 without the switch to `mri_coreg`. That can be done in 21.0.0.

Doing this first so we can get the mri_coreg fix into the next 21.0 RC.

cc @HippocampusGirl

### DIFF
--- a/fmriprep/workflows/bold/registration.py
+++ b/fmriprep/workflows/bold/registration.py
@@ -652,7 +652,8 @@ for distortions remaining in the BOLD reference.
     wm_mask = pe.Node(niu.Function(function=_dseg_label), name='wm_mask')
     wm_mask.inputs.label = 2  # BIDS default is WM=2
     flt_bbr_init = pe.Node(FLIRTRPT(dof=6, generate_report=not use_bbr,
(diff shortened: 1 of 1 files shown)


mgxd:
should we port this into `nibabies` as well?

effigies:
Are we using FLIRT instead of always running FreeSurfer?

mgxd:
Yes, in the current version `mri_coreg` was disabled

effigies:
It will only impact BOLD series with 0.75mm voxels or smaller, though, and I don't think I've seen anybody pushing those bounds on infants. Still no harm in doing it, either.

eilidhmacnicol:
I've just came across this PR because I was having trouble registering some bold images to T2w structurals in rats (we had taken out most freesurfer dependencies because most rodent researchers are unlikely to use it). Tried adding the argument and it wholly improved the flirt results, so thanks for documenting! cc @oesteban

PR #2630, opened 2021-11-18
FIX: Threshold boldref resampled values to be non-negative

effigies:
https://neurostars.org/t/negative-data-following-multi-echo-optimal-combination-in-fmriprep/20698

The simplest place to do this seemed to be post-resample, pre-merge. This way it applies to single-echo and multi-echo data identically, with no possibility of missed branches during workflow construction.

I chose to make a simple Python interface rather than use fslmaths just because I didn't want to reason about possible type coercion.

Should we be doing this in all resampling workflows?


fmriprep/workflows/bold/resampling.py
```diff
@@ -620,7 +628,8 @@ def init_bold_preproc_trans_wf(
                                      (("bold_file", _first), "reference_image")]),
         (inputnode, merge, [("name_source", "header_source")]),
         (merge_xforms, bold_transform, [("out", "transforms")]),
-        (bold_transform, merge, [("out_files", "in_files")]),
+        (bold_transform, threshold, [("out_files", "in_file")]),
```
oesteban, line 631: This would actually threshold the "preprocessed" echos, right? (which is actually the fix to the problem, as the boldref is not the target here).

oesteban (commented):
Looks good in principle, although we may have this interface already in niworkflows (or something really close) and we may have a patched version of antsApplyTransforms that runs this thresholding too - worth having a look and confirming I'm wrong before merging.

fmriprep/interfaces/maths.py
```diff
@@ -0,0 +1,44 @@
...
+    in_file = File(exists=True, mandatory=True, desc="Input imaging file")
+    out_file = File(desc="Output file name")
+    threshold = traits.Float(0., usedefault=True, desc="Values under this become 0")
+
+
+class ThresholdOutputSpec(TraitedSpec):
+    out_file = File(desc="Output file name")
+
+
+class Threshold(SimpleInterface):
```
mgxd, line 16: I think if we are having a `Threshold` interface, it needs to handle upper/lower bounds. Alternatively, this could be named `LowerThreshold` instead to avoid any confusion.

fmriprep/interfaces/maths.py
```diff
@@ -0,0 +1,44 @@
+import os
+from nipype.interfaces.base import SimpleInterface, TraitedSpec, traits, File
+from nipype.utils.filemanip import fname_presuffix
+
+
+class ThresholdInputSpec(TraitedSpec):
+    in_file = File(exists=True, mandatory=True, desc="Input imaging file")
+    out_file = File(desc="Output file name")
+    threshold = traits.Float(0., usedefault=True, desc="Values under this become 0")
```
mgxd, line 9: Personally, not a fan of having a default for this interface - makes it more unclear from an initialization perspective

```
thresh = Threshold()
```
vs
```
thresh = Threshold(threshold=0)
```

fmriprep/workflows/bold/resampling.py
effigies: Yes, this thresholds the preprocessed echos (or the preprocessed BOLD, if single echo).

effigies:
@mgxd Switched to a clipping interface.

@oesteban We use MultiApplyTransforms, which definitely doesn't threshold. I would rather keep it explicit than patch.

Should we also be thresholding after `MultiApplyTransforms` when resampling to T1w and template?

mgxd (approved):
LGTM, just need to add in the missing imports

fmriprep/interfaces/maths.py
```diff
@@ -0,0 +1,48 @@
...
+    If no values are outside the bounds, nothing is done and the in_file is passed
+    as the out_file without copying.
+    """
+    input_spec = ClipInputSpec
+    output_spec = ClipOutputSpec
+
+    def _run_interface(self, runtime):
+        import nibabel as nb
+        img = nb.load(self.inputs.in_file)
+        data = img.get_fdata()
```
mgxd, line 32: won't this potentially convert data to `float64`?

mgxd (commented):
Actually, that may have been pre-emptive...

fmriprep/interfaces/maths.py
effigies: Sure, but it's only one volume at a time. It will not save the outputs as float64, since we preserve the header.

mgxd: ah, then it's fine as is

PR #2647, opened 2021-12-02
MAINT: Upgrade Docker to FSL 6

mgxd:
Local test passed, so let's see how this goes.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -69,7 +69,7 @@ RUN curl -sSL https://surfer.nmr.mgh.harvard.edu/pub/dist/freesurfer/6.0.1/frees
     --exclude='freesurfer/trctrain'
 
 # Simulate SetUpFreeSurfer.sh
-ENV FSL_DIR="/opt/fsl-5.0.11" \
+ENV FSL_DIR="/opt/fsl-6.0.5.1" \
     OS="Linux" \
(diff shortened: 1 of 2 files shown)


mgxd:
For reference, here are a few reports of a subject using `flirt -bbr`

current master (5929041): https://3428-53175327-gh.circle-artifacts.com/0/tmp/ds210/derivatives/fmriprep/sub-02.html
fsl6: https://3427-53175327-gh.circle-artifacts.com/0/tmp/ds210/derivatives/fmriprep/sub-02.html

I don't think this is as controversial as initially believed.

oesteban:
The problem is FSL bet - it easily gets stuck with long paths if I recall correctly. I thought there was an open issue on nipype, but I can't seem to find it. Was the problem perhaps addressed?

mgxd:
does d3a4d33 address the name concerns, or were you referring to something different?

effigies:
@oesteban Just a bump on this. Should be a very quick yes/no.

effigies:
@mgxd Can you open PRs (they won't need to be merged) in niworkflows, sdcflows and smriprep? Let's give their CIs a chance to find errors, too.

effigies:
Okay, I think we've tested with all dependencies...

PR #2649, opened 2021-12-03
MAINT: Reduce redundant processing of ds005

mgxd:
The `ds005` circle runs are currently taking > 3h to process, and we recently started hitting a cap at 3h.

This attempts to lighten the processing load, and indirectly cache size.


mgxd:
decreases the build by an hour, getting us passed the current cap. merging

PR #2656, opened 2021-12-09
FIX: Re-group and space boilerplate text

### LINKED ISSUE #2526: Verify generated boilerplate accuracy
See: https://github.com/nipreps/fmriprep/pull/2392#issuecomment-910408135

effigies:
The big issue was that the functional heading was attached to the anat postproc. Prepending it to all functional workflows ensures that it's always kept local. The fmap workflow needs an extra newline to be placed correctly, as well.

While testing this, I ran across the problem that the PKG_PATH in the wrapper didn't keep up with the changes in the docker file. I thought I'd fixed that in another PR, but I either never included it or it hasn't been merged yet.

Closes #2526.

### DIFF
--- a/fmriprep/workflows/base.py
+++ b/fmriprep/workflows/base.py
@@ -356,8 +356,7 @@ It is released under the [CC0]\
 
     # Append the functional section to the existing anatomical exerpt
     # That way we do not need to stream down the number of bold datasets
-    anat_preproc_wf.__postdesc__ = (anat_preproc_wf.__postdesc__ or '') + """
-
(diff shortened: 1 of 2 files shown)


effigies:
ds210 shows correct order. ds054 shows fmap before anatomical. Probably due to the non-SyN fmaps not being dependent on anat, so the topological sort can be arbitrary.

mgxd (approved):
I think this is a good workaround - ultimately we should add some mechanism of ordering to `LiterateWorkflow`.

PR #2684, opened 2022-01-13
FIX: Re-enable reverse-PE EPI fieldmaps

effigies:
Requires nipreps/sdcflows#258.

Started testing on ds000223, and the workflow looks reasonable, but the affines are slightly different between EPI and BOLD. I'll look for another dataset or just hack the affines for now.

### DIFF
--- a/fmriprep/workflows/base.py
+++ b/fmriprep/workflows/base.py
@@ -439,20 +439,20 @@ Setting-up fieldmap "{estimator.bids_id}" ({estimator.method}) with \
         if estimator.method in (fm.EstimatorType.MAPPED, fm.EstimatorType.PHASEDIFF):
             continue
 
-        suffices = set(s.suffix for s in estimator.sources)
+        suffices = [s.suffix for s in estimator.sources]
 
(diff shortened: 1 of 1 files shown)


fmriprep/workflows/base.py
```diff
@@ -439,20 +439,20 @@ def init_single_subject_wf(subject_id):
...
-            ]
-            getattr(fmap_wf.inputs, f"in_{estimator.bids_id}").metadata = [
-                s.metadata for s in estimator.sources
-            ]
-
-        elif estimator.method == fm.EstimatorType.PEPOLAR:
-            raise NotImplementedError(
-                "Sophisticated PEPOLAR schemes are unsupported."
-            )
+        suffices = [s.suffix for s in estimator.sources]
```
mgxd, line 442: just so i'm clear - this was changed to ensure there is only 1 bold / fmap pair in the estimation?

mgxd (approved):
lgtm

fmriprep/workflows/base.py
effigies: Yes, that or all EPI, which we previously allowed.

PR #2685, opened 2022-01-13
FIX: Pass boolean use_syn_sdc to find_estimators

effigies:
Alternative to #2681. I think the reason we were hitting this edge case is that there was a type change.


effigies:
This gets past the workflow building stage, so I'd say my fix was correct. Going to merge to get other PRs working.

PR #2715, opened 2022-02-15
BF(DOC): use a BIDS compliant _mask instead of old BEP03 _roi to point to lesion mask

yarikoptic:
Not sure if the specific name is the best one, since does not point
explicitly to T1w as the orig of that mask, and either label should
may be made extended instead of abbreviated L.

But in either case, old doc example is clearly wrong thus this fix

Attn @michael-sun

### DIFF
--- a/docs/workflows.rst
+++ b/docs/workflows.rst
@@ -121,8 +121,8 @@ ANTs will use this mask to minimize warping of healthy tissue into damaged
 areas (or vice-versa).
 Lesion masks should be binary NIfTI images (damaged areas = 1, everywhere else = 0)
 in the same space and resolution as the T1 image, and follow the naming convention specified in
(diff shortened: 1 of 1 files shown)


effigies:
Before we can update the docs, we need to accept the standard ROI masks.

yarikoptic:
oh, right... I have tried to spot which code is relying on those `_roi` files but failed to find any.

effigies:
It's in smriprep. Apologies for the web of dependencies; the idea is to make identical anatomical preprocessing available to dmriprep or aslprep or whatevercomesnextprep.

effigies:
Docs are being updated to match behavior in #2943. They will be updated again when behavior conforms to BIDS Derivatives spec, targeting 23.1.0.

PR #2724, opened 2022-02-28
DOC: clarify calculation of confounding signals

dillonplunkett:
A few words seemed to be missing in the documentation about how confounding signals are calculated. I believe this is the intended meaning. (But please double check me!)


effigies:
Good catch. Thanks!

PR #2731, opened 2022-03-03
Diagnose ds210 failure

effigies:
Starting by setting a random seed to verify that it can be reproduced.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,6 +12,7 @@ jobs:
       # upgrade Docker version
       image: ubuntu-2004:202104-01
     working_directory: /tmp/src/fmriprep
+    resource_class: large
     steps:
       - checkout
       - run:
@@ -354,6 +355,7 @@ jobs:
     machine:
       image: ubuntu-2004:202104-01
     working_directory: /tmp/ds005
+    resource_class: large
     environment:
       - FS_LICENSE: /tmp/fslicense/license.txt
       - DATASET: ds005
@@ -448,8 +450,8 @@ jobs:
                     --fs-subjects-dir /tmp/${DATASET}/freesurfer \
                     --skull-strip-template OASIS30ANTs:res-1 \
                     --output-spaces MNI152NLin2009cAsym MNI152NLin6Asym \
-                    --sloppy --write-graph --mem_mb 4096 \
-                    --nthreads 2 --anat-only -vv
+                    --sloppy --write-graph --mem-mb 14336 \
+                    --nthreads 4 --anat-only -vv
             fi
       - run:
           name: Clean-up after anatomical run
@@ -490,11 +492,11 @@ jobs:
                 --fs-subjects-dir /tmp/${DATASET}/freesurfer \
                 ${FASTRACK_ARG} \
                 --output-layout legacy \
-                --sloppy --write-graph --mem_mb 4096 \
+                --sloppy --write-graph --mem-mb 14336 \
                 --skull-strip-template OASIS30ANTs:res-1 \
                 --output-spaces fsaverage5 fsnative \
                 --use-plugin /home/circleci/src/fmriprep/.circleci/legacy.yml \
-                --nthreads 2 -vv
+                --nthreads 4 -vv
       - store_artifacts:
           path: /tmp/ds005/derivatives
       - run:
@@ -534,9 +536,9 @@ jobs:
                 --config $PWD/nipype.cfg -w /tmp/${DATASET}/work_bids \
                 /tmp/data/${DATASET}/ /tmp/${DATASET}/bids participant \
                 ${FASTRACK_ARG} \
-                --sloppy --write-graph --mem_mb 4096 \
+                --sloppy --write-graph --mem-mb 14336 \
                 --output-spaces fsaverage5 \
-                --nthreads 2 -vv
+                --nthreads 4 -vv
       - run:
           name: Checking outputs of partial fMRIPrep run
           command: |
@@ -565,10 +567,10 @@ jobs:
                 --fs-subjects-dir /tmp/${DATASET}/freesurfer \
                 ${FASTRACK_ARG} \
                 --output-layout legacy \
-                --sloppy --write-graph --use-syn-sdc --mem_mb 4096 \
+                --sloppy --write-graph --use-syn-sdc --mem-mb 14336 \
                 --output-spaces MNI152NLin2009cAsym fsaverage5 fsnative MNI152NLin6Asym anat \
                 --aroma-melodic-dimensionality 2 --use-aroma \
-                --nthreads 2 --cifti-output -vv
+                --nthreads 4 --cifti-output -vv
       - run:
           name: Checking outputs of partial fMRIPrep run
           command: |
@@ -605,6 +607,7 @@ jobs:
     machine:
       image: ubuntu-2004:202104-01
     working_directory: /tmp/ds054
+    resource_class: large
     environment:
       - FS_LICENSE: /tmp/fslicense/license.txt
       - DATASET: ds054
@@ -692,7 +695,7 @@ jobs:
                     --output-layout legacy \
                     --fs-no-reconall --sloppy --write-graph \
                     --output-spaces MNI152NLin2009cAsym \
-                    --mem_mb 4096 --nthreads 2 --anat-only -vv
+                    --mem-mb 14336 --nthreads 4 --anat-only -vv
             fi
       - run:
           name: Clean-up after anatomical run
@@ -721,7 +724,7 @@ jobs:
                 --output-layout legacy \
                 --fs-no-reconall --sloppy \
                 --output-spaces MNI152NLin2009cAsym:res-2 anat func \
-                --mem_mb 4096 --nthreads 2 -vv --debug compcor
+                --mem-mb 14336 --nthreads 4 -vv --debug compcor
       - run:
           name: Checking outputs of fMRIPrep
           command: |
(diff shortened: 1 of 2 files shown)


effigies:
Looks like that does it. Rerunning to confirm. Will try to reproduce locally.

mgxd:
@effigies were you able to replicate this? I just tried locally (with the same random seed) and it completed..

effigies:
Sorry, should have updated. I did try locally and couldn't replicate. If we upload the entire working directory as artifacts, that might be the best bet for tracking it down.

mgxd:
Update - I reran, this time using the fast-track anatomicals provided, and was able to replicate the crash. As predicted, the BOLD mask is awful. After looking into this, I think the failure point is `final_boldref_wf.enhance_and_skullstrip_bold_wf.n4_correct` (see attachment)
Screen Shot 2022-04-14 at 9 03 22 AM

Two things:
- N4Correct sets dimensionality to 3, but the input is a 4D file
- Shouldn't [RobustAverage](https://github.com/nipreps/niworkflows/blob/e1f4267eb5fd878b2a001f184e1bddbb3f4a6843/niworkflows/func/util.py#L190) be producing a single volume?

cc @effigies @oesteban

oesteban:
Wow great job catching it.

Your two hypotheses are correct. The question is what is wrong in the fast track that ends up giving N4 a 4D file (?)

Can you check what are the inputs to `final_boldref_wf.enhance_and_skullstrip_bold_wf` that change between fast-track / regular-track?

mgxd:
> Two things:
>
> - N4Correct sets dimensionality to 3, but the input is a 4D file
> - Shouldn't [RobustAverage](https://github.com/nipreps/niworkflows/blob/e1f4267eb5fd878b2a001f184e1bddbb3f4a6843/niworkflows/func/util.py#L190) be producing a single volume?

I think I got confused across working directories, because this didn't seem to be the case after retesting. But I think I have found a simple solution - we have not been passing the already calculated bold mask into the final boldref workflow. I submitted a quick fix (https://github.com/nipreps/fmriprep/pull/2731/commits/3911638439c1970fd13d73f6c6e036311223f75e) for the particular pathway our ds210 test takes (ME, SDC), but the others are unaccounted for and thus failing.

mgxd:
@effigies i can't request you as a reviewer since it's your PR, but would appreciate a quick look (if you find some time)

fmriprep/workflows/bold/base.py
```diff
@@ -993,6 +994,9 @@ def init_func_preproc_wf(bold_file, has_fieldmap=False):
             (bold_hmc_wf, bold_bold_trans_wf, [
                 ("outputnode.xforms", "inputnode.hmc_xforms"),
             ]),
+            (initial_boldref_wf, final_boldref_wf, [
```
effigies, line 997: I thought the idea of having a second go-round was to get a better mask.

oesteban: This could be IMHO okay, but to me, the problem is that we do not know yet what's making the fast-track fall into this while the regular track is fine for the given seed.

We first need to bisect the issue and determine the point where inputs and (or) outputs deviate from the regular track.

We may live with this just fine, but if we do this not knowing why this happened, the regression will be almost certain.

mgxd: the (silent) failure point is [n4_correct](https://github.com/nipreps/niworkflows/blob/e1f4267eb5fd878b2a001f184e1bddbb3f4a6843/niworkflows/func/util.py#L392-L400), though from looking at the inputs it is not very clear why one case is failing while the other is fine. I've compiled the working directories of fast-track(ft)/no fast-track(noft) within a tarball ([ds210_n4_error.tar.gz](https://github.com/nipreps/fmriprep/files/8496251/ds210_n4_error.tar.gz)) if either of you would like to take a look.

effigies:
@mgxd Very unlikely today, as we'll be wrapping our sprint. I'll have some time on the plane and train, though.

oesteban:
The masks are different. That doesn't justify the crash (as in, it's crazy the mask makes such a big difference for N4), but at least we know where the problem is coming from, right?

Screen Shot 2022-04-15 at 4 57 43 PM

Some thoughts that I'm skeptical will fully solve the issue but that together will give more reliability to the full process:

- Using the weights input here: https://github.com/nipreps/niworkflows/blob/e1f4267eb5fd878b2a001f184e1bddbb3f4a6843/niworkflows/func/util.py#L526 would be more reliable, can you give it a quick try?
- This float setting (and the one on ``ApplyTransforms`` later) could also be biting at us: https://github.com/nipreps/niworkflows/blob/master/niworkflows/data/epi_atlasbased_brainmask.json#L4
- If we change the N4 setting to weights, then we could also remove this binarization: https://github.com/nipreps/niworkflows/blob/e1f4267eb5fd878b2a001f184e1bddbb3f4a6843/niworkflows/func/util.py#L499

Only when these things have been tested and the outputs of N4 in both conditions are more similar, I would then consider whether we also want to use a prior mask to avoid these steps altogether.

However, these steps are going to happen anyways in the first time around, so we want to make sure the workflow is more reliable.

WDYT?

oesteban:
I'm digging up more on the inputs.

- The masks are float32 (which is not great for a mask).
- The `_average` input images are very close to one another, but not exactly the same. It would be good to trace that back to the fast-track divergence.
- The `_average` images should be clipped for N4 to perform appropriately. Currently they have a range between 0 through 12000, with a median of ~6.5. I'm sure reducing the dynamic range to 0-255 would basically compress all those low intensities within one bin, making the job easier for N4.

That all said, there is also appeal in giving SynthStrip a go, and replace all of this if it works.

oesteban:
> * The `_average` input images are very close to one another, but not exactly the same. It would be good to trace that back to the fast-track divergence.

Q: Is this happening with single-echo too?

mgxd:
I would guess the difference is because the anatomical derivatives have been run on the high res ds210, whereas we're calculating the T1w/MNI transform using the downsampled outputs (and using `--sloppy`). I tried testing with the latest ANTs version (2.3.5), but no luck. I'm rerunning now using your suggestions (https://github.com/nipreps/fmriprep/pull/2731#issuecomment-1100166061)

> That all said, there is also appeal in giving SynthStrip a go, and replace all of this if it works.

Yes, it would be good to test. But we'd still want to backport a fix for `21.0.x`

> Q: Is this happening with single-echo too?

Potentially (see #2761) but I haven't personally replicated / seen it

oesteban:
Are we using FSL 6? Could the reason for these errors bubble up an upgrade and this code https://github.com/nipreps/niworkflows/blob/e1f4267eb5fd878b2a001f184e1bddbb3f4a6843/niworkflows/func/util.py#L378-L386 introducing weirdness?

Please check some of the ideas in nipreps/niworkflows#707.

I'm pretty positive we want to do the clipping before registration and N4, but don't have time now to have a stab at it.

mgxd:
I just ran with the following changes: https://github.com/nipreps/niworkflows/compare/maint/1.4.x...mgxd:dbg/ds210-failure

and was able to successfully complete, with normal looking reports.

oesteban:
> and was able to successfully complete, with normal looking reports.

Yup, that's the weights instead of the mask, I'm sure of that.

I have updated my PR with further changes (esp. removing the binarization and binary dilation for the no-premask pathway).

mgxd:
interestingly, we're now running into

```
220420-17:06:37,576 nipype.workflow ERROR:
	 Saving crash info to /out/fmriprep/sub-02/log/20220420-161238_95a4bf05-53b1-4a12-965b-3f1ff395c91f/crash-20220420-170637-UID1001-skullstrip_first_pass-4731a275-68da-44b0-8314-e8ce701ce2f9.txt
  File "/opt/conda/lib/python3.8/site-packages/nipype/pipeline/plugins/multiproc.py", line 67, in run_node
  File "/opt/conda/lib/python3.8/site-packages/nipype/pipeline/engine/nodes.py", line 516, in run
  File "/opt/conda/lib/python3.8/site-packages/nipype/pipeline/engine/nodes.py", line 635, in _run_interface
  File "/opt/conda/lib/python3.8/site-packages/nipype/pipeline/engine/nodes.py", line 741, in _run_command
  File "/opt/conda/lib/python3.8/site-packages/nipype/interfaces/base/core.py", line 428, in run
  File "/opt/conda/lib/python3.8/site-packages/nipype/interfaces/fsl/preprocess.py", line 163, in _run_interface
    runtime = super(BET, self)._run_interface(runtime)
  File "/opt/conda/lib/python3.8/site-packages/nipype/interfaces/base/core.py", line 822, in _run_interface
    self.raise_exception(runtime)
  File "/opt/conda/lib/python3.8/site-packages/nipype/interfaces/base/core.py", line 749, in raise_exception
    raise RuntimeError(
RuntimeError: Command:
bet vol0000_xform-00000_clipped_merged_average_corrected.nii vol0000_xform-00000_clipped_merged_average_corrected_brain.nii.gz -f 0.20 -m
Standard output:
/opt/fsl-6.0.5.1/bin/bet failed during command:vol0000_xform-00000_clipped_merged_average_corrected.nii vol0000_xform-00000_clipped_merged_average_corrected_brain.nii.gz -f 0.20 -m
Standard error:
/opt/fsl-6.0.5.1/bin/bet: line 399:  1528 Segmentation fault      (core dumped) ${FSLDIR}/bin/bet2 $IN $OUT $bet2opts
Return code: 1
```

mgxd:
@effigies @oesteban I think the latest niworkflows changes have fixed this. Are we fine with merging (after removing the random seed) and cutting 21.0.2?

mgxd:
merging since i'd like to get a release in before the weekend

effigies:
Thanks for going ahead. No objections.

Issue #2735, opened 2022-03-05
Don't we want to test the ``--cifti-output`` integration?


oesteban:
Since this 8421f89701d2f474d788e3f24d695f5393c4a9bb, ds005 is run without ``--cifti-output``. We probably want to take it back, right? /cc @effigies

effigies:
I thought we only removed it in the rerun test.

oesteban:
I haven't looked in depth but I don't think it's in the config file at all.

I spotted it because the report doesn't have anything about CIFTI

PR #2736, opened 2022-03-06
MAINT: Upload artifacts after each step of Circle's workflow

oesteban:
Resolves: #2735.


oesteban:
Okay, this makes visible the reports with CIFTI.

PR #2746, opened 2022-03-25
CI: Update package builds to use python -m build

effigies:
* ~~Run tests/doctests on Python 3.10 in GitHub Actions~~ No vtk for Python 3.10, so we can't install niworkflows.
* Build sdists/wheels with `python -m build`, which provides build isolation, so we don't need to use our own on CircleCI
* Remove the `VERSION` file from sdist/wheel builds. We check whether the packages have the correct version added by versioneer.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1060,18 +1060,11 @@ jobs:
     steps:
       - checkout
       - run:
+          name: Update build tools
+          command: python -m pip install --user pip build twine
+      - run:
           name: Build fMRIPrep
-          command: |
-            python -m pip install --user twine  # For use in checking distributions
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -1187,10 +1166,6 @@ workflows:
 
       - test_deploy_pypi:
           filters:
-            branches:
-              only:
-                - /rel\/.*/
-                - /maint\/.*/
```
effigies, line 1193: @mgxd Vaguely recall you were the one who preferred to keep this off of every PR. Should I restore these filters now that we've tested?

mgxd: I think so, that'll give us more flexibility during development cycles.

PR #2751, opened 2022-03-30
ENH: Compare T2* map to BOLD reference

effigies:

We had discussed using the closest echo to standard BOLD, but given that we're using the shortest echo to produce a reference image, it seems appropriate to compare to the reference image. If in the future we select an echo other than the shortest, then we will get that automatically.

The reference image being used should be selected from the shortest echo and preprocessed somewhat to improve gray/white contrast. If STC or SDC is applied, it will have already been applied to this image, just as these will have already been applied to the individual echos that are used to calculate the T2* map. (HMC will also have been applied to the individual echos, but does not apply to the reference volume.)

cc OHBM coauthors: @eilidhmacnicol @oesteban @eurunuela @javiergcas @handwerkerd @tsalo @emdupre @mgxd @jbteves

Haven't tested locally; using CI.


effigies:
Example figure:

sub-02_task-cuedSGT_run-01_desc-t2scomp_bold

oesteban (approved):
I think this new visualization is very helpful. One thing I have just realized (cc/ @celprov) is that the T2\* map and the optimal combination of echos we are passing into the edge mask have been brain-masked before.

For the regressors and the "royal" carpet plot, we can either calculate the optimal combination within the edge mask or change them to represent perhaps the closest echo to the standard TE of single echo EPI.

That said, I think this PR is a good idea.

effigies:
RE carpet plots, what do you think about including [Tedana's carpet plots](https://tedana.readthedocs.io/en/latest/outputs.html#carpet-plots) directly, which include combined, denoised, accepted, rejected?

Issue #2659, opened 2021-12-09
DS210 CI failing intermittently in confounds workflow


effigies:
### What happened?

DVARS and RenameACompCor nodes are failing. Looking into this.

### What command did you use?

```shell
RUNNING: docker run --rm -e DOCKER_VERSION_8395080871=20.10.6 -it -e FMRIPREP_DEV=1 -u 1001:1002 -v /tmp/fslicense/license.txt:/opt/freesurfer/license.txt:ro -v /tmp/data/ds210:/data:ro -v /tmp/ds210/derivatives:/out -v /tmp/ds210/smriprep-0.6:/opt/smriprep/subjects -v /tmp/ds210/work:/scratch -v /tmp/ds210/nipype.cfg:/home/fmriprep/.nipype/nipype.cfg:ro nipreps/fmriprep:latest /data /out participant --output-layout legacy --me-output-echos --fs-no-reconall --use-syn-sdc --ignore slicetiming --dummy-scans 1 --sloppy --write-graph --mem_mb 4096 --nthreads 2 -vv --anat-derivatives /opt/smriprep/subjects -w /scratch --output-spaces MNI152NLin2009cAsym
```

### What version of fMRIPrep are you running?

21.0.0rc2+27.g6b110981

### How are you running fMRIPrep?

Docker

### Is your data BIDS valid?

Yes

### Are you reusing any previously computed results?

Anatomical derivatives

### Please copy and paste any relevant log output.

```shell
211209-18:17:53,25 nipype.workflow INFO:
	 [Node] Setting-up "fmriprep_wf.single_subject_02_wf.func_preproc_task_cuedSGT_run_01_echo_1_wf.bold_confounds_wf.dvars" in "/scratch/fmriprep_wf/single_subject_02_wf/func_preproc_task_cuedSGT_run_01_echo_1_wf/bold_confounds_wf/dvars".
211209-18:17:53,35 nipype.workflow INFO:
	 [Node] Running "dvars" ("nipype.algorithms.confounds.ComputeDVARS")
211209-18:17:53,77 nipype.workflow INFO:
	 [MultiProc] Running 2 tasks, and 13 jobs ready. Free memory (GB): 3.99/4.10, Free processors: 0/2.
                     Currently running:
                       * fmriprep_wf.single_subject_02_wf.func_preproc_task_cuedSGT_run_01_echo_1_wf.bold_confounds_wf.dvars
[...]

PR #2754, opened 2022-04-06
FIX: Preserve `collect_data` behavior by using named args

mgxd:
See https://github.com/nipreps/nibabies/pull/208


mgxd:
failing test is because of #2659 - merging this in as master is broken without it.

PR #2768, opened 2022-04-27
Explain better SDC and B0FieldSource requirement.

toomanycats:
Explain that `sdcflows` has it's own code base but it's still included in `fmriprep`.
I'm actually unsure about this but I think this is correct: Tell users to add the `B0FieldSource` key and value to all fieldmap images.

### DIFF
--- a/docs/workflows.rst
+++ b/docs/workflows.rst
@@ -368,8 +368,15 @@ the scanner.
     fieldmap estimation.
 
 Please note that all routines for susceptibility-derived distortion correction
-have been excised off of *fMRIPrep* for utilization on other projects
-(e.g., `dMRIPrep <https://www.nipreps.org/dmriprep>`__).
+have been moved into their own code base called `sdcflows`. No action is
+required by user as this module is included in `fmriprep`.
+
+Details about the BIDS specification for field maps can be found at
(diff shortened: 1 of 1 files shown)


docs/workflows.rst
```diff
@@ -368,8 +368,15 @@ the scanner.
...
-have been excised off of *fMRIPrep* for utilization on other projects
-(e.g., `dMRIPrep <https://www.nipreps.org/dmriprep>`__).
+have been moved into their own code base called `sdcflows`. No action is
+required by user as this module is included in `fmriprep`.
+
+Details about the BIDS specification for field maps can be found at
+<https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/01-magnetic-resonance-imaging-data.html#types-of-fieldmaps>
+
+**NOTE** There is a deviation from the BIDS spec, in that all fieldmap image side
+car files must include the `B0Fielsource` key and appropriate value.
```
effigies, line 378: I don't think this is true. We should respect `IntendedFor` (and where we don't it's a bug); it just might be ambiguous.

oesteban:
```suggestion
**NOTE** There is a deviation from the BIDS spec, in that all fieldmap image side
car files must include the `B0FieldIdentifier` key and appropriate value.
```

Two notes:
- I believe this should say `B0FieldIdentifier` (as this is the field you define on the fieldmap metadata)
- I'm skeptical that this is true, we use PyBIDS to query the metadata, and it should follow general inheritance rules.

Issue #2728, opened 2022-03-03
Issue preprocessing multiple multi echo / multi band data with fmriprep update


njr175:
### What happened?

We updated fMRIPrep to version 21.0.0. We have been using fMRIPrep for a few years now, and we have preprocessed multiple datasets with this pipeline. After making the upgrade, we wanted to re preprocess 2 of the datasets. We were able to re-preprocess the first no problem (NOT multi echo data). We went to preprocess the 2nd dataset - multiband/multiecho - and we can't preprocess more than 1 participant at a time. As I mentioned, these data have been preprocessed with fMRIPrep in the past no problem; I could run 10 at a time. (All work directories have been deleted, 'old' data have been moved out of the fMRIPrep path.) I am testing with 2-3 participants at a time, and have tried the following flags: 
--low-mem
--mem-mb55000
--use-plugin plugin.yml (made a corresponding file - plugin: LegacyMultiProc
plugin_args: {maxtasksperchild: 1, memory_gb: 60, n_procs: 10, raise_insufficient: false}

Without the flags, and with the '--low-mem' flag, I get a 'broken process pull' and fMRIPrep doesn't finish. With the other 2 flags, I get a Node Name error: fmriprep_wf.single_subject_3156_wf.func_preproc_ses_1_task.... _ echo_1_wf.bold_t2smap_wf.t2nmap_node
(full error below)
I get this node error for each participant, different number of nodes. 
If I run the participants separately, they run no problem.

### What command did you use?

```shell
singularity run --cleanenv -B /data/MNA/data /data/MNA/fmriprep_21.0.0.simg /data/MNA/data/Nifti /data/MNA/data/Preprocessed participant --participant_label sub-3158 3159 3162 --ignore slicetiming sbref --fs-license-file /home/nicoler/license.txt

('original' command used. I also tried the various flags I outlined above)
```

### What version of fMRIPrep are you running?

21.0.0

### How are you running fMRIPrep?

Singularity

### Is your data BIDS valid?

Yes

### Are you reusing any previously computed results?

No

### Please copy and paste any relevant log output.

```shell
When fMRIPrep doesn't finish/just hangs error is as follows: 
ile "/opt/conda/lib/python3.8/concurrent/futures/_base.py", line 437, in result
   return self.__get_result()
File "/opt/conda/lib/python3.8/concurrent/futures/_base.py", line 389, in __get_result raise         self._exception
[...]

PR #2776, opened 2022-05-09
ENH: Save T2starmap files in all requested output spaces, if calculated

effigies:

Resample T2* map (calculated in BOLD space) to T1w and standard spaces if requested. Sink any of these as functional derivatives.

Unclear what would happen with fsaverage, so will need to test this case.

Need to update expected outputs.

### DIFF
--- a/.circleci/ds210_fasttrack_outputs.txt
+++ b/.circleci/ds210_fasttrack_outputs.txt
@@ -29,5 +29,6 @@ fmriprep/sub-02/func/sub-02_task-cuedSGT_run-01_space-MNI152NLin2009cAsym_desc-b
 fmriprep/sub-02/func/sub-02_task-cuedSGT_run-01_space-MNI152NLin2009cAsym_desc-brain_mask.nii.gz
 fmriprep/sub-02/func/sub-02_task-cuedSGT_run-01_space-MNI152NLin2009cAsym_desc-preproc_bold.json
 fmriprep/sub-02/func/sub-02_task-cuedSGT_run-01_space-MNI152NLin2009cAsym_desc-preproc_bold.nii.gz
+fmriprep/sub-02/func/sub-02_task-cuedSGT_run-01_space-MNI152NLin2009cAsym_T2starmap.nii.gz
 fmriprep/sub-02.html
 /tmp/ds210/derivatives
--- a/.circleci/ds210_fasttrack_outputs.txt
+++ b/.circleci/ds210_fasttrack_outputs.txt
@@ -29,6 +29,7 @@ fmriprep/sub-02/func/sub-02_task-cuedSGT_run-01_space-MNI152NLin2009cAsym_desc-b
 fmriprep/sub-02/func/sub-02_task-cuedSGT_run-01_space-MNI152NLin2009cAsym_desc-brain_mask.nii.gz
 fmriprep/sub-02/func/sub-02_task-cuedSGT_run-01_space-MNI152NLin2009cAsym_desc-preproc_bold.json
 fmriprep/sub-02/func/sub-02_task-cuedSGT_run-01_space-MNI152NLin2009cAsym_desc-preproc_bold.nii.gz
+fmriprep/sub-02/func/sub-02_task-cuedSGT_run-01_space-MNI152NLin2009cAsym_T2starmap.json
 fmriprep/sub-02/func/sub-02_task-cuedSGT_run-01_space-MNI152NLin2009cAsym_T2starmap.nii.gz
 fmriprep/sub-02.html
 /tmp/ds210/derivatives
--- a/.circleci/ds210_outputs.txt
+++ b/.circleci/ds210_outputs.txt
@@ -48,6 +48,7 @@ fmriprep/sub-02/func/sub-02_task-cuedSGT_run-01_space-MNI152NLin2009cAsym_desc-b
 fmriprep/sub-02/func/sub-02_task-cuedSGT_run-01_space-MNI152NLin2009cAsym_desc-brain_mask.nii.gz
 fmriprep/sub-02/func/sub-02_task-cuedSGT_run-01_space-MNI152NLin2009cAsym_desc-preproc_bold.json
 fmriprep/sub-02/func/sub-02_task-cuedSGT_run-01_space-MNI152NLin2009cAsym_desc-preproc_bold.nii.gz
+fmriprep/sub-02/func/sub-02_task-cuedSGT_run-01_space-MNI152NLin2009cAsym_T2starmap.json
 fmriprep/sub-02/func/sub-02_task-cuedSGT_run-01_space-MNI152NLin2009cAsym_T2starmap.nii.gz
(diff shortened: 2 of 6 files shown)


fmriprep/workflows/bold/resampling.py
```diff
@@ -462,6 +464,7 @@ def init_bold_std_trans_wf(
     output_names = [
         "bold_mask_std",
         "bold_std",
+        "t2star_std",
```
mgxd, line 467: i think the problem is this is always being joined on, rather than only conditionally being in this list when multiecho is True

effigies: Ah, I was wondering why we made the FreeSurfer ones conditional rather than just letting them be `Undefined`.

.circleci/ds210_fasttrack_outputs.txt
```diff
@@ -29,5 +29,6 @@ fmriprep/sub-02/func/sub-02_task-cuedSGT_run-01_space-MNI152NLin2009cAsym_desc-b
 fmriprep/sub-02/func/sub-02_task-cuedSGT_run-01_space-MNI152NLin2009cAsym_desc-brain_mask.nii.gz
 fmriprep/sub-02/func/sub-02_task-cuedSGT_run-01_space-MNI152NLin2009cAsym_desc-preproc_bold.json
 fmriprep/sub-02/func/sub-02_task-cuedSGT_run-01_space-MNI152NLin2009cAsym_desc-preproc_bold.nii.gz
+fmriprep/sub-02/func/sub-02_task-cuedSGT_run-01_space-MNI152NLin2009cAsym_desc-tedana_T2starmap.nii.gz
```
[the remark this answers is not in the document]
tsalo: By metadata, do you mean specific fields for the sidecar JSON file, or general documentation about this derivative?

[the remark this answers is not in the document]
effigies: Sidecar fields.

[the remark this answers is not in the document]
tsalo: T2* maps _must_ be in seconds, but it might be nice to have a `Units` field anyway. [The qMRI appendix](https://bids-specification.readthedocs.io/en/stable/99-appendices/11-qmri.html#metadata-requirements-for-qmri-maps) also recommends including `BasedOn`, `EstimationReference`, and `EstimationAlgorithm` fields. 

I think fMRIPrep covers `BasedOn` already, but `EstimationReference` and `EstimationAlgorithm` might be useful to include. I think @handwerkerd would be able to say what those fields should be set to better than me though.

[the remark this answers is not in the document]
handwerkerd: Not sure how much detail can go into `EstimationAlgorithm` but a long option would be:
"Fit to monoexponential T2* decay model [across the mean value across time for each voxel and echo]"
We don't explicitly reference anything. My understanding is the math could go back to some really early NMR papers. I I think the first paper to average across time for multi-echo data is: Poser et al "BOLD contrast sensitivity enhancement and artifact reduction with multiecho EPI: parallel-acquired inhomogeneity-desensitized fMRI" MRM 2006 https://doi.org/10.1002/mrm.20900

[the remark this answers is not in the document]
tsalo: fMRIPrep could hypothetically support multiple t2smap options (e.g., see #2728), but for the current implementation a short algorithm description might be "nonlinear monoexponential".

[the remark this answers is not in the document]
effigies: To be clear, I don't think we need a "first paper", I think we need a meaningful reference that would make the actual algorithm clear. If it makes sense to use one of the citations on https://tedana.readthedocs.io/, I'm happy to do that.

fmriprep/workflows/bold/outputs.py
```diff
@@ -169,6 +169,11 @@ def init_func_derivatives_wf(
     # BOLD series will generally be unmasked unless multiecho,
     # as the optimal combination is undefined outside a bounded mask
     masked = multiecho
+    t2star_meta = {
+        'Units': 's',
+        'EstimationReference': 'doi:10.1002/mrm.20900',
+        'EstimationAlgorithm': 'monoexponential decay model',
+    }
```
effigies, line 176: This + `RawSources` will be the contents of each metadata file.

docs/outputs.rst
```diff
@@ -235,6 +235,38 @@ are saved::
...
+For multi-echo datasets, the output ``_bold`` series are "optimally combined" by
+`tedana`_ to better estimate the BOLD signal.
+This process also generates a T2\* map, which is resampled into every requested output
+space.
+
+::
+
+  sub-<subject_label>/
+    func/
+      sub-<subject_label>_[specifiers]_T2starmap.nii.gz
```
effigies, line 248:
```suggestion
      sub-<subject_label>_[specifiers]_desc-tedana_T2starmap.nii.gz
```

.circleci/ds210_outputs.txt
```diff
@@ -48,5 +48,7 @@ fmriprep/sub-02/func/sub-02_task-cuedSGT_run-01_space-MNI152NLin2009cAsym_desc-b
 fmriprep/sub-02/func/sub-02_task-cuedSGT_run-01_space-MNI152NLin2009cAsym_desc-brain_mask.nii.gz
 fmriprep/sub-02/func/sub-02_task-cuedSGT_run-01_space-MNI152NLin2009cAsym_desc-preproc_bold.json
 fmriprep/sub-02/func/sub-02_task-cuedSGT_run-01_space-MNI152NLin2009cAsym_desc-preproc_bold.nii.gz
+fmriprep/sub-02/func/sub-02_task-cuedSGT_run-01_space-MNI152NLin2009cAsym_desc-tedana_T2starmap.json
+fmriprep/sub-02/func/sub-02_task-cuedSGT_run-01_space-MNI152NLin2009cAsym_desc-tedana_T2starmap.nii.gz
```
effigies, line 52: Just to bring this up, are we happy with this naming convention? I could use a different `desc` if that would be preferred.

effigies: Since we always use `space-`, it is not necessary.

fmriprep/workflows/bold/outputs.py
```diff
@@ -244,6 +250,20 @@ def init_func_derivatives_wf(
             (raw_sources, ds_bold_mask_native, [('out', 'RawSources')]),
         ])
 
+        if multiecho:
+            ds_t2star_bold = pe.Node(
+                DerivativesDataSink(base_directory=output_dir, desc='tedana', space='boldref',
```
effigies, line 255: Are we happy with `space-boldref` when in the BOLD space? Or maybe just `bold`?

effigies:
@mgxd final review?

mgxd (commented):
Looks good - only thing is we should update the docstring "Inputs/Outputs" if they are being altered

PR #2778, opened 2022-05-16
ENH: Plot histogram of T2* values in gray-matter mask

### LINKED ISSUE #2541: Multi-echo reporting: histogram of T2* values in gray-matter mask
via @javiergcas

I assume a FAST or FreeSurfer gray-matter mask resampled into BOLD space would be appropriate. Are there example figures that we could base the reportlet on?

effigies:

Closes #2541. I think we should also show the gray-matter mask in the T2*-boldref map immediately above the histogram.

Working on the OHBM poster should provide some images/text that can be added to the docs.

### DIFF
--- a/fmriprep/data/reports-spec.yml
+++ b/fmriprep/data/reports-spec.yml
@@ -103,6 +103,10 @@ sections:
       of the T2\* map and the BOLD reference map used for BOLD-T1w coregistration.
     static: false
     subtitle: T2\* map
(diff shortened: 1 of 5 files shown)


effigies:
I do have WM and CSF masks as well, if it would be helpful, rather than cluttering to plot these histograms together. See https://seaborn.pydata.org/generated/seaborn.histplot.html for options for dealing with multiple histograms.

effigies:
Fixes can come in future PRs.

PR #2781, opened 2022-05-20
ENH: Clip T2* values at 100ms to keep consistent histogram axes

### LINKED ISSUE #2780: Clip and re-bin T2* histograms

The T2* histograms are set to a certain number of bins, so a single outlier far to the right of the mode will lose a lot of specificity.

t2star_full_diag

I think we should clip at a reasonably high number (100ms?) and let outliers pile up there to indicate something might have gone wrong while preserving some resolution in the modal range.

In the BOLD space there's already a lower bound at 0. I haven't made an effort to clip the resampled images in T1w/standard spaces, which do get some negative values around the edges due to ringing, as I don't think these are used for calculations.

Yes

_No response_

effigies:
Closes #2780.

### DIFF
--- a/fmriprep/data/reports-spec.yml
+++ b/fmriprep/data/reports-spec.yml
@@ -106,7 +106,8 @@ sections:
     subtitle: T2* map
   - bids: {datatype: figures, desc: t2starhist, suffix: bold}
     caption: A histogram of estimated T2* values within the anatomically-derived gray-matter mask
-      shown in the previous plot.
+      shown in the previous plot. Note that values are clipped at 100ms, so extreme outliers will
+      all appear in the 100ms bin.
     static: false
     subtitle: T2* gray-matter values
   - bids: {datatype: figures, desc: flirtnobbr, suffix: bold}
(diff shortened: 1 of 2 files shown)


effigies:
t2star_hist_clipped

@mgxd Any thoughts?

fmriprep/data/reports-spec.yml
```diff
@@ -106,7 +106,8 @@ sections:
     subtitle: T2* map
   - bids: {datatype: figures, desc: t2starhist, suffix: bold}
     caption: A histogram of estimated T2* values within the anatomically-derived gray-matter mask
-      shown in the previous plot.
+      shown in the previous plot. Note that values are clipped at 100ms, so extreme outliers will
+      all appear in the 100ms bin.
```
mgxd, line 110: Slight rewording
```suggestion
      shown in the previous plot. Note that values are clipped at 100ms, so any extreme outliers will
      appear in the 100ms bin.
```

mgxd (approved):
LGTM

effigies:
Only affects ds210, not waiting for ds005 to finish.

PR #2817, opened 2022-07-13
ENH: Add migas telemetry in addition to sentry

mgxd:
This adds `migas` telemetry in addition to sentry, to allow us to compare +/-'s

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -384,6 +384,7 @@ jobs:
       - FS_LICENSE: /tmp/fslicense/license.txt
       - DATASET: ds005
       - CHECK_DS005: true
+      - MIGAS_OPTOUT: 1
     steps:
       - checkout:
           path: *src
@@ -593,6 +594,7 @@ jobs:
       - FS_LICENSE: /tmp/fslicense/license.txt
       - DATASET: ds054
       - CHECK_DS054: true
+      - MIGAS_OPTOUT: 1
     steps:
       - checkout:
           path: *src
@@ -759,7 +761,7 @@ jobs:
                     --output-layout legacy \
                     --fs-no-reconall --sloppy --write-graph \
                     --output-spaces MNI152NLin2009cAsym \
-                    --mem-mb 14336 --nthreads 4 --anat-only -vv
+                    --mem-mb 14336 --nthreads 4 --anat-only -vv --notrack
             fi
       - run:
           name: Clean-up after anatomical run
--- a/fmriprep/cli/run.py
+++ b/fmriprep/cli/run.py
@@ -71,7 +71,7 @@ def main():
 
         sentry_setup()
         setup_migas(init_ping=True)
-        atexit.register(_migas_exit)
+        atexit.register(migas_exit)
 
     # CRITICAL Save the config to a file. This is necessary because the execution graph
     # is built as a separate process to keep the memory footprint low. The most
--- a/fmriprep/cli/run.py
+++ b/fmriprep/cli/run.py
@@ -24,6 +24,8 @@
 """fMRI preprocessing workflow."""
 from .. import config
 
+EXITCODE: int = -1
+
 
 def main():
     """Entry point."""
@@ -61,11 +63,15 @@ def main():
 
     sentry_sdk = None
     if not config.execution.notrack and not config.execution.debug:
+        import atexit
+
         import sentry_sdk
 
-        from ..utils.sentry import sentry_setup
+        from ..utils.telemetry import sentry_setup, setup_migas
 
         sentry_setup()
+        setup_migas(init_ping=True)
+        atexit.register(_migas_exit)
 
     # CRITICAL Save the config to a file. This is necessary because the execution graph
     # is built as a separate process to keep the memory footprint low. The most
@@ -91,7 +97,8 @@ def main():
     else:
         retval = build_workflow(str(config_file), {})
 
-    retcode = retval.get("return_code", 0)
+    global EXITCODE
+    EXITCODE = retval.get("return_code", 0)
     fmriprep_wf = retval.get("workflow", None)
 
     # CRITICAL Load the config from the file. This is necessary because the ``build_workflow``
@@ -100,14 +107,14 @@ def main():
     config.load(config_file)
 
     if config.execution.reports_only:
-        sys.exit(int(retcode > 0))
+        sys.exit(int(EXITCODE > 0))
 
     if fmriprep_wf and config.execution.write_graph:
         fmriprep_wf.write_graph(graph2use="colored", format="svg", simple_form=True)
 
-    retcode = retcode or (fmriprep_wf is None) * EX_SOFTWARE
-    if retcode != 0:
-        sys.exit(retcode)
+    EXITCODE = EXITCODE or (fmriprep_wf is None) * EX_SOFTWARE
+    if EXITCODE != 0:
+        sys.exit(EXITCODE)
 
     # Generate boilerplate
     with Manager() as mgr:
@@ -118,7 +125,7 @@ def main():
         p.join()
 
     if config.execution.boilerplate_only:
-        sys.exit(int(retcode > 0))
+        sys.exit(int(EXITCODE > 0))
 
     # Clean up master process before running workflow, which may create forks
     gc.collect()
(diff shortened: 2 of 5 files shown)


fmriprep/utils/sentry.py
```diff
@@ -180,3 +180,27 @@ def _chunks(string, length=CHUNK_SIZE):
     """
     return (string[i:i + length]
             for i in range(0, len(string), length))
+
+
+def ping_migas(status='pending'):
```
effigies, line 185: I think we need a few tests here. Can we simulate what happens with `--notrack` set and with no network access?

effigies, line 185: Also, let's go ahead and type this...

```suggestion
def ping_migas(status: str='pending') -> None:
```

I'm not sure what `res` is on L200, but we never use the return value. I'm guessing it's a `dict`, so if we think we want this function to return a `dict`, then let's change L192 to `return {}` and set:

```suggestion
def ping_migas(status: str='pending') -> dict:
```

.circleci/config.yml
```diff
@@ -794,7 +794,7 @@ jobs:
                 --fs-no-reconall --use-syn-sdc --ignore slicetiming \
                 --dummy-scans 1 --sloppy --write-graph \
                 --output-spaces MNI152NLin2009cAsym \
-                --mem-mb 14336 --nthreads 4 -vv
+                --mem-mb 14336 --nthreads 4 -vv --notrack
```
oesteban, line 797: How much effort would it be to split debug/production pings for migas? It seems useful that we cross-compare with sentry on these debug heartbeats.

fmriprep/cli/run.py
```diff
@@ -43,6 +44,7 @@ def main():
         from ..utils.sentry import sentry_setup
 
         sentry_setup()
+        ping_migas()
```
oesteban, line 47: Is in this first ping when migas will check the user and figure out whether it is cached?

If not, I would favor to replicate sentry's design and have a `migas_setup()` call here.

I would imagine one day we will feed arguments here to help apps customize the settings:
```Python
migas_setup(userid="<some identifier generated by the app>", appid="override/fmriprep")
```

fmriprep/cli/run.py
```diff
@@ -71,13 +73,15 @@ def main():
     config.load(config_file)
 
     if config.execution.reports_only:
+        ping_migas(status="success" if retcode == 0 else "error")
```
oesteban, line 76: May we abide by OS' standards and just:
```Python
ping_migas(status=retcode)
```
Then let migas do the mapping.

fmriprep/cli/run.py
```diff
@@ -71,13 +73,15 @@ def main():
...
         sys.exit(int(retcode > 0))
 
     if fmriprep_wf and config.execution.write_graph:
         fmriprep_wf.write_graph(graph2use="colored", format="svg", simple_form=True)
 
     retcode = retcode or (fmriprep_wf is None) * EX_SOFTWARE
     if retcode != 0:
+        ping_migas(status="error")
```
oesteban, line 84: What about replacing all these `ping_migas` with status with an exit hook? - https://docs.python.org/3/library/atexit.html

Since the hook does not allow to access the exit code, we probably want to:
1. Add a new `status` or `exitcode` config setting (perhaps under the `execution` section). Initialize it at -1.
2. Define some codes: e.g., 1001 (I haven't really thought this through, this might not be a good code) is set right before building the workflow, then 1002 is set after it is built and before calling `run()` and finally 0 is set when it is complete and successful. An error in the pipeline overrides the exitcode too.
3. The `atexit()` hook will not send a ping if the code is still -1, and otherwise, submit whatever code it has.

fmriprep/utils/sentry.py
oesteban: I think @mgxd  said that type annotations are not very well supported by Python 3.7, right?

fmriprep/utils/sentry.py
```diff
@@ -180,3 +180,27 @@ def _chunks(string, length=CHUNK_SIZE):
...
+    """Communicate with the migas telemetry server."""
+    import os
+
+    import migas
+
+    if config.execution.notrack:
+        return
+
+    os.environ['ENABLE_MIGAS'] = 'yes'
```
oesteban, line 194:
```suggestion
```

fmriprep/utils/sentry.py
```diff
@@ -180,3 +180,27 @@ def _chunks(string, length=CHUNK_SIZE):
...
             for i in range(0, len(string), length))
+
+
+def ping_migas(status='pending'):
+    """Communicate with the migas telemetry server."""
+    import os
+
+    import migas
+
+    if config.execution.notrack:
```
oesteban, line 191:
```suggestion
    if (
        config.execution.notrack
        or os.getenv("MIGAS_DISABLE", "no").lower() in ("yes", "on", "1", "true")
    ):
```

fmriprep/utils/sentry.py
```diff
@@ -180,3 +180,27 @@ def _chunks(string, length=CHUNK_SIZE):
     """
     return (string[i:i + length]
             for i in range(0, len(string), length))
+
+
+def ping_migas(status='pending'):
+    """Communicate with the migas telemetry server."""
+    import os
+
+    import migas
```
oesteban, line 189: delay after the following if branch?

fmriprep/utils/sentry.py
```diff
@@ -180,3 +180,27 @@ def _chunks(string, length=CHUNK_SIZE):
...
+    if config.execution.notrack:
+        return
+
+    os.environ['ENABLE_MIGAS'] = 'yes'
+    session_id = None
+    if config.execution.run_uuid:
+        session_id = config.execution.run_uuid.split('_')[1]
+
+    # TODO: check for bad versions
+    res = migas.add_project(
```
oesteban, line 200: this looks like what `migas_setup()` would look like inside.

oesteban (commented):
Added some suggestions, not particularly strong on them -- but thanks for considering.

fmriprep/utils/sentry.py
mgxd: nope, they are supported - they are just much more verbose or require `typing-extensions`

fmriprep/cli/run.py
```diff
@@ -25,6 +25,9 @@
 from .. import config
 
 
+EXITCODE: int = -1
```
effigies, line 28: You check for `> 0` in several places. Should they be `!= 0` or should this be something >0, e.g. 255?

mgxd: -1 is just a placeholder, it should never be this value

fmriprep/utils/telemetry.py
```diff
@@ -182,3 +183,27 @@ def _chunks(string, length=CHUNK_SIZE):
...
+    """
+    # generate session UUID from generated run UUID
+    session_id = None
+    if config.execution.run_uuid:
+        session_id = config.execution.run_uuid.split('_', 1)[-1]
+
+    migas.setup(session_id=session_id)
+    if init:
+        # send initial status ping
+        send_breadcrumb(status='R', status_desc='workflow start')
```
effigies, line 201:
```suggestion
def setup_migas(init_ping: bool = True) -> None:
    """
    Prepare the migas python client to communicate with a migas server.
    If ``init_ping`` is ``True``, send an initial breadcrumb.
    """
    # generate session UUID from generated run UUID
    session_id = None
    if config.execution.run_uuid:
        session_id = config.execution.run_uuid.split('_', 1)[-1]

    migas.setup(session_id=session_id)
    if init_ping:
        send_breadcrumb(status='R', status_desc='workflow start')
```

fmriprep/cli/run.py
```diff
@@ -61,11 +63,15 @@ def main():
...
+        import atexit
+
         import sentry_sdk
 
-        from ..utils.sentry import sentry_setup
+        from ..utils.telemetry import sentry_setup, setup_migas
 
         sentry_setup()
+        setup_migas(init_ping=True)
+        atexit.register(_migas_exit)
```
effigies, line 74:
```suggestion
        atexit.register(migas_exit)
```

effigies: @mgxd Went ahead and committed this on my own. If you'd rather it be `_migas_exit` in both locations, that's also fine with me.

PR #2821, opened 2022-07-21
REL: 22.0.0rc4

effigies:
Okay, think we're really just about there. One last RC and then we'll pull the trigger on Monday.

These upgrades resolve DVARS errors in extremely low-variance voxels and GIFTI metadata errors with oblique surfaces.


effigies:
Waiting on smriprep to finish deploying...

PR #2853, opened 2022-09-14
DOCKER: Pin conda environment more strictly

effigies:
Discovered while fixing some bugs in the release. Might as well see if the LTS build is holding up.


effigies:
@mgxd @oesteban @bpinsard It's becoming difficult to have a consistent Python environment. I'm doing a full pin based on the environment in the 20.2.6 container, minus the nipreps. We're going to need to figure out how to walk dependencies forward while verifying their behavior.

Curious if others have better ideas that an `environment.yml`.

effigies:
Bump @oesteban @bpinsard for thoughts. (I see @mgxd gave this a :+1:.)

oesteban (approved):
This feels like it's heading in a good direction. I guess updating dependencies will become a bit more involved. Let's see how it goes and write some update script if necessary down the line.

PR #2857, opened 2022-09-22
MAINT: Add `pre-commit`, dev installation for consistent styling

mgxd:
Adds `pre-commit` hooks for styling to repo, which can be leveraged after running `pre-commit install` at the project root.

Also runs another round of styling across the repo, with the addition of `# fmt:on/off` comments before/after nipype workflow connect statements.

### DIFF
--- a/fmriprep/interfaces/confounds.py
+++ b/fmriprep/interfaces/confounds.py
@@ -171,11 +171,10 @@ class RenameACompCor(SimpleInterface):
         a_orig = a_comp_cor["component"]
         a_new = [f"a_comp_cor_{i:02d}" for i in range(len(a_orig))]
 
-        (
-            components.rename(columns=dict(zip(c_orig, c_new)))
-            .rename(columns=dict(zip(w_orig, w_new)))
-            .rename(columns=dict(zip(a_orig, a_new)))
-        ).to_csv(self._results["components_file"], sep='\t', index=False)
+    final_components = components.rename(columns=dict(zip(c_orig, c_new)))
+    final_components.rename(columns=dict(zip(w_orig, w_new)), inplace=True)
+    final_components.rename(columns=dict(zip(a_orig, a_new)), inplace=True)
+    final_components.to_csv(self._results["components_file"], sep='\t', index=False)
 
         metadata.loc[c_comp_cor.index, "component"] = c_new
         metadata.loc[w_comp_cor.index, "component"] = w_new
--- a/fmriprep/interfaces/confounds.py
+++ b/fmriprep/interfaces/confounds.py
@@ -171,10 +171,10 @@ class RenameACompCor(SimpleInterface):
         a_orig = a_comp_cor["component"]
         a_new = [f"a_comp_cor_{i:02d}" for i in range(len(a_orig))]
 
-    final_components = components.rename(columns=dict(zip(c_orig, c_new)))
-    final_components.rename(columns=dict(zip(w_orig, w_new)), inplace=True)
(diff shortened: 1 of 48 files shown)


fmriprep/workflows/base.py
```diff
@@ -335,24 +372,23 @@ def init_single_subject_wf(subject_id):
         )
 
         if config.workflow.use_syn_sdc and not fmap_estimators:
-            message = ("Fieldmap-less (SyN) estimation was requested, but "
-                       "PhaseEncodingDirection information appears to be "
-                       "absent.")
+            message = (
+                "Fieldmap-less (SyN) estimation was requested, but "
+                "PhaseEncodingDirection information appears to be "
+                "absent."
```
effigies, line 378:
```suggestion
                "Fieldmap-less (SyN) estimation was requested, but PhaseEncodingDirection "
                "information appears to be absent."
```

fmriprep/workflows/base.py
```diff
@@ -335,24 +372,23 @@ def init_single_subject_wf(subject_id):
...
                 raise ValueError(message)
 
-        if (
-            "fieldmaps" in config.workflow.ignore
-            and [f for f in fmap_estimators
-                 if f.method != fm.EstimatorType.ANAT]
-        ):
+        if "fieldmaps" in config.workflow.ignore and [
+            f for f in fmap_estimators if f.method != fm.EstimatorType.ANAT
+        ]:
```
effigies, line 386: I think this should be `==`. Does this match a known bug? Also, either way, it seems cleaner to use `any()` than to make a list and check it's non-empty.

```suggestion
        if "fieldmaps" in config.workflow.ignore and any(
            f.method == fm.EstimatorType.ANAT for f in fmap_estimators
        ):
```

effigies (approved):
Looks good. Some suggestions and also I think I spotted a bug, but please have a look.

fmriprep/workflows/base.py
mgxd: I think you're right, this looks like a bug

fmriprep/interfaces/confounds.py
```diff
@@ -163,10 +171,10 @@ def _run_interface(self, runtime):
...
         a_new = [f"a_comp_cor_{i:02d}" for i in range(len(a_orig))]
 
-        (components.rename(columns=dict(zip(c_orig, c_new)))
-                   .rename(columns=dict(zip(w_orig, w_new)))
-                   .rename(columns=dict(zip(a_orig, a_new)))
-         ).to_csv(self._results["components_file"], sep='\t', index=False)
+    final_components = components.rename(columns=dict(zip(c_orig, c_new)))
+    final_components.rename(columns=dict(zip(w_orig, w_new)), inplace=True)
+    final_components.rename(columns=dict(zip(a_orig, a_new)), inplace=True)
+    final_components.to_csv(self._results["components_file"], sep='\t', index=False)
```
effigies, line 177: Sorry, looks like I gave you a dedented suggestion.

```suggestion
        final_components = components.rename(columns=dict(zip(c_orig, c_new)))
        final_components.rename(columns=dict(zip(w_orig, w_new)), inplace=True)
        final_components.rename(columns=dict(zip(a_orig, a_new)), inplace=True)
        final_components.to_csv(self._results["components_file"], sep='\t', index=False)
```

effigies:
@mgxd Canceled the circle build. Please restart if you think it should be run all the way through.

effigies (approved):

PR #2859, opened 2022-09-26
DOCKER: Build wheel and install in two-stage build

effigies:

This PR aims to make the Docker build less fragile by building the package wheel in an initial stage with full git information, and then copying only the wheel over to the main build.

Assuming that the COPY at L29 doesn't break the caching of everything afterwards, then we should be good. Otherwise we can switch to something like:

```Dockerfile
FROM ubuntu:... as base
...

FROM python:slim as wheel
...

FROM base
COPY --from=wheel ...
...
```

This will also allow us to stop using a custom versioneer.


.dockerignore
```diff
@@ -12,31 +12,3 @@ fmriprep.egg-info/**/*
...
-out/**/*
-out/
-
-# CI, etc.
-.circleci
-.circleci/**/*
-.zenodo.json
-.travis.yml
-.readthedocs.yml
-CONTRIBUTING.md
```
effigies, line 42: These things were removed to avoid getting a dirty repository. We now do not do a `COPY .` outside of the wheel building stage, so this should not pollute the image.

Dockerfile
```diff
@@ -22,6 +22,13 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
+FROM python:slim
+RUN pip install build
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends git
+COPY . /src/fmriprep
+RUN python -m build /src/fmriprep
```
effigies, line 30: I've tested locally. Modifying the repository will make this rerun but will not cause a cache miss on L33+ until there's another change in the Dockerfile or a `COPY`/`ADD` command fails a hash check.

Dockerfile
```diff
@@ -22,6 +22,13 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
+FROM python:slim
```
mgxd, line 25: Named stages make it a little clearer
```suggestion
FROM python:slim AS src
```

Dockerfile
```diff
@@ -283,12 +290,8 @@ RUN /opt/conda/bin/python fetch_templates.py && \
     find $HOME/.cache/templateflow -type f -exec chmod go=u {} +
 
 # Installing FMRIPREP
-COPY . /src/fmriprep
-ARG VERSION
-# Force static versioning within container
-RUN echo "${VERSION}" > /src/fmriprep/fmriprep/VERSION && \
-    echo "include fmriprep/VERSION" >> /src/fmriprep/MANIFEST.in && \
-    /opt/conda/bin/python -m pip install --no-cache-dir "/src/fmriprep[all]"
+COPY --from=0 /src/fmriprep/dist/*.whl .
```
mgxd, line 293:
```suggestion
COPY --from=src /src/fmriprep/dist/*.whl .
```

fmriprep/workflows/bold/outputs.py
```diff
@@ -77,8 +78,9 @@ def prepare_timing_parameters(metadata):
     ...     prepare_timing_parameters(dict(RepetitionTime=2, SliceTiming=[0.0, 0.2, 0.4, 0.6]))
     {'RepetitionTime': 2, 'SliceTimingCorrected': False, 'DelayTime': 1.2}
     >>> with mock.patch("fmriprep.config.workflow.ignore", ["slicetiming"]):
-    ...     prepare_timing_parameters(dict(VolumeTiming=[0.0, 1.0, 2.0, 5.0, 6.0, 7.0],
-    ...                                    SliceTiming=[0.0, 0.2, 0.4, 0.6, 0.8]))
+    ...     prepare_timing_parameters(
+    ...         dict(VolumeTiming=[0.0, 1.0, 2.0, 5.0, 6.0, 7.0],
+    ...              SliceTiming=[0.0, 0.2, 0.4, 0.6, 0.8]))  #doctest: +NORMALIZE_WHITESPACE
```
mgxd, line 83: Should we just add a pytest section to `pyproject.toml`?

```
[tool.pytest]
doctest_optionflags = NORMALIZE_WHITESPACE
```

mgxd (approved):
+1 on removing the versioneer hack

fmriprep/workflows/bold/outputs.py
effigies: That won't impact testing the installed version with `--pyargs` without pulling the repository in. We do have that in `setup.cfg` already.

mgxd (approved):

PR #2871, opened 2022-10-06
MAINT: Add a ``--debug pdb`` to allow easier line-by-line debugging

oesteban:
Makes setting traces and debugging errors easier.

### DIFF
--- a/fmriprep/cli/run.py
+++ b/fmriprep/cli/run.py
@@ -35,11 +35,18 @@ def main():
 
     from ..utils.bids import write_bidsignore, write_derivative_description
     from .parser import parse_args
+    from .workflow import build_workflow
 
     parse_args()
 
+    if "pdb" in config.execution.debug:
+        from fmriprep.utils.debug import setup_exceptionhook
+
+        setup_exceptionhook()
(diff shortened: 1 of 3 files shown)


fmriprep/cli/run.py
```diff
@@ -56,16 +63,19 @@ def main():
     # CRITICAL Call build_workflow(config_file, retval) in a subprocess.
     # Because Python on Linux does not ever free virtual memory (VM), running the
     # workflow construction jailed within a process preempts excessive VM buildup.
-    with Manager() as mgr:
```
effigies, line 59: I had a vague memory that the numpy OMP_NUM_THREADS turned out to be the problem that this was trying to resolve? Do we still need this?

effigies (approved):
Looks good overall.

fmriprep/cli/run.py
oesteban: It was the main culprit. This could probably be removed - but it would be nice to check how much memory it builds up if made within the main thread - probably not much.

oesteban:
The failing test looks like Circle is having the hiccups - I'm going to risk and merge.

PR #2876, opened 2022-10-11
FIX: Bug introduced in ``--debug pdb`` PR (#2871)

oesteban:
Only breaks when running with ``--debug pdb`` because `p.exitcode` will not be defined at that point (old line 77).

### DIFF
--- a/fmriprep/cli/run.py
+++ b/fmriprep/cli/run.py
@@ -140,13 +140,13 @@ def main():
                 for crashfile in crashfolder.glob("crash*.*"):
                     process_crashfile(crashfile)
 
-            if "Workflow did not execute cleanly" not in str(e):
+            if sentry_sdk is not None and "Workflow did not execute cleanly" not in str(e):
                 sentry_sdk.capture_exception(e)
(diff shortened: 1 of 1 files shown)


mgxd:
I'm not sure what is causing the ds054 crash, but it seems consistent...

oesteban:
> I'm not sure what is causing the ds054 crash, but it seems consistent...

It seems to be reaching https://github.com/nipreps/fmriprep/blob/ec4c21bc37a0e035f22fe958aa216a81ee4ac76c/fmriprep/cli/run.py#L148 without issues, and then we don't seem to handle very well the problem, which must happen in relation to the citation boilerplate.

oesteban:
I might have broken the report generation?

https://github.com/nipreps/fmriprep/blob/ec4c21bc37a0e035f22fe958aa216a81ee4ac76c/fmriprep/cli/run.py#L189

effigies (approved):
ds054 passing

oesteban:
Good catch with the sentry problem :)

PR #2879, opened 2022-10-14
FIX: Remove cortex masking during vol2surf sampling

mgxd:
This experiments with removing the cortical masking during the BOLD volume -> surface sampling, as suggested in https://github.com/nipreps/fmriprep/issues/2364#issuecomment-1275400866

The point is if we need to apply this mask in a later point, it should be trivial given the information is present in 
`?h.cortex.label` files.

### DIFF
--- a/fmriprep/workflows/bold/resampling.py
+++ b/fmriprep/workflows/bold/resampling.py
@@ -37,7 +37,7 @@ from nipype.pipeline import engine as pe
 from ...config import DEFAULT_MEMORY_MIN_GB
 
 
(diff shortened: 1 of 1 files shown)


effigies (approved):
I think this is fine as-is. We do medial wall masking separately. The commit (bbcda3d83291b5da291d7c46465006636fb3e5e6) where this was added came with no justification and there's none in PR #391 where it was introduced.

This is a derivatives change, though, so it would be good to put this in the next minor release unless there's some reason to hold off?

effigies:
I'm confident enough in this to let the tests run on master instead of waiting here...

PR #2881, opened 2022-10-19
DEPRECATE: --topup-max-vols

mgxd:
`pepolar.flatten` was removed in the latest SDCFlows, as it now averages any 4D input into a single 3D volume.


mgxd:
reusing anatomicals is going to fail now that smriprep outputs have changed (nipreps/smriprep#305) - I'm just going to pin sdcflows in this PR

PR #2894, opened 2022-11-21
MAINT/CI: Update fast track outputs, use latest smriprep

mgxd:
Anatomicals were generated using smriprep built off https://github.com/nipreps/smriprep/commit/05a762b8b11d6df90fae72ef0f43a27c78e72610


mgxd:
This is working as expected, installation problems seem unrelated and likely related to new `pip` updates (#2895)

effigies:
It doesn't seem to be new pip. Installing pip 22.1 in a Python 3.7 environment produces the same issue for me. I think there must be something in the dependency chain that's causing a resolution failure in 3.7. Pip's resolution mechanism is to find older and older versions of a tool, and we eventually found a nitime that is so old that it imports the entire module and breaks because numpy isn't a build dependency.

I think the thing to do is to drop Python 3.7. [NEP29](https://numpy.org/neps/nep-0029-deprecation_policy.html) recommended dropping it almost a year ago, and I assume that what we're seeing is the consequences of the broader ecosystem no longer testing on it. It's probably not worth a lot of effort tracking this down.

I would suggest we take one of the following routes:

1) Adhere to NEP29 and drop Python minor releases along with the rest of the ecosystem.
2) Continue testing and advertise support for non-EOL Pythons until the fixes become non-obvious.

effigies:
Squashing as this will be more coherent as a single commit. LGTM.

PR #2895, opened 2022-11-23
CI: Test on Python 3.10, bump actions versions

mgxd:


.github/workflows/stable.yml
```diff
@@ -21,7 +21,7 @@ jobs:
     strategy:
       matrix:
         os: ['ubuntu-latest']
-        python-version: [3.7, 3.8, 3.9]
+        python-version: [3.7, 3.8, 3.9, '3.10']
```
effigies, line 24:
```suggestion
        python-version: [3.8, 3.9, '3.10']
```

effigies (commented):
Posted my comment at https://github.com/nipreps/fmriprep/pull/2894#issuecomment-1329308035, but meant to do it here:

> It doesn't seem to be new pip. Installing pip 22.1 in a Python 3.7 environment produces the same issue for me. I think there must be something in the dependency chain that's causing a resolution failure in 3.7. Pip's resolution mechanism is to find older and older versions of a tool, and we eventually found a nitime that is so old that it imports the entire module and breaks because numpy isn't a build dependency.
>
> I think the thing to do is to drop Python 3.7. [NEP29](https://numpy.org/neps/nep-0029-deprecation_policy.html) recommended dropping it almost a year ago, and I assume that what we're seeing is the consequences of the broader ecosystem no longer testing on it. It's probably not worth a lot of effort tracking this down.
>
> I would suggest we take one of the following routes:
>
> 1) Adhere to NEP29 and drop Python minor releases along with the rest of the ecosystem.
> 2) Continue testing and advertise support for non-EOL Pythons until the fixes become non-obvious.

effigies:
Skipping over Circle builds.

PR #2922, opened 2022-12-20
FIX: Ensure version installed in Docker file is clean

effigies:

The last build generated a bad version number in the docker file:

```console
$ docker run --rm -it nipreps/fmriprep:22.1.0 --version
fMRIPrep v22.1.0+0.gce344b39.dirty
```

Building locally, it doesn't seem to have an issue, so it seems to be a CircleCI problem. I've resolved what I think was causing the dirty repository and also added a safeguard that should ensure future CI issues don't affect version determination.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -167,7 +167,6 @@ jobs:
               echo "the command ``git fetch --tags --verbose`` and push"
               echo "them to your fork with ``git push origin --tags``"
             fi
-            sed -i -E "s/(__version__ = )'[A-Za-z0-9.-]+'/\1'${CIRCLE_TAG:-$THISVERSION}'/" wrapper/fmriprep_docker.py
             # Build docker image
             e=1 && for i in {1..5}; do
               docker build --rm \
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -178,6 +177,8 @@ jobs:
                 --build-arg VERSION="${CIRCLE_TAG:-$THISVERSION}" . \
               && e=0 && break || sleep 15
             done && [ "$e" -eq "0" ]
+            # Set wrapper version to match
+            sed -i -E "s/(__version__ = )'[A-Za-z0-9.-]+'/\1'${CIRCLE_TAG:-$THISVERSION}'/" wrapper/fmriprep_docker.py
```
mgxd, line 181: I think we can just drop this, since we do it before publishing the wrapper to PyPI
https://github.com/nipreps/fmriprep/blob/ce344b3963caacea367d512787f1a9065932fbc5/.circleci/config.yml#L966

Dockerfile
```diff
@@ -27,6 +27,8 @@ RUN pip install build
 RUN apt-get update && \
     apt-get install -y --no-install-recommends git
 COPY . /src/fmriprep
+# Ensure clean build
+RUN git -C /src/fmriprep reset --hard && git -C /src/fmriprep clean -fxd
```
mgxd, line 31: This makes our build too rigid - we should still allow building with uncommitted changes

mgxd (commented):
I'm worried we're moving towards rigidity - WDYT about just adding a container version check here before deploying?
https://github.com/nipreps/fmriprep/blob/ce344b3963caacea367d512787f1a9065932fbc5/.circleci/config.yml#L887-L891

.circleci/config.yml
```diff
@@ -178,6 +177,23 @@ jobs:
...
+      - run:
+          name: Check Docker image and bump wrapper version
+          no_output_timeout: 60m
+          command: |
+            export PY3=$(pyenv versions | grep '3\.' |
+                         sed -e 's/.* 3\./3./' -e 's/ .*//')
+            pyenv local $PY3
+            # Get version, update files.
+            THISVERSION=$( python3 get_version.py )
+            BUILT_VERSION=$( docker run --rm -it nipreps/fmriprep:latest --version )
```
mgxd, line 189: interactive/tty option can clutter the output
```suggestion
            BUILT_VERSION=$( docker run --rm nipreps/fmriprep:latest --version )
```

mgxd (approved):
lgtm, left some final comments

.circleci/config.yml
```diff
@@ -178,6 +177,21 @@ jobs:
                 --build-arg VERSION="${CIRCLE_TAG:-$THISVERSION}" . \
               && e=0 && break || sleep 15
             done && [ "$e" -eq "0" ]
+      - run:
+          name: Check Docker image
+          no_output_timeout: 60m
```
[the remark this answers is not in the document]
effigies: Good catch. Was blindly copied from the previous step. Should probably just remove it:
```suggestion
```

effigies:
Build passing. Skipping tests.

PR #2937, opened 2023-01-17
RF: Use DataFrame.rename instead of ad hoc process

### LINKED ISSUE #2904: Remove Pandas Index.ravel()
```
/opt/conda/lib/python3.9/site-packages/fmriprep/interfaces/confounds.py:558: FutureWarning: Index.ravel returning ndarray is deprecated; in a future version this will return a view on self.
  colnames = data.columns.ravel().tolist()
```

effigies:

We had a use of deprecated `Index.ravel()` functionality. It turns out this was to do the same thing as `DataFrame.rename(columns=mapper)`, so I just replaced that.

Closes #2904.

### DIFF
--- a/fmriprep/interfaces/confounds.py
+++ b/fmriprep/interfaces/confounds.py
@@ -555,12 +555,7 @@ class FMRISummary(SimpleInterface):
         else:
             data = dataframe[headers]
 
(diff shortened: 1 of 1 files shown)


mgxd (approved):
much more efficient

PR #2938, opened 2023-01-17
CI: Weekly docker build from scratch

### LINKED ISSUE #2916: CI action to build the Docker image with no layer caching

We try to ensure that we reuse layers across fMRIPrep versions as much as possible. This was especially important when we were releasing multiple times a month, but it will still save people disk space.

However, our Docker images should be buildable without any cached layers, and we should have a CI job (I would suggest a cron job, not on push) just to warn us about bit rot in the Dockerfile.

Yes

_No response_

effigies:

Closes #2916.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -95,6 +95,31 @@ orbs:
   docker: circleci/docker@2.1.4
 
 jobs:
+  build_clean:
+    <<: *machine_defaults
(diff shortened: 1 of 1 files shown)


effigies:
This didn't break the CI config (I cancelled, though). I think we just merge and see if it works next Monday...

PR #2943, opened 2023-01-27
DOC: Update lesion ROI documentation, warn in docs and app about upcoming changes

### LINKED ISSUE #2892: lesion mask: inconsistent documentation ?
I think there is an inconsistency in the documentation regarding how lesion mask should be named.

### From the normalization section

https://fmriprep.org/en/stable/workflows.html#cost-function-masking-during-spatial-normalization

> Lesion masks should be binary NIfTI images (damaged areas = 1, everywhere else = 0) in the same space and resolution as the T1 image, and follow the naming convention specified in [BIDS Extension Proposal 3: Common Derivatives](https://docs.google.com/document/d/1Wwc4A6Mow4ZPPszDIWfCUCRNstn7d_zzaWPcfcHmgI4/edit#heading=h.9146wuepclkt) (e.g., sub-001_T1w_label-lesion_roi.nii.gz). This file should be placed in the sub-*/anat directory of the BIDS dataset to be run through fMRIPrep. Because lesion masks are not currently part of the BIDS specification, it is also necessary to include a .bidsignore file in the root of your dataset directory. This will prevent [bids-validator](https://github.com/bids-standard/bids-validator#bidsignore) from complaining that your dataset is not valid BIDS, which prevents fMRIPrep from running. Your .bidsignore file should include the following line:
>
>`*lesion_roi.nii.gz`

The name `sub-001_T1w_label-lesion_roi.nii.gz` looks unBIDSish to me.

Shouldn't it be: `sub-001_label-lesion_roi.nii.gz` ?

### From the BEP003 google doc

Linked in the normalization section

https://docs.google.com/document/d/1Wwc4A6Mow4ZPPszDIWfCUCRNstn7d_zzaWPcfcHmgI4/edit#heading=h.9146wuepclkt

```text
manual_masks/
     sub-001/
         anat/
            sub-001_T1w_label-desc_mask.nii.gz
```

This one specifies a mask suffix.

Also shouldn't the doc point to the specification of the derivatives for this? If it makes sense to keep this link here?

### From the bids filter file FAQ section

https://fmriprep.org/en/stable/faq.html#how-do-i-select-only-certain-files-to-be-input-to-fmriprep

```json
{
  "fmap": {"datatype": "fmap"},
  "bold": {"datatype": "func", "suffix": "bold"},
  "sbref": {"datatype": "func", "suffix": "sbref"},
  "flair": {"datatype": "anat", "suffix": "FLAIR"},
  "t2w": {"datatype": "anat", "suffix": "T2w"},
  "t1w": {"datatype": "anat", "suffix": "T1w"},
  "roi": {"datatype": "anat", "suffix": "roi"}
}
```

### Conclusion

So from reading all this I would not be sure how to name my lesion mask.

I would be tempted to make the bids filter file the final authority on this and hence have a file like this one would be fine.

```text
manual_masks/
     sub-001/
         anat/
            sub-001_roi.nii.gz
```

If that is the case, should I edit the doc to fix this?

effigies:

* Update docs to better match BIDS conventions for lesion ROIs, but do not change existing behavior.
* Emit warning that a change will be coming and users working with this feature should check the docs before doing further upgrades.

Closes #2892.

* Added notice that lesion ROI detection will change in an upcoming version (not 23.0).

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -226,7 +226,7 @@ To do so, follow the next steps.
 After getting the resources you'll need, you will just need to make sure your
 runtime environment is able to access the filesystem, at the location of your
 *TemplateFlow home* directory.
-If you are a Singularity user, please check out `TemplateFlow and Singularity
+If you are an Apptainer (formerly Singularity) user, please check out `TemplateFlow and Singularity
 <https://www.nipreps.org/apps/singularity/#templateflow-and-singularity>`__.
 
 How do I select only certain files to be input to *fMRIPrep*?
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -226,7 +226,8 @@ To do so, follow the next steps.
 After getting the resources you'll need, you will just need to make sure your
 runtime environment is able to access the filesystem, at the location of your
 *TemplateFlow home* directory.
-If you are a Singularity user, please check out :ref:`singularity_tf`.
+If you are a Singularity user, please check out `TemplateFlow and Singularity
+<https://www.nipreps.org/apps/singularity/#templateflow-and-singularity>`__.
 
 How do I select only certain files to be input to *fMRIPrep*?
 -------------------------------------------------------------
--- a/docs/workflows.rst
+++ b/docs/workflows.rst
@@ -79,6 +79,7 @@ single reference template (see `Longitudinal processing`_).
         ]),
         skull_strip_fixed_seed=False,
         t1w=['sub-01/anat/sub-01_T1w.nii.gz'],
+        t2w=[],
     )
 
 .. important::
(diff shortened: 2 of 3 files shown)


effigies:
image

docs/faq.rst
```diff
@@ -226,7 +226,8 @@ To do so, follow the next steps.
 After getting the resources you'll need, you will just need to make sure your
 runtime environment is able to access the filesystem, at the location of your
 *TemplateFlow home* directory.
-If you are a Singularity user, please check out :ref:`singularity_tf`.
+If you are a Singularity user, please check out `TemplateFlow and Singularity
```
mgxd, line 229:
```suggestion
If you are an Apptainer (formerly Singularity) user, please check out `TemplateFlow and Singularity
```

mgxd (approved):

PR #2950, opened 2023-02-13
DOC: update the description of the carpetplot in the sample report

celprov:


effigies:
Is this taken directly from somewhere? I thought we took an actual output report here, so I'm surprised this is the one change.

celprov:
@effigies  While visualizing a report from fMRIPrep from v22.0.0, I noticed the edge is not mentioned in the text describing the carpetplot that is why I thought of quickly adding it. Does it reply to your question ?

Issue #2962, opened 2023-03-10
Comments on HTML report version 23.0.0rc0


julfou81:
Many thanks for making fmriprep such a useful tool, and maintaining it to such high standards!

I gave a shot at FMRIPREP version 23.0.0rc0,and compared it with v22.1.1 for the same subject. I thought I would report what I noticed in the HTML report, before the release of v23.0.0.

- Alignement of functional and anatomical: strange red square around the moving `boldref`:
<img width="1486" alt="Capture d’écran 2023-03-10 à 23 53 57" src="https://user-images.githubusercontent.com/32452352/224443335-e439bfeb-627b-4bfc-b226-4046f60dc944.png">

I also preferred to see the anatomical segmentation visible on those images, as it is shown on the report for v22.1.1:
<img width="1491" alt="Capture d’écran 2023-03-10 à 23 55 52" src="https://user-images.githubusercontent.com/32452352/224443526-8d4c55fc-a0a1-454f-a9b9-32681dd2967e.png">

- On the carpet plot, the legend changed compared to v22.1.1, one category being sWM+sCSF, the other one being dWM+dCSF and the color does not match what is said in the description above the carpet plot:
<img width="1474" alt="Capture d’écran 2023-03-10 à 23 57 35" src="https://user-images.githubusercontent.com/32452352/224443896-8f64d0ae-39e3-423d-bc49-8be4bfbc3d39.png">

- The correlation matrix among nuisance regressors in v23.0.0rc0 has a different ordering of the regressors than v22.1.1:0rc0:
v23.0.0rc0:
<img width="1498" alt="Capture d’écran 2023-03-11 à 00 01 09" src="https://user-images.githubusercontent.com/32452352/224444203-7c5cce04-6a10-4279-9950-f49684137a7a.png">

v22.1.1:
<img width="1494" alt="Capture d’écran 2023-03-11 à 00 01 25" src="https://user-images.githubusercontent.com/32452352/224444257-3a994ef4-c384-48b4-8320-1e1cc33d4da1.png">

effigies:
Thanks for the feedback. The contour problem in your first couple plots has been resolved and will be fixed in the full release.

The confound ordering is now deterministic where it had been unhelpfully random. Another ordering might be more useful, and we're happy to hear suggestions, but we won't go back to random.

I'll have a look at the carpet plot when I can. On my phone right now...

julfou81:
Thank you for your answer!
[...]

PR #2964, opened 2023-03-13
REL: 23.0.0

effigies:

Updated niworkflows, smriprep and sdcflows dependencies over the weekend.

@mgxd Any recommendations for running pip-tools here? Or should setting up that process be a goal for 23.1.0? (Note I would like to release today, rather than iterate for a couple days.)

@julfou81 At the prompting of #2962, I've updated the carpetplot caption to be a bit more general and not tie itself to specific colors, instead explaining how to read the legend. LMK what you think.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,5 @@
-23.0.0 (To be determined)
-=========================
+23.0.0 (March 13, 2023)
+=======================
 New feature release in the 23.0.x series.
 
 This release adds improvements for workflows targeting the fsLR grayordinate space.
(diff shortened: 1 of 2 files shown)


mgxd:
In migas, I've been pushing a `stable-requirements.txt` (generated via `pip-sync`) and then using a Docker BUILDARG to control whether to install with the latest dependencies (`pip install *whl`) or fixed (`pip-compile stable-requirements.txt`). For fmriprep, I think first shifting away from conda would go a long way to facilitating this change - I can work on that for 23.1.0

julfou81:
The carpet plot looks good like this and the description seems good as well! Nice job!

effigies:
Thanks for the feedback. There should be a full release on PyPI and DockerHub by the end of the day.

PR #2982, opened 2023-04-05
MNT: Refactor Docker build process

effigies:

* Break Docker build into stages, with each significant (and often slow) download getting its own stage
* Switch conda environment to being micromamba-managed
* Create `env.yml` with the pinned conda environment, keeping this information outside of the Dockerfile
   * Due to FSL relying on portions of the scientific python stack, we continue to install that from conda
* Use `pip-compile` to pin our dependency graph and install that after the conda environment

This means that `env.yml` and `requirements.txt` need to be updated in lock-step if the dependency affects `env.yml`. We can probably script this.


effigies:
@mgxd @oesteban Here's a working proposal. ds005 is expected to fail due to ICA-AROMA and FSL 6.0.6.2.

effigies:
Test failures: nipreps/niworkflows#795.

mgxd:
Looks like something is wrong with ds210's SDC

```
Traceback:
	  File "/opt/conda/envs/fmriprep/lib/python3.10/site-packages/nipype/interfaces/base/core.py", line 398, in run
	  File "/opt/conda/envs/fmriprep/lib/python3.10/site-packages/niworkflows/interfaces/nibabel.py", line 467, in _run_interface
	    self._results["out_file"] = _advanced_clip(
	  File "/opt/conda/envs/fmriprep/lib/python3.10/site-packages/niworkflows/interfaces/nibabel.py", line 705, in _advanced_clip
	    a_min = np.percentile(
	  File "<__array_function__ internals>", line 200, in percentile
	  File "/opt/conda/envs/fmriprep/lib/python3.10/site-packages/numpy/lib/function_base.py", line 4205, in percentile
	    return _quantile_unchecked(
	  File "/opt/conda/envs/fmriprep/lib/python3.10/site-packages/numpy/lib/function_base.py", line 4473, in _quantile_unchecked
	    return _ureduce(a,
	  File "/opt/conda/envs/fmriprep/lib/python3.10/site-packages/numpy/lib/function_base.py", line 3752, in _ureduce
	    r = func(a, **kwargs)
	  File "/opt/conda/envs/fmriprep/lib/python3.10/site-packages/numpy/lib/function_base.py", line 4639, in _quantile_ureduce_func
	    result = _quantile(arr,
	  File "/opt/conda/envs/fmriprep/lib/python3.10/site-packages/numpy/lib/function_base.py", line 4745, in _quantile
	    take(arr, indices=-1, axis=DATA_AXIS)
	  File "<__array_function__ internals>", line 200, in take
	  File "/opt/conda/envs/fmriprep/lib/python3.10/site-packages/numpy/core/fromnumeric.py", line 190, in take
	    return _wrapfunc(a, 'take', indices, axis=axis, out=out, mode=mode)
	  File "/opt/conda/envs/fmriprep/lib/python3.10/site-packages/numpy/core/fromnumeric.py", line 57, in _wrapfunc
	    return bound(*args, **kwds)
	IndexError: cannot do a non-empty take from an empty axes.
```

effigies:
I guess we can try bumping caches...

effigies:
@mgxd I canceled ds005 just because it takes so long. Otherwise this is passing. Note the codecov removal is due to https://community.codecov.com/t/codecov-yanked-from-pypi-all-versions/4259, and has no additional impact since we haven't been using that package.

Dockerfile
```diff
@@ -22,147 +22,122 @@
...
 
-# Use Ubuntu 22.04 LTS
-FROM ubuntu:jammy-20221130
+#
+# Download stages
+#
 
-# Prepare environment
+# Utilities for downloading packages
+FROM ubuntu:jammy-20221130 as downloader
```
mgxd, line 40: What about setting a build arg for the base OS:

```suggestion
# Utilities for downloading packages
ARG BASE_OS=ubuntu:jammy-20221130
FROM ${BASE_OS} as downloader
```

Dockerfile
```diff
@@ -22,147 +22,122 @@
...
+
+COPY requirements.txt /tmp/requirements.txt
+RUN /opt/conda/envs/fmriprep/bin/pip install --no-cache-dir -r /tmp/requirements.txt
+
+#
+# Main stage
+#
+
+# Use Ubuntu 22.04 LTS
+FROM ubuntu:jammy-20221130 as fmriprep
```
mgxd, line 117: then we could avoid any desync here
```suggestion
FROM ${BASE_OS} as fmriprep
```

requirements.txt
```diff
@@ -0,0 +1,337 @@
+#
+# This file is autogenerated by pip-compile with Python 3.10
```
mgxd, line 2: Is this handling the extras too? It should account for `[telemetry,test]`

Dockerfile
```diff
@@ -174,127 +149,112 @@ RUN apt-get update -qq \
...
+
+# Unless otherwise specified each process should only use one thread - nipype
+# will handle parallelization
+ENV MKL_NUM_THREADS=1 \
+    OMP_NUM_THREADS=1
+
 # Installing FMRIPREP
 COPY --from=src /src/fmriprep/dist/*.whl .
-RUN /opt/conda/bin/python -m pip install --no-cache-dir $( ls *.whl )[all]
+RUN pip install --no-cache-dir $( ls *.whl )[telemetry,test]
```
mgxd, line 246: Alternatively, we could just install 

```
pip install --no-cache-dir $( ls *.whl )[telemetry]
```
as a base build step, and then create a testing build step that installs the testing dependencies. But that may be excessively complicated for a very minor size improvement.

mgxd (approved):
A few suggestions, but this is really nice! 

One consideration (not needed for this PR) would be to create these build stages into their own dedicated repos in nipreps-containers, so other tools could easily grab what they need with a simple `COPY --from=<nipreps/tool> ...`

Closes https://github.com/nipreps/fmriprep/issues/2441

requirements.txt
effigies: I'm okay adding telemetry, but the function of the tool does not depend on the versions of test dependencies, so I'd rather not set hard pins.

Dockerfile
```diff
@@ -22,147 +22,123 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
```
mgxd, line 24: ah, the build arg needs to be defined before any FROM instructions to be reused.
```suggestion
ARG BASE_OS=ubuntu:jammy-20221130
```

Dockerfile
```diff
@@ -22,147 +22,123 @@
...
 
-# Use Ubuntu 22.04 LTS
-FROM ubuntu:jammy-20221130
+#
+# Download stages
+#
 
-# Prepare environment
+# Utilities for downloading packages
+ARG BASE_OS=ubuntu:jammy-20221130
```
mgxd, line 40:
```suggestion
```

mgxd (commented):
should fix it, my bad!

Dockerfile
effigies: Yeah, been playing with this. It only shaves off 20MB. I think not worth it.

Dockerfile
effigies: I fixed that as well.

PR #3012, opened 2023-05-20
CI: Use registry for layer caching

effigies:

Getting tired of build times again, so experimented with layer caching.

Some references for how I got here:

* https://docs.docker.com/build/cache/backends/registry/
* https://docs.dagger.io/1237/persistent-cache-with-dagger/


effigies:
@mgxd @oesteban This shaves about 50% of our build time out, and we're definitely getting caching. The mamba environment seems to be rebuilt every time, but FreeSurfer isn't re-downloading.

PR #3011, opened 2023-05-18
ENH: Resample BOLD to fsLR directly, dropping fsaverage intermediate

effigies:
## Changes proposed in this pull request

1) Remove goodvoxels and CIFTI-specific code from `mri_vol2surf` resampling workflow. fsnative and fsaverage will remain the same for now.
2) Removes `fsaverage` as an implicit resampling target for `--cifti-output`. `fsaverage` and `fsLR` are now entirely independent.
3) Create a subject-specific cortical ROI mask in GIFTI space. Not currently output, as it's only needed internally; it's not a great mask but is intended to be more liberal than the fsLR mask applied later.
4) Adds a number of `wb_command` commands, including an OpenMP mixin that allows `OMP_NUM_THREADS` to be set in the environment. I think most commands can take advantage, but I only tagged slow ones (take >10s) that I could verify actually used them.
5) Depends on https://github.com/nipreps/smriprep/pull/339 and https://github.com/nipreps/niworkflows/pull/806.

TODO:

* [x] Restore literate workflow code.
* [x] Restore saving goodvoxels mask, if desired.

## Documentation that should be reviewed

TODO

---

Closes #2990.

PR #3022, opened 2023-06-02
FIX: Update sloppy/debug parameters in anatomical base workflow

mgxd:
Reflect changes in https://github.com/nipreps/smriprep/pull/344


effigies:
I think this will fail because we don't have the pin. Going to merge the smriprep PR and push a commit here to pin smriprep master.

effigies:
Just folded this into #3011, since that has some fixes needed for pinning git+https dependencies.

PR #3030, opened 2023-06-08
Add `--me-t2s-fit-method` parameter

### LINKED ISSUE #3029: Add option to control t2smap fittype

Users have reported memory issues with the nonlinear T2*/S0 fitting setting (e.g., #2935 and #2827), so it might be worthwhile to give them control over the setting. I don't think it merits a command-line parameter, but is it perhaps something that could be controlled via the config file?

Yes

Alternatively, I could change the fittype from nonlinear to the less memory-intensive, but potentially less accurate, `loglin` option.

tsalo:
Closes #3029.

- Add a new command line parameter, `--me-t2s-fit-method`, to control how T2* and S0 will be estimated with tedana. The default, "curvefit", is slower and more memory intensive, which has led to out-of-memory errors for some users. The new option, "loglin", shouldn't result in the same memory issues, but it also may produce less accurate estimates.

I added a couple of sentences on the new parameter to the "T2*-driven echo combination" section of `workflows.rst`.

### DIFF
--- a/fmriprep/workflows/bold/t2s.py
+++ b/fmriprep/workflows/bold/t2s.py
@@ -89,12 +89,19 @@ def init_bold_t2s_wf(
     from niworkflows.interfaces.morphology import BinaryDilation
 
     workflow = Workflow(name=name)
-    workflow.__desc__ = """\
+    if config.workflow.me_t2s_fit_method == "curvefit":
+        fit_str = (
+            "nonlinear regression. "
+            "The T2<sup>★</sup>/S<sub>0</sub> estimates from a log-linear regression fit "
+            "were used for initial values"
+        )
+    else:
+        fit_str = "log-linear regression"
+
+    workflow.__desc__ = f"""\
 A T2<sup>★</sup> map was estimated from the preprocessed EPI echoes, by voxel-wise fitting
 the maximal number of echoes with reliable signal in that voxel to a monoexponential signal
-decay model with nonlinear regression.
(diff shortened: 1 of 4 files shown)


effigies (approved):
Nice and straightforward. Unfortunately, I suspect that the ds210 failure that seems to be happening on most runs now will prevent us from verifying that it runs clean. Hopefully #3031 resolves the issue and we can retest this.

fmriprep/workflows/bold/t2s.py
```diff
@@ -109,7 +116,7 @@ def init_bold_t2s_wf(
     dilate_mask = pe.Node(BinaryDilation(radius=2), name='dilate_mask')
 
     t2smap_node = pe.Node(
-        T2SMap(echo_times=list(echo_times)),
+        T2SMap(echo_times=list(echo_times), fittype=config.workflow.me_t2s_fit_method),
         name='t2smap_node',
         mem_gb=2.5 * mem_gb * len(echo_times),
```
mgxd, line 121: reading that curvefit is more memory intensive, will this estimate need to be adjusted?

tsalo: The estimate there came from #2728, wherein I profiled t2smap with curvefit, so I think it should be accurate for curvefit. Would having mem_gb depend on the parameter (e.g., have a lower value for loglin) be a good idea?

mgxd: ah, if we're covering the upper bound then no need to get in the weeds.

PR #3031, opened 2023-06-08
PIN: nipreps/sdcflows#366

effigies:
Integration test for nipreps/sdcflows#366. Do not merge.


effigies:
Canceled ds005 just to save time. This PR has served its purpose.

PR #3038, opened 2023-06-14
MNT: Python 3.11 should be supported

DimitriPapadopoulos:

Test with Python 3.11, show Python 3.11 as supported.

\-

### DIFF
--- a/.github/workflows/pre-release.yml
+++ b/.github/workflows/pre-release.yml
@@ -29,7 +29,7 @@ jobs:
     strategy:
       matrix:
         os: ['ubuntu-latest']
-        python-version: [3.9, '3.10']
+        python-version: [3.9, '3.10', '3.11']
         install: ['pip']
         check: ['tests']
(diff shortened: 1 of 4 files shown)


.github/workflows/pre-release.yml
```diff
@@ -29,7 +29,7 @@ jobs:
     strategy:
       matrix:
         os: ['ubuntu-latest']
-        python-version: [3.9, '3.10']
+        python-version: [3.9, '3.10', '3.11']
```
DimitriPapadopoulos, line 32: Keep all versions of just a few, such as 3.9 and 3.11?

effigies: This works. Compared to the CircleCI jobs, these take no time, so I'm fine risking over-coverage.

effigies:
Thanks! We've been updating the dependency stack, but it took a while to percolate here and I kind of forgot...

PR #3040, opened 2023-06-14
CI: Clean up pre-release builds

effigies:
Pre-release tests on GitHub actions are taking a long time, almost all in installation. It turns out the Pandas is uploading sdists, but not a full set of wheels, to nightly. This adds `--prefer-binary` to just get the latest wheel they uploaded.

Also, we don't plan to update dependencies during maintenance phases, so I'm disabling pre-release tests for `maint/` branches. If this is merged, I'll cherry-pick onto `maint/20.2.x` and `maint/23.1.x`.

### DIFF
--- a/.github/workflows/pre-release.yml
+++ b/.github/workflows/pre-release.yml
@@ -4,11 +4,9 @@ on:
   push:
     branches:
       - master
-      - maint/*
   pull_request:
     branches:
       - master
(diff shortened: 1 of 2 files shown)


effigies:
Merging and cherry-picking back to `maint/23.1.x`. Can revert if y'all disagree.

PR #3073, opened 2023-08-17
MNT: Update environment

effigies:
We've got a couple dependabot nudges. Running `pip-compile`, but with the following manual edits:

1) No change to nipreps dependencies. This also means skipping `importlib_resources`.
2) Avoiding scipy 1.11 for now, as networkx has pinned `!= 1.11.0, != 1.11.1`.

Assuming this passes, I'm going to see about relaxing the tedana pin.


mgxd:
looks like we're hitting https://github.com/matplotlib/matplotlib/issues/26468, so we'll have to wait for a new matplotlib release

effigies:
Merging so we can keep it moving in the morning. Feel free to review post hoc and we can fix.

PR #3076, opened 2023-08-22
MNT: Pin sdcflows 2.5.x

effigies:

#3075 shows that the refactor necessary to use the new `unwarp_wf` will be bigger than just adding a separate unwarp workflow for the reference. Not sure it's worth the effort, given that sdcflows `master` is in flux and I'm working with the new workflow in `next`.


effigies:
Combined into #3073

PR #3079, opened 2023-08-24
CI: Stop testing legacy layout

effigies:
It's been long enough, and a good time to change the outputs is when we've disabled checking that they don't change.


effigies:
ds005 is down to 20 minutes with a two-BOLD run and a one-BOLD re-run + SyN-SDC.

I promise I'll give y'all more chance to review in the future, but I don't want to keep wasting CI time with pointless work, especially when tests are passing and can start meaning something again.

PR #3081, opened 2023-08-27
ENH: Generate anatomical derivatives useful for resampling

effigies:
Adds a `--level` flag to permit going beyond minimal. First thing beyond minimal is surface resampling outputs.

Also factored out the fieldmap estimator discovery into its own function.

### DIFF
--- a/.github/workflows/pre-release.yml
+++ b/.github/workflows/pre-release.yml
@@ -27,7 +27,7 @@ jobs:
     strategy:
       matrix:
         os: ['ubuntu-latest']
-        python-version: [3.9, '3.10', '3.11']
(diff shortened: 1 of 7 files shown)


effigies:
cc @feilong As of this merge, you should be able to get out the anatomical bits. Still not generating the weights from @HippocampusGirl's update, but this should be good enough to get you going. Will also push to the Docker tag `nipreps/fmriprep:next`, so you can test it out. Use `--level full` just to make everything happen.

effigies:
The docker image should be good to pull now.

feilong:
Just pulled it with apptainer/singularity, will test it soon.

PR #3116, opened 2023-10-18
ENH: Restore resampling to T1w target

effigies:
## Changes proposed in this pull request

Adds T1w resampling. Unconditionally resamples in T1w space, but conditionally outputs.

PR #3121, opened 2023-10-26
ENH: Restore resampling BOLD to volumetric templates

effigies:
Builds on #3116.

### DIFF
--- a/fmriprep/utils/transforms.py
+++ b/fmriprep/utils/transforms.py
@@ -43,41 +43,34 @@ def load_ants_h5(filename: Path) -> nt.TransformChain:
     return nt.TransformChain([warp_transform, nt.Affine(affine)])
 
 
+FIXED_PARAMS = np.array([
+    193.0, 229.0, 193.0,
+    96.0, 132.0, -78.0,
+    1.0, 1.0, 1.0,
+    -1.0, 0.0, 0.0,
+    0.0, -1.0, 0.0,
+    0.0, 0.0, 1.0,
+])  # fmt:skip
+
+
 def parse_combined_hdf5(h5_fn, to_ras=True):
     # Borrowed from https://github.com/feilong/process
     # process.resample.parse_combined_hdf5()
     h = h5py.File(h5_fn)
     xform = ITKCompositeH5.from_h5obj(h)
     affine = xform[0].to_ras()
+    transform2 = h['TransformGroup']['2']
     # Confirm these transformations are applicable
-    assert (
-        h['TransformGroup']['2']['TransformType'][:][0] == b'DisplacementFieldTransform_float_3_3'
-    )
-    assert np.array_equal(
-        h['TransformGroup']['2']['TransformFixedParameters'][:],
-        np.array(
-            [
-                193.0,
-                229.0,
-                193.0,
-                96.0,
-                132.0,
-                -78.0,
-                1.0,
-                1.0,
-                1.0,
-                -1.0,
-                0.0,
-                0.0,
-                0.0,
-                -1.0,
-                0.0,
-                0.0,
-                0.0,
-                1.0,
-            ]
-        ),
-    )
+    if transform2['TransformType'][:][0] != b'DisplacementFieldTransform_float_3_3':
+        msg = 'Unknown transform type [2]\n'
+        for i in h['TransformGroup'].keys():
+            msg += f'[{i}]: {h["TransformGroup"][i]["TransformType"][:][0]}\n'
+        raise ValueError(msg)
+    if not np.array_equal(transform2['TransformFixedParameters'], FIXED_PARAMS):
+        msg = 'Unexpected fixed parameters\n'
+        msg += f'Expected: {FIXED_PARAMS}\n'
+        msg += f'Found: {transform2["TransformFixedParameters"][:]}'
+        raise ValueError(msg)
     warp = h['TransformGroup']['2']['TransformParameters'][:]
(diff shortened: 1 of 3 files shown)


fmriprep/utils/transforms.py
```diff
@@ -43,50 +44,60 @@ def load_ants_h5(filename: Path) -> nt.TransformChain:
...
-            [0.0, 0.0, 1.0, -78.0],
-            [0.0, 0.0, 0.0, 1.0],
-        ]
-    )
+
+    warp_affine = np.eye(4)
+    warp_affine[:3, :3] = fixed_params[9:].reshape((3, 3))
+    warp_affine[:3, 3] = fixed_params[3:6]
+    lps_to_ras = np.eye(4) * np.array([-1, -1, 1, 1])
+    warp_affine = lps_to_ras @ warp_affine
```
effigies, line 89: @feilong FYI this appears to be the conversion from fixed parameters to warp/affine modifications. At the very least, it works for the fixed parameters you expect.

It seems we test with ones that don't quite match, but should have enough coverage to not be worrisome.

```
        ValueError: Unexpected fixed parameters
        Expected: [193. 229. 193.  96. 132. -78.   1.   1.   1.  -1.   0.   0.   0.  -1.
           0.   0.   0.   1.]
        Found: [182. 218. 182.  91. 126. -72.   1.   1.   1.  -1.   0.   0.   0.  -1.
           0.   0.   0.   1.]
```

effigies: See also https://itk.org/Doxygen/html/classitk_1_1DisplacementFieldTransform.html#a869379345c48a09a274b023fb0de7b68.

feilong: Thanks for taking care of this!  It's reassuring to know that it works for other parameters.

effigies:
@mgxd This is ready for review.

fmriprep/utils/transforms.py
```diff
@@ -36,57 +37,76 @@ def load_transforms(xfm_paths: list[Path], inverse: list[bool]) -> nt.base.Trans
...
+    # This should be upstreamed into nitransforms
+    h = h5py.File(filename)
     xform = ITKCompositeH5.from_h5obj(h)
-    affine = xform[0].to_ras()
+
+    # nt.Affine
+    transforms = [nt.Affine(xform[0].to_ras())]
+
+    if '2' not in h['TransformGroup']:
+        return transforms[0]
```
mgxd, line 68: this is returning an np.array, not a TransformChain

fmriprep/workflows/bold/base.py
```diff
@@ -381,6 +388,59 @@ def init_bold_wf(
             (bold_anat_wf, ds_bold_t1_wf, [('outputnode.bold_file', 'inputnode.bold')]),
         ])  # fmt:skip
 
+    if spaces.get_spaces(nonstandard=False, dim=(3,)):
+        # Missing:
+        #  * Clipping BOLD after resampling
+        #  * Resampling parcellations
```
mgxd, line 394: separate pr?

fmriprep/utils/transforms.py
```diff
@@ -36,57 +37,76 @@ def load_transforms(xfm_paths: list[Path], inverse: list[bool]) -> nt.base.Trans
...
+        raise ValueError(msg)
+
+    fixed_params = transform2['TransformFixedParameters'][:]
+    if not np.array_equal(fixed_params, FIXED_PARAMS):
+        msg = 'Unexpected fixed parameters\n'
+        msg += f'Expected: {FIXED_PARAMS}\n'
+        msg += f'Found: {fixed_params}'
+        if not np.array_equal(fixed_params[6:], FIXED_PARAMS[6:]):
+            raise ValueError(msg)
+        warnings.warn(msg)
```
mgxd, line 86: is the purpose of warning than erroring to allow use of customized xform parameters?

mgxd (commented):
A few questions / comments

fmriprep/utils/transforms.py
effigies: Actually, an `nt.Affine()`, but fair point. I'll relax the return type.

fmriprep/utils/transforms.py
```diff
@@ -36,57 +37,76 @@ def load_transforms(xfm_paths: list[Path], inverse: list[bool]) -> nt.base.Trans
...
+    1.0, 1.0, 1.0,        # Spacing
+    -1.0, 0.0, 0.0,       # Directions
+    0.0, -1.0, 0.0,
+    0.0, 0.0, 1.0,
+])  # fmt:skip
 
 
-def parse_combined_hdf5(h5_fn, to_ras=True):
+def load_ants_h5(filename: Path) -> nt.TransformChain:
+    """Load ANTs H5 files as a nitransforms TransformChain"""
```
effigies, line 51:
```suggestion
def load_ants_h5(filename: Path) -> nt.base.TransformBase:
    """Load ANTs H5 files as a nitransforms TransformChain"""
```

fmriprep/utils/transforms.py
effigies: The purpose is for debugging. I *think* we know how to deal with unusual parameters, but it would be nice to get the hint that this could be the source of problems, if something goes wrong.

fmriprep/workflows/bold/base.py
effigies: Yes, if we want to restore them.

effigies:
@mgxd Good to go, or do you want to review again?

mgxd (approved):

PR #3120, opened 2023-10-26
ENH: Restore confound generation

effigies:
Builds on #3116. Just check the last two commits.

PR #3126, opened 2023-11-03
ENH: Restore resampling to surface GIFTIs

effigies:

This PR moves the datasinks for GIFTI files into the surface resampling workflow, and adds the resampling workflow to the main BOLD workflow.

Previously we used these GIFTI files as intermediates for CIFTI, which required collation across multiple sub-workflows. With that now being done by other means, we have the opportunity to remove several metadata-collation nodes. We are also able to get rid of a `JoinNode` altogether, as iteration is all we need.

This workflow might now belong in `workflows.bold.outputs`, but that would make the overall changes harder to assess.

This builds on #3120. It also has a cleanup commit that removes portions of the derivatives workflow that have been reimplemented elsewhere.


effigies:
@mgxd Ready for review.

fmriprep/workflows/bold/resampling.py
```diff
@@ -172,47 +180,43 @@ def select_target(subject_id, space):
...
             ("subjects_dir", "subjects_dir")
         ]),
         (inputnode, targets, [("subject_id", "subject_id")]),
         (inputnode, itk2lta, [
-            ("source_file", "reference"),
+            ("bold_t1w", "moving"),
             ("fsnative2t1w_xfm", "in_xfms"),
         ]),
-        (get_fsnative, itk2lta, [("T1", "moving")]),
+        (get_fsnative, itk2lta, [("T1", "reference")]),
```
effigies, line 207: Note that this reversal fixes a bug introduced in 66cb0bb34a3233e6bf2827a43b01ea8b8039f889, which is only part of `:next`. The bug was not discovered because this workflow was not hooked up before now.

PR #3129, opened 2023-11-08
ENH: Restore CIFTI-2 generation

effigies:
Builds on #3126.

This one splits out the selection of the MNI template for the CIFTI-2 generation from the rest of the resampling to volumetric templates. If someone requests MNI152NLin6Asym and CIFTI, then we will resample twice in the working directory, but given that resampling is significantly faster now and does not scale files per-time point, I'm okay with that. I think it's a pretty rare case, and I suspect we'll want to resample a smaller ROI than the full brain at some point, so a separate pathway for that seems okay.

### DIFF
--- a/fmriprep/workflows/bold/base.py
+++ b/fmriprep/workflows/bold/base.py
@@ -213,6 +213,9 @@ def init_bold_wf(
                 "std_cohort",
                 "std_t1w",
                 "std_mask",
+                # MNI152NLin6Asym warp, for CIFTI use
+                "anat2mni6_xfm",
+                "mni6_mask",
             ],
         ),
         name="inputnode",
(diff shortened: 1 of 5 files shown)


effigies:
This is ready for review. Will move onto carpetplots building on this.

effigies:
Tests passing. @mgxd do you have time to review?

fmriprep/workflows/bold/base.py
```diff
@@ -462,6 +473,87 @@ def init_bold_wf(
...
+
+        ds_bold_cifti = pe.Node(
+            DerivativesDataSink(
+                base_directory=fmriprep_dir,
+                space='fsLR',
+                density=config.workflow.cifti_output,
+                suffix='bold',
+                compress=False,
+                TaskName=all_metadata[0].get('TaskName'),
+                **prepare_timing_parameters(all_metadata[0]),
```
mgxd, line 507: not necessarily related to this PR, but is this the only time we are outputing this metadata? seems like something we should only calculate once and then distribute to any bold outputs

mgxd (approved):
just a question

fmriprep/workflows/bold/base.py
effigies: No, we also calculate it in the volumetric workflow:

https://github.com/nipreps/fmriprep/blob/c2940289829c2aec96c144f3a4938744c33a04d5/fmriprep/workflows/bold/outputs.py#L686-L737

So it gets recalculated several times. We could refactor the function to have a cacheable core (can't `@lru_cache` with a dictionary argument), but I'm not sure saving <1s (estimate, not measured) at workflow build is really worth it.

PR #3135, opened 2023-11-16
CI: Test on Python 3.12

DimitriPapadopoulos:


effigies:
We're stuck on https://github.com/ME-ICA/tedana/issues/976.

DimitriPapadopoulos:
I see. The error is about an old `versioneer.py` incompatible with Python 3.12, bundled with that old version of bokeh bundled with tedana. Wait and see...

effigies:
Done in #3141

PR #3138, opened 2023-11-20
chore: update to latest migas api

mgxd:
Strips away error handling logic, which has been ported to `migas.track_exit`.


effigies (approved):
This looks good to me. I'm a big fan of less code.

PR #3147, opened 2023-11-21
STY: Apply pyupgrade suggestions

DimitriPapadopoulos:

`pyupgrade --py310-plus`

Also manually change a few `.format()` calls to f-strings.


effigies:
I want to hold off on this until the full release, then we'll probably just do what we're doing over in smriprep. We can take care of pyupgrade, flake8, and all that in a single PR.

effigies:
I think this should be covered by ruff now.

DimitriPapadopoulos:
Not entirely. For example, switching to f-strings is not covered by ruff (pyupgrade rules).

effigies:
Ah, do you want to go ahead and re-upgrade? Note that `wrapper/` needs to be excluded, since we're still maintaining Python 2.7 compatibility for that package.

DimitriPapadopoulos:
Let me have a look first:
- exclude wrapper
- perhaps add other ruff rules that may be associated to the pyupgrade suggestions

PR #3153, opened 2023-11-21
MNT: Update smriprep to 0.13.1

smeisler:


effigies:
Thanks for this @smeisler, but we'll update this with `pip-compile` as part of our release process. We do need to update `pyproject.toml` to require `smriprep >=0.13.1`, however.

smeisler:
Got it, I can do that to on my branch too but if you were planning on doing that in a separate PR I can close this. Is there anyway this can be a quick patch so I can test the alpha, or would you rather wait until the remaining issues are addressed for the full release?

effigies:
I'm waiting on #3152, which has already caught a couple bugs like the one in smriprep. Once that's green and merged, I'll merge `master` into `maint/23.2.x` and cut a new release. I may get time tonight, or it might be in the morning.

I will bump the dependency before releasing. You can propose it or not, up to you.

The next `:unstable` tag to hit DockerHub will have smriprep 0.13.1 in it, if you want to test it out, but it's probably worth it to you to wait on #3152, too.

effigies (approved):
Will merge at the same time as #3152 to avoid spamming the CI more than we have to.

PR #3171, opened 2023-12-06
Exclude non-steady-state volumes from confound correlation plot

### LINKED ISSUE #2614: Exclude non-steady-state time points from confound correlations
A large non-steady-state effect will dominate many signals, making GS, WM, CSF correlate highly with the first non-steady-state outlier. Computing the correlations on the steady-state portions of the BOLD series seems more useful.

confoundcorr

tsalo:
Closes #2614 and is related to https://github.com/nipreps/niworkflows/pull/843.

- Feed `skip_vols` into confounds correlation plot. Since niworkflows is pinned to `master` in `pyproject.toml`, the new parameter from https://github.com/nipreps/niworkflows/pull/843 can be used without waiting for a niworkflows release.


tsalo:
👍 That makes sense!

PR #3176, opened 2023-12-07
MAINT: load reportlets interfaces from nireports rather than niworkflows

celprov:
I don't understand what the test complains about. Help, please?


effigies:
Ah, it was isort, not black. Apologies. This sort of thing will be simpler once we move to ruff.

PR #3177, opened 2023-12-07
RF: Load report assembler from nireports

celprov:
This transition is inspired by how MRIQC defines `generate_reports` using _nireports_' Report()

mnt: load Report from nireports instead of niworkflows
mnt: adapt function arguments accordingly
mnt: delete run_reports function
mnt: delete Report patch subclass that is no longer necessary

### DIFF
--- a/fmriprep/cli/run.py
+++ b/fmriprep/cli/run.py
@@ -223,6 +223,12 @@ def main():
         write_derivative_description(config.execution.bids_dir, config.execution.fmriprep_dir)
         write_bidsignore(config.execution.fmriprep_dir)
 
+        if failed_reports:
+            config.loggers.cli.error(
+                "Report generation was not successful for the following participants : %s.",
+                ", ".join(failed_reports),
+            )
+
         if sentry_sdk is not None and failed_reports:
             sentry_sdk.capture_message(
                 "Report generation was not successful for the following participants : %s.",
--- a/fmriprep/cli/run.py
+++ b/fmriprep/cli/run.py
@@ -225,10 +225,11 @@ def main():
 
         if sentry_sdk is not None and failed_reports:
             sentry_sdk.capture_message(
-                "Report generation failed for %d subjects" % failed_reports,
+                "Report generation was not successful for the following participants : %s.",
+                ", ".join(failed_reports),
                 level="error",
             )
-        sys.exit(int((errno + failed_reports) > 0))
+        sys.exit(int((errno + len(failed_reports)) > 0))
 
 
 if __name__ == "__main__":
--- a/fmriprep/cli/run.py
+++ b/fmriprep/cli/run.py
@@ -218,8 +218,7 @@ def main():
             config.execution.participant_label,
             config.execution.fmriprep_dir,
             config.execution.run_uuid,
-            config=data.load("reports-spec.yml"),
-            packagename="fmriprep",
+            config=data.load("reports-spec.yml")
         )
         write_derivative_description(config.execution.bids_dir, config.execution.fmriprep_dir)
         write_bidsignore(config.execution.fmriprep_dir)
--- a/fmriprep/cli/run.py
+++ b/fmriprep/cli/run.py
@@ -218,7 +218,7 @@ def main():
             config.execution.participant_label,
             config.execution.fmriprep_dir,
             config.execution.run_uuid,
-            config=data.load("reports-spec.yml")
+            config=data.load("reports-spec.yml"),
         )
         write_derivative_description(config.execution.bids_dir, config.execution.fmriprep_dir)
         write_bidsignore(config.execution.fmriprep_dir)
--- a/fmriprep/reports/core.py
+++ b/fmriprep/reports/core.py
@@ -22,7 +22,7 @@
 #
 from pathlib import Path
 
-from niworkflows.reports.core import Report as _Report
+from nireports.reports.core import Report as _Report
 
 # This patch is intended to permit fMRIPrep 20.2.0 LTS to use the YODA-style
 # derivatives directory. Ideally, we will remove this in 20.3.x and use an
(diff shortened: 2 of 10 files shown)


celprov:
@oesteban two problems:

1. How do give `layout.get()` not only the subject label but also the file path? or how do I construct entities differently so that it indicates the subject_label but also all the other info it needs to indicate?

3. What should I do about [those lines](https://github.com/nipreps/fmriprep/blob/f3109502a987ad89e2acecc08c04bf6232907a39/fmriprep/reports/core.py#L108C2-L121C17) given that `generate_report()` always return 0 ?

effigies:
If it's not a drop-in replacement, let's postpone this until the next release.

celprov:
I think the failure in Contribution checks is triggered by the fact that the colorama python package is not installed in the CI.

Secondly, what should I do about the ModuleNotFoundError "pkg_resources"  that triggers the failure of the other tests?

effigies:
Yes, the isort problem is waiting on a new release. We can ignore it for now.

It looks like nireports requires `pkg_resources`, which is distributed with `setuptools`. We should not add it as a runtime dependency, we should purge it in favor of `importlib.resources`. I'll open a PR over there.

effigies:
pkg_resources fix: https://github.com/nipreps/nireports/pull/85

fmriprep/reports/core.py
```diff
@@ -34,7 +34,6 @@ def _load_config(self, config):
         from yaml import safe_load as load
 
         settings = load(config.read_text())
```
effigies, line 36: I think we can get rid of this subclass and use `nireports.assembler.report.Report` directly.

celprov:
The test now passes, and the report is correctly generated.
I'm not entirely sure about how errors related to the failure in report generation are handled. Feedback on that part is more than welcome.

fmriprep/reports/core.py
```diff
@@ -22,90 +22,38 @@
...
-        for subject_label in subject_list
-    ]
 
-    errno = sum(report_errors)
+        # Count nbr of subject for which report generation failed
+        errno = 0
+        try:
+            robj.generate_report()
+        except:
+            errno += 1
```
effigies, line 55: The point of this previously was to count errors that were added to the report, not count failures to generate reports. It seems that functionality was removed in the transition to nireports, or perhaps it didn't exist in MRIQC and that was used as the model for nireports.

@oesteban might be able to share the history here...

oesteban: Yes, apologies for the slow response.

Indeed, the new version of the report generation does not return the crashfiles count. It feels a bit clunky that pipeline errors (crashfiles) and report generation errors are mixed up together. For example, if tomorrow we figure (and now with fit+transform this scenario is more likely) a function mode where reports are not generated - we still want to know if crashfiles were generated.

So I would leave this as it stands now, but also add the counting of crashfiles after pipeline execution, so that we do not miss that error managing bit the report generation was doing for us.

Does that make sense?

oesteban:
> If it's not a drop-in replacement, let's postpone this until the next release.

Does this mean "please do not merge into master just yet"?

celprov:
I corrected how the error generated by rob.generate_reports() is logged following the Mathias' suggestion.
I also opened issue #3207 so we keep in mind that counting the number of crash files needs to be implemented.
@oesteban @effigies, please review this PR, let me know if I'm missing anything, or merge if all is good.
Thanks!

fmriprep/reports/core.py
```diff
@@ -22,100 +22,45 @@
...
+            error_list = error_list + f"{subject_label}, "
```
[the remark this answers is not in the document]
oesteban: On top of @effigies' suggestion, I think we also want to dump in the log the stack trace information somewhere to make it easier for the user to report errors.

```suggestion
        try:
            robj.generate_report()
        except:
            import sys
            import traceback

            errors.append(subject_label)
            traceback.print_exception(
                *sys.exc_info(),
                file=str(Path(output_dir) / "logs" / f"report-{run_uuid}-{subject_label}.err"),
            )
```

fmriprep/reports/core.py
```diff
@@ -22,100 +22,45 @@
...
         import logging
 
         logger = logging.getLogger("cli")
-        error_list = ", ".join(
-            f"{subid} ({err})" for subid, err in zip(subject_list, report_errors) if err
-        )
-        logger.error(
-            "Preprocessing did not finish successfully. Errors occurred while processing "
-            "data from participants: %s. Check the HTML reports for details.",
+        logger.debug(
```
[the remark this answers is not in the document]
oesteban: `config` here is not the `fmriprep.config` module. I would just return the list of errored subjects and report it outside.

fmriprep/reports/core.py
```diff
@@ -22,100 +22,45 @@
...
 
-    errno = sum(report_errors)
+        # Count nbr of subject for which report generation failed
+        errno = 0
+        try:
+            robj.generate_report()
+        except:
+            errno += 1
+            error_list = error_list + f"{subject_label}, "
+
```
oesteban, line 57: See commient about the config singleton (it is an unrelated object at this point). Let's just return here.

```suggestion

    return errors
```

IMPORTANT - please note that returning thing will require you to remove the remainder of the code in this function -- I can't do that with the suggestion tool, apparently.

fmriprep/cli/run.py
```diff
@@ -219,7 +219,6 @@ def main():
             config.execution.fmriprep_dir,
             config.execution.run_uuid,
             config=data.load("reports-spec.yml"),
-            packagename="fmriprep",
         )
```
oesteban, line 222:
```suggestion
        )
        if failed_reports:
            config.loggers.cli.error(
                "Report generation was not successful for the following participants : %s.",
                ", ".join(failed_reports),
            )
```

celprov: @oesteban l.226-231 below already log the error if failed reports, but using sentry_sdk, which logger should we keep sentry_sdk or config.loggers?

oesteban: sentry is doing a different thing: it sends telemetry. So we need both.

fmriprep/reports/core.py
```diff
@@ -22,100 +22,49 @@
...
+        except:
+            import sys
+            import traceback
+
+            errors.append(subject_label)
+            traceback.print_exception(
+                *sys.exc_info(),
+                file=str(Path(output_dir) / "logs" / f"report-{run_uuid}-{subject_label}.err"),
+            )
+
```
oesteban, line 64: Apologies, I made the suggestion in the wrong place.
```suggestion
    return errors
```

celprov:
@oesteban @effigies I've applied your suggestions; let me know what you think.

effigies (approved):

effigies:
We need to fix the ds054 tests ASAP, but we agreed that would be outside this PR.

PR #3184, opened 2023-12-11
MNT: Load reportlets interfaces from nireports rather than niworkflows

effigies:
Reverts nipreps/fmriprep#3182, restoring #3176. There's an issue with the CompCor reportlets to resolve before we can merge.

cc @celprov


celprov:
Issue should be solved by nipreps/nireports#84

effigies:
@celprov Needed to pin to a branch instead of a version: af2a495

This should pass now. And you can do the same thing in #3177.

celprov:
LGTM

PR #3186, opened 2023-12-14
ENH: Allow --ignore fmap-jacobian to disable Jacobian determinant modulation during fieldmap correction

effigies:
There is some indication (https://github.com/nipreps/sdcflows/issues/413) that fieldmap correction can lead to weird patches of intensity variation. This is almost certainly caused to modulation by the Jacobian determinant image, although the reason that this is not working as expected in these cases is unclear.

In any case, this allows the behavior to be disabled with `--ignore fmap-jacobian`.

### DIFF
--- a/fmriprep/cli/parser.py
+++ b/fmriprep/cli/parser.py
@@ -285,7 +285,7 @@ def _build_parser(**kwargs):
         action="store",
         nargs="+",
         default=[],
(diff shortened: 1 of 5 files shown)


effigies:
Merged to get this and sdcflows 2.7.0 pushed as `nipreps/fmriprep:unstable`.

I don't think I'm going to try to do the final release before January.

PR #3202, opened 2024-01-15
Re-enable anat fasttrack for dataset without t1w

bpinsard:
## Changes proposed in this pull request

fixes #3201 
Re-enables the possibility to run fMRIPrep on a fMRI only dataset using anat derivatives from another (anatomical) dataset with the same participant. 

Tries to implement integration test for that feature.

related to #2201 

## Documentation that should be reviewed

NA

PR #3208, opened 2024-01-18
ENH: Leverage T2w if available for BOLD -> anat coregistration

mgxd:
This PR adds a new parameter `use_t2w` to `init_bold_reg_wf`, which enables/disables using a T2w reference as an intermediate for bbregister.

As long as a T2w is found in the dataset, this option is enabled, as smriprep has already preprocessed and brought the T2w to T1w space.

This PR only changes the behavior of the `bbreg_wf` fork, the FSL workflow remains using the T1w reference.

cc @madisoth as this was inspired based on some comparison work from him


fmriprep/interfaces/patches.py
```diff
@@ -77,3 +79,26 @@ def _run_interface(self, runtime):
...
+        "aparc+aseg.mgz`` unless otherwise specified)",
+    )
+
+
+class MRICoreg(fs.MRICoreg):
+    """
+    Patched that allows setting both a reference file and the subjects dir.
+    """
+
+    input_spec = MRICoregInputSpec
```
mgxd, line 104: this patch is needed to include the reference file + subject dir/id information in the outputted LTA

fmriprep/workflows/bold/fit.py
```diff
@@ -211,10 +212,15 @@ def init_bold_fit_wf(
...
         bold_series,
-        entity_overrides=config.execution.get().get('bids_filters', {}).get('sbref', {}),
+        entity_overrides=bids_filters.get('sbref', {}),
         layout=layout,
     )
 
+    # T2w is not required, but if available, use to improve BOLD -> anat coreg
+    has_t2w = bool(
+        layout.get(suffix='T2w', extension=['.nii', '.nii.gz'], **bids_filters.get('t2w', {}))
+    )
```
effigies, line 222: I would rather that we determine `has_t2w` (or whatever) at the `workflows.base` and pass it up. That will allow us to check whether `t2w` appears in the anatomical directory or `t2w_preproc` appears in precomputed derivatives. Hitting the layout and re-applying filters in here feels fragile.

fmriprep/workflows/bold/registration.py
```diff
@@ -362,14 +385,25 @@ def init_bbreg_wf(
             (mri_coreg, transforms, [('out_lta_file', 'in2')]),
```
effigies, line 385: Here you're still using `mri_coreg`. One problem here is going to be that `t2w_preproc` may be off from `T1.mgz` by `fsnative2t1w_xfm`.

I think it might be cleaner to change the `--bold2t1w-init <register|header>` option to `--bold2anat-init <auto|t1w|t2w|header>`. If we get `auto`, then we look for a T2w image in the inputs and assume `T2.mgz` will be present in the FreeSurfer directory (we can always improve this check later). Then we add a `FreeSurferSource` here to find `T2.mgz` and pass it to `mri_coreg` if `bold2anat_init == 't2w'`.

fmriprep/cli/parser.py
```diff
@@ -313,19 +334,32 @@ def _slice_time_ref(value, parser):
...
     g_conf.add_argument(
         "--bold2t1w-dof",
+        action=DeprecatedAction,
+        choices=[6, 9, 12],
+        type=int,
+        help="Deprecated - use `--bold2anat-dof` instead.",
+    )
+    g_conf.add_argument(
+        "--bold2anat-init",
+        choices=["auto", "t1w", "header"],
```
effigies, line 350: Do we want to allow forcing t2w as well? It's simple enough to implement.

```suggestion
        choices=["auto", "t1w", "t2w", "header"],
```

fmriprep/cli/parser.py
```diff
@@ -313,19 +334,32 @@ def _slice_time_ref(value, parser):
...
+        default="auto",
+        help="Method of BOLD to anatomical coregistration. If `auto`, a T2w image is used if "
+        "available, otherwise the T1w image. `t1w` forces use of the T1w, and `header` uses "
+        "the T1w header information.",
```
effigies, line 354:
```suggestion
        help="Method of initial BOLD to anatomical coregistration. If `auto`, a T2w image is used if "
        "available, otherwise the T1w image. `t1w` forces use of the T1w, and `header` uses "
        "the BOLD header information without an initial registration.",
```

fmriprep/cli/parser.py
```diff
@@ -40,6 +40,27 @@ def _build_parser(**kwargs):
 
     from .version import check_latest, is_flagged
 
+    deprecations = {
+        # parser attribute name: (replacement flag, version slated to be removed in)
+        'use_aroma': (None, '24.0.0'),
+        'aroma_melodic_dim': (None, '24.0.0'),
+        'aroma_err_on_warn': (None, '24.0.0'),
```
effigies, line 47: 24.0 is next. We can just delete these ones...

fmriprep/config.py
```diff
@@ -525,11 +525,12 @@ class workflow(_Config):
...
-    """Degrees of freedom of the BOLD-to-T1w registration steps."""
-    bold2t1w_init = "register"
-    """Whether to use standard coregistration ('register') or to initialize coregistration from the
-    BOLD image-header ('header')."""
+    bold2anat_dof = None
+    """Degrees of freedom of the BOLD-to-anatomical registration steps."""
+    bold2anat_init = "auto"
+    """Method of BOLD to anatomical coregistration. The target anatomical (``'t1w'``, ``'t2w'``)
+    can be specified, otherwise ``'auto'`` will prefer a T2w image but fall back to T1w if none
+    are available. Alternatively, ``'header'`` will use the T1w header information."""
```
effigies, line 533: Needs updating to match parser help text.

fmriprep/workflows/base.py
```diff
@@ -623,6 +623,10 @@ def init_single_subject_wf(subject_id: str):
         num_bold=len(bold_runs)
     )
 
+    # Before initializing BOLD workflow, select/verify anatomical target for coregistration
+    if config.workflow.bold2anat_init in ('auto', 't2w'):
+        config.workflow.bold2anat_init = 't2w' if subject_data['t2w'] else 't1w'
```
effigies, line 628: #3202 allowed us to proceed with derivatives even when the raw anatomicals are missing. Might as well keep it up.

```suggestion
        config.workflow.bold2anat_init = 't2w' if subject_data['t2w'] or 't2w_preproc' in anatomical_cache else 't1w'
```

fmriprep/workflows/base.py
```diff
@@ -623,6 +623,10 @@ def init_single_subject_wf(subject_id: str):
         num_bold=len(bold_runs)
     )
 
+    # Before initializing BOLD workflow, select/verify anatomical target for coregistration
+    if config.workflow.bold2anat_init in ('auto', 't2w'):
```
effigies, line 627: IMO if someone forces it, they should not get silent fallback.

```suggestion
    if config.workflow.bold2anat_init == 'auto':
```

fmriprep/workflows/bold/registration.py
```diff
@@ -266,23 +268,30 @@ def init_bbreg_wf(
...
         reason=''
-        if bold2t1w_dof == 6
+        if bold2anat_dof == 6
         else 'to account for distortions remaining in the BOLD reference',
     )
 
+    use_t2w = bold2anat_init == 't2w'
+    if use_t2w:
+        workflow.__desc__ += (
+            " A T2w reference was used as an intermediate volume to improve tissue contrast."
```
effigies, line 292:
```suggestion
            " The aligned T2w image was used for initial co-registration."
```

fmriprep/workflows/bold/registration.py
```diff
@@ -302,36 +311,39 @@ def init_bbreg_wf(
         name='outputnode',
     )
 
-    if bold2t1w_init not in ("register", "header"):
-        raise ValueError(f"Unknown BOLD-T1w initialization option: {bold2t1w_init}")
+    if bold2anat_init not in ty.get_args(RegistrationInit):
+        raise ValueError(f"Unknown BOLD-anat initialization option: {bold2anat_init}")
```
effigies, line 315:
```suggestion
        raise ValueError(f"Unknown BOLD-to-anatomical initialization option: {bold2anat_init}")
```

fmriprep/workflows/bold/registration.py
```diff
@@ -379,13 +396,16 @@ def init_bbreg_wf(
         (bbregister, transforms, [('out_lta_file', 'in1')]),
     ])  # fmt:skip
 
+    if use_t2w:
+        workflow.connect(fssource, 'T2', bbregister, 'intermediate_file')
+
```
effigies, line 401: Does not do anything with `--init-reg` set, and nothing useful with `--init-header` set.

```suggestion
```

fmriprep/workflows/bold/registration.py
```diff
@@ -344,6 +356,8 @@ def init_bbreg_wf(
     concat_xfm = pe.Node(ConcatenateXFMs(inverse=True), name='concat_xfm')
 
     workflow.connect([
+        (inputnode, fssource, [('subjects_dir', 'subjects_dir'),
+                               ('subject_id', 'subject_id')]),
```
effigies, line 360: I would combine this with the connection in L380, not run it unconditionally.

fmriprep/workflows/base.py
```diff
@@ -669,6 +673,7 @@ def init_single_subject_wf(subject_id: str):
                     f'outputnode.sphere_reg_{"msm" if msm_sulc else "fsLR"}',
                     'inputnode.sphere_reg_fsLR',
                 ),
+                ('outputnode.t2w_preproc', 'inputnode.t2w_preproc'),  # Optional
```
effigies, line 676: I think we no longer actually connect this up in deeper workflows. It could make sense to pass for FSL registration, but I don't really want to put more effort into that.

effigies (commented):
Overall this looks good!

fmriprep/cli/parser.py
mgxd: The way I see it, `auto` signifies "use T2w but don't fail if there isn't one", whereas a dedicated `t2w` option should fail if there isn't one available - do you think this would be useful?

fmriprep/cli/parser.py
mgxd: I figured we could remove these (along with the arguments themselves) in a single, separate PR

fmriprep/cli/parser.py
effigies: The use-case I can imagine is "I want to use t2w, but I would rather get an error than silent fallback if it's unavailable." I'm not sure how common that need would be, but I don't see a good reason to disable it, and adding it later would require another release.

fmriprep/workflows/tests/test_base.py
```diff
@@ -105,7 +105,7 @@ def bids_root(tmp_path_factory):
 
 
 def _make_params(
-    bold2t1w_init: str = "register",
+    bold2anat_init: str = "t1w",
```
effigies, line 108: The default is "auto", so that's what we should do here:

```suggestion
    bold2anat_init: str = "auto",
```

fmriprep/workflows/tests/test_base.py
```diff
@@ -163,12 +163,12 @@ def _make_params(
     ),
     [
         _make_params(),
-        _make_params(bold2t1w_init="header"),
+        _make_params(bold2anat_init="header"),
```
effigies, line 166:
```suggestion
        _make_params(bold2anat_init="t1w"),
        _make_params(bold2anat_init="t2w"),
        _make_params(bold2anat_init="header"),
```

effigies (approved):
LGTM. Small comments at this point.

PR #3210, opened 2024-01-18
TST: Update test to reflect new report generation behavior

### LINKED ISSUE #3209: The test for crash file count on ds054 requires fixing.
In nireports, generate_report() no longer counts the number of crash files but the number of subjects for which the report generation failed. (cf #3177)
As such, the test in ds054 needs to be adapted to the new behavior.

Related to #3207

mgxd:
Closes #3209 

The exit code for `--reports-only` should be an indicator of report generation status, not whether or not errors were present in the workflow (those should be reported separately, at the time they are encountered during processing).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -648,27 +648,6 @@ jobs:
                 --mem-mb 14336 --nthreads 4 -vv --debug compcor
       - run: *check_outputs
       - run:
-          name: Generate report with one artificial error
-          command: |
-            set -x
-            sudo mv /tmp/${DATASET}/fmriprep/sub-100185.html \
-                    /tmp/${DATASET}/fmriprep/sub-100185_noerror.html
-            UUID=$(grep uuid /tmp/${DATASET}/work/*/config.toml | cut -d\" -f 2 | tail -n 1)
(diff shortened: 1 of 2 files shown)


effigies:
I think we do still want to exit non-zero if there were errors detected. It's just that the error counting was removed from the report generation. I was about to work on this, but would you rather?

mgxd:
I'm going back and forth on this - now that we're examining it, it feels like there are separate cases. 

1) During the regular workflow processing, absolutely, the return code should be nonzero as something went wrong, regardless of report generation result.

2) If running `--reports-only`, (1) had to have been run and the errors made apparent. So asking fmriprep to regenerate reports, which if it does, feels like it should return a "successful" code. 

Perhaps we should expand the return code options and add a new one that specifies report generation was successful, but errors were present?

effigies:
Ugh. I forgot that we split up the reports-only and the main implementation. Yes, if we're doing reports-only and it's not just a dry run with reports at the end, then sure, that can exit with 0 as long as the reports actually generate.

IMO the CI test is significantly less useful than it was when written, now that we duplicate report generation in two branches. We could just drop it or at least change it to grep for the error text in the HTML, rather than just checking the exit code.

effigies (approved):

PR #3236, opened 2024-02-23
Ignore unselected subjects in BIDSLayoutIndexer

### LINKED ISSUE #3235: Ignore all subjects except the requested ones in BIDSLayout initialization

I believe the BIDSLayout initialization might take less time on large datasets if the `ignore` patterns included a regular expression to select any participant folders except the ones requested by the user.

```python
ignore_patterns = [
    'code',
    'stimuli',
    'sourcedata',
    'models',
    re.compile(r'^\.'),
    re.compile(
        r'sub-[a-zA-Z0-9]+(/ses-[a-zA-Z0-9]+)?/(beh|dwi|eeg|ieeg|meg|perf)'
    ),
]
if cls.participant_label:
    # Ignore any subjects who aren't the requested ones.
    ignore_patterns.append(re.compile(r'sub-(?!' + '|'.join(cls.participant_label) + r')\w+'))

_indexer = BIDSLayoutIndexer(
    validate=False,
    ignore=ignore_patterns
)
```

Yes

This stems from a similar request in QSIPrep (https://github.com/PennLINC/qsiprep/issues/513).

tsalo:
Closes #3235.

- Incorporate a regular expression in the `ignore` list of the `BIDSLayoutIndexer` that flags subjects who are _not_ in `Config.participant_label` (when provided).


fmriprep/config.py
```diff
@@ -471,18 +471,23 @@ def init(cls):
...
+                'sourcedata',
+                'models',
+                re.compile(r'^\.'),
+                re.compile(r'sub-[a-zA-Z0-9]+(/ses-[a-zA-Z0-9]+)?/(beh|dwi|eeg|ieeg|meg|perf)'),
+            ]
+            if cls.participant_label:
+                # Ignore any subjects who aren't the requested ones.
+                ignore_patterns.append(
+                    re.compile(r'sub-(?!' + '|'.join(cls.participant_label) + r')\w+')
+                )
```
effigies, line 486: I think you probably don't want to do this if `cls.bids_database_dir is not None`. This would cause a database file to be written that is missing large portions of the dataset and then reused the next time.

tsalo: I thought the BIDSLayout's `reset_database` setting meant that the database would be left untouched if `bids_database_dir` was set?

tsalo: Otherwise, wouldn't it be a problem if users wanted to run both fMRIPrep (which ignores `dwi` folders) and dMRIPrep/QSIPrep (which needs them) on the same dataset?

effigies: The issue is that if `--bids-database-dir` is set to a path where there does not already exist a file, it will be created and then frozen with the first round.

fmriprep/config.py
```diff
@@ -472,18 +472,24 @@ def init(cls):
...
+                'sourcedata',
+                'models',
+                re.compile(r'^\.'),
+                re.compile(r'sub-[a-zA-Z0-9]+(/ses-[a-zA-Z0-9]+)?/(beh|dwi|eeg|ieeg|meg|perf)'),
+            ]
+            if cls.participant_label and cls.bids_database_dir is None:
+                # Ignore any subjects who aren't the requested ones.
+                # This is only done if the database is written out to a run-specific folder.
+                ignore_patterns.append(
+                    re.compile(r'sub-(?!' + '|'.join(cls.participant_label) + r')\w+')
```
effigies, line 487: Unfortunately, this doesn't ignore `sub-013` if `sub-01` is selected. That's probably okay, as it's an unlikely case and all that happens is indexing more than expected, not less. I'm not sure how to fix up this regex, though. Negative lookaheads are tricky.

effigies (approved):
I'm okay with this as-is. LMK if you want to try to improve the regex.

fmriprep/config.py
tsalo: I don't think I want to fix it in this PR, because I honestly have no clue how to do so, but I want to update the comments to make it clear that this is an issue.

effigies: Oh, I think I got it:

```suggestion
                    re.compile(r'sub-(?!(' + '|'.join(cls.participant_label) + r')\b)')
```

tsalo: Thanks!

effigies:
Okay, it looks like we are using `re.search` not `re.match` in pybids, which means we need to consider `_`, which is considered a word character.

PR #3242, opened 2024-02-29
MNT: Add flake8-logging-format (G) rules to ruff

DimitriPapadopoulos:

```
ruff check --select G
```

### DIFF
--- a/fmriprep/_warnings.py
+++ b/fmriprep/_warnings.py
@@ -36,9 +36,9 @@ def _warn(message, category=None, stacklevel=1, source=None, **kwargs):
         category = category.replace('type', 'WARNING')
 
     if kwargs:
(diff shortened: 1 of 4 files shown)


effigies:
I'm not sure why we should care about this.

DimitriPapadopoulos:
The linter recommends that because that's what many programmers suggest: defer string interpolation to the log function so that string interpolation doesn't occur unless the log function does print at that log level. So that would be a performance issue. The rationale is that the speed gain when deferring string interpolation with terse logging outwheighs the speed gain of f-strings with verbose logging. Other programmers don't agree.

DimitriPapadopoulos:
That's not really about- f-strings.

Just to make sure, this is not about using
```python
logger.info("info: %s" % s)
```
vs.
```python
logger.info(f"info: {s}")
```

It's about
```python
logger.info("info: %s", s)
```

PR #3245, opened 2024-03-02
Add metadata for motion parameters

### LINKED ISSUE #3231: Add motion parameter units to confounds sidecar

I think it would be nice if the confounds JSON file included the units of motion-related columns.

This could be done for all motion-related confounds, but I think it's most important for the rotation columns.

Yes

This stems from https://github.com/PennLINC/xcp_d/issues/1053. Basically, a user wants me to include metadata for the motion parameters in XCP-D. I'm happy to do that, but I thought it would make more sense to inherit the metadata from the preprocessing pipeline (i.e., fMRIPrep).

tsalo:
Closes #3231.

- Add metadata, including Units, for motion parameters (6 basic parameters and FD).
    - I included FD because it might be good to log the head radius that is used to calculate it at some point (e.g., if https://github.com/nipreps/nibabies/issues/198 moves forward)

none

### DIFF
--- a/fmriprep/workflows/bold/confounds.py
+++ b/fmriprep/workflows/bold/confounds.py
@@ -27,6 +27,7 @@ Calculate BOLD confounds
 .. autofunction:: init_bold_confs_wf
 
 """
+
 from nipype.algorithms import confounds as nac
 from nipype.interfaces import utility as niu
 from nipype.pipeline import engine as pe
--- a/fmriprep/workflows/bold/confounds.py
+++ b/fmriprep/workflows/bold/confounds.py
@@ -412,10 +412,21 @@ the edge of the brain, as proposed by [@patriat_improved_2017].
         name='crowncc_metadata_fmt',
     )
 
+    # Combine all confounds metadata
     mrg_conf_metadata = pe.Node(
-        niu.Merge(3), name='merge_confound_metadata', run_without_submitting=True
+        niu.Merge(4), name='merge_confound_metadata', run_without_submitting=True
     )
+    # Tissue mean time series
     mrg_conf_metadata.inputs.in3 = {label: {'Method': 'Mean'} for label in signals_class_labels}
+    # Movement parameters
+    mrg_conf_metadata.inputs.in4 = {
+        "trans_x": {"Units": "mm"},
+        "trans_y": {"Units": "mm"},
+        "trans_z": {"Units": "mm"},
+        "rot_x": {"Units": "rad"},
+        "rot_y": {"Units": "rad"},
+        "rot_z": {"Units": "rad"},
+    }
     mrg_conf_metadata2 = pe.Node(
         DictMerge(), name='merge_confound_metadata2', run_without_submitting=True
     )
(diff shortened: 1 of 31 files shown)


fmriprep/workflows/bold/confounds.py
```diff
@@ -412,10 +412,22 @@ def init_bold_confs_wf(
...
+    # Tissue mean time series
     mrg_conf_metadata.inputs.in3 = {label: {'Method': 'Mean'} for label in signals_class_labels}
+    # Movement parameters
+    mrg_conf_metadata.inputs.in4 = {
+        'trans_x': {'Description': 'Translation Along X Axis', 'Units': 'mm'},
+        'trans_y': {'Description': 'Translation Along Y Axis', 'Units': 'mm'},
+        'trans_z': {'Description': 'Translation Along Z Axis', 'Units': 'mm'},
+        'rot_x': {'Description': 'Rotation Along X Axis', 'Units': 'rad'},
+        'rot_y': {'Description': 'Rotation Along Y Axis', 'Units': 'rad'},
+        'rot_z': {'Description': 'Rotation Along Z Axis', 'Units': 'rad'},
```
effigies, line 428: Should be rotation "about". And could we indicate the x axis is left-right, and so on?

fmriprep/workflows/bold/confounds.py
```diff
@@ -412,10 +412,22 @@ def init_bold_confs_wf(
...
+        'framewise_displacement': {'Units': 'mm / TR'},
```
effigies, line 429: It strikes me as odd to call it a velocity instead of a distance. Unless there's precedent in the literature you're following?

```suggestion
        'framewise_displacement': {'Units': 'mm'},
```

tsalo: Isn't it actually a measure of instantaneous velocity though? One could scale it based on seconds instead of TRs (i.e., as something like "secondwise_displacement"), and the units would be different, even if the distance unit (mm) was the same.

fmriprep/workflows/bold/confounds.py
tsalo: Absolutely can do. Just to make sure, the motion parameters don't depend on the phase encoding direction? Meaning they don't directly map onto i, j, and k at this point?

effigies: Hmmm, would be good to check.

fmriprep/workflows/bold/confounds.py
effigies: I think to be precise, it's a change in displacement with respect to the index: $\Delta x_i / \Delta i$. Indices are unitless so the quantity is in the original units. Granted, for most BOLD sequences, $t_i = t_0 + T_R*i$, but it is not true in all cases, and it plays no role in the calculation.

Apart from the pedantic point that I'm not 100% confident in, I don't think I've ever seen it written as `mm / TR`, so this would be something we're introducing; and BIDS units are supposed to be SI, which makes this awkward if anybody ever tries to validate it.

tsalo: Ah, I see. I'll change it in that case.

Regarding units though, TR _is_ a special value for SamplingFrequency in BEP012, so it seems like we should have it as a unit in BIDS at some point. Or maybe support SamplingFrequency as a unit.

Issue #3261, opened 2024-03-25
Track immediate `Sources` when possible


tsalo:
### What would you like to see added in fMRIPrep?

1. Standard-space and T1w-space outputs do not include `Sources`.
    - The `Sources` should probably be the boldref-space outputs, when those are requested. Otherwise, they should include the original data, the HMC transforms, the coreg transforms, the ref2anat transform, and the anat2std transform.
2. Echo-wise outputs do not include `Sources`.
3. Surface files don't seem to include `Sources`.

### Do you have any interest in helping implement the feature?

Yes

tsalo:
@effigies do you think it would be worth it to try building the TemplateFlow templates into the DatasetLinks and Sources?

effigies:
Yes, I think including templateflow would be very helpful both for sources and spatial references.

tsalo:
Want to note that `from-orig_to-boldref_mode-image_desc-hmc_xfm.txt` isn't in the `Sources`.

`from-boldref_to-auto00000_mode-image_xfm.txt` isn't either, but I can't tell if that's accounted for by including `from-boldref_to-T1w_mode-image_xfm.txt` in the Sources. @effigies do you know? If `orig-to-boldref`, `boldref-to-fmap` and `boldref-to-T1w` are _all_ strung together, then I think `boldref-to-T1w` might be mislabeled.

effigies:
Each of these is an atomic transform. orig-to-boldref + boldref-T1w are strung together to get the individual volume there, but boldref-to-fmap is used to transform the fieldmap into the boldref space and calculate a voxel-shift-map. We do not save the VSM, so it would be appropriate to include both the fieldmap and the transform as inputs. The exact provenance would be better modeled using `prov.jsonld`.

tsalo:
Ah, thanks! I think SDCFlows will need to pass its fmap file from its derivatives workflow in order to use it as a source in fMRIPrep.

effigies:
I'd forgotten about this and opened #3534. I made it a sub-issue so it's linked prominently.

PR #3263, opened 2024-03-28
Track proximal sources of functional GIFTIs

tsalo:
Related to #3261.

- Add an outputnode to `init_ds_volumes_wf` with the saved `bold` path.
- Add `sources` field to `init_bold_surf_wf`'s inputnode.
- Infer Sources for functional GIFTI outputs.
    - If the user has requested T1w-space outputs, then this should be `bold` from the associated `ds_volumes_wf` and `fsnative2t1w_xfm`
    - If the user didn't want T1w-space volumetric outputs, then this should be the raw BOLD file, `motion_xfm`, `boldref2anat_xfm`, `fsnative2t1w_xfm`.


tsalo:
The GIFTI metadata from ds005 look good:

```
{
  "RepetitionTime": 2.0,
  "SliceTimingCorrected": false,
  "Sources": [
    "bids:raw:sub-01/func/sub-01_task-mixedgamblestask_run-01_bold.nii.gz",
    "bids::sub-01/func/sub-01_task-mixedgamblestask_run-01_from-orig_to-boldref_mode-image_desc-hmc_xfm.txt",
    "bids::sub-01/func/sub-01_task-mixedgamblestask_run-01_from-boldref_to-T1w_mode-image_desc-coreg_xfm.txt",
    "bids:deriv-0:sub-01/anat/sub-01_from-fsnative_to-T1w_mode-image_xfm.txt"
  ],
  "TaskName": "mixed-gambles task"
}
```

fmriprep/workflows/bold/base.py
```diff
@@ -514,6 +514,37 @@ def init_bold_wf(
             (bold_anat_wf, bold_surf_wf, [('outputnode.bold_file', 'inputnode.bold_t1w')]),
         ])  # fmt:skip
 
+        if nonstd_spaces.intersection(('anat', 'T1w')):
+            # Source file should be output T1w-space volumetric file and fsnative2t1w_xfm
```
effigies, line 518: I disagree. I see the use of this output as an implementation detail, and one I hope to move away from. We should be able to resample directly from BOLD to surface eventually.

fmriprep/workflows/bold/base.py
```diff
@@ -514,6 +514,37 @@ def init_bold_wf(
...
+                niu.Merge(2),
+                name='merge_surface_sources',
+                run_without_submitting=True,
+            )
+            workflow.connect([
+                (ds_bold_t1_wf, merge_surface_sources, [('outputnode.bold', 'in1')]),
+                (inputnode, merge_surface_sources, [('fsnative2t1w_xfm', 'in2')]),
+            ])  # fmt:skip
+        else:
+            # sources are bold_file, motion_xfm, boldref2anat_xfm, fsnative2t1w_xfm
```
effigies, line 529: I would include the white-matter and pial surfaces for each hemisphere, since those determine the coordinates at which the BOLD is sampled. If that's difficult to do at the moment, I'm fine with postponing it to another PR.

fmriprep/workflows/bold/outputs.py
```diff
@@ -746,6 +746,14 @@ def init_ds_volumes_wf(
...
         name='inputnode',
     )
+    outputnode = pe.Node(
+        niu.IdentityInterface(
+            fields=[
+                'bold',
+            ],
+        ),
+        name='outputnode',
+    )
```
effigies, line 756: I'm okay keeping this, even if we don't end up using it. `IdentityInterface` nodes get optimized away when building the execution graph. But this is a lot of lines:

```suggestion
    outputnode = pe.Node(niu.IdentityInterface(fields=['bold']), name='outputnode')
```

fmriprep/workflows/bold/base.py
tsalo: Fair enough. I'll drop that part.

fmriprep/workflows/bold/outputs.py
tsalo: I'm just going to remove it, since it won't be used in this PR and I want to minimize the diff.

fmriprep/workflows/bold/base.py
tsalo: The hemisphere loop happens within `init_bold_surf_wf`, so I need to add those sources there, but I can't pin down where exactly those two files are grabbed in the workflow. I think it'll have to wait for another PR.

PR #3264, opened 2024-03-28
ENH: Support named derivative paths

### LINKED ISSUE #3260: Allow `--derivatives` to be named

In #3252, we started creating DatasetLinks for derivative datasets passed through the `--derivatives` parameter. However, because there's no easy way to infer what each derivative dataset actually is, we're labeling them as `deriv-<index>` in the DatasetLinks. I think the DatasetLinks and BIDS URIs throughout the derivatives would be more interpretable if users could label their input derivative datsets.

Yes

I was thinking something like:

```
fmriprep ... --derivatives smriprep=/path/to/smriprep
```

tsalo:
Closes #3260.

- Copy over `ToDict` action from fmriprep-docker over to the main CLI parser.
- Modify the Config to handle named derivative paths.
    - E.g., `--derivatives anat=/path/to/smriprep` --> `DatasetLinks = {'anat': '/path/to/smriprep'}`
- Infer derivative name from the folder name of the derivative path if it isn't labeled.
    - E.g., `--derivatives /path/to/smriprep` --> `DatasetLinks = {'smriprep': '/path/to/smriprep'}`

### DIFF
--- a/fmriprep/cli/parser.py
+++ b/fmriprep/cli/parser.py
@@ -68,6 +68,7 @@ def _build_parser(**kwargs):
             for i_kv, kv in enumerate(values):
                 if '=' not in kv:
                     k = f'deriv-{i_kv}'
+                    v = kv
                 else:
                     k, v = kv.split('=')
                 d[k] = Path(v)
--- a/fmriprep/cli/parser.py
+++ b/fmriprep/cli/parser.py
@@ -236,7 +236,7 @@ def _build_parser(**kwargs):
         action=ToDict,
         metavar='PACKAGE=PATH',
         type=str,
-        nargs='*',
+        nargs='+',
         help=(
             'Search PATH(s) for pre-computed derivatives. '
             'These may be provided as named folders '
--- a/fmriprep/cli/parser.py
+++ b/fmriprep/cli/parser.py
@@ -71,7 +71,7 @@ def _build_parser(**kwargs):
                     k = f'deriv-{i_kv}'
                 else:
                     k, v = kv.split('=')
-                d[k] = os.path.abspath(v)
+                d[k] = Path(v)
             setattr(namespace, self.dest, d)
 
     def _path_exists(path, parser):
@@ -236,7 +236,7 @@ def _build_parser(**kwargs):
         '--derivatives',
         action=ToDict,
         metavar='PACKAGE=PATH',
-        type=Path,
+        type=str,
         nargs='*',
         help=(
             'Search PATH(s) for pre-computed derivatives. '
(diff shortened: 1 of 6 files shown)


fmriprep/cli/parser.py
```diff
@@ -222,11 +233,15 @@ def _slice_time_ref(value, parser):
     g_bids.add_argument(
         '-d',
         '--derivatives',
-        action='store',
-        metavar='PATH',
-        type=Path,
+        action=ToDict,
+        metavar='PACKAGE=PATH',
+        type=str,
         nargs='*',
```
tsalo, line 239: @effigies is there a reason `--derivatives` is set with `nargs='*'` instead of `nargs='+'`?

fmriprep/cli/parser.py
```diff
@@ -62,6 +62,18 @@ def __call__(self, parser, namespace, values, option_string=None):
...
+    class ToDict(Action):
+        def __call__(self, parser, namespace, values, option_string=None):
+            d = {}
+            for i_kv, kv in enumerate(values):
+                if '=' not in kv:
+                    k = f'deriv-{i_kv}'
+                    v = kv
+                else:
+                    k, v = kv.split('=')
+                d[k] = Path(v)
```
effigies, line 74: Not a huge fan of `deriv-[0..n]`. What if we take the basename by default? And then raise an error if there are multiple with the same name (whether passed explicitly or not)?

```suggestion
            d = {}
            for spec in values:
                try:
                    name, loc = spec.split('=')
                    loc = Path(loc)
                except ValueError:
                    loc = Path(spec)
                    name = loc.name

                if name in d:
                    raise ValueError(f"Received duplicate derivative name: {name}")
                d[k] = Path(v)
```

tsalo: I like it. I can't think of any reasonable cases where there would be duplicate derivative folder names, so I don't think there's a problem with automatically inferring the name.

wrapper/src/fmriprep_docker/__main__.py
```diff
@@ -521,7 +521,7 @@ def main():
         unknown_args.append('--derivatives')
         for deriv, deriv_path in opts.derivatives.items():
             command.extend(['-v', '{}:/deriv/{}:ro'.format(deriv_path, deriv)])
-            unknown_args.append('/deriv/{}'.format(deriv))
+            unknown_args.append('{}=/deriv/{}'.format(deriv, deriv))
```
effigies, line 524: This is now redundant:

```suggestion
            unknown_args.append('/deriv/{}'.format(deriv))
```

effigies (approved):
LGTM. Small suggestion.

wrapper/src/fmriprep_docker/__main__.py
tsalo: Good point!

PR #3287, opened 2024-05-14
[FIX] Add "double" type to allowed DisplacementFieldTransform

kjamison:
The ANTs utility `CompositeTransformUtil --assemble` outputs its displacement field as 'DisplacementFieldTransform_double_3_3', but utils/transform/load_ants_h5 throws an error unless it is 'DisplacementFieldTransform_float_3_3'. Presumably this is just a check to make sure it is a displacement field. The code after this check works fine for "double".

Just changing `!= b'DisplacementFieldTransform_float_3_3'` to `not in (b'DisplacementFieldTransform_float_3_3', b'DisplacementFieldTransform_float_3_3')`

None


effigies:
Thanks!

PR #3292, opened 2024-05-18
FIX: Write out boldref-space brain mask with minimal level

### LINKED ISSUE #3290: boldref-space brain mask appears to be missing from minimal outputs
### What happened?

Per the documentation, the boldref-space functional brain mask should be created with `--level minimal`. However, this file appears to be missing from the minimal outputs.

```
sub-<subject_label>/
  func/
    sub-<subject_label>_[specifiers]_space-<space_label>_desc-brain_mask.nii.gz
```

### What command did you use?

I haven't checked this myself- I'm just following up on a conversation from the NiPreps hackathon.

23.2.2

Docker

### Is your data BIDS valid?

Yes

No

```shell
n/a
```

n/a

tsalo:
Closes #3290.

Changes proposed in this pull request
- Create init_ds_boldmask_wf workflow. This might be overkill.
- Write out the boldref-space brain mask in init_bold_fit_wf.
- Update expected outputs.

### DIFF
--- a/fmriprep/workflows/bold/fit.py
+++ b/fmriprep/workflows/bold/fit.py
@@ -47,6 +47,7 @@ from ...utils.misc import estimate_bold_mem_usage
 # BOLD workflows
 from .hmc import init_bold_hmc_wf
 from .outputs import (
+    init_ds_boldmask_wf,
     init_ds_boldref_wf,
     init_ds_hmc_wf,
     init_ds_registration_wf,
@@ -467,6 +468,12 @@ def init_bold_fit_wf(
             desc='coreg',
             name='ds_coreg_boldref_wf',
         )
+        ds_boldmask_wf = init_ds_boldmask_wf(
+            bids_root=layout.root,
+            output_dir=config.execution.fmriprep_dir,
+            desc='brain',
+            name='ds_boldmask_wf',
+        )
 
         # fmt:off
         workflow.connect([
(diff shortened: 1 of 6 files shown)


effigies:
Traveling so can't look closely, but does this avoid double-sinking in cases where boldref outputs are requested?

tsalo:
Ah good point. It doesn't yet, so I'll switch this to a draft until I've got everything working and there's no double-sinking.

fmriprep/workflows/bold/outputs.py
```diff
@@ -466,6 +466,51 @@ def init_ds_boldref_wf(
...
+    )
+    outputnode = pe.Node(niu.IdentityInterface(fields=['boldmask']), name='outputnode')
+
+    raw_sources = pe.Node(niu.Function(function=_bids_relative), name='raw_sources')
+    raw_sources.inputs.bids_root = bids_root
+
+    ds_boldmask = pe.Node(
+        DerivativesDataSink(
+            base_directory=output_dir,
+            datatype='func',
```
tsalo, line 491: I'm not sure if this is a good idea, but for some reason the files were written out to `anat`.

fmriprep/workflows/bold/fit.py
```diff
@@ -467,13 +468,21 @@ def init_bold_fit_wf(
...
+            name='ds_boldmask_wf',
+        )
 
         # fmt:off
         workflow.connect([
             (hmcref_buffer, fmapref_buffer, [("boldref", "boldref_files")]),
             (fmapref_buffer, enhance_boldref_wf, [("out", "inputnode.in_file")]),
             (fmapref_buffer, ds_coreg_boldref_wf, [("out", "inputnode.source_files")]),
             (ds_coreg_boldref_wf, regref_buffer, [("outputnode.boldref", "boldref")]),
+            (fmapref_buffer, ds_boldmask_wf, [("out", "inputnode.source_files")]),
```
effigies, line 484: Other workflows just set 

```
        ds_hmc_wf.inputs.inputnode.source_files = [bold_file]
```

I don't think I would worry about getting sbref vs bold straight here, as it's just a mask.

fmriprep/workflows/bold/outputs.py
effigies: I believe this will be fixed by passing in an original bold file.

effigies (commented):
This seems fine. It does raise the question of what to do if a pre-computed mask or coregistration transform is found, but that can be handled in a separate issue. This one is sufficient for a bug-fix.

effigies (approved):

PR #3296, opened 2024-06-03
FIX: Ensure shape matches template during resampling

mgxd:
The reverse of the TransformFixedParameters should match the template dimensions.

This fell through the cracks because the templates tested (MNI152 variants) all have the same X and Z dimensions. Only when testing on a template with distinct values for shape did this become evident.

Additionally, this removes any `FIXED_PARAMS` checks, as that is template specific and more useful for debugging the initial implementation.

### DIFF
--- a/fmriprep/utils/transforms.py
+++ b/fmriprep/utils/transforms.py
@@ -71,9 +71,10 @@ def load_ants_h5(filename: Path) -> nt.base.TransformBase:
 
     fixed_params = transform2['TransformFixedParameters'][:]
 
-    shape = tuple(fixed_params[:3].astype(int)[::-1])
-    warp = h['TransformGroup']['2']['TransformParameters'][:]
-    warp = warp.reshape((*shape, 3)).transpose(2, 1, 0, 3)
+    shape = tuple(fixed_params[:3].astype(int))
+    # ITK stores warps in Fortran-order, where the vector components change fastest
+    # Nitransforms expects 3 volumes, not a volume of three-vectors, so transpose
+    warp = np.reshape(transform2['TransformParameters'], (3, *shape), order='F').transpose(1, 2, 3, 0)
     warp *= np.array([-1, -1, 1])
 
     warp_affine = np.eye(4)
--- a/fmriprep/utils/transforms.py
+++ b/fmriprep/utils/transforms.py
@@ -71,6 +71,16 @@ def load_ants_h5(filename: Path) -> nt.base.TransformBase:
 
     fixed_params = transform2['TransformFixedParameters'][:]
 
+    spacing = fixed_params[6:9]
+    direction = fixed_params[9:]
+
+    # Supported spacing
+    if not np.array_equal(spacing, np.array([1.0, 1.0, 1.0])):
+        raise ValueError(f'Unexpected spacing: {spacing}')
+
+    if not np.array_equal(direction, np.diag([-1, -1, -1])):
+        raise ValueError(f'Asymmetric direction matrix: {direction}')
+
     shape = tuple(fixed_params[:3].astype(int))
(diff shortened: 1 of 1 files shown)


fmriprep/utils/transforms.py
```diff
@@ -81,15 +70,8 @@ def load_ants_h5(filename: Path) -> nt.base.TransformBase:
...
-        msg += f'Expected: {FIXED_PARAMS}\n'
-        msg += f'Found: {fixed_params}'
-        if not np.array_equal(fixed_params[6:], FIXED_PARAMS[6:]):
-            raise ValueError(msg)
-        warnings.warn(msg, stacklevel=1)
 
-    shape = tuple(fixed_params[:3].astype(int))
+    shape = tuple(fixed_params[:3].astype(int)[::-1])
     warp = h['TransformGroup']['2']['TransformParameters'][:]
     warp = warp.reshape((*shape, 3)).transpose(2, 1, 0, 3)
```
effigies, line 76: This is a really unintuitive change that deserves a comment, so I've thought it through. I've decided we actually want to modify the reshape, not the shape.

What's going on here is that we have a Fortran-ordered array, so the first index changes fastest, and the order of indices is `(component, i, j, k)`. Then we want three NIfTI volumes, which are also Fortran ordered, but that means we want to transpose to `(i, j, k, component)`. We've been treating the vector component as if it's in the right place, and everybody else is out of step. Simpler to just move the vector component to the end:

```suggestion
    shape = tuple(fixed_params[:3].astype(int))
    # ITK stores warps in Fortran-order, where the vector components change fastest
    # Nitransforms expects 3 volumes, not a volume of three-vectors, so transpose
    warp = np.reshape(transform2['TransformParameters'], (3, *shape), order='F').transpose(1, 2, 3, 0)
```

mgxd: This seems much more readable - let's try it

fmriprep/utils/transforms.py
```diff
@@ -38,16 +37,6 @@ def load_transforms(xfm_paths: list[Path], inverse: list[bool]) -> nt.base.Trans
     return chain
 
 
-FIXED_PARAMS = np.array([
-    193.0, 229.0, 193.0,  # Size
-    96.0, 132.0, -78.0,   # Origin
-    1.0, 1.0, 1.0,        # Spacing
```
effigies, line 44: We now use size (`[:3]`), origin (`[3:6]`) and directions (`[9:15]`). We don't deal with the case where Spacing (`[6:9]`) is not 1x1x1, and we are desperately hoping that the directions will never be anything but orthonormal LPS.

I think we probably do want to at least keep the check that `fixed_params[6:]` isn't something we don't know exactly what to do with. (What's currently a `ValueError`.)

fmriprep/utils/transforms.py
```diff
@@ -71,6 +71,16 @@ def load_ants_h5(filename: Path) -> nt.base.TransformBase:
 
     fixed_params = transform2['TransformFixedParameters'][:]
 
+    spacing = fixed_params[6:9]
+    direction = fixed_params[9:]
```
effigies, line 75:
```suggestion
    direction = fixed_params[9:].reshape((3, 3))
```

fmriprep/utils/transforms.py
```diff
@@ -71,6 +71,16 @@ def load_ants_h5(filename: Path) -> nt.base.TransformBase:
...
+
+    # Supported spacing
+    if not np.array_equal(spacing, np.array([1.0, 1.0, 1.0])):
+        raise ValueError(f'Unexpected spacing: {spacing}')
+
+    if not np.array_equal(direction, np.diag([-1, -1, -1])):
```
effigies, line 81:
```suggestion
    if not np.array_equal(direction, direction.T):
```

PR #3297, opened 2024-06-04
Use BIDSURI in init_ds_boldmask_wf

### LINKED ISSUE #3252: Track sources with DatasetLinks and BIDS-URIs

I was thinking it would be nice to output `Sources` metadata fields for all outputs, using `DatasetLinks` and BIDS URIs to track the files.

`DatasetLinks` and BIDS URIs were added in 1.8.0, so we would need to increase the `BIDSVersion` to at least 1.8.0.

Yes

I've had some success doing this in XCP-D, and I think it would be even easier in fMRIPrep since we could put a `DatasetLinks` field in the Config object.

tsalo:
Closes #3252.

- Replace `RawSources` in `init_ds_boldmask_wf` with `Sources` determined using `BIDSURI`.

### DIFF
--- a/fmriprep/workflows/bold/fit.py
+++ b/fmriprep/workflows/bold/fit.py
@@ -474,7 +474,6 @@ def init_bold_fit_wf(
             name='ds_coreg_boldref_wf',
         )
         ds_boldmask_wf = init_ds_boldmask_wf(
-            bids_root=layout.root,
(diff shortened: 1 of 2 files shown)


effigies (approved):
Looks sensible. I'm looking into the CI failures...

PR #3300, opened 2024-06-05
RF: Fix ITK warp conversion to nitransforms format

effigies:
Alternative to (and builds on) #3296.

Closes #3296.

### DIFF
--- a/fmriprep/utils/transforms.py
+++ b/fmriprep/utils/transforms.py
@@ -7,6 +7,7 @@ import nibabel as nb
 import nitransforms as nt
 import numpy as np
 from nitransforms.io.itk import ITKCompositeH5
+from tranforms3d.affines import compose as compose_affine
 
 
 def load_transforms(xfm_paths: list[Path], inverse: list[bool]) -> nt.base.TransformBase:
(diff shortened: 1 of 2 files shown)


fmriprep/utils/transforms.py
```diff
@@ -1,13 +1,13 @@
...
 
-import warnings
 from pathlib import Path
 
 import h5py
 import nibabel as nb
 import nitransforms as nt
 import numpy as np
 from nitransforms.io.itk import ITKCompositeH5
+from tranforms3d.affines import compose as compose_affine
```
effigies, line 10:
```suggestion
from transforms3d.affines import compose as compose_affine
```

mgxd (approved):
LGTM - I have not tested it, but can appreciate the improved clarity.
The only concern I have is this code is copied in a number of places (here, nibabies, resampler package) - we should focus some energy on having a single distribution and importing from there (maybe an ohbm hackathon project?)

effigies:
Actually just looked at the MNI-resampled boldref in https://app.circleci.com/pipelines/github/nipreps/fmriprep/2204/workflows/b0c88b0f-7a65-4904-a5d4-7dc26b8233f4/jobs/11504/artifacts, and it's degraded relative to your PR and master. Let me try to understand this...

effigies:
Hmm. I loaded the file with my function and yours, and they seem identical. I'm going to re-run the test, in case it's just a one-off weirdness.

effigies:
Okay, it appears to have been a one-off bad boldmask, not a problem with this function at all.

effigies:
> The only concern I have is this code is copied in a number of places (here, nibabies, resampler package) - we should focus some energy on having a single distribution and importing from there (maybe an ohbm hackathon project?)

And yes. We should get this fixed up in nitransforms so we can stop having our own special loader.

PR #3326, opened 2024-07-16
chore(dev): Move to tox to simplify test/CI setup

effigies:
Had this in a branch for a bit. I think it's working...


.github/workflows/stable.yml
```diff
@@ -31,40 +35,38 @@ jobs:
       matrix:
         os: ['ubuntu-latest']
         python-version: ['3.10', '3.11', '3.12']
-        install: ['pip']
-        check: ['tests']
-        pip-flags: ['']
+        dependencies: ['', 'pre']
+        include:
+          - os: ubuntu-latest
+            python-version: 3.10
```
effigies, line 41:
```suggestion
            python-version: '3.10'
```

effigies:
All coverage losses are a result of using branch tracing so that `if`s are only considered fully covered if both the `True` and `False` paths are hit.

image

effigies:
Any objections? Otherwise will merge today.

PR #3338, opened 2024-07-30
RF: Replace `config.execution.fmriprep_dir` uses in BIDSURI calls with `output_dir` parameter

### LINKED ISSUE #3337: Replace `config.execution.fmriprep_dir` uses in BIDSURI calls with `output_dir` parameter

All of the workflows in fMRIPrep that currently use `BIDSURI` to track sources currently rely on `config.execution.fmriprep_dir` for the output directory, and yet all of those functions have an `output_dir` parameter.

It would be easier for other preps that use fMRIPrep as a dependency (i.e., ASLPrep) if we used `output_dir` instead of the config object's package-specific `fmriprep_dir` field.

Yes

I think I completely overlooked the `output_dir` parameter when I started implementing `BIDSURI`-based source tracking (#3255, #3262, #3263, #3297) because I originally wanted to reference the config object directly _within_ the `BIDSURI` interface, which ended up not working.

tsalo:
Closes #3337.

- Use `output_dir` function argument for `BIDSURI` objects instead of the `config.execution.fmriprep_dir` field. This will let other preps (e.g., ASLPrep) use these workflow functions without having to mock up `fmriprep_dir` fields in their config objects.

### DIFF
--- a/fmriprep/workflows/bold/outputs.py
+++ b/fmriprep/workflows/bold/outputs.py
@@ -441,7 +441,7 @@ def init_ds_boldref_wf(
         BIDSURI(
             numinputs=1,
             dataset_links=config.execution.dataset_links,
-            out_dir=str(config.execution.fmriprep_dir.absolute()),
+            out_dir=str(output_dir),
         ),
         name='sources',
     )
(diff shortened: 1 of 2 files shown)


effigies (approved):

effigies:
Are there any of these in smriprep?

tsalo:
Not yet. I think I was planning to add BIDSURIs to smriprep and sdcflows at some point, but I haven't gotten around to it yet.

PR #3374, opened 2024-10-08
update fsl-topup for reported bug fixes

### LINKED ISSUE #3373: Update fsl-topup from 2203.2 -> 2203.5?

It sounds like the version of `topup` that is [included in the fmriprep docker](https://github.com/nipreps/fmriprep/blob/3e69454c7ac3d9bc4aea767882a8f897282dd187/env.yml#L37) has a "buglet". From the [topup changelog for 2203.4
](https://fsl.fmrib.ox.ac.uk/fsl/docs/#/development/history/changelog-6.0.7.12?id=fsl-topup-22032-gt-22034)

> even in the absence of movement, the most basal slice would get replicated at the very top. This had been reported as a bug, but I had not seen it because of the masking that is done. But it did mean that there would always be a missing slice at the top, which could have been a problem if the FOV was very tight in the slice direction. And it prevented me from outputting an unwarped image that wasn't truncated at the top and/or bottom. I have now fixed the bug, which means that my_fname_fieldmap_mag will have full FOV. I also added writing of a mean unwarped image as part of the --iout output, also with full FOV. This will be useful for creating a mask for eddy and/or dtifit. It means that for data with very tight FOV the estimated field will be slightly different, especially in the topmost slice. But the field generated by the new version will be, slightly, more accurate.

Another issue was resolved [in 2203.5](https://fsl.fmrib.ox.ac.uk/fsl/docs/#/development/history/changelog-6.0.7.13?id=fsl-topup-22034-gt-22035)

> There is a sanity check on readout times in the --datain/--acqp files. I recently coordinated that with eddy's sanity check, to ensure they have the same view of what is sane. But that meant that registering to a distortion free image no longer worked. And that is a valid thing to do, for example in the context of synb0-DISCO. I have now fixed it so that it allows one volume to have a readout time of zero, or almost zero.

Yes

_No response_

psadil:

Closes #3373

### DIFF
--- a/env.yml
+++ b/env.yml
@@ -34,7 +34,7 @@ dependencies:
   - fsl-fugue=2201.4
   - fsl-mcflirt=2111.0
   - fsl-miscmaths=2203.2
-  - fsl-topup=2203.2
+  - fsl-topup=2203.5
   - pip
   - pip:
     - -r requirements.txt


psadil:
FWIW, those failures look like https://github.com/scipy/scipy/issues/21623#issuecomment-2394945406

effigies:
Can you look up the latest FSL release and see if the other pins we have are up to date? I'd rather move them as a group, as I expect they are tested together.

psadil:
```
$ mapfile -t pinned < <(grep fsl- env.yml)
version_pattern="[0-9]+\.[0-9]*"
for fsl_version in 6.0.7.8 6.0.7.9 6.0.7.10 6.0.7.11 6.0.7.12 6.0.7.13; do
    echo "checking FSL ${fsl_version}"
    manifest=$(curl https://git.fmrib.ox.ac.uk/fsl/conda/manifest/-/raw/"${fsl_version}"/fsl-release.yml\?ref_type\=tags 2>/dev/null)    
    for pin in "${pinned[@]}"; do
        package_name=$(echo "${pin}" | grep -oE "fsl-[a-z0-9]+" )
        if [[ $manifest =~ $package_name ]]; then
            manifest_version=$(echo "${manifest}" | grep -oE "${package_name} ${version_pattern}" | grep -oE "${version_pattern}")
            manifest_package_w_version=$(printf "%s=%s" "${package_name}" "${manifest_version}")
            if [[ ! "${pin}" =~ ${manifest_package_w_version} ]]; then
                printf "env has %s but FSL %s has %s\n" "${pin}" "${fsl_version}" "${manifest_package_w_version}"
            fi
        fi
    done
done

checking FSL 6.0.7.8
env has   - fsl-bet2=2111.4 but FSL 6.0.7.8 has fsl-bet2=2111.7
checking FSL 6.0.7.9
env has   - fsl-bet2=2111.4 but FSL 6.0.7.9 has fsl-bet2=2111.7
checking FSL 6.0.7.10
env has   - fsl-bet2=2111.4 but FSL 6.0.7.10 has fsl-bet2=2111.7
checking FSL 6.0.7.11
env has   - fsl-bet2=2111.4 but FSL 6.0.7.11 has fsl-bet2=2111.7
checking FSL 6.0.7.12
env has   - fsl-bet2=2111.4 but FSL 6.0.7.12 has fsl-bet2=2111.8
env has   - fsl-topup=2203.2 but FSL 6.0.7.12 has fsl-topup=2203.4
checking FSL 6.0.7.13
env has   - fsl-bet2=2111.4 but FSL 6.0.7.13 has fsl-bet2=2111.8
env has   - fsl-topup=2203.2 but FSL 6.0.7.13 has fsl-topup=2203.5
```

effigies:
So if I understand correctly, this just means that these lines

https://github.com/nipreps/sdcflows/blob/656a51d4a734d2d8d66550e9e317386f677801a3/sdcflows/workflows/fit/pepolar.py#L144-L145

become unnecessary?

Given that we aren't patching the TOPUP interface directly, this seems fairly safe. @mgxd Were you going to test on a dataset that has an odd number of slices?

mgxd:
No, I don't believe so, the fix for the odd # of slices has to do with a configuration option (`--subsamp`). We decided against changing the configuration based on even/odd slices (see https://github.com/nipreps/sdcflows/pull/217)

From my understanding this fix will only be relevant/noticed on images with tight FOVs

PR #3376, opened 2024-10-09
fix: Remove checks for unit zooms and symmetric rotations in template warp

effigies:
A user with a custom template is generating warps that have spacing `[1.5 1.5 1.5]`: https://neurostars.org/t/use-custom-dartel-template-in-fmriprep/30670/8

The check for unit spacing was mostly intended to avoid weird behavior where we might generate a bad affine. The interpretation of isotropic spacing is straightforward, so I think we can relax this. Creating a docker image to allow for testing.

### DIFF
--- a/fmriprep/utils/transforms.py
+++ b/fmriprep/utils/transforms.py
@@ -49,7 +49,8 @@ def load_ants_h5(filename: Path) -> nt.base.TransformBase:
     #   * Always return a nitransforms TransformBase
     #   * Construct warp affine from fixed parameters
     #
-    # This should be upstreamed into nitransforms
+    # This has been upstreamed into nitransforms.
+    # Future versions should switch to using nitransforms directly.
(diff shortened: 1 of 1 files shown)


effigies:
I resampled the MNI T1w image to one with a shape of `(193, 192, 229)` and an affine of:

```
[[   0.     1.1    0.  -105.6]
 [   0.     0.    -1.2  118.8]
 [   1.     0.     0.   -78. ]
 [   0.     0.     0.     1. ]]
```

After registering the original image to the resampled image with `antsRegistrationSyNQuick.sh` (modified to produce a composite `.h5` output), I was able to load it and verify that the reference it generated had the same shape and affine using the method we have here. I'm comfortable removing these checks.

That method is also reproduced in nitransforms 24+ (verified again), so we can move to that in 24.2.

PR #3379, opened 2024-10-10
CI: Build docker image on GHA

effigies:
Trying to use GHCR as a cache.


effigies:
`buildx` seems to be trying to push to a Docker registry as well, causing a crash.

https://github.com/docker/buildx/issues/2731

PR #3381, opened 2024-10-12
CI: Fetch tags and 200 commits to support describe

effigies:
This tests the behavior of PRs with the docker build. It should reuse the `ghcr.io/nipreps/fmriprep:master` cache and create `ghcr.io/nipreps/fmriprep:pull-3381` or similar.

It also checks out enough history to properly create a version string.


effigies:
Ah right, I turned off building for PRs. We can disable that switch in PRs we want to publish versions for, or figure out some other condition. We can't use the `docker/*` approach without modifying the tags ourselves.

PR #3387, opened 2024-10-24
FIX: Ensure fieldmap is resampled correctly in report

effigies:
While attempting to reproduce a bug report, I ran across a crash. Fieldmaps are always in RAS, while the BOLD references and masks are in the space of the original BOLD. In the case of ds000221, that's LSP. The reportlet assumes all input images are in the same space, and so can fail if the field-of-view is not a cube (and would be bad otherwise).

I was unable to reproduce the reported bug, but it's possible it was confused for this. For searchability's sake, the error was:

```
2024-10-21 12:21:37,963 [CRITICAL] fMRIPrep failed: Traceback (most recent call last):
  File "/home/chris/mambaforge/envs/fmriprep/lib/python3.11/site-packages/nipype/pipeline/plugins/multiproc.py", line 67, in run_node
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/chris/mambaforge/envs/fmriprep/lib/python3.11/site-packages/nipype/pipeline/engine/nodes.py", line 527, in run
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/chris/mambaforge/envs/fmriprep/lib/python3.11/site-packages/nipype/pipeline/engine/nodes.py", line 645, in _run_interface
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/chris/mambaforge/envs/fmriprep/lib/python3.11/site-packages/nipype/pipeline/engine/nodes.py", line 771, in _run_command
    raise NodeExecutionError(msg)
nipype.pipeline.engine.nodes.NodeExecutionError: Exception raised while executing Node sdecreg_report.

Traceback:
          File "/home/chris/mambaforge/envs/fmriprep/lib/python3.11/site-packages/nipype/interfaces/base/core.py", line 398, in run
            runtime = self._post_run_hook(runtime)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
          File "/home/chris/mambaforge/envs/fmriprep/lib/python3.11/site-packages/nipype/interfaces/mixins/reporting.py", line 50, in _post_run_hook
            self._generate_report()
          File "/home/chris/Projects/nipreps/sdcflows/sdcflows/interfaces/reportlets.py", line 103, in _generate_report
            abs(np.percentile(fmapdata[maskdata], 99.8)),
                              ~~~~~~~~^^^^^^^^^^
        IndexError: boolean index did not match indexed array along dimension 1; dimension is 88 but corresponding boolean dimension is 64
```

I'm calling this a bug-fix because the workflow change occurred in a non-critical path (reports) and there is no risk of pollution from an existing working directory.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,10 +12,11 @@ Bug fix release in the 24.1.x series.
 Precomputed functional derivatives were not being correcly detected,
 and a couple fixes for rare issues.
 
-  * FIX: Remove checks for unit zooms and symmetric rotations in template warp (#3376)
-  * FIX: Stop excluding FS minc_modify_header used during fallback registration (#3372)
-  * FIX: Repair search for precomputed bold references (#3370)
-  * FIX: Repair search for precomputed transforms (#3369)
+* FIX: Remove checks for unit zooms and symmetric rotations in template warp (#3376)
+* FIX: Stop excluding FS minc_modify_header used during fallback registration (#3372)
+* FIX: Repair search for precomputed bold references (#3370)
+* FIX: Repair search for precomputed transforms (#3369)
+
 
 24.1.0 (September 16, 2024)
 ===========================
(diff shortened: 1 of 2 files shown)


mgxd (approved):
lgtm - opened https://github.com/nipreps/sdcflows/issues/467 to either catch or address this in the reportlet

PR #3393, opened 2024-11-03
ENH: Flexibilize "sophisticated" pepolar to allow monomodal execution

oesteban:

### DIFF
--- a/fmriprep/workflows/base.py
+++ b/fmriprep/workflows/base.py
@@ -586,12 +586,8 @@ Setting-up fieldmap "{estimator.bids_id}" ({estimator.method}) with \
                 #    (typically, more than two EPI PE directions), or
                 # 2. Two modalities are involved, with at most two images to pass
                 #    into FSL TOPUP.
-                if (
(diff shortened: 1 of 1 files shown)


effigies:
What use case is this intended to support?

oesteban:
I have a dataset with all four AP/PA/LR/RL EPIs. This should be easy to handle by topup.

effigies:
What about len(suffixes) >= 2, then? Or we could drop the length check altogether and rely on sdcflows to not create estimators with too few inputs.

oesteban:
> What about len(suffixes) >= 2, then? Or we could drop the length check altogether and rely on sdcflows to not create estimators with too few inputs.

The idea is that the four images are of the same modality. That would allow for epi + bold (which should be capture by the other condition, and disallow more than 2 images)

effigies:
Got it. Let's add a comment, as I'm sure I'll forget.

oesteban:
Sure, I'll add it later today

effigies (approved):

Issue #3406, opened 2024-12-05
Wrong registration between fMRI data and fieldmap


1RuobingLiu:
### What happened?

Hi every one, recently I am working on a fMRIPrep for

this example dataset for subeject 33 session Y01: original -containning two imaginary and two real fieldmaps apart from two magnetitude fieldmaps under fmap folder; and processed – I merged imaginary and real fieldmaps into two phase fieldmaps first using Python Nib.arctan function.
I also tried run on my local pc, same results.

The fmriprep-output shows that there is something wrong with registration.
![Screenshot 2024-12-05 at 1 52 09 PM](https://github.com/user-attachments/assets/ed8dbd3c-894c-4e51-affb-eff77bbd0510)

There is something wrong with bold brain mask.
This is the bold mask if we apply fieldmap correction:
![image](https://github.com/user-attachments/assets/a6921f44-ef82-49fe-8d1f-8e89bf87a3de)
This is the bold mask if we do not apply fieldmap:
![image](https://github.com/user-attachments/assets/58a9b8f3-a29e-4b0e-934a-86b924fae219)

Can you help me on this, please? How can I fix the registration problem? I have been stuck on this for a long while. Thank you! Appreciate it if you could help!

### What command did you use?

```shell
#hcc platform
apptainer exec docker://nipreps/fmriprep fmriprep $inPath $outPath participant --participant-label sub-NCANDA00033 --output-spaces MNI152NLin2009cAsym MNI152NLin6Asym:res-2 anat -w $cachePath --fs-license-file $licFile --skip_bids_validation --nthreads 5 --low-mem --cifti-output --ignore slicetiming
#local pc
python -m fmriprep /Users/ruoliu/Downloads/NCANDA-sub95/raw /Users/ruoliu/Downloads/NCANDA-sub95/derivatives/fmriprep participant \
  --participant-label sub-NCANDA00095 \
  --output-spaces MNI152NLin2009cAsym MNI152NLin6Asym:res-2 anat \
  -w /Users/ruoliu/Downloads/NCANDA-sub95/cache/fmriprep \
  --fs-license-file /Users/ruoliu/Downloads/NCANDA-sub95/raw/license.txt \
  --skip_bids_validation \
  --nthreads 5 \
  --low-mem \
  --cifti-output
```

### What version of fMRIPrep are you running?

24.2.0

### How are you running fMRIPrep?

Docker

### Is your data BIDS valid?

Yes

### Are you reusing any previously computed results?

[...]

PR #3415, opened 2024-12-18
fix: Use consistent skull-stripping pre- and post- SDC

effigies:
Overall seems safe, the `skullstrip_wf` is extracted from the `enhance_and_skullstrip_wf`, skipping reruns.

Running locally now. We'll see if does address #3406.

### DIFF
--- a/fmriprep/workflows/bold/fit.py
+++ b/fmriprep/workflows/bold/fit.py
@@ -27,7 +27,7 @@ import bids
 import nibabel as nb
 from nipype.interfaces import utility as niu
 from nipype.pipeline import engine as pe
-from niworkflows.func.util import init_enhance_and_skullstrip_bold_wf
+from niworkflows.func.util import init_enhance_and_skullstrip_bold_wf, init_skullstrip_bold_wf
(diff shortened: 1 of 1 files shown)


mgxd (commented):
Curious if you think [sdcflows' brainextraction workflow](https://github.com/nipreps/sdcflows/blob/03076bddae3ffff257b4c4176a067de4de671817/sdcflows/workflows/ancillary.py#L32-L112) could be an alternative

effigies:
That's what this replaces.

mgxd:
Ah, I missed that in the connections - perhaps we should change this directly in sdcflows then, to keep it consistent across nibabies/nirodents/etc

effigies:
I was doing this out here because I didn't want to change that workflow and risk impacting use cases where that wishes well. Really, I think I should just go ahead and use the resampler for unwarping and get rid of the unwarp workflow altogether.

1RuobingLiu:
Hi everyone! Could you merge this version to the master please? Thank you!

1RuobingLiu:
Hi everyone, could you create a docker sif for this version please? Thank you

PR #3421, opened 2025-01-14
Upgrade to ruff 0.9.1

DimitriPapadopoulos:

Upgrade ruff to the latest 0.9.1. Enable `ISC001` as it now documented as compatible with the formatter.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -178,7 +178,6 @@ extend-select = [
 ]
 ignore = [
   "S311",  # We are not using random for cryptographic purposes
-  "ISC001",
   "S603",
 ]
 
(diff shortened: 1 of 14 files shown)


pyproject.toml
```diff
@@ -123,10 +123,6 @@ version-file = "fmriprep/_version.py"
 # Developer tool configurations
 #
 
-# Disable black
-[tool.black]
-exclude = ".*"
-
```
effigies, line 129: Please leave this. It's specifically to prevent IDEs that auto-apply black from modifying files at save.

wrapper/pyproject.toml
```diff
@@ -63,10 +63,6 @@ universal = true
 # Developer tool configurations
 #
 
-# Disable black
-[tool.black]
-exclude = ".*"
-
```
effigies, line 69: And this.

PR #3423, opened 2025-01-29
feat: Add flag to fallback to Estimated* metadata or a passed value for TotalReadoutTime

### LINKED ISSUE #3009: Ensure per-BOLD SyN-SDC does not require TotalReadoutTime

fMRIPrep 23.1.0 will not lump BOLD images together for running SyN-SDC. This should eliminate the need for a dataset-defined TotalReadoutTime, and we can impute something in order to ensure that a plausible fieldmap. 1/16 (0.0625) or 1/32 (0.03125) seconds are reasonable, and being powers of two should mostly avoid rounding errors.

Yes

_No response_

effigies:
This flag provides two additional ways to retrieve TotalReadoutTime. If `EstimatedTotalReadoutTime` or `EstimatedEffectiveEchoSpacing` are defined, the user can pass `--fallback-total-readout-time estimated` to direct fMRIPrep to use these.

This could be turned on by default, but I don't think this is prudent. The reason that `dcm2niix` does not simply say `TotalReadoutTime` or `EffectiveEchoSpacing` is not to misrepresent the confidence that these are correct. Previously, users had to actively remove the `Estimated` to accept the estimates, and I don't think we want to eliminate the conscious decision. However, we can make it so that users do not need to modify their datasets to accept them.

Closes #3009.

### DIFF
--- a/fmriprep/workflows/base.py
+++ b/fmriprep/workflows/base.py
@@ -615,15 +615,26 @@ It is released under the [CC0]\
         from sdcflows.workflows.base import init_fmap_preproc_wf
 
         fallback_trt = config.workflow.fallback_total_readout_time
-        fmap_wf = init_fmap_preproc_wf(
-            use_metadata_estimates=fallback_trt == 'estimated',
-            fallback_total_readout_time=fallback_trt if isinstance(fallback_trt, float) else None,
-            debug='fieldmaps' in config.execution.debug,
-            estimators=fmap_estimators,
-            omp_nthreads=omp_nthreads,
-            output_dir=fmriprep_dir,
-            subject=subject_id,
-        )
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -34,7 +34,7 @@ dependencies = [
     "psutil >= 5.4",
     "pybids >= 0.16",
     "requests >= 2.27",
-    "sdcflows >= 2.11.0",
+    "sdcflows @ git+https://github.com/nipreps/sdcflows.git@refs/pull/479/head",
```
effigies, line 37:
```suggestion
    "sdcflows @ git+https://github.com/nipreps/sdcflows.git@master",
```

PR #3424, opened 2025-01-29
rf: Reconstruct motion confounds from minimal derivatives

effigies:
After an exploration in https://github.com/nipreps/fmripost-aroma/issues/94, we now know how to reconstruct the original motion parameters (to at least 0.001 mm/rad precision) from the ITK transforms we generate.

To ensure consistency between single-shot and fit+apply runs, I propose to use this function in place of the original outputs.

I will add a regression test before merging, to ensure we notice if a change in nitransforms breaks this calculation.


effigies:
Timeseries are being generated, the outputs are close to those from master, and tests confirm the validity of the changes.

PR #3440, opened 2025-03-05
chore(deps): Update pinned environment

effigies:
Starting by updating things that are not pinned in env.yml. After testing with this, will try upgrading env.yml.


effigies:
Okay, this is looking good.

PR #3446, opened 2025-03-25
REL: 25.0.0

effigies:
Alright, time for a release. I think we're in pretty good shape.

@keithcallenberg Would you like to be included in the author listing? Do you have an ORCID and affiliation you would like us to use?


keithcallenberg:
Sure, thanks for asking!
My ORCID is 0000-0001-7786-9300
The relevant affiliation is Department of Psychiatry, University of
Pittsburgh

PR #3464, opened 2025-05-19
feat: Replace c3d_affine_tool with a ConvertAffine interface

effigies:
Upgrading convert3d leads to segfaults, which means that we are playing a game of cat and mouse with ANTs, libitk and convert3d. It's not worth it just to convert FSL affines to ITK, when we can do this with nitransforms.

The existing niworkflows `ConcatenateXFMs` interface does not quite do the job, since it assumes that `.mat` is an ITK binary affine, so I create a `ConvertAffine` interface.

This replaces the `fsl2itk_fwd`, `fsl2itk_inv` and `invt_bbr` nodes, and moves the `lta_to_fsl` node to the section it is connected in.

This PR could go further and stop using `LTAConvert` as well, but I intend this to be a narrow replacement. The main logical difference is that I only convert the `mri_coreg` output to FSL format if FLIRT is going to run; otherwise, I convert its output directly to ITK to avoid converting twice.

cc @mgxd We might want to do this in nibabies as well.


effigies:
@mgxd This is working, and we can look at the visual reports to confirm we didn't do anything crazy:

* https://output.circle-artifacts.com/output/job/9c2f3901-e0c5-44a7-b1ac-7de14cabcc0d/artifacts/0/tmp/ds054/fmriprep/sub-100185.html#datatype-figures_desc-summary_run-1_subject-100185_suffix-bold_task-machinegame
* https://output.circle-artifacts.com/output/job/1f15b589-0be8-41c9-8446-dba391f91827/artifacts/0/tmp/ds210/fmriprep/sub-02.html#datatype-figures_desc-summary_echo-None_run-1_subject-02_suffix-bold_task-cuedSGT

Could also write some unit tests, but I'd appreciate your input first. Worried I'll dive into a rabbit hole and spend way too much time on the tests and maybe not even test what needs testing.

fmriprep/interfaces/nitransforms.py
```diff
@@ -0,0 +1,99 @@
...
+                inv_affine.reference = moving
+
+            out_inv = fname_presuffix(
+                self.inputs.in_xfm,
+                suffix=f'_inv.{ext}',
+                newpath=runtime.cwd,
+                use_ext=False,
+            )
+            self._results['out_inv'] = out_inv
+            affine.to_filename(out_inv, moving=reference, fmt=self.inputs.out_fmt)
```
mgxd, line 97:
```suggestion
            inv_affine.to_filename(out_inv, moving=reference, fmt=self.inputs.out_fmt)
```

mgxd (commented):
Looks good, definitely agree some unit tests would help!

effigies:
@mgxd This is ready to go. The bugs I was hitting were in my test code...

effigies:
Note the test `.nii.gz` files are just compressed headers, no data.

mgxd (approved):
LGTM

PR #3467, opened 2025-05-22
rf: Separate fieldmap registration from coreg ref generation

effigies:
Porting https://github.com/PennLINC/aslprep/pull/531 over.

This should have no impact on the workflow structure for full runs, but allows runs with precomputed derivatives to display SDC reports. I therefore think this should be a patch candidate.

@mgxd Could you have a look? I'm happy to port this over to nibabies as well.

### DIFF
--- a/fmriprep/workflows/bold/fit.py
+++ b/fmriprep/workflows/bold/fit.py
@@ -349,8 +349,7 @@ def init_bold_fit_wf(
         )
 
     func_fit_reports_wf = init_func_fit_reports_wf(
-        # TODO: Enable sdc report even if we find coregref
-        sdc_correction=not (coreg_boldref or fieldmap_id is None),
+        sdc_correction=fieldmap_id is not None,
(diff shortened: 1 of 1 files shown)


fmriprep/workflows/bold/fit.py
```diff
@@ -462,10 +462,82 @@ def init_bold_fit_wf(
     else:
         config.loggers.workflow.info('Found motion correction transforms - skipping Stage 2')
 
-    # Stage 3: Create coregistration reference
-    # Fieldmap correction only happens during fit if this stage is needed
+    # Stage 3: Register fieldmap to boldref and reconstruct in ASL space
```
mgxd, line 465:
```suggestion
    # Stage 3: Register fieldmap to boldref and reconstruct in BOLD space
```

mgxd (approved):
Looks reasonable, and agreed that it can be included in a patch release

Issue #3468, opened 2025-05-29
<<>> in B0FieldIdentifier


michaelf500:
### What happened?

With <<>> (as generated by bidscoin) in the B0FieldIdentifier json field, an error is thrown: 

### What command did you use?

```shell
fmriprep-docker ./ppmi/ out/ participant --participant-label 100017 --boilerplate-only --fs-license-file /home/michael/P/MJF/fmriprep/license.txt
RUNNING: docker run --rm -e DOCKER_VERSION_8395080871=24.0.2 -it -v /home/michael/P/MJF/fmriprep/license.txt:/opt/freesurfer/license.txt:ro -v /home/michael/fmriprep/ppmi:/data:ro -v /home/michael/fmriprep/out:/out nipreps/fmriprep:25.1.1 /data /out participant --participant-label 100017 --boilerplate-only
bids-validator@1.14.10
```

### What version of fMRIPrep are you running?

25.1.1

### How are you running fMRIPrep?

Docker

### Is your data BIDS valid?

Yes

### Are you reusing any previously computed results?

No

### Please copy and paste any relevant log output.

```shell
ValueError: [Workflow|Node] name "out_fmap_rest<<ses20201222>>" is not valid.
```

### Additional information / screenshots

the B0 tag is
    "B0FieldIdentifier": "fmap_rest<<ses20201222>>"

marcelzwiers:
For reference, see https://github.com/Donders-Institute/bidscoin/issues/270

michaelf500:
Hi, thanks. 

I tried that with command: 

`docker run --rm -e DOCKER_VERSION_8395080871=27.5.1 -it -v /home/michael/P/MJF/fmriprep/license.txt:/opt/freesurfer/license.txt:ro -v /home/michael/fmriprep/test:/data:ro -v /home/michael/fmriprep/out:/out nipreps/fmriprep:unstable /data /out participant --participant-label 100017 --boilerplate-only`

and I get an error message: 

`         B0 field inhomogeneity map will be estimated with the following 1 estimator(s): [<EstimatorType.PEPOLAR: 2>].
250617-08:53:45,80 nipype.workflow INFO:
[...]

PR #3470, opened 2025-06-17
FIX: Reflect sanitized changes from sdcflows

mgxd:
Addresses #3468 

Changes were made in sdcflows, but forgotten here...


effigies:
@mgxd Do you want to add/update a test where we have a `B0FieldIdentifier` fitting the erroring form?

mgxd:
@effigies the changes in [`b7536d9`](https://github.com/nipreps/fmriprep/pull/3470/commits/b7536d9eebe2907b778f0c49f2e998cad0f09119) (which corrects the layout to be bids compliant by using `dir` rather than `direction` for the entity label) exposed an invalid test - let me know what you think.

PR #3494, opened 2025-08-06
Update transforms.py according to new tranform chain of nitransforms

AxelCouturierCIBM:
https://github.com/nipy/nitransforms/commit/23a425d1b8e45a039a98f91bca4b07e9c561f02a

Without this, we get an error like "TypeError: Input file is not in X5 format"

### DIFF
--- a/fmriprep/utils/transforms.py
+++ b/fmriprep/utils/transforms.py
@@ -20,7 +20,7 @@ def load_transforms(xfm_paths: list[Path], inverse: list[bool]) -> nt.base.Trans
         path = Path(path)
         if path.suffix == '.h5':
             # Load as a TransformChain
-            xfm = nt.manip.load(path)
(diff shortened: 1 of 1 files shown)


fmriprep/utils/transforms.py
```diff
@@ -20,7 +20,7 @@ def load_transforms(xfm_paths: list[Path], inverse: list[bool]) -> nt.base.Trans
         path = Path(path)
         if path.suffix == '.h5':
             # Load as a TransformChain
-            xfm = nt.manip.load(path)
+            xfm = nt.manip.load(path, fmt="h5")
```
effigies, line 23:
```suggestion
            xfm = nt.manip.load(path, fmt='h5')
```

effigies (approved):

PR #3495, opened 2025-08-06
ENH: Add dedicated session filtering, alternative anatomical template options

mgxd:
Builds off https://github.com/nipreps/smriprep/pull/480

Adds new arguments:
- `--session-label` for a more streamlined way to filter sessions
- `--subject-anatomical-reference` to determine how the reference anatomical should be constructed.


fmriprep/config.py
```diff
@@ -434,8 +440,12 @@ class execution(_Config):
     """Only build the reports, based on the reportlets found in a cached working directory."""
     run_uuid = f'{strftime("%Y%m%d-%H%M%S")}_{uuid4()}'
     """Unique identifier of this particular run."""
+    processing_groups = None
+    """List of tuples (participant, session(s)) that will be preprocessed."""
```
effigies, line 444: This is derived from `participant_label`, `session_label` and `subject_anatomical_reference`. Might be handy for reporting, but it does mean that the config object state can be inconsistent.

IMO it would be better just to calculate the groups when needed with a well-isolated function rather than deal with (de)serialization and consistency.

mgxd: i also considered omitting, but decided to include it since having that provenance is nice. `session_label` is only populated if `--session-label` is used, otherwise all sessions pertaining to that subject are queried from the layout (and that information is not currently being stored). the config object should never be changed, just the dict representation of it (and when `dump`ing to toml).

docs/workflows.rst
```diff
@@ -174,15 +174,22 @@ Your ``.bidsignore`` file should include the following line::
...
-merged into a single template image using FreeSurfer's `mri_robust_template`_.
-This template may be *unbiased*, or equidistant from all source images, or
-aligned to the first image (determined lexicographically by session label).
+In the case of multiple T1w images (across sessions and/or runs), *fMRIPrep* provides
+a few choices on how to generate the reference anatomical space.
+
+If ``--subject-anatomical-reference first-lex`` is used, all T1w images are
+merged into a single template image using FreeSurfer's `mri_robust_template`_, aligned
+aligned to the first image (determined lexicographically by session label). This is
+the default behavior.
```
tsalo, line 183:
```suggestion
If ``--subject-anatomical-reference first-lex`` is used, all T1w images are
merged into a single template image using FreeSurfer's `mri_robust_template`_,
aligned to the first image (determined lexicographically by session label). This is
the default behavior.
```

docs/workflows.rst
```diff
@@ -174,15 +174,22 @@ Your ``.bidsignore`` file should include the following line::
...
+
+If ``--subject-anatomical-reference unbiased`` is used, all T1w images are merged into
+a an *unbiased* template, or equidistant from all source images.
```
tsalo, line 186:
```suggestion
If ``--subject-anatomical-reference unbiased`` is used, all T1w images are merged into
an *unbiased* template, equidistant from all source images.
```

docs/workflows.rst
```diff
@@ -174,15 +174,22 @@ Your ``.bidsignore`` file should include the following line::
...
 For two images, the additional cost of estimating an unbiased template is trivial,
 but aligning three or more images is too expensive to justify being the default behavior.
-For consistency, in the case of multiple images, *fMRIPrep* constructs
-templates aligned to the first image, unless passed the ``--longitudinal``
-flag, which forces the estimation of an unbiased template.
+
+If ``--subject-anatomical-reference sessionwise`` is used, a reference template will be
+generated for each session independently. If multiple T1w images are found within a session,
+other images will align to the first lexicographical image.
```
tsalo, line 192:
```suggestion
If ``--subject-anatomical-reference sessionwise`` is used, a reference template will be
generated for each session independently. If multiple T1w images are found within a session,
the images will be aligned to the first lexicographical image from that session.
```

fmriprep/config.py
```diff
@@ -589,7 +599,7 @@ class workflow(_Config):
     level = None
     """Level of preprocessing to complete. One of ['minimal', 'resampling', 'full']."""
     longitudinal = False
-    """Run FreeSurfer ``recon-all`` with the ``-logitudinal`` flag."""
+    """*Deprecated* - run FreeSurfer ``recon-all`` with the ``-logitudinal`` flag."""
```
tsalo, line 602: Isn't the old description incorrect? I thought the longitudinal Freesurfer workflow wasn't supported.

tsalo (commented):
I just have a couple of minor documentation-related suggestions for now.

fmriprep/config.py
mgxd: changed, although as i think about it as long as we keep the parser argument and remap to unbiased, this can be removed...

fmriprep/workflows/base.py
```diff
@@ -91,12 +91,30 @@ def init_fmriprep_wf():
...
+        log_dir = log_dir / 'log' / config.execution.run_uuid
+
+        wf_name = '_'.join(
+            (
+                'sub',
+                subject_id,
+            )
+            + (('ses', ses_str) if ses_str else ())
+            + ('wf',)
         )
```
effigies, line 114: This is cleaner with unpacking. Might still need reformatting.

```suggestion
        wf_name = '_'.join(['sub', subject_id, *(('ses', ses_str) if ses_str else ()), 'wf'])
```

effigies (commented):
Sorry for taking so long. I think this is mostly good to go.

PR #3513, opened 2025-08-29
ENH: Write out goodvoxels mask

### LINKED ISSUE #3510: save goodvoxels mask

Just curious if the fmriprep output saves the goodvoxels mask if the --project-good-voxels is set to true? Or whether there is a way to do it such that we can leave that in the output for --level minimal in case we need to use it later?

Yes

_No response_

tsalo:
Closes #3510. I'm happy to move the DerivativesDataSink into `init_goodvoxels_bold_mask_wf` if the current approach clutters up the workflow too much.

- Add a DerivativesDataSink to `init_bold_wf` to write out the goodvoxels mask.
- Update the expected outputs for relevant tests.


effigies (approved):
Seems fine.

PR #3524, opened 2025-09-16
MNT: Avoid using non-existing, deprecated contributor JSON files

jhlegarreta:

Avoid using non-existing, deprecated contributor JSON files:
- `contributors.json` was transitioned to `CONTRIBUTORS.md` in commit 059b59c, but the file was not updated in `paper_author_list.py`, probably because it is not run with every release.
- Prefer using `FORMER.md` over `former.json` in line with the adopted latest convention.
- Add the necessary functions to read contributors from Markdown tables in `paper_author_list.py`.

Ensure that reading contributors from a Markdown file table does not fail if the "Affiliation" and "ORCID" data are missing, which is the case for the current record in `FORMER.md`. Use `None` if no data exists for those columns.

Read the `FORMER.md` file at a single place in `update_zenodo.py` and store its contents in a variable that can be reused.

N/A

### DIFF
--- a/.maint/paper_author_list.py
+++ b/.maint/paper_author_list.py
@@ -17,14 +17,45 @@ def _aslist(inlist):
     return inlist
 
 
+def loads_table_from_markdown(s):
+    """Read the first table from a Markdown text."""
+    table = []
+    header = None
+    for line in s.splitlines():
+        if line.startswith('|'):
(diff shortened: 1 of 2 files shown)


effigies:
Dropped these files in favor of an updated `update_authors.py`.

PR #3556, opened 2025-10-17
feat: Ignore non-steady-state volumes in T2*/S0 estimation

### LINKED ISSUE #3507: T2* estimation appears to use dummy volumes
### What happened?

The T2* estimation and optimal combination step for multi-echo data appears to include non-steady-state volumes in the T2* estimation part. While it makes sense to optimally combine all volumes, the non-steady-state volumes could have some (almost certainly very small) effect on the T2* and S0 estimates.

### What command did you use?

```shell
n/a
```

25.1.4 (current)

Docker

### Is your data BIDS valid?

Yes

No

```shell

```

Related to https://github.com/ME-ICA/tedana/issues/1247.

tsalo:
Closes #3507. As of 25.1.0, tedana has an `--exclude` parameter in its t2smap CLI, which allows users to ignore certain volumes when estimating T2* and S0. This PR bumps the tedana version and leverages that parameter to ignore dummy scans. If it proves useful, we could also ignore high-motion volumes in the future.

- Increase minimum tedana version from 25.0.0 to 25.1.0.
- Use estimated number of dummy scans to set `exclude` parameter in t2smap call, so that those volumes do not contribute to the estimation of T2*, which is in turn used to optimally combine the data across echoes. I doubt the dummy scans actually have much effect on T2* estimation, but it still stands to reason that they are not _helping_.

### DIFF
--- a/fmriprep/interfaces/multiecho.py
+++ b/fmriprep/interfaces/multiecho.py
@@ -69,7 +69,7 @@ class T2SMapInputSpec(CommandLineInputSpec):
         ),
     )
     n_exclude = traits.Int(
-        argstr='--exclude %s',
+        argstr='--exclude 0:%d',
         position=5,
         desc='Number of volumes from the beginning of the run to exclude from T2*/S0 estimation.',
     )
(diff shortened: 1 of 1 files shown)


effigies:
This seems to work: https://output.circle-artifacts.com/output/job/c3ab52e1-5c68-4a65-ade6-430a130d4c32/artifacts/0/tmp/ds210/fmriprep/sub-02.html#datatype-figures_desc-summary_echo-None_run-1_subject-02_suffix-bold_task-cuedSGT

@tsalo Is this ready to merge? Do you need to test it on any other datasets?

Issue #3364, opened 2024-09-18
Process two-echo data like multi-echo without echo combination


tsalo:
### What would you like to see added in fMRIPrep?

#3362 got me thinking that it might be good to process two-echo data similarly to multi-echo data, except without any tedana-based echo combination steps. _Or_, maybe we could do a basic unweighted average when there's only two echoes?

1. Motion correct using the first echo.
2. Generate the boldref image either from one of the echoes _or_ just take an average.
3. Apply the same motion parameters to both echoes.
4. Output the preprocessed echoes in target spaces as separate files, as if you were running fMRIPrep twice with `--echo-idx` set to each of the echoes.

I think we could have a new `--me-t2s-fit-method` option called `mean` to just take the average across echoes. `--me-output-echos` might still be applicable, depending on if we wanted to just output the mean combined data or if we wanted to stick to outputting individual echoes.

**NB**: I'm not actually sure what the use-cases for two-echo protocols are. I know dual-echo denoising ([Bright & Murphy (2013)](https://www.sciencedirect.com/science/article/pii/S1053811912009536?via%3Dihub)) is a thing, but AFAIK that's only feasible in very limited cases where you can get the first echo early enough to not have any BOLD signal.

### Do you have any interest in helping implement the feature?

Yes

PR #3570, opened 2025-11-24
Raise error if two-echo data are provided

tsalo:
Replaces #3569. Relates to #3364. This acts as a short-term patch until we figure out how best to process two-echo fMRI data.

- If two-echo fMRI data are provided and `--echo-idx` is not, raise a RuntimeError.


effigies (approved):
LGTM

PR #3577, opened 2025-12-02
fix: Ensure multi-echo processing and slice timing correction text is in boilerplate

### LINKED ISSUE #3576: Multi-echo processing-related text is not included in boilerplate
### What happened?

I _think_ the problem is that the parent workflow where `init_bold_t2smap_wf` is run (`init_bold_native_wf`) is using a vanilla Nipype workflow object instead of the Niworkflows LiterateWorkflow. My guess is that having a regular Workflow between LiterateWorkflows prevents the embedded LiterateWorkflows’ descriptions from being compiled (see https://neurostars.org/t/no-multi-echo-t2-etc-in-the-boilerplate-text/34629/7?u=tsalo).

### What command did you use?

```shell
n/a
```

25.2.2

Docker

### Is your data BIDS valid?

Yes

No

```shell

```

It looks like other boilerplate text that should be compiled from workflows called within `init_bold_native_wf` is missing, namely slice timing correction information from `init_bold_stc_wf`.

tsalo:
Closes #3576.

- Change `init_bold_native_wf` from a Nipype Workflow to a Niworkflows LiterateWorkflow to make descriptions from workflows called within it accessible.
- Add a reference for tedana.

### DIFF
--- a/fmriprep/workflows/bold/t2s.py
+++ b/fmriprep/workflows/bold/t2s.py
@@ -100,7 +100,7 @@ def init_bold_t2s_wf(
 
     workflow.__desc__ = f"""\
 A T2<sup>★</sup> map was estimated from the preprocessed EPI echoes using tedana's t2smap
-workflow [@tedana], by voxel-wise fitting the maximal number of echoes with reliable signal in
+workflow [@DuPre2021], by voxel-wise fitting the maximal number of echoes with reliable signal in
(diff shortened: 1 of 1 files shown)


effigies (approved):
https://output.circle-artifacts.com/output/job/e531206e-1905-405a-85a0-c6320779308d/artifacts/0/tmp/ds210/fmriprep/logs/CITATION.html looks good:

image

image

PR #3588, opened 2025-12-16
fix collect freesurfer  

bpinsard:

Currently when preprocessing fMRI only dataset using precomputed sMRIPrep+freesurfer (`--fs-no-resume`)  and passing `--session-label yy`, the `fs_subject_id`  ends up being `sub-xx_ses-yy` using the fMRI selected session, and thus it picks files from that folder. I noticed it because my jobs was only pre-fetching with datalad `freesurfer/sub-xx` folder from a precomputed longitudinal freesurfer, but the pipeline tried using freesurfer files from `sub-xx_ses-yy` instead, which were broken symlinks. 

Not sure what is the best approach, but it seems that it should only try to get session-specific freesurfer when using `--subject-anatomical-reference sessionwise`, but it might require another case for `first-lex`. 

I am not 100% sure what are the outputs when you using this option:
- session-wise -> freesurfer/sub-xx_ses-yy/ ?
- first-lex -> freesurfer/sub-xx or same as sessionwise
- unbiased -> freesurfer/sub-xx I assume

### DIFF
--- a/fmriprep/workflows/base.py
+++ b/fmriprep/workflows/base.py
@@ -397,10 +397,7 @@ It is released under the [CC0]\
         ]),
         (bidssrc, src_file, [('out_dict', 'bids_info')]),
         (src_file, bids_info, [('source_file', 'in_file')]),
-        (bids_info, create_fs_id, [
-            ('subject', 'subject_id'),
-            ('session', 'session_id'),
-        ]),
+        (bids_info, create_fs_id, [('subject', 'subject_id')]),
         (create_fs_id, anat_fit_wf, [('subject_id', 'inputnode.subject_id')]),
         # Reporting connections
         (inputnode, summary, [('subjects_dir', 'subjects_dir')]),
@@ -409,6 +406,11 @@ It is released under the [CC0]\
         (summary, ds_report_summary, [('out_report', 'in_file')]),
         (about, ds_report_about, [('out_report', 'in_file')]),
     ])  # fmt:skip
+
+    if config.workflow.subject_anatomical_reference == 'sessionwise':
+        workflow.connect([
+            (bids_info, create_fs_id, [('session', 'session_id')]),
+        ])  # fmt:skip
 
     # Set up the template iterator once, if used
     template_iterator_wf = None


bpinsard:
Do you mean that this change would break the pipeline when not using precomputed freesurfer?

If I understand correctly the code, if using `unbiased` or `first-lex` `BIDSSourceFile` will generate a `source_file` that removes the `ses-` entity if multiple T1w (T2w) are used (even if all from the same session), so the sMRIPrep and freesurfer outputs will not contain that entity (even for `first-lex` !). 
So reusing these outputs for both cases for a new fMRI only session would fail without this fix, as it will try to get freesurfer output with that session entity. 
So the way I see it is that `--session-label`  is for fMRI only, while `--subject-anatomical-reference` (and bids_filters) control what is used to build the anatomical reference. 

In my case I cannot simply use symlinks, as these session-level folder already exists as it is  a true longitudinal freesurfer (including `sub-xx_ses-yy.long`  too) that was run separately (though using sMRIPrep brain extraction as it is more robust).

Covering all these cases with tests would be tricky, but important.

bpinsard:
> I think the current changes would break things when using `--subject-anatomical reference unbiased` or `first-lex` with `--session-label`. I don't think `--session-label` is supposed to only filter fMRI sessions, but I could be wrong. I can imagine people wanting to run fMRIPrep on a subset of sessions, using the anatomical data from just those sessions.

I think it would not break that usecase. 

Trying to branch all the possibility, this is complex. 
- if no precomputed sMRIPrep+freesurfer (haven't tested it):
  - "unbiased" or "first-lex" : if multiple T1w from one or multiple sessions, with or without `--session-label` -> fs_id has no `ses-` entity currently (related to comment here https://github.com/nipreps/fmriprep/pull/3583#discussion_r2619756855 ). 
  - "sessionwise": with one or multiple T[12]w per session `_create_multi_source_file` keeps the `ses-` entity, freesurfer output gets name accordingly.
- if precomputed + sessionwise 
- if precomputed given with a freesurfer `subjects-dir/sub-xx` (output from `first-lex` or `unbiased` above or custom):
  -  "unbiased" or "first-lex":
     - with or without `--session-label`
     - if no anat in BIDS and a single bold file `_create_multi_source_file`  keeps the `-ses` entity in the filename, if multiple bold runs drops it (including for multi-echo with the latest fix) generating an invalid bids path (ses- folder, no ses- in file), so as @mgxd mentioned this function is problematic. Either way `niworkflows.interfaces.bids.BIDSInfo` extract the `session` entity from the path, from the first BOLD (or anat) file that is then used to generated the `fs_subject_id`, which points to non-existent folder. Unless I am missing something that logic is flawed. 
```
> niworkflows.interfaces.bids.BIDSInfo(in_file='/path/to/bids/sub-01/ses-001/func/sub-01_bold.nii.gz',bids_validate=True).run().outputs

acquisition = <undefined>
reconstruction = <undefined>
run = <undefined>
session = 001
subject = 01
suffix = bold
task = <undefined>
``` 
 (BTW: `bids_validate` is not used anywhere in that interface)

Anyway, sorry I am getting lost in the multiverse, just to say that the logic from command-line options -> select files -> select single representative file / generic source (with the broken `_create_multi_source_file` ) -> extract BIDS entities -> generate fs_subject_id is very convoluted, brittle and might result in questionable output naming that are not reusable for "apply".

AFAIK the freesurfer outputs naming is not covered by tests in either fMRIPrep or sMRIPrep for circleCI or pytests, but I know running full freesurfer in tests would be a pain, unless we use `--fs-no-resume` and check that the workflow finds what it needs.

bpinsard:
superseded by #3614

Issue #3587, opened 2025-12-16
aCompCor mask creation depends on Freesurfer, but sMRIPrep uses FAST


tsalo:
### What happened?

In sMRIPrep, FAST is used to calculate the T1w TPMs even when Freesurfer reconstruction is enabled. However, in fMRIPrep's [`acompcor_masks`](https://github.com/nipreps/fmriprep/blob/55087d535c3101aa20bb36e913875ebb1e50e875/fmriprep/utils/confounds.py#L60), it looks like the aCompCor masks will still be created as if the TPMs were binary files from aseg when Freesurfer reconstruction is enabled.

If I'm missing something, please let me know.

### What command did you use?

```shell
n/a
```

### What version of fMRIPrep are you running?

master @ 55087d5

### How are you running fMRIPrep?

Docker

### Is your data BIDS valid?

Yes

### Are you reusing any previously computed results?

No

### Please copy and paste any relevant log output.

```shell

```

tsalo:
@mgxd @effigies any chance one of you could take a look at this this week?

PR #3589, opened 2025-12-16
Use continuous TPM-based procedure to calculate aCompCor masks

tsalo:
Cloes #3587.

- Assume that anatomical TPMs come from FAST and not Freesurfer's aseg output, since sMRIPrep always uses FAST. This means setting `freesurfer` to False in `init_bold_confs_wf`.
- Update boilerplate in `init_bold_confs_wf` to reflect changes and what is actually happening.

### DIFF
--- a/fmriprep/utils/confounds.py
+++ b/fmriprep/utils/confounds.py
@@ -26,8 +26,10 @@
 def mask2vf(in_file, zooms=None, out_file=None):
     """
     Convert a binary mask on a volume fraction map.
+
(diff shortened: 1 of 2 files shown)


mgxd (commented):
This looks correct to me - the TPMs passed in are not discrete, regardless of whether or not freesurfer is used. I can see some uses where the FS conversion logic would still be useful (i.e. infants), so my inclination is to preserve it and just set `freesurfer=False` when defining `init_bold_confs_wf()`

PR #3592, opened 2026-01-12
Apply Repo-Review rules

DimitriPapadopoulos:

* PC170: Uses PyGrep hooks (only needed if rST present)
* PC191: Ruff show fixes if fixes enabled
* PC903: Specified pre-commit CI schedule
* ~~PP006: The dev dependency group should be defined~~ (it looks like the tox/uv combination cannot support PEP 735 out-of-the-box)


effigies:
Thanks!

PR #3593, opened 2026-01-12
Get rid of black

DimitriPapadopoulos:

We don't use black any more. Explicitly disabling it is confusing, it gives the wrong impression.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -178,10 +178,6 @@ git-annex = ">=10.20250721"
 # Developer tool configurations
 #
 
-# Disable black
-[tool.black]
-exclude = ".*"
-
 [tool.pytest.ini_options]
 minversion = "6"
 testpaths = ["fmriprep"]
--- a/wrapper/pyproject.toml
+++ b/wrapper/pyproject.toml
@@ -63,10 +63,6 @@ universal = true
 # Developer tool configurations
 #
 
-# Disable black
-[tool.black]
-exclude = ".*"
-
 [tool.ruff]
 line-length = 99
 


effigies:
vscode turned on black by default some years back, and it became common to get PRs with whole files reformatted for minor changes. The purpose this serves is to disable autoformatting from any tool that wants to auto-apply black, which is mostly likely a new contributor and someone who doesn't need their first interaction with the project to be quibbling over formatting. The cost of keeping this in is lower than the cost of removing it, IMO.

DimitriPapadopoulos:
I believe Visual Studio Code **never** turned on Black by default. What you actually saw was contributors installing the Black extension without understanding what they do.

See [Formatter extension for Visual Studio Code using the Black formatter](https://marketplace.visualstudio.com/items?itemName=ms-python.black-formatter):
> ### Usage and Features
> 
> The Black extension for Visual Studio Code provides formatting support for your Python files. Check out the [Settings section](https://marketplace.visualstudio.com/items?itemName=ms-python.black-formatter#settings) for more details on how to customize the extension.
> * Integrated formatting: Once this extension is installed in VS Code, Black will be automatically available as a formatter for Python. This is because the extension ships with a Black binary. You can ensure VS Code uses Black by default for all your Python files by setting the following in your User settings (**View > Command Palette...** and run **Preferences: Open User Settings (JSON)**):

> ### Disabling formatting with Black
> 
> If you want to disable Black formatter, you can [disable this extension](https://code.visualstudio.com/docs/editor/extension-marketplace#_disable-an-extension) per workspace in Visual Studio Code.

The same could happen with other Studio Code extensions. It doesn't make sense to disable every possible extension just in case. Contributors have to know what they're doing...

Issue #3581, opened 2025-12-13
--derivatives and --session-label and bids filters conflicts


bpinsard:
### What happened?

I am trying to replicate what we were doing with the 20LTS for the specific usecase of neuromod but the 25.xx
- run smriprep on the anat dataset (T1w, T2w)
- run fmriprep on a fMRI only dataset, each session separately to distribute more easily on a HPC.

When running fmriprep with bids filters and `--derivatives` to input smriprep results and also trying to use `--session-label` all end up crashing at different stages.

```
fmriprep w ./workdir --participant-label 01 --session-label 001 --
derivatives sourcedata/smriprep --fs-subjects-dir sourcedata/smriprep/sourcedata/*freesurfer* --bids-filter-file code/fmriprep_study-cneuromod.emotion-videos_sub-01_ses-001_bids_filters.json --output-layout bids --ignore slicetiming --use-syn-sdc warn --me-output-echos --output-spaces MNI152NLin2009cAsym T1w:res-iso2mm --cifti-output 91k --notrack --write-graph --skip_bids_validation --omp-nthreads 8 --nprocs 12 --mem_mb 45056 --fs-license-file code/freesurfer.license  --track-carbon --stop-on-first-crash sourcedata/cneuromod.emotion-videos ./ participant
``` 
gives 
```
          File "/app/.pixi/envs/fmriprep/lib/python3.12/site-packages/fmriprep/interfaces/bids.py", line 148, in _creat
e_multi_source_file
            p = Path(filename_to_list(in_files)[0])
                     ~~~~~~~~~~~~~~~~~~~~~~~~~~^^^
        IndexError: list index out of range
``` 
because it passes the `session-label`  intended for fMRI selection to the precomputed anat derivatives collection here 
https://github.com/nipreps/fmriprep/blob/55087d535c3101aa20bb36e913875ebb1e50e875/fmriprep/workflows/base.py#L281
which will not collect anything as smriprep output does not have sessions (~averages all), unless maybe when using the new `sessionwise` option.
Maybe the simplest solution would be to only pass `session_id` when `--subject-anatomical-reference sessionwise` ?

Not using `--session-label` and relying on `--bids-filters` to select the session for func/fmap results in another error triggered here .
https://github.com/nipreps/niworkflows/blob/2591b8a5fc90a0f4809ac3288d6fdce5f9aaa227/niworkflows/utils/bids.py#L279-L281
[...]

PR #3606, opened 2026-02-10
feat: Add --fs-legacy-id flag

effigies:

This adds a `--fs-legacy-id` flag to disable adding `_ses-<label>` to FreeSurfer subject IDs. This can't be turned on by default for backwards compatibility reasons. Separately, we need to discuss what the best way forward is in general.

I think this is related to #3581, but that appears not to be limited to FreeSurfer. Need to think more there, as well.


mgxd (dismissed):
I think this is the best solution with minimal friction for the 25.2.X series - I've opened up #3607 to discuss changes for the 26.0.x series

effigies:
Are there other impacts on derivative reuse besides FS?

Issue #3595, opened 2026-01-14
Use config.toml to determine settings when creating full results from minimal mode


jbwexler:
### What would you like to see added in fMRIPrep?

There are many use cases in which users would want to have a generic command that can get full results from minimal mode results, without having to look back at the original command and include the correct settings used (ie `--use-syn-sdc` or `--ignore fmap-jacobian`). Perhaps there could be a setting like `--reuse-settings` that would pull settings from config.toml.

### Do you have any interest in helping implement the feature?

Maybe?

jbwexler:
It's turned out to be surprisingly complicated to automate the expansion of something like `sub-01/log/*/fmriprep.toml` as an input in my workflow (using ReproMan). My best option is probably to add it to the command array after `reproman run`, but before the job starts. I wonder if others would find it useful to have an option for `--config-file` that looks in the default location for that subject?

jbwexler:
Also, fmriprep.toml uses absolute paths and not all the paths can be overridden with command line arguments. For example, `log_dir`.

jbwexler:
After some more testing, it seems that `log_dir` isn't an issue, as it gets determined by the `fmriprep_dir`. The problem is that the `fmriprep_dir`  from the config file isn't getting overridden by the command line argument. Conversely, the `notrack=false` that is present in the config file doesn't get utilized, instead reverting to the default `notrack=true`.

PR #3625, opened 2026-03-09
FIX: Reuse fewer config options when using `--config-file`

mgxd:
Addresses #3595

Some configuration options make reusing configuration files difficult, either due to:
- being set at run time, not when parsing arguments (`fmriprep_dir`, `log_dir`)
- not having off toggles (`anat-only`, `notrack`)

This is a smell, but rather than a big refactor this adds `config.default_reuse_skips()` which manually tracks them.


effigies (approved):
Made a small fix directly.

PR #3627, opened 2026-03-18
Remove cohort entity and append cohort to space

### LINKED ISSUE #3365: Remove cohort entity and append cohort to space

I think surface and volumetric space both need to be indicated in filenames, as the current convention of using shorthand names for combinations of surface template and volume template (i.e., fsLR standing in for fsLR + MNI152NLin6Asym) doesn't extend well to arbitrary combinations.

In order to indicate both the surface and volume templates, I originally proposed adding `cohort`, `volspace`, and `volcohort` to filenames. However, most folks weren't a fan of that, so we've started going in the direction of using `+` signs in entities to incorporate multiple values (see bids-standard/bids-specification#1926). The thing is, in cases where one or both of the spaces in a CIFTI file (surface and/or volume) have an associated cohort value, the cohort entity doesn't really work. To use an example from https://github.com/nipreps/niworkflows/issues/881, if you wanted dhcpAsym:cohort-42 as the surface space and MNIInfant:cohort-1 as the volume space, the current approach would have an output like `space-dhcpAsym+MNIInfant_cohort-42+1`, which isn't very interpretable. Instead, appending the cohort directly to the template name might be clearer: `space-dhcpAsym42+MNIInfant1`.

Yes

This is related to #3330, #3340, bids-standard/bids-specification#1926, https://github.com/nipreps/niworkflows/issues/881, and https://github.com/nipreps/niworkflows/issues/883.

tsalo:
Closes #3365. Reproduces nipreps/nibabies#524. Depends on nipreps/smriprep#526.

- Remove uses of `cohort` and just pass `space` along.

### DIFF
--- a/fmriprep/workflows/base.py
+++ b/fmriprep/workflows/base.py
@@ -443,7 +443,6 @@ It is released under the [CC0]\
                     ('outputnode.std_t1w', 'inputnode.ref_file'),
                     ('outputnode.anat2std_xfm', 'inputnode.anat2std_xfm'),
                     ('outputnode.space', 'inputnode.space'),
-                    ('outputnode.cohort', 'inputnode.cohort'),
                     ('outputnode.resolution', 'inputnode.resolution'),
                 ]),
             ])  # fmt:skip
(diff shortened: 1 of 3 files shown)


tsalo:
I was able to test this locally with MNIPediatricAsym cohort-5 (with sMRIPrep master mounted as well) and the outputs have `space-MNIPediatricAsym+5` as desired.

PR #3640, opened 2026-04-25
FIX: Remove redundant fmap_boldref resample in desc-fmapCoreg_bold.svg report

### LINKED ISSUE #3639: Fieldmap overlay in desc-fmapCoreg_bold.svg is resampled twice, may render outside the brain
### What happened?

The fieldmap-in-Hz overlay in `desc-fmapCoreg_bold.svg` ("Alignment between the anatomical reference of the fieldmap and the BOLD reference") is resampled twice on its way into the reportlet, so the colored overlay is plotted at the wrong location relative to the brain. The size of the misalignment depends on the acquisition geometry: when BOLD and fieldmap have similar FOV anchors the shift is small (~mm) and falls inside the brain so the figure looks fine, but when they differ substantially the overlay can render entirely outside the brain.

The actual SDC correction is **not** affected — the buggy node is on a reports-only branch. Only the figure misleads.

#### ses-03 — visible

| Current `desc-fmapCoreg_bold.svg` | With proposed fix |
|---|---|
| current | fixed |

The fieldmap overlay sits anterior of (and largely outside) the brain in the current report — clearest in sagittal slices `x=-14` through `x=40` and coronal slices `y=25` through `y=64`. With the proposed fix it sits in the orbitofrontal and inferior-temporal regions where susceptibility-induced fields physically are.

> The "current" panel is a verbatim rasterization of the SVG fmriprep produced for this run. The "with proposed fix" panel was generated by invoking `FieldmapReportlet` with the same inputs except `inputnode.fieldmap` set to `boldref_fmap.out_file` (i.e. the `boldref2fmap_xfm` resample applied once instead of twice).

#### ses-01 — bug present, masked by FOV overlap

| Current | With proposed fix |
|---|---|
| current | fixed |

Same workflow, same code path, same bug. For this session the BOLD and fieldmap acquisitions had similar FOV anchors so `boldref2fmap_xfm` is small (~2.6 mm). The overlay still falls inside the brain in both panels, so the bug is invisible — presumably why it has not been reported in the year between #3387 and #3467.

### What command did you use?

```shell
fmriprep $BIDS_DIR $OUT_DIR participant \
  --participant-label s8 \
  --output-spaces "MNI152NLin2009cAsym:res-2 fsnative func anat" \
  --bids-filter-file filter.json \
  --notrack --skip-bids-validation
```

with `filter.json`:

```json
{"bold": {"task": "stroop"}}
```

25.2.5; bug also present on `master` HEAD `acdb0fde`.

Singularity

### Is your data BIDS valid?

Yes

FreeSurfer

```shell
# fMRIPrep does not error; the workflow completes and the SVG is generated.
# (Run exits 1 with "fMRIPrep finished successfully!" — likely #3611, not this bug.)
```

**Centroid trace through the SDC-report pipeline (sub-s8 ses-03; BOLD ref shape `(80, 80, 51)`, voxel `2.8 mm`):**

| Stage | World Y centroid (mm) |
|---|---|
| 1. fieldmap coefficients (raw, fmap-space) | -54.4 |
| 2. fieldmap magnitude (raw, fmap-space) | -55.3 |
| 3. `boldref_fmap.out_file` (B-spline reconstructed onto BOLD grid) | **-5.0** ← matches BOLD ref |
| 4. `fmap_boldref.output_image` (extra `antsApplyTransforms` — fed to `sdcreg_report`) | **+20.2** ← shifted ~25 mm |
| 5. `fmapref_boldref.output_image` (parallel path for magnitude) | -4.9 ← matches BOLD ref |
| 6. `desc-coreg_boldref.nii.gz` (rendering reference) | -3.7 |

The Y shift in stage 4 corresponds to the magnitude of `boldref2fmap_xfm`. For ses-03 this transform contains a ~50 mm Y translation; for ses-01 it is only ~2.6 mm — same code path, much smaller visible error.

**Root cause**

`boldref_fmap` (a `ReconstructFieldmap` node added in #3467) already produces a fieldmap-in-Hz on the BOLD reference grid. Its output is connected as `func_fit_reports_wf.inputnode.fieldmap` at [`fmriprep/workflows/bold/fit.py#L500`](https://github.com/nipreps/fmriprep/blob/25.2.5/fmriprep/workflows/bold/fit.py#L500):

```python
(boldref_fmap, func_fit_reports_wf, [('out_file', 'inputnode.fieldmap')]),
```

Inside `init_func_fit_reports_wf`, that already-aligned image is then resampled a second time by `fmap_boldref` (added in #3387) at [`fmriprep/workflows/bold/outputs.py#L380-L385`](https://github.com/nipreps/fmriprep/blob/25.2.5/fmriprep/workflows/bold/outputs.py#L380-L385):

```python
(inputnode, fmap_boldref, [
    ('fieldmap', 'input_image'),
    ('coreg_boldref', 'reference_image'),
    ('boldref2fmap_xfm', 'transforms'),
]),
```

`fmap_boldref` was correct at the time of #3387 because `inputnode.fieldmap` carried a raw RAS fieldmap then. After #3467 rewired the connection, the resample became redundant — but `fmap_boldref` was not removed.

**Proposed fix**

Remove `fmap_boldref` and connect `inputnode.fieldmap` directly to `sdcreg_report.fieldmap`. `grep -rn fmap_boldref` confirms there are no other consumers in fmriprep. A PR follows.

**Possibly related but separate**

- #3550 asks about the conceptual purpose of `desc-fmapCoreg_bold.svg`; not this bug.
- #3611 / #3426 — exit-1 with "fMRIPrep finished successfully!" — observed in this run but unrelated to the SVG content.

lobennett:
Closes #3639.

- Remove the `fmap_boldref` `ApplyTransforms` node from `init_func_fit_reports_wf`.
- Connect `inputnode.fieldmap` directly to `sdcreg_report.fieldmap`.

This restores the wiring that existed before #3387, which is now correct because #3467 rewired `func_fit_reports_wf.inputnode.fieldmap` to receive `boldref_fmap.out_file` — a fieldmap already reconstructed onto the BOLD reference grid.

## Why

`fmap_boldref` was added in #3387 to resample a raw RAS fieldmap into BOLD space for the `FieldmapReportlet`. After #3467 introduced `boldref_fmap` (a `ReconstructFieldmap` node) and rewired its output into the reports workflow at `fit.py:500`, the resample became redundant. Removing `fmap_boldref` restores correct rendering. Actual SDC is unaffected — `unwarp_boldref` (`fit.py:614`) and the time-series `resample` (`apply.py`) use `ReconstructFieldmap` outputs directly. See the linked issue for figures and a centroid trace.

## Impact

- **Reports:** `desc-fmapCoreg_bold.svg` now plots the fieldmap at the correct location regardless of BOLD/fmap acquisition geometry.
- **Workflow:** one node removed, two connections deleted, one connection added. `grep -rn fmap_boldref` returns zero matches after the change.
- **Other outputs:** none affected. `fmap_boldref` was reports-only.

## Testing

- Built `init_func_fit_reports_wf` under `sdc_correction=True/False` × `freesurfer=True/False` and walked the workflow graph: `fmap_boldref` is gone, and `sdcreg_report.fieldmap` is connected directly from `inputnode.fieldmap` post-fix; the other three required inputs (`reference`, `moving`, `mask`) remain wired as before.
- Reproduced the bug on a real subject with fmriprep `25.2.5` (figures in the linked issue) and verified the fix produces correctly-located overlays by invoking `FieldmapReportlet` with the proposed wiring.
- `ruff check`, `ruff format --check`, and `codespell` pass on the modified file.

None.

### DIFF
--- a/fmriprep/workflows/bold/outputs.py
+++ b/fmriprep/workflows/bold/outputs.py
@@ -313,18 +313,6 @@ def init_func_fit_reports_wf(
             mem_gb=1,
         )
 
-        fmap_boldref = pe.Node(
-            ApplyTransforms(
-                dimension=3,
-                default_value=0,
-                float=True,
-                invert_transform_flags=[True],
-                interpolation='LanczosWindowedSinc',
-            ),
-            name='fmap_boldref',
-            mem_gb=1,
-        )
-
         # SDC1
         sdcreg_report = pe.Node(
             FieldmapReportlet(
@@ -377,17 +365,12 @@ def init_func_fit_reports_wf(
                 ('coreg_boldref', 'reference_image'),
                 ('boldref2fmap_xfm', 'transforms'),
             ]),
-            (inputnode, fmap_boldref, [
-                ('fieldmap', 'input_image'),
-                ('coreg_boldref', 'reference_image'),
-                ('boldref2fmap_xfm', 'transforms'),
-            ]),
             (inputnode, sdcreg_report, [
                 ('sdc_boldref', 'reference'),
+                ('fieldmap', 'fieldmap'),
                 ('bold_mask', 'mask'),
             ]),
             (fmapref_boldref, sdcreg_report, [('output_image', 'moving')]),
-            (fmap_boldref, sdcreg_report, [('output_image', 'fieldmap')]),
             (sdcreg_report, ds_sdcreg_report, [('out_report', 'in_file')]),
             (inputnode, sdc_report, [
                 ('sdc_boldref', 'before'),
--- a/fmriprep/workflows/bold/outputs.py
+++ b/fmriprep/workflows/bold/outputs.py
@@ -163,25 +163,33 @@ def init_func_fit_reports_wf(
 
     Inputs
     ------
-    std_t1w
-        T1w image resampled to standard space
-    std_mask
-        Mask of skull-stripped template
+    sdc_boldref
+        BOLD reference before SDC, in BOLD space
+    coreg_boldref
+        BOLD reference after SDC, in BOLD space
+    boldref2anat_xfm
+        Affine transform from BOLD reference to anatomical space
+    boldref2fmap_xfm
+        Affine transform from BOLD reference to fieldmap space
+    t1w_preproc
+        The T1w reference map, which is calculated as the average of bias-corrected
+        and preprocessed T1w images, defining the anatomical space.
+    t1w_mask
+        Brain (binary) mask estimated by brain extraction.
+    t1w_dseg
+        Segmentation in T1w space
+    fieldmap
+        Reconstructed fieldmap, in BOLD space
+    fmap_ref
+        Fieldmap reference image, in fieldmap space
     subject_dir
         FreeSurfer SUBJECTS_DIR
     subject_id
         FreeSurfer subject ID
-    t1w_conform_report
-        Conformation report
-    t1w_preproc
-        The T1w reference map, which is calculated as the average of bias-corrected
-        and preprocessed T1w images, defining the anatomical space.
-    t1w_dseg
-        Segmentation in T1w space
-    t1w_mask
-        Brain (binary) mask estimated by brain extraction.
-    template
-        Template space and specifications
+    summary_report
+        HTML snippet summarizing BOLD processing decisions
+    validation_report
+        HTML snippet indicating whether affine metadata were overwritten
 
     """
     from nireports.interfaces.reporting.base import (


effigies:
Thanks for the detailed analysis and PR! I've triggered the tests. I still need to digest these reports and the referenced PRs, but this is much very much appreciated.

effigies (approved):
This is correct, but a bit disturbing. In general, we've tried to indicate the type and space of the thing being passed between workflows via the name, and it seems that I didn't follow that in #3467, which makes me worry that we aren't being consistent elsewhere.

I've updated the deeply outdated docstring so at least the space is documented here.

cc @mgxd as a heads up.

Issue #3336, opened 2024-07-29
Minimally preprocess BOLD phase data


tsalo:
### What would you like to see added in fMRIPrep?

I think phase reconstruction is going to become more popular over time as tools become available to leverage it. It would be great if fMRIPrep could apply minimal preprocessing to these data. The following steps would probably be enough:

1. Phase unwrapping, preferably leveraging multi-echo information when available, using ROMEO.
2. Motion correction (no STC necessary, AFAIK)
3. Distortion correction

### Do you have any interest in helping implement the feature?

Yes

### Additional information / screenshots

This is related to #3308 (NORDIC uses phase data, when available), https://github.com/nipreps/sdcflows/issues/36 (dynamic distortion correction include phase unwrapping), and https://github.com/vanandrew/warpkit/issues/12 (warpkit will hopefully output unwrapped phase time series soon).

Some potential applications of the phase data:

- phase regression (e.g., [Curtis et al. (2014)](https://linkinghub.elsevier.com/retrieve/pii/S1053811914004649))
- phase jolt time series ([OHBM 2024 poster](https://ww6.aievolution.com/hbm2401/index.cfm?do=abs.viewAbstract&style=1&abstractID=3633), [code](https://github.com/layerfMRI/LAYNII/blob/master/src/LN2_PHASE_JOLT.cpp))
- improved T2* estimation ([Cohen-Adad et al. (2012)](https://linkinghub.elsevier.com/retrieve/pii/S1053811912000705))
- complex-valued ICA (e.g., [Adali and Calhoun (2007)](http://ieeexplore.ieee.org/document/4350234/))
- complex-valued modeling (e.g., [Yu et al. (2022)](https://onlinelibrary.wiley.com/doi/10.1111/biom.13631))

I think that most processing could be pulled from fMRIPrep derivatives (e.g., transforms) and applied to the phase data in an fMRIPost workflow, but the one application I'm concerned about is the T2* estimation, which has implications for the optimal combination done within fMRIPrep.

I have created an fMRIPost repo ([`nipreps/fmripost-phase`](https://github.com/nipreps/fmripost-phase)) to handle at least some of the processing/applications I've proposed.

PR #3645, opened 2026-05-08
ENH: Add optional warpkit MEDIC support for multi-echo BOLD distortion correction

DVSneuro:

I think this is a little related to #3336 and maybe some other work from @tsalo , but this PR addresses a narrower use case than that issue: optional MEDIC-based susceptibility distortion correction for compatible multi-echo BOLD runs with `part-phase` data, using `warpkit` following the logic from Andrew Van's ( @vanandrew ) preprint on bioRxiv: https://www.biorxiv.org/content/10.1101/2023.11.28.568744v1

When `--me-use-warpkit` is enabled, fMRIPrep will:
- identify eligible multi-echo BOLD runs with corresponding `part-phase` data
- estimate a time-varying B0 fieldmap series with MEDIC from phase evolution across echoes
- use that fieldmap series during resampling
- report the run as `MEDIC (warpkit)` in the functional summary

This path is opt-in and does not change default behavior for existing workflows.

**Main changes in this PR:**
- add a new CLI/config option: `--me-use-warpkit`
- add an interface for calling `warpkit.api.medic`
- wire MEDIC into the multi-echo BOLD workflow for eligible runs
- ensure MEDIC-managed runs do not also use standard fmap selection for the same run
- add support for applying a 4D fieldmap series volume-by-volume during resampling
- add reportlets for a representative fieldmap view and an SDC before/after panel
- add citation/boilerplate text describing the MEDIC-based correction path
- add tests covering parser/config behavior, workflow wiring, BIDS helpers, and resampling behavior

**Current scope / limitations:**
- this feature is opt-in
- it is currently intended for compatible multi-echo datasets with `part-phase` companions
- MEDIC-managed runs currently prefer the MEDIC path rather than falling back to standard fieldmaps for the same run
- report output currently includes representative static views rather than a time-resolved QC plot of the full 4D fieldmap series

**Validation performed:**
- targeted unit/workflow tests for the new CLI flag, interface wiring, BIDS helpers, and 4D fieldmap resampling
- container build including `warpkit`
- smoke testing of the containerized import path
- a real multi-echo dataset with phase data from OpenNeuro (https://openneuro.org/datasets/ds005123/versions/1.1.3), confirming successful execution and report generation

In the HTML report, warpkit-enabled runs showed:
- `Susceptibility distortion correction: MEDIC (warpkit)`
- a representative fieldmap reportlet
- an SDC before/after reportlet

This is my first contribution to fMRIPrep, so I tried to keep the scope focused and follow existing workflow patterns closely.

I used OpenAI Codex to help draft and iterate on portions of the code and tests, and I manually reviewed and tested the resulting changes as carefully as I could, but I wouldn't say I'm an expert on this (maybe a little better than vibe coding, but not by much). 

This PR does not add standalone user documentation pages, but it does update user-facing/report-facing text in a few places:
- a new CLI option, `--me-use-warpkit`
- functional reportlets for the MEDIC/warpkit path
- citation/boilerplate text describing the MEDIC-based distortion-correction procedure

### DIFF
--- a/fmriprep/interfaces/resampling.py
+++ b/fmriprep/interfaces/resampling.py
@@ -378,6 +378,15 @@ async def resample_series_async(
         The resampled array, with shape ``coordinates.shape[1:] + (N,)``,
         where N is the number of volumes in ``data``.
     """
+    fmap_is_series = fmap_hz.ndim > 3
+    if data.ndim == 3 and fmap_is_series:
+        raise ValueError('A 3D source image cannot be resampled with a 4D fieldmap series.')
+    if fmap_is_series and fmap_hz.shape[-1] != data.shape[-1]:
+        raise ValueError(
+            'Fieldmap series and source series must have matching numbers of volumes '
+            f'(got {fmap_hz.shape[-1]} and {data.shape[-1]}).'
+        )
+
     if data.ndim == 3:
         return resample_vol(
             data,
@@ -385,7 +394,7 @@ async def resample_series_async(
             pe_info[0],
             jacobian,
             hmc_xfms[0] if hmc_xfms else None,
-            fmap_hz,
+            fmap_hz[..., 0] if fmap_is_series else fmap_hz,
             output_dtype,
             order,
             mode,
@@ -409,7 +418,7 @@ async def resample_series_async(
                     pe_info=pe_info[volid],
                     jacobian=jacobian,
                     hmc_xfm=hmc_xfms[volid] if hmc_xfms else None,
-                    fmap_hz=fmap_hz,
+                    fmap_hz=fmap_hz[..., volid] if fmap_is_series else fmap_hz,
                     output=out_array[..., volid],
                     order=order,
                     mode=mode,
--- a/.maint/CONTRIBUTORS.md
+++ b/.maint/CONTRIBUTORS.md
@@ -60,6 +60,7 @@ Before every release, unlisted contributors will be invited again to add their n
 | Schaefer          | Theo A.J.    |  | 0000-0003-4102-559X | Max Planck Institute for Human Cognitive and Brain Sciences, Leipzig, Germany                                                                |
 | Seeley            | Saren        | @sarenseeley | 0000-0002-9493-8512 | Department of Psychiatry, Icahn School of Medicine at Mount Sinai |
 | Sitek             | Kevin R.     |  | 0000-0002-2172-5786 | Speech & Hearing Bioscience & Technology Program, Harvard University                                                                         |
+| Smith             | David V.     | @dvsmith.bsky.social | 0000-0001-5754-9633 | Department of Psychology and Neuroscience, Temple University |
 | Smith             | Robert E.    |  | 0000-0003-3636-4642 | Florey Institute of Neuroscience and Mental Health                                                                                           |
 | Sneve             | Markus H.    |  | 0000-0001-7644-7915 | Center for Lifespan Changes in Brain and Cognition, University of Oslo                                                                       |
 | Stojić            | Hrvoje       |  | 0000-0002-9699-9052 | Max Planck UCL Centre for Computational Psychiatry and Ageing Research, University College London                                            |
--- a/Dockerfile
+++ b/Dockerfile
@@ -40,6 +40,7 @@ ARG PIXI_LOCK_FLAGS=--frozen
 #   - ...
 #
 FROM ghcr.io/prefix-dev/pixi:0.53.0 AS build
+ARG PIXI_LOCK_FLAGS=--frozen
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
                     ca-certificates \
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,6 +23,7 @@
 # SOFTWARE.
 
 ARG BASE_IMAGE=ghcr.io/nipreps/fmriprep-base:20251006
+ARG PIXI_LOCK_FLAGS=--frozen
 
 #
 # Build pixi environment
@@ -51,7 +52,7 @@ RUN pixi config set --global run-post-link-scripts insecure
 RUN mkdir /app
 COPY pixi.lock pyproject.toml /app
 WORKDIR /app
-RUN --mount=type=cache,target=/root/.cache/rattler pixi install -e fmriprep -e test --frozen --skip fmriprep
+RUN --mount=type=cache,target=/root/.cache/rattler pixi install -e fmriprep -e test ${PIXI_LOCK_FLAGS} --skip fmriprep
 RUN --mount=type=cache,target=/root/.npm pixi run --as-is -e fmriprep npm install -g svgo@^3.2.0 bids-validator@1.14.10
 # Note that PATH gets hard-coded. Remove it and re-apply in final image
 RUN pixi shell-hook -e fmriprep --as-is | grep -v PATH > /shell-hook.sh
@@ -59,7 +60,7 @@ RUN pixi shell-hook -e test --as-is | grep -v PATH > /test-shell-hook.sh
 
 # Finally, install the package
 COPY . /app
-RUN --mount=type=cache,target=/root/.cache/rattler pixi install -e fmriprep -e test --frozen
+RUN --mount=type=cache,target=/root/.cache/rattler pixi install -e fmriprep -e test ${PIXI_LOCK_FLAGS}
 
 #
 # Pre-fetch templates
--- a/fmriprep/cli/parser.py
+++ b/fmriprep/cli/parser.py
@@ -103,6 +103,13 @@ def _build_parser(**kwargs):
             raise parser.error("Argument can't be less than one.")
         return value
 
+    def _min_zero(value, parser):
+        """Ensure an argument is not lower than 0."""
+        value = int(value)
+        if value < 0:
+            raise parser.error("Argument can't be negative.")
+        return value
+
     def _to_gb(value):
         scale = {'G': 1, 'T': 10**3, 'M': 1e-3, 'K': 1e-6, 'B': 1e-9}
         digits = ''.join([c for c in value if c.isdigit()])
@@ -177,6 +184,7 @@ def _build_parser(**kwargs):
     PathExists = partial(_path_exists, parser=parser)
     IsFile = partial(_is_file, parser=parser)
     PositiveInt = partial(_min_one, parser=parser)
+    NonnegativeInt = partial(_min_zero, parser=parser)
     BIDSFilter = partial(_bids_filter, parser=parser)
     SliceTimeRef = partial(_slice_time_ref, parser=parser)
     FallbackTRT = partial(_fallback_trt, parser=parser)
@@ -493,6 +501,17 @@ https://fmriprep.readthedocs.io/en/{currentv.base_version if is_release else 'la
             'Use warpkit MEDIC for susceptibility distortion correction of compatible '
             'multi-echo BOLD runs. Requires phase companions for each echo and a '
             'warpkit installation (for example, `fmriprep[warpkit]`) on Python 3.11+.'
+        ),
+    )
+    g_conf.add_argument(
+        '--me-warpkit-noise-frames',
+        action='store',
+        type=NonnegativeInt,
+        default=0,
+        help=(
+            'Number of trailing non-imaging/noise frames to trim from each '
+            'multi-echo magnitude and phase file before running warpkit MEDIC. '
+            'Only applies when --me-use-warpkit is enabled.'
         ),
     )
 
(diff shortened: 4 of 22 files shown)


effigies:
This is extremely interesting and coincides with some work by @vanandrew in sdcflows (note to self: link PR). I've been meaning to look into how it would integrate and haven't found the time.

Have you run datasets with this workflow? Do the results look good to you? Do you have comparisons with other fieldmap types?

I will try to look at this in some detail soon.

DVSneuro:
I've only ran a couple of subjects from our dataset (https://openneuro.org/datasets/ds005123/versions/1.1.3), and here's one example visualization for a single time point, which looks pretty good to me. 

sub-10886_ses-01_task-sharedreward_run-1_part-mag_desc-fieldmap_bold

I haven't compared it directly with standard fieldmaps, though we have those in the dataset above. 

We were drawn to warpkit/medic since some of our participants move a lot (older adults) and we had a few cases where the participant needed to get out of the scanner for a break but we forgot to get a new fieldmap when they got back in. So, the idea of making the fieldmaps for each frame using the multi-echo phase data was very appealing to us! But when we first started using medic/warpkit a year or so ago, I couldn't figure out how to get it integrated into fmriprep, so we'd generate a fieldmap for the first time point of each run and use that in the normal fmriprep workflow and the results have been good. I'm assuming the framewise implementation should be even better, especially for high-motion participants -- but hopefully I didn't break anything or do anything too silly with this PR!

vanandrew:
I got you @effigies: https://github.com/nipreps/sdcflows/pull/541 :)

I haven't reviewed @DVSneuro PR yet, but my vague plan was to have MEDIC be called from `sdcflows` via `init_fmap_preproc_wf` (this [line](https://github.com/nipreps/fmriprep/blob/4e948806237854639e486e957329d772cd2c137c/fmriprep/workflows/base.py#L698) in fmriprep).

On the fmriprep side, the biggest changes would need to be in how the field map is applied: in MEDIC, you get a field map for every frame of your scan, so the one step resampling would need to be adjusted to account for this. Given that @DVSneuro has some results, I assume your PR already makes a change like this?

Awesome job getting this working btw :).

DVSneuro:
Thanks @vanandrew ! I had plenty of help from Codex. And yes, this PR does update the fMRIPrep side so a 4D MEDIC fieldmap series is applied volume-by-volume during one-step resampling. The current implementation calls MEDIC directly from fMRIPrep, but I’d be open to trying to restructure that piece if the preferred long-term architecture is to generate the MEDIC estimates in sdcflows and then consume them from fMRIPrep.

I'm not seeing any other multi-echo datasets on OpenNeuro that have the phase data, so the testing here has been somewhat limited.

tsalo:
@DVSneuro, I suggest checking https://me-ica.github.io/open-multi-echo-data/. I try to keep it up to date, and you can filter by if there's complex-valued data or not.

There are nine on OpenNeuro that I'm aware of:

- https://openneuro.org/datasets/ds007637
- https://openneuro.org/datasets/ds007486
- https://openneuro.org/datasets/ds006926
- https://openneuro.org/datasets/ds006707
- https://openneuro.org/datasets/ds006131
- https://openneuro.org/datasets/ds006072
- https://openneuro.org/datasets/ds005123
- https://openneuro.org/datasets/ds005085
- https://openneuro.org/datasets/ds002278

I think the goal should be to have MEDIC within SDCFlows so that dynamic distortion correction is treated like static distortion correction, though it would be cool if fMRIPrep could use `wk-unwrap-phase` to produce unwrapped phase data in any desired output spaces.

One change that might be necessary on fMRIPrep's side is (optionally?) swapping the order of HMC and SDC when MEDIC is used, since I think dynamic distortion correction should improve motion correction.

vanandrew:
For datasets, I've mostly been using this one for my own tests: https://openneuro.org/datasets/ds006926/versions/1.0.0 (Thanks @tsalo for the list of mag/phase datasets).

Getting into some of the more technical details:

- The plan is to have https://github.com/nipreps/sdcflows/pull/541 emit [`fmap_dynamic*`](https://github.com/vanandrew/sdcflows/blob/c06669f766ea8bc5200c241538745273ccd99bff/sdcflows/workflows/base.py#L105) fields from [`init_fmap_workflow_wf`](https://github.com/vanandrew/sdcflows/blob/c06669f766ea8bc5200c241538745273ccd99bff/sdcflows/workflows/base.py#L36). 
- This fieldmap dynamic file is a 4D nifti, where each frame contains a fieldmap in Hz (so extracting a single frame would be equivalent to the regular `fmap` output).
- @tsalo suggestion for swapping HMC/SDC, that's definitely worth considering. I don't think I looked into this in my original paper. I'm not sure how much work that is though, so it might be worth shifting into another PR.

Side question: I noticed that both nipreps project currently support python version 3.10+, but I've made `warpkit` support 3.11+ (mainly because 3.10 is due to be deprecated in October this year). It is worth lowering the python version requirement for warpkit? I wasn't sure how long you guys support deprecated python versions.

DVSneuro:
Thanks, @tsalo and @vanandrew ! I'm happy to try and help with the sdcflows approach. In the meantime, I ran additional validation with an Apptainer image built from this branch using `--me-use-warpkit` on several public OpenNeuro datasets suggested above. I skipped a couple of the datasets since they are from my own lab and use the same parameters as my original test (`ds007486` and `ds006707`).

Successful runs so far:

| Dataset | Subject | Notes |
| --- | --- | --- |
| `ds007637` | `sub-01` | Completed successfully; logs show MEDIC/warpkit for eligible multi-echo runs and standard PEPOLAR handling for other runs. |
| `ds006926` | `sub-a01` | Completed successfully with MEDIC/warpkit workflows. |
| `ds006131` | `sub-20188` | Completed successfully with MEDIC/warpkit workflows. |
| `ds002278` | `sub-Blossom` | Completed successfully with multiple MEDIC/warpkit workflows. |
| `ds005085` | `sub-10015` | Completed successfully; this subject/run is from my lab and has a mix of single echo and multi-echo data, so it's a nice compatibility check. |

One remaining test case is `ds006072`. `sub-P1` was not eligible because it lacked matching `part-phase` BOLD files. `sub-P1R` did start MEDIC workflows, but failed inside warpkit phase unwrapping with an `IndexError` message -- I'm not sure what's happening there. I had to throttle the resources a lot since it's a big dataset and kept running out of memory (here's my testing script in case it's helpful for anyone: https://github.com/DVS-Lab/rf1-sra-linux2/blob/main/code/fmriprep-warpkit.sh).

vanandrew:
> sub-P1R did start MEDIC workflows, but failed inside warpkit phase unwrapping with an IndexError message

@DVSneuro If it's an error in `warpkit`, can you create a issue ticket [here](https://github.com/vanandrew/warpkit/issues) and post the exception stack? Thanks in advance!

DVSneuro:
Thanks, @vanandrew! Just tying up this loose end: your suggestion about the `noise_frames` parameter resolved the `ds006072/sub-P1R` failure I noted above and reported in https://github.com/vanandrew/warpkit/issues/24.

I pushed a small follow-up commit exposing warpkit’s `noise_frames` parameter as `--me-warpkit-noise-frames` in the current fMRIPrep-side implementation: (https://github.com/nipreps/fmriprep/pull/3645/commits/4eb308f034ad8986923a0d9244fa2ec7cb261d51). I rebuilt the container and reran `ds006072` v1.0.8, `sub-P1R`, limited to `ses-1`, with:

`--me-use-warpkit --me-warpkit-noise-frames 3`

This resolved the previous warpkit phase-unwrapping failure. The log shows the expected trimming message, both MEDIC nodes completed, and fMRIPrep finished successfully. [ds006072_v1_0_8_sub-P1R_run-20260511-132325_fmriprep-warpkit.log](https://github.com/user-attachments/files/27614520/ds006072_v1_0_8_sub-P1R_run-20260511-132325_fmriprep-warpkit.log)

I also saw the draft changes in https://github.com/DVSneuro/fmriprep/pull/1, including the move toward routing MEDIC through SDCFlows. Thanks for doing that, and I'm happy to test whenever it's ready! I understand that move may make `--me-use-warpkit` unnecessary, and that the exact home/name for a noise-frame option may change or become moot depending on how MEDIC is exposed through SDCFlows. I’m very open to whatever structure the maintainers think is best -- I'm mostly I’m excited about the possibility of getting this working cleanly in fMRIPrep!

DVSneuro:
Validation update after testing @vanandrew 's correctness-fix branch from DVSneuro/fmriprep#1.

I rebuilt the container and reran the public validation datasets I’ve been using. MEDIC workflows ran and completed for:

- ds002278/sub-Blossom
- ds005085/sub-10017
- ds006131/sub-20188
- ds006926/sub-a01
- ds007637/sub-01

For ds005085/sub-10017, which includes a mix of single-echo and multi-echo runs, the single-echo runs followed the standard PHASEDIFF fieldmap path, while the eligible 4-echo runs used MEDIC/warpkit. I previously tested with ds005085/sub-10015, which didn't have any multi-echo data (an oversight on my part), but that test case didn't fail, which may be undesirable with the `--me-use-warpkit` option?

I also separately reran ds006072/sub-P1R/ses-1 with `--me-warpkit-noise-frames 3`, which resolved the prior warpkit phase-unwrapping failure from trailing noise frames. But, I did notice that the the masking might be getting messed up when using the `--me-warpkit-noise-frames 3` option. This happened on both my tests, and I just missed it last time. The contour lines look ok, so maybe it's just image they are overlaid onto?
sub-P1R_ses-1_task-BOLDREST1_dir-PA_run-1_part-mag_desc-rois_bold

Otherwise, across these runs, I did not see crashes, tracebacks, fieldmap-length mismatches, or OOM failures in the logs.

One small reporting/QC question I noticed while reviewing the HTML: the MEDIC fieldmap reportlet uses the sdcflows FieldmapReportlet hover/toggle behavior, and the `Fieldmap (Hz)` layer can visually resemble the BOLD/magnitude underlay because the fieldmap is displayed over the reference image. I’m not sure this is incorrect, but the labeling/caption may be worth a maintainer look to make sure it is clear to users.

Thanks again for helping with this! @effigies -- looking forward to your comments and feedback when you get a chance!

Best wishes,
David

effigies:
Hey @DVSneuro @vanandrew @tsalo I wonder if we could catch up on a call this week. I'm pretty slammed at the moment, so I'm not sure when I can review this thoroughly and don't want to leave you hanging. A call where you could generally walk us through it all might be helpful.

tsalo:
@effigies that would be great. I can make time for a call.

vanandrew:
@effigies No worries. I'm also pretty busy this week, but can do next week or later if we want to do a call.

Where I think things stand: MEDIC needs to land in SDCflows first (https://github.com/nipreps/sdcflows/pull/541), or at least be stable enough to build off of. I wouldn't merge this PR as-is, it predates the SDCflows work and would lock in a parallel MEDIC path we're about to rework.

Once https://github.com/nipreps/sdcflows/pull/541 is in, we can start the fMRIPrep side. @DVSneuro has a good start here, and pieces of it carry over (the part-phase BIDS filtering, the framewise-4D ResampleSeries path, boilerplate/citations). The main thing to revisit is that with SDCflows owning MEDIC as a first-class estimator, we can simplify by routing it through the normal fieldmap pipeline rather than as a multi-echo-specific branch.

DVSneuro:
Thanks, @effigies. I’d be happy to chat through it, though no pressure on timing from my end. This isn’t holding anything up for us, and I’m also pretty swamped with a couple of grants right now, so the first week of June or later would work well for me!

I agree with @vanandrew’s read here. If the preferred long-term path is for MEDIC to land in SDCFlows first via nipreps/sdcflows#541, then I’m very happy to treat this PR as a working prototype/validation branch rather than something that should merge as-is.

When the timing makes sense, I’d be happy to walk through what this branch currently does, the datasets I tested, and the pieces that might carry over to the fMRIPrep side, especially the `part-phase` discovery, framewise/4D resampling path, and reporting/boilerplate pieces. And I’m happy to help refactor or retest once the SDCFlows side settles!

effigies:
@vanandrew @DVSneuro Sounds good. I've created a when2meet for next week and filled in my availability: https://www.when2meet.com/?36851069-sVt5P

If we can't make that, I will next have some availability after OHBM.

cc @mgxd @tsalo @oesteban (Oscar, I assume you're not available, but I think you'd be interested if you are.)

effigies:
Meeting invitation sent

oesteban:
Sorry, I missed this. When's the meeting, finally?

effigies:
3pm ET on Monday. There is also a Thursday slot everyone can make.

fmriprep/interfaces/resampling.py
```diff
@@ -378,14 +378,23 @@ async def resample_series_async(
...
+
     if data.ndim == 3:
         return resample_vol(
             data,
             coordinates,
             pe_info[0],
             jacobian,
             hmc_xfms[0] if hmc_xfms else None,
-            fmap_hz,
+            fmap_hz[..., 0] if fmap_is_series else fmap_hz,
```
tsalo, line 397: I don't think you need the inline if/else since line 382 raises an error if `fmap_is_series` and `data.ndim == 3`. So this should only be `fmap_hz`.

PR #3667, opened 2026-08-11
chore: adopt spec0, test on 3.14

mgxd:


fmriprep/utils/asynctools.py
```diff
@@ -1,11 +1,8 @@
 import asyncio
 from collections.abc import Callable
-from typing import TypeVar
 
-R = TypeVar('R')
 
-
-async def worker(job: Callable[[], R], semaphore: asyncio.Semaphore) -> R:
+async def worker[R](job: Callable[[], R], semaphore: asyncio.Semaphore) -> R:
```
effigies, line 5: I don't think this is valid syntax. I think there's an open PEP to add this to functions and not just classes.

mgxd: [PEP 695](https://peps.python.org/pep-0695/#summary-examples) adds this syntax to functions and classes - now that 3.12 is the min we can freely use it

effigies: Ah, I misunderstood the scope of https://peps.python.org/pep-0718/.

effigies (approved):

PR #3670, opened 2026-08-20
doc: Add note on StartTime metadata

effigies:
This PR adds a note on StartTime metadata. We previously warned that STC impacts modeling decisions, but did not indicate that the necessary metadata to automatically adjust onsets was included in recent versions in the documentation.

* [Functional derivatives](https://fmriprep--3670.org.readthedocs.build/en/3670/outputs.html#functional-derivatives)

cc @jmumford


jmumford:
Looks good, thanks!

PR #3668, opened 2026-08-19
ENH: Enable `--bold-coreg-level` to select a shared coregistration template

mgxd:
This adds alternative BOLD coregistration target spaces, selected with ``--bold-coreg-level {run,session,subject}``.

Historically, fMRIPrep coregistered each BOLD run to the anatomical reference independently; this remains the default (``--bold-coreg-level run``). With ``--bold-coreg-level session`` or ``subject``, the run references are first combined into a common boldref template, which is coregistered to the anatomical reference once. This writes the template image alongside ``from-run_to-<session|subject>`` and ``from-<session|subject>_to-T1w`` transforms.

Additionally, this renames several functional outputs so the ``space`` and ``from``/``to`` entities name the reference explicitly (`run|session|subject`).

PR #3679, opened 2026-08-27
Bold coreg level review

effigies:
Changes on top of #3668. Trying to understand how everything fits together. Biggest change is attempting to use `{run,template}_{boldref,mask}` consistently for the sake of clarity, but this led to a couple additional changes:

* Pass `hmc_boldref` to `bold_apply_wf` so that the FSL motion outlier interface can estimate center of gravity using the same file.
* Generate per-run reportlets in run space, rather than template space. This will produce more visual difference between runs within a report, but should also be more consistent within run between fmriprep calls (different versions, different flags).

I dropped a level of indirection in `init_bold_anat_coreg_wf`, but I'm not committed to this change.

I haven't finished my review, just posting this to trigger discussion if anything here needs it and also get a full CI run.


fmriprep/workflows/base.py
```diff
@@ -991,10 +991,10 @@ def _jacobian_for(fieldmap_id):
                         'bold_file',
                         'template2anat_xfm',
                         'run2template_xfm',
-                        'coreg_boldref',
-                        'bold_mask',
+                        'template_boldref',
+                        'template_mask',
```
mgxd: we output but never use/ingest this - since we just apply the template transform to each run mask, i'm ok with dropping this, and the datasink (unless we want to provide via `--debug coreg`

fmriprep/workflows/bold/base.py
```diff
@@ -755,8 +756,8 @@ def init_bold_apply_wf(
         (inputnode, bold_confounds_wf, [
             ('t1w_tpms', 'inputnode.t1w_tpms'),
             ('t1w_mask', 'inputnode.t1w_mask'),
-            ('orig_bold_mask', 'inputnode.bold_mask'),
-            ('run_boldref', 'inputnode.hmc_boldref'),
+            ('run_mask', 'inputnode.run_mask'),
+            ('hmc_boldref', 'inputnode.hmc_boldref'),
```
mgxd, line 759: good catch!

fmriprep/workflows/bold/coreg.py
```diff
@@ -45,7 +45,7 @@
     'fsnative2t1w_xfm',
 ]
 OUTPUT_FIELDS = [
-    'coreg_boldrefs',
+    'template_boldrefs',
     'bold_masks',
```
mgxd: potential removal

fmriprep/workflows/bold/outputs.py
```diff
@@ -253,6 +255,15 @@ def init_func_fit_reports_wf(
         mem_gb=config.DEFAULT_MEMORY_MIN_GB,
     )
 
+    anat2run_xfms = pe.Node(
+        niu.Merge(2), name='anat2run_xfms', run_without_submitting=True, mem_gb=1
```
mgxd: shouldn't require much memory

```suggestion
        niu.Merge(2), name='anat2run_xfms', run_without_submitting=True, mem_gb=config.DEFAULT_MEMORY_MIN_GB
```

.github/workflows/test.yml
```diff
@@ -11,6 +11,7 @@ on:
       - master
       - maint/*
       - next
+      - enh/bold-coreg-level
```
mgxd: just flagging this for removal once we're happy

fmriprep/workflows/bold/outputs.py
```diff
@@ -795,8 +796,8 @@ def init_ds_volumes_wf(
                 'source_files',
                 'ref_file',
                 'bold',  # Resampled into target space
-                'bold_mask',  # boldref space
-                'bold_ref',  # boldref space
+                'run_mask',  # boldref space
```
mgxd, line 803: since this is now the run_boldref, we need to make sure `boldref2target` (perhaps also rename to run2target for clarity) accounts for run2template as well.

fmriprep/workflows/bold/outputs.py
effigies: I think that was a copilot injection I didn't notice.

fmriprep/workflows/base.py
effigies: YAGNI

PR #3682, opened 2026-09-02
rf: Reorganize template coregistration

effigies:
Review PR of #3668. There are a few minor things here, but the big changes I'm proposing are:

1) Consolidate checking for precomputed inputs and defining/connecting plumbing nodes (buffers, merges, expansions) at the front of the workflow. This allows the logic of the different workflow branches to be less cluttered. It also drops the `skip_*` variables; this allows the conditions to speak for themselves (is the precomputed value present?) rather than remembering why we skipped things.
2) When there are multiple branches, prioritize the no-precomputed-derivatives first, follow up with all-precomputed-derivatives, and finish with the partial case. It's easier for me to follow the logic, and I hope you'll agree.

This also includes some fixes:

1) Buffer nodes receive sunk derivatives, not intermediates, which ensures better alignment between the clean and partial-precomputed cases.


fmriprep/workflows/bold/coreg.py
```diff
@@ -234,25 +232,25 @@ def init_bold_run_coreg_wf(
...
             output_dir=output_dir,
             source=coreg_space,
             dest=reference_anat,
             desc='coreg',
-            name=f'ds_template2anat_{bold_id}',
+            name=f'ds_template2anat_{bold_id}_wf',
         )
 
         workflow.connect([
+            (inputnode, reg_wf, ANAT_REG_INPUTS),
```
mgxd: `workflow.connect` will actually consume the lists, so best to make a copy given we iterate through this.

```suggestion
            (inputnode, reg_wf, list(ANAT_REG_INPUTS)),
```

fmriprep/workflows/bold/coreg.py
```diff
@@ -428,106 +481,28 @@ def init_bold_template_coreg_wf(
...
-            # Pass single template->anat transform to all runs
-            workflow.connect([
-                (reg_buffer, merge_template2anat, [('template2anat', f'in{i + 1}')]),
-                (reg_buffer, merge_run2anat, [('template2anat', 'in2')]),
-            ])  # fmt:skip
-
-        workflow.connect([
-            (merge_run2anat, concat, [('out', 'in_xfms')]),
-            (concat, merge_run2anat_xfms, [('out_xfm', f'in{i + 1}')]),
+            (inputnode, boldref_reg_wf, ANAT_REG_INPUTS),
```
mgxd:
```suggestion
            (inputnode, boldref_reg_wf, list(ANAT_REG_INPUTS)),
```

fmriprep/workflows/bold/template.py
```diff
@@ -62,8 +61,6 @@ def init_bold_template_wf(
     -------
     boldref
         The computed session-level BOLD reference.
```
mgxd:
```suggestion
        The computed BOLD template reference.
```

mgxd (commented):
this looks reasonable, but let's get CI passing

effigies:
The workflows have progressed to the point they should finish. Will push to your branch and keep things moving.

PR #3687, opened 2026-09-11
rf: Transition to nipost derivatives collection and resampling

effigies:
Depends on:

* nipreps/nipost#1
* nipreps/smriprep#585


fmriprep/data/fmap_spec.yml
```diff
@@ -0,0 +1,25 @@
+fieldmaps:
```
mgxd: isn't this already in nipost?

